Issue #20, opened 2020-04-17
Improve Semicolons


b-studios:
There is two things we should do:

1. Allow trailing Semicolons
2. Implement semicolon inference

For the later, we probably need to move to manual whitespace handling in the parser

b-studios:
See #21

PR #21, opened 2020-04-18
First stab at automatic semicolon insertion by looking back (#20)

b-studios:
I draft implemented automatic semicolon insertion in the following way:

the Kiama parser combinators already consume all whitespace when matching a token. Since newlines are considered white space, those are skipped and cannot be matched on.

Instead of refactoring the whole parser and manually treating whitespace I implemented a semicolon parser that checks:

- whether at the current position is a semicolon
- or whether preceding the current position is some newline

That is, starting from
```
val x = 4\n  val y = 5
             ^
```
we go backwards, skiping whitespaces to end up at the newline
```
val x = 4\n  val y = 5
          ^
```
We accept this newline as a replacement for a semicolon.


b-studios:
@phischu could you maybe experiment with the insertion and see whether it matches your expected behavior? That would be helpful

b-studios:
Trailing semicolons still do not work. For this, we need to change the grammar from semicolons being separators to semicolons being delimiters. Additionally, semicolons then also need to be inferred if there is a lookahead `}` or `case`.

PR #26, opened 2020-06-26
Case insensitivity of the import statement

edding4500:
The import statement allowed import of files without checking for cases in file names.
This branch should fix the issue (by aborting translation when the name of the import file is unequal to the file name of the imported module itself) and introduce an according test scenario.

### DIFF
--- a/jvm/src/test/scala/effekt/RegressionTests.scala
+++ b/jvm/src/test/scala/effekt/RegressionTests.scala
@@ -10,7 +10,6 @@ import scala.util.matching._
 import org.scalatest.funspec.AnyFunSpec
 
 import scala.language.implicitConversions
-import scala.util.control.NonFatal
 
 class RegressionTests extends AnyFunSpec with TestUtils {
 
--- a/jvm/src/test/scala/effekt/RegressionTests.scala
+++ b/jvm/src/test/scala/effekt/RegressionTests.scala
@@ -10,6 +10,7 @@ import scala.util.matching._
 import org.scalatest.funspec.AnyFunSpec
 
 import scala.language.implicitConversions
+import scala.util.control.NonFatal
 
 class RegressionTests extends AnyFunSpec with TestUtils {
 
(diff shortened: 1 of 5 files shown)


jvm/src/test/scala/effekt/RegressionTests.scala
```diff
@@ -10,6 +10,7 @@ import scala.util.matching._
 import org.scalatest.funspec.AnyFunSpec
 
 import scala.language.implicitConversions
+import scala.util.control.NonFatal
```
b-studios, line 13: Could you remove the unused import?

shared/src/main/scala/effekt/util/ColoredMessaging.scala
```diff
@@ -21,10 +21,10 @@ class ColoredMessaging(positions: Positions) extends Messaging(positions) {
       case Some(pos) =>
         val severity = severityToWord(message.severity)
         val context = util.Highlight(pos.optContext.getOrElse(""))
-        s"[$severity] ${pos.format} ${homogenizePath(message.label)}\n$context\n"
+        s"[$severity] ${homogenizePath(pos.format)} ${message.label}\n$context\n"
```
b-studios, line 24: Perfect. I don't know how I ended up homogenizing the message label...

shared/src/main/scala/effekt/Namer.scala
```diff
@@ -45,6 +45,10 @@ class Namer extends Phase[Module, Module] { namer =>
     val imports = mod.decl.imports map {
       case im @ source.Import(path) => Context.at(im) {
         val modImport = Context.moduleOf(path)
+        // compare path with modImport.path to ensure correct handling of case sensitivity
+        if (modImport.decl.path != path) {
+          Context.abort(s"Could find an import, but has wrong module path.\nPath declared in the module: ${modImport.decl.path}")
```
b-studios, line 50: Maybe we can be even more specific and say "Path declared in the *imported* module:".
WDYT?

b-studios:
@edding4500 the CI test fails on unix systems since there it (apparently) already has the desired semantics. 

It is a bit fragile, but you could change the error message to 'Cannot find source for examples/pos/liStS" to match the behavior on unix.

b-studios:
Superseded by #158

PR #28, opened 2020-09-17
Add types of temporaries to DB.

phischu:


shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -93,8 +93,9 @@ class Transformer extends Phase[Module, core.ModuleDecl] {
     case source.DefStmt(d, rest) =>
       transform(d, transform(rest))
 
-    case source.ExprStmt(e, rest) =>
-      Val(Wildcard(C.module), ANF { transform(e).map(Ret) }, transform(rest))
+    case source.ExprStmt(e, rest) => {
+      Val(freshWildcardFor(tree), ANF { transform(e).map(Ret) }, transform(rest))
```
b-studios, line 97: We could also have a companion object:

```
object Wildcard {
  def apply(tree)
}
```

b-studios: Same for Tmp

shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -235,11 +236,28 @@ class Transformer extends Phase[Module, core.ModuleDecl] {
     StateEffect(eff, get, put)
   }
 
+  def freshTmpFor(e: source.Tree)(implicit C: TransformerContext): Tmp = {
+    val x = Tmp(C.module)
+    C.typeOf(e) match {
+      case Some(t / _) => C.assignType(x, t)
+      case _           => C.abort("Internal Error: Missing type of source expression.")
+    }
```
b-studios, line 244: You could replace this with:

```
val (tpe / eff) = C.typeOf(e).getOrElse { C.abort("Internal Error: Missing type of source expression.") }
C.assignType(x, tpe)
```
but it is mostly personal taste. In your version the side effect `assignType` is a tiny bit hidden.

shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -23,8 +23,8 @@ package object symbols {
   sealed trait TermSymbol extends Symbol
 
   // the two universes of values and blocks
-  sealed trait ValueSymbol extends TermSymbol
-  sealed trait BlockSymbol extends TermSymbol
+  trait ValueSymbol extends TermSymbol
+  trait BlockSymbol extends TermSymbol
```
b-studios, line 27: Ah, that's a bit sad because this way we loose exhaustivity checks.

b-studios (approved):

b-studios:
LGTM overall, see the smaller comments.

PR #29, opened 2020-10-15
Add support for bidirectional effects

long-long-float:
This PR adds the feature of bidirectional control flow to the Effekt language. The feature is inspired by the paper

> "Handling bidirectional control flow".
>  Yizhou Zhang, Guido Salvaneschi, and Andrew C. Myers. OOPSLA'20 ([PDF](https://dl.acm.org/doi/10.1145/3428207)).


b-studios:
Could you maybe add a couple of example files with expected output check files to the `examples/pos` directory?

shared/src/main/scala/effekt/Parser.scala
```diff
@@ -426,7 +426,10 @@ class Parser(positions: Positions) extends Parsers(positions) with Phase[Source,
     idRef ~ maybeTypeArgs ~ some(args) ^^ Call
 
   lazy val resumeExpr: P[Expr] =
-    (`resume` ^^^ IdRef("resume")) ~ valueArgs ^^ { case r ~ args => Call(r, Nil, List(args)) withPositionOf r }
+    (
+      ((`resume` ^^^ IdRef("resume")) ~ valueArgs ^^ { case r ~ args => Call(r, Nil, List(args)) withPositionOf r })
+    | ((`resume` ^^^ IdRef("resume")) ~ (`{` ~> stmts <~ `}`) ^^ { case r ~ body => Call(r, Nil, List(BlockArg(ValueParams(Nil), body))) withPositionOf r })
+    )
```
b-studios, line 432:
```suggestion
(`resume` ^^^ IdRef("resume")) ~ args ^^ { case r ~ args => Call(r, Nil, List(args)) withPositionOf r }
```

shared/src/main/scala/effekt/Namer.scala
```diff
@@ -113,7 +113,8 @@ class Namer extends Phase[Module, Module] { namer =>
           val name = Context.freshTermName(id)
           Context scoped {
             val tps = tparams map resolve
-            val tpe = Effectful(resolve(ret), Effects(List(effectSym)))
+            val effs = ret.eff.effs map resolve
+            val tpe = Effectful(resolve(ret.tpe), Effects(List(effectSym) ::: effs))
             val op = EffectOp(Name(id), tps, params map resolve, Some(tpe), effectSym)
```
b-studios, line 118:
```suggestion
            val Effectful(tpe, otherEffs) = resolve(ret)
            val retResolved = Effectful(tpe, otherEffs + effectSym)
            val op = EffectOp(Name(id), tps, params map resolve, Some(retResolved), effectSym)
            Context.define(id, op)
```

shared/src/main/scala/effekt/Typer.scala
```diff
@@ -115,7 +115,9 @@ class Typer extends Phase[Module, Module] { typer =>
               val effectOp = d.definition
               val bt = Context.blockTypeOf(effectOp)
               val ps = checkAgainstDeclaration(op.name, bt.params, params)
-              val resumeType = BlockType(Nil, List(List(effectOp.ret.get.tpe)), ret / Pure)
+              val effs = effectOp.ret.get.effects - effectOp.effect
+              val resumeArgType = BlockType(Nil, List(Nil), Effectful(effectOp.ret.get.tpe, effs))
+              val resumeType = BlockType(Nil, List(List(resumeArgType)), ret / Pure)
```
b-studios, line 120:
```suggestion
              val BlockType(_, pms, tpe / effs) = Context.blockTypeOf(effectOp)
              val ps = checkAgainstDeclaration(op.name, pms, params)
 
              val resumeType = if (effectOp.isBidirectional) {
                // resume { e }
                BlockType(Nil, List(List(BlockType(Nil, List(Nil), tpe / effectOp.otherEffects))), ret / Pure)
              } else {
                // resume(v)
                BlockType(Nil, List(List(tpe)), ret / Pure)
              }
```

We should not allow `resume(e)` syntax for bidirectional effects. Otherwise this will lead to a runtime error "f.apply is not a function".

The methods `isBidirectional` and `otherEffects` should be added to the EffectOp symbol, like you did previously:

```
def otherEffects: Effects = ret.get.effects - effect
def isBidirectional: Boolean = otherEffects.nonEmpty
```

shared/src/main/scala/effekt/Typer.scala
```diff
@@ -557,7 +559,7 @@ class Typer extends Phase[Module, Module] { typer =>
     //   BlockArg: foo { n => println("hello" + n) }
     //     or
     //   BlockArg: foo { (n: Int) => println("hello" + n) }
-    def checkBlockArgument(tpe: BlockType, arg: source.BlockArg): Unit = Context.at(arg) {
+    def checkBlockArgument(tpe: BlockType, arg: source.BlockArg, isResume: Boolean): Unit = Context.at(arg) {
```
b-studios, line 562: I would revert the changes to `checkBlockArgument`. Instead for now it would suffice to reject `resume(v)` for an bidirectional operation and `resume { e }` for a non-bidirectional one. It would be great to have an error in that case that explains the situation well.

shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -165,6 +165,9 @@ class Transformer extends Phase[Module, core.ModuleDecl] {
           val params = ps.params.map { v => core.ValueParam(v.symbol) }
           val caps = p.ret.effects.userDefined.toList.map { core.BlockParam }
           pure { List(BlockLit(params ++ caps, transform(body))) }
+        case (source.BlockArg(ps, body), _) if sym.isInstanceOf[ResumeParam] =>
+          val params = ps.params.map { v => core.ValueParam(v.symbol) }
+          pure { List(BlockLit(params, transform(body))) }
```
b-studios, line 170: That should not be necessary, if I am not mistaken.

lib/effekt_runtime.js
```diff
@@ -152,8 +152,20 @@ const $runtime = (function() {
       for (var op in h) {
         const impl = h[op];
         cap[op] = function() {
+          // split two kinds of arguments, parameters of the operation and capabilities
           const args = Array.from(arguments);
-          return shift(p)(k => impl.apply(null, args.concat([k])))
+          const olen = impl.length
+          const oargs = args.slice(0, olen - 1)
+          const cargs = args.slice(olen - 1, args.length)
```
b-studios, line 159:
```suggestion
          const arity = impl.length - 1
          const oargs = args.slice(0, arity)
          const caps = args.slice(arity)
```

lib/effekt_runtime.js
```diff
@@ -152,8 +152,20 @@ const $runtime = (function() {
...
+          var r = shift(p)(k => {
+            return impl.apply(null, oargs.concat([k]))
+          })
```
b-studios, line 162:
```suggestion
          var r = shift(p)(k => impl.apply(null, oargs.concat([k])))
```

lib/effekt_runtime.js
```diff
@@ -152,8 +152,20 @@ const $runtime = (function() {
...
+          if (cargs.length > 0) {
+            return r.then(f => f.apply(null, cargs))
+          }
+          else {
+            return r
+          }
```
b-studios, line 168:
```suggestion
          // resume { caps => e}
          if (caps.length > 0) {
            return r.then(f => f.apply(null, caps))
          // resume(v)
          } else {
            return r
          }
```

b-studios (changes requested):
Looks already pretty good to me. I added a few suggested changes that would be great if you could incorporate them. As already mentioned, it would also be great to have a few examples that illustrate the new feature.

In the changes I propose to reject programs where `resume` is used in the wrong way. I think that is easiest for now. If you would like to stick with the automatic adaption to the correct kind, I think we need another, consistent way that actually changes the terms.

b-studios:
_closing and reopening to trigger CI_

b-studios:
Thank you @long-long-float for your contributions! As you can see, there are still a few tests that do not pass and I cannot merge your feature without the CI going through.

The Effekt language has multiple backends and your bidirectionality feature only supports one of them (JavaScript), which is fine. It is expected that your new test fails on all other backends. You can see how the bidirectionality test can be disabled in the other backends here:

https://github.com/effekt-lang/effekt/blob/581b1113eee97dabe0ab4e9a0a0118cef9d81b3d/jvm/src/test/scala/effekt/ChezSchemeTests.scala#L18-L28

However, all other tests that are not concerned with bidirectionality should still pass.  Maybe you can look into this? If you have any questions, I am happy to assist.

shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -100,7 +100,7 @@ case class VarDef(id: IdDef, annot: Option[ValueType], binding: Stmt) extends De
 case class EffDef(id: IdDef, ops: List[Operation]) extends Def {
   type symbol = symbols.UserEffect
 }
-case class Operation(id: Id, tparams: List[Id], params: List[ValueParams], ret: Effectful) extends Definition {
+case class Operation(id: IdDef, tparams: List[Id], params: List[ValueParams], ret: Effectful) extends Definition
```
b-studios, line 103: Oh.. I forgot the opening brace here, when editing this in the online editor. @long-long-float could you fix this in your branch?

long-long-float:
@b-studios The CI fails and says an error `[error] examples/neg/existential_effect_leaks.effekt:22:1: Unhandled effects: GiveInt`. My modification may effect this. Should I fix this?

b-studios:
Yes, it should warn about an effect leaving its scope as you can see in the check file. Would be great if you could look into this

long-long-float:
@b-studios I fixed the test case of `examples/neg/existential_effect_leaks.effekt`, but other cases are faild.

```
[info]   - unbound_type.effekt (lift) *** FAILED *** (145 milliseconds)
[info]     "... pos: Position): A
[info]     ^[^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^]
[info]     " did not equal "... pos: Position): A
[info]     ^[]
[info]     " (ChezSchemeTests.scala:62)
[info]     Analysis:
[info]     "... pos: Position): A
[info] ^[^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^]
[info] " -> "... pos: Position): A
[info] ^[]
```

I Investigated the cause and guess that the position of the type of an operator (`ret: Effectful`) is not set (but I don't know how to set it.)

b-studios:
Don't worry -- just update the test file to reflect the new position. We can fix and improve positions later.

b-studios:
Actually, while you are at it: You could update the methods `resolveTerm`, `resolveType`, and `resoleCallTarget` to something like:

```
  private[namer] def resolveType(id: Id): TypeSymbol = at(id) {
  //                                                    ^^^^^
    val sym = scope.lookupType(id.name)
    assignSymbol(id, sym)
    sym
  }
```
This will improve the error positions of the namer

b-studios:
Update: I just modified Namer in master so you only need to rebase. This should hopefully also solve the problem you described here.

shared/src/main/scala/effekt/Typer.scala
```diff
@@ -571,7 +579,13 @@ class Typer extends Phase[Module, Module] { typer =>
       effs = (effs ++ (stmtEffs -- handled))
     }
 
-    (params zip args) foreach { case (ps, as) => checkArgumentSection(ps, as) }
+    (params zip args) foreach {
+      // Special case - treat 1st param type () => T as T
```
b-studios, line 583: Maybe you could expand a bit on this comment (and maybe add an example?). This part is really the most tricky bit, I believe.

long-long-float: I reconsidered about it, it is unnecessary. So I removed it.

long-long-float:
@b-studios All tests are passed! Thanks for your advice.

b-studios (approved):

b-studios:
Thanks for your contributions @long-long-float, which I'll merge now!

If you want, you could add more examples of the feature, potentially translating the examples from the original paper? The paper is presented this week at OOPSLA and having translated their examples could facilitate discussion.

long-long-float:
Of course I'm creating more examples for the paper 😄

b-studios:
@long-long-float Just to be clear: What I meant is: could you take examples from the existing paper by Zhang er al., translate them to Effekt and add them to this repository?

long-long-float:
Yes, I'm translating examples from the existing paper by Zhang er al., to Effekt.
* Generators: I'm translating it.
* Async-await: I considered that it cannot be because Effekt doesn't have first-class function.
* Ping and Pong: I translated it as `examples/pos/bidirectional.effekt`.

b-studios:
Hey, thanks! I added a quick translation of the iterator example [here](https://github.com/effekt-lang/effekt/blob/master/examples/pos/bidirectional/iterators.effekt) to talk about it with people at the OOPLSA conference.

long-long-float:
Cool! I have just translated the same example [here](https://gist.github.com/long-long-float/bddd5cbdf8a76c7a49b38870713e7811).

b-studios:
That's great. Since you are the expert on bidirectional effects in the Effekt language, do you want to document it on the website? We have a separate repository for that. I added a quick stub, but there could be much detailed explanation about what bidirectional effects motivates, how they work and how they interact with effect polymorphism:

https://github.com/effekt-lang/effekt-website/blob/master/docs/bidirectional.md

long-long-float:
Sure, I want to do. I'll write it with my master thesis.

PR #31, opened 2020-10-21
Modules

anfark:
Added examples for possible module system syntax

### DIFF
--- a/examples/modules/01-nested-modules/example.effekt
+++ b/examples/modules/01-nested-modules/example.effekt
@@ -1,0 +1,15 @@
+module hello {
+    def greet(name: String) = {
+        println(concat("Hello ", name))
+    }
+
+    module world {
+        def greet() = {
+            greet("World")
+        }
+    }
+}
+
+def main() = {
+    hello.world.greet()
+}
--- a/examples/modules/05b-partial-import/example.effekt
+++ b/examples/modules/05b-partial-import/example.effekt
@@ -1,0 +1,6 @@
+import hello.world.{greet goodbye}
+
+def main() = {
+    greet()
+    goodbye()
+}
--- a/examples/modules/06a-rename-import/example.effekt
+++ b/examples/modules/06a-rename-import/example.effekt
@@ -1,0 +1,5 @@
+import hello.world as hw
+
+def main() = {
+    hw.greet()
+}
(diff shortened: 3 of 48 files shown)


examples/modules/01-nested-modules/example.effekt
```diff
@@ -0,0 +1,15 @@
+module hello {
+    def greet(name: String) = {
+        println(concat("Hello ", name))
+    }
+
+    module world {
+        def greet() = {
+            greet("World")
```
anfark, line 8: Design Decision: Should nested modules implicitly import their parent?

lib/text/string.effekt
```diff
@@ -28,4 +28,7 @@ extern pure def unsafeCharAt(str: String)(n: Int): String =
   "str[n]"
 
 extern pure def unsafeToInt(str: String): Int =
-  "Number.isNaN(parseInt(str)) ? undefined : parseInt(str)"
\ No newline at end of file
+  "Number.isNaN(parseInt(str)) ? undefined : parseInt(str)"
+
+extern pure def concat(str1: String, str2: String): String =
+  "str1.concat(str2)"
```
b-studios, line 34: It is great that you add what you need, but this is already defined in the prelude :) 

https://github.com/effekt-lang/effekt/blob/98ba53335e8fa53c9ebc72f5fc18142a2fc52457/lib/effekt.effekt#L37-L38

examples/modules/05b-partial-import/example.effekt
```diff
@@ -0,0 +1,6 @@
+import hello.world.{greet goodbye}
```
b-studios, line 1: Since we borrow a lot of Syntax from Scala, we could think about also using a comma to separate different imports. That is: 

```suggestion
import hello.world.{greet, goodbye}
```

examples/modules/06a-rename-import/example.effekt
```diff
@@ -0,0 +1,5 @@
+import hello.world as hw
```
b-studios, line 1: Scala uses the following syntax for renaming imports:

```suggestion
import hello.{ world => hw }
```

An advantage is that this way you can import multiple names in one line. That is,

```
import foo.bar as fb
import foo.bam as fm
```
becomes
```
import foo. { bar => fb, bam => fm}
```

We could also use the keyword `as` instead of overloading the `=>` operator. So this looks like 
```
import foo. { bar as fb, bam as fm}
```

examples/modules/01-nested-modules/example.effekt
b-studios: I would go with yes, it is just like nested scopes.

anfark:
Moved branch to [feature/modules](https://github.com/effekt-lang/effekt/tree/feature/modules)

PR #32, opened 2020-10-28
Refactoring usage of IDs

anfark:
IDs can be references or definitions. Currently the property `id` of `effekt.source.Definition` and `effekt.source.Reference` is just typed as `Id`. 

To ensure the correct usage of both traits `Definition.id` and `Reference.id` should be typed as `IdDef` and `IdRef` respectively.

### DIFF
--- a/shared/src/main/scala/effekt/Parser.scala
+++ b/shared/src/main/scala/effekt/Parser.scala
@@ -238,7 +238,7 @@ class Parser(positions: Positions) extends Parsers(positions) with Phase[Source,
   lazy val effectDef: P[Def] =
     ( `effect` ~> effectOp ^^ {
         case op =>
-          EffDef(IdRef(op.id.name) withPositionOf op.id, List(op))
+          EffDef(IdDef(op.id.name) withPositionOf op.id, List(op))
       }
     | `effect` ~> idDef ~ (`{` ~/> some(`def` ~> effectOp)  <~ `}`) ^^ EffDef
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/effekt/Parser.scala
```diff
@@ -238,7 +238,7 @@ class Parser(positions: Positions) extends Parsers(positions) with Phase[Source,
   lazy val effectDef: P[Def] =
     ( `effect` ~> effectOp ^^ {
         case op =>
-          EffDef(IdRef(op.id.name) withPositionOf op.id, List(op))
+          EffDef(IdDef(op.id.name) withPositionOf op.id, List(op))
```
anfark, line 241: Not sure if this change will break something. If yes this pull request is obsolete.

shared/src/main/scala/effekt/Parser.scala
```diff
@@ -238,7 +238,7 @@ class Parser(positions: Positions) extends Parsers(positions) with Phase[Source,
   lazy val effectDef: P[Def] =
     ( `effect` ~> effectOp ^^ {
         case op =>
-          EffDef(IdRef(op.id.name) withPositionOf op.id, List(op))
```
b-studios, line 241: Ha! good catch.

b-studios:
_closed and reopened for CI_

b-studios (approved):

PR #33, opened 2020-10-28
Refactor Name trait

anfark:
Refactor Name trait to better handle qualified names.

### DIFF
--- a/jvm/src/main/scala/effekt/Server.scala
+++ b/jvm/src/main/scala/effekt/Server.scala
@@ -79,7 +79,7 @@ trait LSPServer extends Driver with Intelligence {
     decl = id // TODO for now we use id as the declaration. This should be improved in SymbolsDB
     kind <- getSymbolKind(sym)
     detail <- getInfoOf(sym)(context)
-  } yield new DocumentSymbol(sym.name.name, kind, rangeOfNode(decl), rangeOfNode(id), detail.header))
+  } yield new DocumentSymbol(sym.name.localName, kind, rangeOfNode(decl), rangeOfNode(id), detail.header))
 
   override def getReferences(position: Position, includeDecl: Boolean): Option[Vector[Tree]] =
     for {
--- a/shared/src/main/scala/effekt/Namer.scala
+++ b/shared/src/main/scala/effekt/Namer.scala
@@ -456,9 +456,9 @@ trait NamerOps { self: Context =>
     scope.define(id.name, s)
   }
 
-  private[namer] def bind(s: TermSymbol): Unit = scope.define(s.name.name, s)
+  private[namer] def bind(s: TermSymbol): Unit = scope.define(s.name.localName, s)
 
-  private[namer] def bind(s: TypeSymbol): Unit = scope.define(s.name.name, s)
+  private[namer] def bind(s: TypeSymbol): Unit = scope.define(s.name.localName, s)
 
   private[namer] def bind(params: List[List[Param]]): Context = {
     params.flatten.foreach { p => bind(p) }
--- a/shared/src/main/scala/effekt/generator/JavaScript.scala
+++ b/shared/src/main/scala/effekt/generator/JavaScript.scala
@@ -3,7 +3,7 @@ package effekt.generator
 import effekt.context.Context
 import effekt.context.assertions._
 import effekt.core._
-import effekt.symbols.{ Module, Name, Symbol, moduleName }
+import effekt.symbols.{ Module, Name, Symbol }
 import effekt.symbols
 import org.bitbucket.inkytonik.kiama
 import kiama.output.ParenPrettyPrinter
@@ -112,7 +112,7 @@ trait JavaScriptPrinter extends JavaScriptBase {
     case Exports(path, exports) =>
       jsCall(
         "module.exports = Object.assign",
-        moduleName(path),
+        jsModuleName(path),
         jsObject(exports.map { e => toDoc(e.name) -> toDoc(e.name) })
       )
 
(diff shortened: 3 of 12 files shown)


shared/src/main/scala/effekt/generator/JavaScript.scala
```diff
@@ -166,7 +167,7 @@ trait JavaScriptBase extends ParenPrettyPrinter {
     case _: symbols.Effect   => toDoc(id.name)
     case _: symbols.EffectOp => "op$" + id.name.toString
     case _ => id.name match {
-      case name: QualifiedName if name.module != C.module => link(name, s"${name.module.name}.${name.name}")
+      case name: NestedName if name.parent != C.module.name => link(name, s"${name.parent.localName}.${name.localName}")
```
anfark, line 170: Not sure if this still works the way it should.

b-studios: The point of this check is that names in the same module should not be referred to by the full qualified name, but the local name should be used instead.

shared/src/main/scala/effekt/symbols/Name.scala
```diff
@@ -4,26 +4,117 @@ import effekt.source.Id
...
+  /**
+   * Computes the qualified version of this name.
+   * @return A String containing the local part of this name prefixed by their parent name. Components are separated by the point character (`.`).
+   *
+   * @note this should only be used for reporting errors, not for code generation.
+   * @note Note to the previous note: Is this still true for the new module-system?
+   */
+  def qualifiedName: String = this match {
+    case ToplevelName(name)       => name
+    case NestedName(parent, name) => s"${parent.qualifiedName}.$localName"
```
b-studios, line 23: Is there a reason this is not a method on Name?

shared/src/main/scala/effekt/symbols/Name.scala
```diff
@@ -4,26 +4,117 @@ import effekt.source.Id
...
-case class QualifiedName(name: String, module: Module) extends Name {
-  def qualified: String = s"${module.name}.${name}"
-  def rename(f: String => String): Name = copy(name = f(name))
-}
+  /**
+   * @return The parent of this name or `None` if this name represents a top-level name.
+   */
+  def parentOption: Option[Name] = this match {
+    case ToplevelName(_)       => None
+    case NestedName(parent, _) => Some(parent)
```
b-studios, line 31: Same here, this could be a method, right?

shared/src/main/scala/effekt/symbols/Name.scala
```diff
@@ -4,26 +4,117 @@ import effekt.source.Id
...
+   * Computes the qualified version of this name.
+   * @return A String containing the local part of this name prefixed by their parent name. Components are separated by the point character (`.`).
+   *
+   * @note this should only be used for reporting errors, not for code generation.
+   * @note Note to the previous note: Is this still true for the new module-system?
+   */
+  def qualifiedName: String = this match {
+    case ToplevelName(name)       => name
+    case NestedName(parent, name) => s"${parent.qualifiedName}.$localName"
+  }
```
b-studios, line 24: I would say it is typically better style (at least more consistent with the remaining code base) to use overwrite `qualifiedName` in the subclass, then pattern matching on `this`.

jvm/src/main/scala/effekt/Server.scala
```diff
@@ -79,7 +79,7 @@ trait LSPServer extends Driver with Intelligence {
     decl = id // TODO for now we use id as the declaration. This should be improved in SymbolsDB
     kind <- getSymbolKind(sym)
     detail <- getInfoOf(sym)(context)
-  } yield new DocumentSymbol(sym.name.name, kind, rangeOfNode(decl), rangeOfNode(id), detail.header))
+  } yield new DocumentSymbol(sym.name.localName, kind, rangeOfNode(decl), rangeOfNode(id), detail.header))
```
b-studios, line 82: 👍  `localName` is a much better name!

shared/src/main/scala/effekt/Namer.scala
```diff
@@ -456,9 +456,9 @@ trait NamerOps { self: Context =>
     scope.define(id.name, s)
   }
 
-  private[namer] def bind(s: TermSymbol): Unit = scope.define(s.name.name, s)
+  private[namer] def bind(s: TermSymbol): Unit = scope.define(s.name.localName, s)
 
-  private[namer] def bind(s: TypeSymbol): Unit = scope.define(s.name.name, s)
+  private[namer] def bind(s: TypeSymbol): Unit = scope.define(s.name.localName, s)
```
b-studios, line 461: I guess, in the long run you want to refactor scope to index by `Name` and not string, right?

shared/src/main/scala/effekt/generator/JavaScript.scala
b-studios: Without having checked it explicitly, your refactoring looks about right here.

b-studios:
_Just closed and reopened to trigger a CI build._

b-studios:
Great Work @anfark! @phischu could you maybe check whether this might be problematic with your LLVM branch?

b-studios:
@anfark It looks like you changes broke some things. Could you check the errors in the CI build? If you have any questions, feel free to ask.

shared/src/main/scala/effekt/symbols/Name.scala
```diff
@@ -4,26 +4,117 @@ import effekt.source.Id
 import effekt.context.Context
 
 trait Name {
```
anfark, line 6: Should we override the equals and hash methods and delegate to `qualifiedName`?

b-studios: They are implemented as case classes that should already do what you want.

b-studios: That is: use the hash of the parents and of their fields. That's good enough, no?

shared/src/main/scala/effekt/symbols/Name.scala
```diff
@@ -107,11 +109,13 @@ object Name {
    */
   def apply(name: String): Name = {
     assert(name.nonEmpty, "Name cannot be empty.")
+    println(s"Convert name $name")
```
b-studios: You probably want to remove those, right?

shared/src/main/scala/effekt/Namer.scala
anfark: Yes this might be one of the use cases.

shared/src/main/scala/effekt/symbols/Name.scala
anfark: Yes that should be enough.

shared/src/main/scala/effekt/symbols/Name.scala
anfark: Yes, just some good old println-debugging.

shared/src/main/scala/effekt/symbols/Name.scala
```diff
@@ -3,27 +3,133 @@ package effekt.symbols
...
+    assert(name.nonEmpty, "Name cannot be empty.")
+
+    // Is nested?
+    if (name.contains('.')) {
+      val segments = name.split('.')
+      val top: Name = ToplevelName(segments.head)
+      segments.drop(1).foldLeft(top)((parent, segment) => parent.nested(segment))
+    } else {
+      ToplevelName(name)
+    }
```
b-studios, line 122: Why do you have to distinguish the two cases? I would assume that something like:

```
  val (top :: rest) = name.split('.').toList
  rest.foldLeft(ToplevelName(top)) { _ nested _ }
```
would do the job.

b-studios (approved):
LGTM!

PR #35, opened 2020-11-21
Fix EffectOp.isBidirectional

long-long-float:
`EffectOp.isBidirectional` has wrong behavior in following program:

```scala
module bug

effect Exn(msg: String) : Unit / { Exn }

def main() = {
  var exit = false
  try {
    try {
      do Exn("throw")
    }
    with Exn { msg =>
      resume { do Exn("re-throw") }
    }
  }
  with Exn { msg => resume { () } }
}
```

The compiler outputs following error:

```
[error] bug.effekt:12:7: Wrong type of argument section
      resume { do Exn("re-throw") }
      ^^^^^^
```

The reason of this error is that `EffectOp.isBidirectional` of `Exn` is `false` because `EffectOp.otherEffects` is empty (`{Exn} - Exn = {}`). We expects that it's `true` because `Exn` is bidirectional effect.

### DIFF
--- a/shared/src/main/scala/effekt/Namer.scala
+++ b/shared/src/main/scala/effekt/Namer.scala
@@ -115,7 +115,7 @@ class Namer extends Phase[Module, Module] { namer =>
             val tps = tparams map resolve
             val Effectful(tpe, otherEffs) = resolve(ret)
             val retResolved = Effectful(tpe, otherEffs + effectSym)
-            val op = EffectOp(Name(id), tps, params map resolve, Some(retResolved), effectSym)
+            val op = EffectOp(Name(id), tps, params map resolve, Some(retResolved), effectSym, otherEffs.nonEmpty)
             Context.define(id, op)
             op
           }
--- a/shared/src/main/scala/effekt/Namer.scala
+++ b/shared/src/main/scala/effekt/Namer.scala
@@ -115,7 +115,7 @@ class Namer extends Phase[Module, Module] { namer =>
             val tps = tparams map resolve
             val Effectful(tpe, otherEffs) = resolve(ret)
             val retResolved = Effectful(tpe, otherEffs + effectSym)
-            val op = EffectOp(Name(id), tps, params map resolve, Some(retResolved), effectSym, otherEffs.nonEmpty)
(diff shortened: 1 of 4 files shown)


b-studios:
Not sure I would call this a bug, since we haven't specified yet fully how this example should behave.

So far
```
effect Exn(msg: String) : Unit
```

would more be equivalent to
```
effect Exn(msg: String) : Unit / { Exc }
```
than to 
```
```
effect Exn(msg: String) : Unit / {}
```
which I agree can be confusing.

**Comment by b-studios** [comment]
The changes you propose in this PR should also be checked to be consistent with the behavior of Typer and Translator.

**Comment by phischu** [comment]
In my opinion, the traditional

```
effect Exn(msg: String) : Unit
```

should be the same as

```
effect Exn(msg: String) : Unit / {}
```

and

```
effect Exn(msg: String) : Unit / { Exn }
```

should be a bidirectional effect that requires `Exn` to be handled at the call-site.

The question is what should happen in the following example:

```
effect Exn(msg: String) : Unit / { Exn }

try {
  do Exn("hello")
} with Exn { (msg) =>
   println(msg); resume { do Exn(msg) }
}
```

And in my opinion it should loop and keep printing `"hello"`.

**Comment by b-studios** [comment]
I think what you propose makes sense and aligns with this PR. As mentioned it would be nice to have some test coverage for corner cases like the ones you showed.

**Comment by long-long-float** [comment]
Thanks for comments. I agree with @phischu for the behavior.
But the following programs does strange behavior. This prints only one `handled`, and this should keep print `handled`. It may be a bug.

```scala
module bug

effect Exn(msg: String) : Unit / { Exn }

def main() = {
  try {
    do Exn("throw")
  }
  with Exn { msg =>
    println("handled")
    resume { do Exn("re-throw") }
  }
}
```

BTW, I'll make a example as a corner case and fix conflicts.

b-studios:
Yes, that looks like a bug. I would expect it to keep printing "handled", like @phischu suggested

b-studios:
I implemented the semantics proposed by @long-long-float in #44 -- it was easier to redo the changes than fully rebasing this branch. I am sorry for losing the commit history here.

PR #38, opened 2020-12-21
Backport refactorings

b-studios:
Some refactorings from working on the first-class-functions branch.

### DIFF
--- a/shared/src/main/scala/effekt/symbols/symbols.scala
+++ b/shared/src/main/scala/effekt/symbols/symbols.scala
@@ -153,15 +153,15 @@ package object symbols {
   }
   case class UserCapability(effect: UserEffect) extends Capability
 
-  case class StateEffect(effect: UserEffect, get: EffectOp, put: EffectOp) extends Capability
-  object StateEffect {
-    def apply(binder: VarBinder)(implicit C: Context): StateEffect = {
+  case class StateCapability(effect: UserEffect, get: EffectOp, put: EffectOp) extends Capability
+  object StateCapability {
+    def apply(binder: VarBinder)(implicit C: Context): StateCapability = {
       val tpe = C.valueTypeOf(binder)
       val eff = UserEffect(binder.name, Nil)
       val get = EffectOp(binder.name.rename(name => "get"), Nil, List(Nil), Some(tpe / Pure), eff)
       val put = EffectOp(binder.name.rename(name => "put"), Nil, List(List(ValueParam(binder.name, Some(tpe)))), Some(builtins.TUnit / Pure), eff)
       eff.ops = List(get, put)
-      StateEffect(eff, get, put)
+      StateCapability(eff, get, put)
     }
   }
 
--- a/shared/src/main/scala/effekt/symbols/symbols.scala
+++ b/shared/src/main/scala/effekt/symbols/symbols.scala
@@ -147,7 +147,10 @@ package object symbols {
   /**
    * A symbol that represents a termlevel capability
    */
-  case class CapabilitySymbol(effect: Effect) extends BlockSymbol { val name = effect.name }
+  //  trait Capability extends BlockSymbol {
+  //    def effect: UserEffect
+  //  }
+  case class Capability(effect: UserEffect) extends BlockSymbol { val name = effect.name }
 
   /**
    * Types
@@ -241,6 +244,8 @@ package object symbols {
     val ret = Some(Effectful(tpe, Pure))
   }
 
+  /** Effects */
+
   sealed trait Effect {
     def name: Name
     def builtin: Boolean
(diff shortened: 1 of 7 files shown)


shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -141,14 +141,37 @@ package object symbols {
    *
    * Refined by typer.
    */
-  case class CallTarget(name: Name, symbols: List[Set[BlockSymbol]]) extends BlockSymbol
+  case class CallTarget(name: Name, symbols: List[Set[BlockSymbol]]) extends Synthetic with BlockSymbol
```
anfark, line 144: Can you explain why `CallTarget` became a `Synthetic` symbol? Does this change how `CallTarget` is/(should be) used in other parts of the code?

b-studios: No, it doesn't change anything. It actually also doesn't have any meaning, besides being a marker (at the moment). Namer introduces `CallTarget`  symbols that don't have a concrete counterpart in the source language. They are only used to represent ambiguous calls, before typer goes over it and replaces it by the correct (resolved) function / blocksymbol.

shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -141,14 +141,37 @@ package object symbols {
...
+  }
 
   /**
    * Types
    */
   sealed trait Type
 
-  // like Params but without name binders
+  /**
+   * like Params but without name binders
```
anfark, line 173: Is there a reason for this change? Should we use only multi-line comments or is this something arbitrary?

shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -233,16 +256,20 @@ package object symbols {
     val ret = Some(Effectful(tpe, Pure))
   }
 
-  sealed trait Effect extends TypeSymbol {
+  /** Effects */
+
+  sealed trait Effect {
```
anfark, line 261: What is the relation between `Effect`, `UserEffect` and `Capabilities`?

shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -233,16 +256,20 @@ package object symbols {
...
+    def name: Name
+    def builtin: Boolean
     // invariant: no EffectAlias in this list
     def dealias: List[Effect] = List(this)
   }
 
-  case class EffectAlias(name: Name, effs: Effects) extends Effect {
+  case class EffectAlias(name: Name, effs: Effects) extends Effect with TypeSymbol {
```
anfark, line 268: Why is `EffectAlias` and `UserEffect` a `TypeSymbol` but `Effect` isn't? This raises two questions in me:

1. Is there a subclass/-type of `Effect` which isn't a `TypeSymbol`?
2. Is an `EffectAlias` more than just a "pointer" to the original `Effect`?

shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -304,6 +336,7 @@ package object symbols {
   }
 
   lazy val Pure = new Effects(Nil)
+
```
anfark, line 339: This seams to be an useless blank line. What's your policy about this? Okay or Nay?

anfark (commented):
Wrote some questions

shared/src/main/scala/effekt/symbols/symbols.scala
b-studios: I just changed to multiline, because it would actually be available as doc comment this way. We don't generate a documentation, but this could help in the future.

shared/src/main/scala/effekt/symbols/symbols.scala
b-studios: That is a **very** good question!

`Effect`: something we want to track
`UserEffect`: something that can be handled by `try ... with ...` -- it is only called "UserEffect" because it is userdefinable, wheras `Console` is a "builtin" and can't be handled. We should have a better name ...
`Capability`: a *termsymbol* representing one particular instance of a usereffect. Only usereffects (and state) have corresponding capabilities.

shared/src/main/scala/effekt/symbols/symbols.scala
b-studios: ha, tricky one. I do not want to make `Effect` a symbol since in the future there will be effects that are not symbols. For instance `State[Int]` is an effect, but not a symbol (`State` is, and `Int` is, but the application `State[Int]` is not a symbol).

shared/src/main/scala/effekt/symbols/symbols.scala
b-studios: No policy, but Effectful is super important and shouldn't be too close to another definition.

PR #40, opened 2020-12-23
Refactor/simplify transformer

b-studios:
It is a nice academic trick to use `Control` for ANF. As long as we do not write the Effekt compiler in Effekt, however, it introduces additional complexity, not necessary. 

This PR rewrites `Transformer` to use mutable state instead of control effects. This works in particular, since `Transfomer` does not perform any sorts of backtracking.


b-studios:
BTW I am keeping the control monad implementation, in case we want to add an interpreter again.

b-studios:
@anfark this PR includes a lot of refactorings. Some of those will also affect you, so please feel free to ask. We can also organize a quick session, where I walk you through them.

PR #41, opened 2021-01-01
Feature/first class functions

b-studios:
This PR adds the feature of first-class functions / lambdas to the language. This feature is still experimental and the error messages are not where they could be. Also the ChezScheme lift backend is not complete with regard to this addition.

Otherwise, it is ready to be reviewed.


b-studios:
@anfark the way I deal with call targets could be relevant for you. So maybe you can look at `CallTarget` in source.Tree and symbols

PR #43, opened 2021-01-05
Add sprinkler-rain-grass model to probabilistic

mm0821:


b-studios:
@mm0821 Sorry for not merging right-away! Is this still up-to-date?

mm0821:
That's my fault. I marked it as draft, since I wanted to check whether I could simplify some comments (they are quite verbose in parts) or maybe even convert this into a case study. But I never found the time. It's quite old and I have to recheck whether it still works with the current version of Effekt.

mm0821:
This should work with the current Effekt version now. I guess the check here failed because the new syntax for block parameters is not yet supported on this branch (as it is quite a bit behind :smile:).

You can merge it, if you like. I don't know when I will have time to revise it and I fear will forget about it again...

mm0821:
Now that there are great case studies for probabilistic programming in Effekt, it is probably not necessary to turn this into a case study (I will likely forget about it anyway). Thus, you can just merge it, in case you want to keep it as a somewhat more elaborated example for testing, or otherwise just close the PR.

b-studios:
I think it is value to keep this. I rebased and updated the comment syntax.

PR #45, opened 2021-01-14
Feature/polymorphic effects

b-studios:
This PR adds the long missing feature of polymorphic effects. Since it is a more complex feature, I would be happy if people could try it out and (ideally) submit their examples as test cases, before we merge into master.


shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -341,16 +342,27 @@ case class IgnorePattern() extends MatchPattern
...
 
+/**
+ * Something that can be resolved by namer
+ *
+ * It does not need to be a symbol
+ */
+sealed trait Resolvable extends Tree {
+  type resolved
+  def resolve(implicit C: Context): resolved
+}
```
b-studios, line 353: This one is new to allow easily obtaining the resolved version of the tree (mostly used for types and effects, which are not symbols.)

shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -341,16 +342,27 @@ case class IgnorePattern() extends MatchPattern
...
 /**
  * Types and Effects
  *
  * TODO generalize to blocks that can take blocks
  */
-sealed trait Type extends Tree {
-  type symbol <: symbols.Type
+sealed trait Type extends Tree with Resolvable {
+  type resolved <: symbols.Type
+  def resolve(implicit C: Context) = C.resolvedType(this)
```
b-studios, line 362: Not all types are symbols, so I am happy to rectify this here (@phischu)

examples/pos/parametrized.effekt
```diff
@@ -0,0 +1,39 @@
+effect State[T] {
+  def get(): T
+  def put[S](n: T, x: S): S
+}
```
b-studios, line 4: Here you can see the new feature. It is now possible that effects are parametrized by type parameters.

examples/neg/lambdas/polymorphicescape.effekt
```diff
@@ -0,0 +1,40 @@
...
+    def put(n, x) = {
+      val h = x;
+      // resume(get()) // ERROR
+      resume(h)
+    }
+  } with State[String] {
+    def get() = resume("")
+    def put(n, x) = {
+      val h = x;
+      put(fun() { resume(h) }, 42) // ERROR
```
b-studios, line 36: here the continuation needs to be run in region `main` and this info leaks into the effect type of main by using `put`.

b-studios:
I will just go ahead and fix bugs in production ;)

PR #52, opened 2021-05-23
Wip/dynamic regions blocks

### LINKED ISSUE #50: Undetected escaping capability when using first class functions
There seems to be a problem with undetected capabilities escaping their delimiters when using first class functions closing over them, if the handler for the corresponding effect is defined in a separate function. In the following example the escaping capability for `Effect2` is detected ([fcfCaughtErr.txt](https://github.com/effekt-lang/effekt/files/6522093/fcfCaughtErr.txt)):
```Scala
effect Effect1(): Unit
effect Effect2(): Unit


def escape { prog: () => Unit / { Effect1, Effect2 } }: Unit = {
  var k: (Unit) => Unit / { escape } = fun (x: Unit) { () }
  try {
    try {
      prog()
    } with Effect1 { () =>
      k = fun (x: Unit) { resume(x) };
      resume(())
    }
  } with Effect2 { () => resume(()) }
  k(())
}

def main() = {
  escape { do Effect1(); do Effect2() }
}
```
However, if the handler for `Effect2` is defined in the function `handle2` as is done here
```Scala
effect Effect1(): Unit
effect Effect2(): Unit


def handle2 { prog: () => Unit / { Effect2 } }: Unit = {
  try { prog() }
  with Effect2 { () => resume(()) }
}

def escape { prog: () => Unit / { Effect1, Effect2 } }: Unit = {
  var k: (Unit) => Unit / { escape } = fun (x: Unit) { () }
  handle2 {
    try {
      prog()
    } with Effect1 { () =>
      k = fun (x: Unit) { resume(x) };
      resume(())
    }
  }
  k(())
}

def main() = {
  escape { do Effect1(); do Effect2() }
}
```
this is not detected and results in a runtime error ([fcfUncaughtErr.txt](https://github.com/effekt-lang/effekt/files/6522097/fcfUncaughtErr.txt)).

b-studios:
Fixes #50.

### DIFF
--- a/examples/neg/issue50.check
+++ b/examples/neg/issue50.check
@@ -1,0 +1,9 @@
+[info] examples/neg/issue50.effekt:16:11: Not allowed: Anonymous function
+      k = fun (x: Unit) { resume(x) };
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+[error] examples/neg/issue50.effekt:16:11: Region mismatch: expected {escape} but got {Anonymous function}
+      k = fun (x: Unit) { resume(x) };
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+[info] examples/neg/issue50.effekt:16:27: Function resume is called, which closes over 'Anonymous function'
+      k = fun (x: Unit) { resume(x) };
+                          ^^^^^^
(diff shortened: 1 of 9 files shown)


mm0821 (commented):
As far as I can judge, this fixes the problem, at least what I tested works. However, I'm not really familiar with regions and the implementation, so I don't know whether someone else should do a review?

PR #59, opened 2022-04-10
Refactor/embed kiama

b-studios:
In this PR I am backporting changes from branch `system-c-steps` to Effekt.

In particular

- [x] integrate kiama into code base
- [x] move effekt into separate subproject

All commits on branch https://github.com/effekt-lang/effekt/compare/system-c-steps that are **newer* than 4a08dc6aa61 (not including it) are ported in this PR. All older commits are yet to be ported...

### TODO for the future
- At some point, we also should backport 088d4e8810bf15d69c48db2b0a41ce270021d108
  However, this is more effort since we need to also consider statically reachable value definitions.
  This refactoring should go hand-in-hand with a module system.
- We should reapply 842369694821cb089a6095511de59e1f736bfca8 and rename Member to Select. However, we still have record field access which is also called Select.
- Commits f4ac8290d9da6bc  and bff36b183 delete some helpers to traverse trees. Maybe we really don't need them anymore.
- Large renaming: fd2c88be3d11baa
- and many more.... commits after fd2c that I haven't yet looked at...

(**Do not merge**)


b-studios:
@phischu this PR might be of interest to you, since you are working on the backend. I moved quite a lot of files to clean up things. This might cause some breakage for you.

PR #61, opened 2022-04-14
Start using effects in the compiler itself

b-studios:
WIP


b-studios:
Next steps:

- [ ] improve loops and conditionals if the involved statements are pure.
- [ ] monomorphize functions (with a region subset of `{ io }`) to generate two versions `io` and `control`. At the callsite, check that all arguments have a region which is a subset of `{ io }` and then choose the `io` version.
- [ ] mark state capabilities (for local state) as tail resumptive and avoid introducing binds. Afterall, they are implemented in the runtime as side-effecting:
  ```
  function Cell(init) {
    var _value = init;
    return {
      "op$get": function() {
        return $effekt.pure(_value) // TAIL!
      },
      "op$put": function(v) {
        _value = v;
        return $effekt.pure($effekt.unit) // TAIL!
      },
      backup: function() {
        var _backup = _value
        var cell = this;
        return () => { _value = _backup; return cell }
      }
    }
  }
  ```

b-studios:
I rebased after merging #79

PR #75, opened 2022-04-25
Refactor/portsystemc

b-studios:
In this PR I am migrating some System C changes back to Effekt.

- [x] new structure of names (#76)
- [ ] mutual blocks
- [x] move lift
- [x] substitution
- [x] type parameters on blocks

**Not ready yet, do not merge**

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Intelligence.scala
+++ b/effekt/shared/src/main/scala/effekt/Intelligence.scala
@@ -72,11 +72,8 @@ trait Intelligence {
   } yield decl
 
   def getDefinitionOf(s: Symbol)(implicit C: Context): Option[Tree] = s match {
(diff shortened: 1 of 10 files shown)


b-studios:
I backported mutual blocks from System C (delimited by `val` or `var` or expressions).
This makes it easier in namer and typer to identify definitions that can be mutually recursive.

Now `val` and `var ` are no "definitions" anymore. 

@phischu There are two decisions we have to make:

1. What should we do in the REPL? Previously users could enter `val x = 4` which would add a toplevel binding (here the binding needed to be pure).
2. What should we do with toplevel constant definitions? There are many tests that now fail (see above) that rely on this feature.

We could add a separate syntactic construct that just "looks" like `ValDef` but doesn't perform sequencing. But this leads to confusion and duplication. WDYT?

PR #79, opened 2022-05-02
Refactor Compiler Pipeline

b-studios:
With the long-term goal of improving testability and reducing side channeling in the compiler, this PR makes intermediate results of phases explicit as instance of `PhaseResult`. It also adds an explicit interface to generators (now "backends").


b-studios:
@phischu I am merging this in order to build some other refactorings on top of it. If you have a counter proposal, please let me know; I would be happy to consider reverting it (partially).

PR #81, opened 2022-05-10
Drop symbols.Effectful

b-studios:
In order to move forward with the typer, as a small intermediate step, in this PR I am dropping `symbols.Effectful` and replace it with individual `ValueType` and `Effects` everywhere.

This prevents constructing and deconstructing instances of `Effectful` all the time using `/` syntax. In `Typer`, most methods now return `TyperResult[T]`, which most of the time is `TyperResult[ValueType]`. It is prepared to add additional fields (like regions, or fresh symbols to be bound) in the future.


b-studios:
This also has the advantage that in many places, we do not use the somewhat cryptic syntax of `/`...

PR #84, opened 2022-05-17
LLVM 20. KW 2022

jfrech:


shared/src/main/scala/effekt/context/Context.scala
```diff
@@ -50,6 +52,10 @@ abstract class Context(val positions: Positions)
   // bring the context itself in scope
   implicit val context: Context = this
 
+  val DEFAULT_LLVM_VERSION: String = "12"
+  var LLVM_VERSION: String = DEFAULT_LLVM_VERSION
```
b-studios, line 56: You can add these options to `Config` and default them to "12".

tests-llvm/ancient/LLVMTests.scala
```diff
@@ -0,0 +1 @@
+../jvm/src/test/scala/effekt/LLVMTests.scala
```
b-studios, line 1: What is this file?

jfrech: A symlink to the relevant testing code.

shared/src/main/scala/effekt/context/Context.scala
jfrech: I hard-coded `"12"` in 0419025ca996340cf6e7370b6c4e43913f269edc for now.

jfrech:
@b-studios @phischu Should we now delete the `singlestack` branch?

tests-llvm/ancient/LLVMTests.scala
b-studios: Usually, you would run `sbt testOnly effekt.LLVMTests` to only run those tests.

b-studios:
> Should we now delete the singlestack branch?

If it only consisted of 50c234c and aad1553, then sure.

tests-llvm/ancient/LLVMTests.scala
jfrech: That is what we test in https://github.com/effekt-lang/effekt/blob/5d4058b8be40296016c8d635cf1157c041aef8fd/.github/workflows/ci.yml#L33; this symlink was a development relic. Removed in 5d4058b.

PR #86, opened 2022-05-20
smoothing `llvm` to make it more mergeable with `master`

jfrech:
Since `llvm` and `master` have diverged this far, some preprocessing on `llvm` is necessary which aims to be cognizant of what happened on `master`. This should pave the way for an eventual `llvm => master` merge.


jfrech:
Abandoning `master`-mergeability in favor of features for now.

PR #109, opened 2022-08-26
stdlib

jfrech:


b-studios:
Is the idea of this PR to introduce a standardlibrary across different backends?

jfrech:
@b-studios Yes, but I could not get the `master`'s JS backend to find it.

b-studios:
The backends are all so different that we decided not to share anything between them. They basically describe different languages where (most of) the details are "implementation defined". Henceforth all attempts to unify the stdlib across backends have been abandoned.

It is even not so simple with programs that can be entirely written in Effekt. Maybe numerics work differently on different backends? Maybe some features are not yet implemented and thus prevent the stdlib from being typechecked / compiled.

I would be very happy if we could have a better approach to our (more or less non-existing) stdlib. But a solution somehow needs to address the above concerns.

b-studios:
> Yes, but I could not get the master's JS backend to find it.

That is strange, adding the path of the stdlib to `--includes` should "just" work.

jfrech:
Ref. https://github.com/effekt-lang/effekt/pull/138

PR #114, opened 2022-08-29
Move kiama to submodule

b-studios:


jfrech (changes requested):
Since kiama will probably not be updated frequently, you could pin it: `branch = 41077399e2630da8bebc5dc3d3d129604dd7a415`.

b-studios:
Thanks for the suggestion, I purposefully left it at master so that I can work on it the submodule's directory :)

jfrech (approved):

PR #118, opened 2022-09-01
Require type annotations on value parameters

jiribenes:
As of right now, the implementation presumes that every value parameter has a specified type annotation.
However, the parser does _not_ force the user to specify a type annotation for a parameter provided that there's a block parameter following it.
Altogether, this results in an unseemly internal error:

Screenshot_20220901_173003

The solution entails:
* _always_ requiring type annotations on value parameters, even if they're followed by block parameters
* reporting a more specific error when an annotation is expected, but not provided by the user


b-studios (approved):
Looks great, thanks for spotting and fixing.

PR #119, opened 2022-09-02
JIT: Use small register instructions instead of one big substitution instruction

marzipankaiser:
Instead of one SUBST instruction, we could use a combination of e.g. MOV, SWAP and DROP instructions in the generated code.

Since the substitution is known statically, we can then generate optimal sequences of those instructions and don't need to create a new array for "double buffering".

This is a proof-of concept implementation, corresponding to https://github.com/se-tuebingen/jitting-effects/tree/feature/small-step-substitution.


b-studios:
What is the intended semantics of `MOVE`? Will it consume the source? If yes, we also might need a `COPY` / `DUP`.

marzipankaiser:
`MOV` actually is more of a (shallow) `COPY` already, much like in some assembly languages. I.e., it reads from the `from` register (leaving it's contents unchanged) and then writes it contents to the `to` register.

If we want to clear the source register, we would have to `DROP` it.

We could easily rename `MOV` to `COPY`.

Issue #121, opened 2022-09-02
JIT: How to deploy the JIT binary


marzipankaiser:
Currently, the binary for the JIT is a blob in the git repository at https://github.com/effekt-lang/effekt/blob/jit/bin/x86_64/rpyeffect-jit.

We should find a better solution for this, that makes updates easier and puts a smaller load on git.

marzipankaiser:
It would be good if we had a solution to this before merging #119 and #122, which both would require an updated version of this binary.

b-studios:
One possible solution would be to store it as a release artifact in the JIT repo (potentially built by CI) and then download it using sbt. That shouldn't be too hard to setup, but would require the other repo to become public, which we should discuss first.

b-studios:
@marzipankaiser can we close this issue or is there something left?

PR #128, opened 2022-09-06
JIT backend: Add sbt target to download jit binary

marzipankaiser:
Currently, this has to be done manually and requires either a GH_TOKEN
environment variable or running `gh auth login` first.

This is a possible (temporary) solution to #121.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,11 @@ jobs:
       with:
         node-version: '10.x'
 
+    - name: Download JIT binary
+      env:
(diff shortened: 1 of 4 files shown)


marzipankaiser:
This needs a GitHub secret named `JITTING_EFFECTS_GH_TOKEN` containing a GitHub token with `repo` access to http://github.com/se-tuebingen/jitting-effects.
Sadly, this currently only seems to work with a personal access token. One possibility would be to create a separate GitHub account for this (or, just live with the full access to all repositories from the token).

marzipankaiser:
Maybe something like [this](https://github.com/marketplace/actions/use-app-token) could also be used.

PR #131, opened 2022-09-07
Recognize when def is used in the place of val

### LINKED ISSUE #130: Unhelpful error message when defining a value
I found an unhelpful error message when I tried to create a variable/value whilst not remembering the grammar:

image

jiribenes:
Resolves #130 by distinguishing a case where the user wrote `def foo = <value>` when they meant `val foo = <value>`.

Internally, it adds a new `UnboxParentDef` annotation which is applied to synthezised `Unbox` nodes
that have a `Def` as a parent.

image

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Typer.scala
+++ b/effekt/shared/src/main/scala/effekt/Typer.scala
@@ -408,7 +408,7 @@ object Typer extends Phase[NameResolved, Typechecked] {
                 // Since this `unbox` was synthesized by the compiler from `def foo = E`,
                 // it's possible that the user simply doesn't know that they should have used the `val` keyword to specify a value
                 // instead of using `def`; see [issue #130](https://github.com/effekt-lang/effekt/issues/130) for more details
-                Context.abort(pretty"Unbox requires a boxed type, but got $vtpe. Maybe try `val` if you're defining a value")
+                Context.abort(pretty"Expected the right-hand side of a `def` binding to be a block, but got a value of type $vtpe.\nMaybe try `val` if you're defining a value.")
               case _ =>
                 Context.abort(pretty"Unbox requires a boxed type, but got $vtpe.")
             }
--- a/effekt/shared/src/main/scala/effekt/Typer.scala
+++ b/effekt/shared/src/main/scala/effekt/Typer.scala
@@ -403,7 +403,15 @@ object Typer extends Phase[NameResolved, Typechecked] {
             usingCapture(capt2)
             Result(btpe, eff1)
           case _ =>
-            Context.abort(pretty"Unbox requires a boxed type, but got $vtpe")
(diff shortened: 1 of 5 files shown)


effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -403,7 +403,15 @@ object Typer extends Phase[NameResolved, Typechecked] {
             usingCapture(capt2)
             Result(btpe, eff1)
           case _ =>
-            Context.abort(pretty"Unbox requires a boxed type, but got $vtpe")
+            Context.annotationOption(Annotations.UnboxParentDef, u) match {
+              case Some(source.DefDef(id, annot, block)) =>
+                // Since this `unbox` was synthesized by the compiler from `def foo = E`,
+                // it's possible that the user simply doesn't know that they should have used the `val` keyword to specify a value
+                // instead of using `def`; see [issue #130](https://github.com/effekt-lang/effekt/issues/130) for more details
+                Context.abort(pretty"Unbox requires a boxed type, but got $vtpe. Maybe try `val` if you're defining a value")
```
jiribenes, line 411: I'm open to bikeshedding regarding the actual contents of the error message. :)

effekt/shared/src/main/scala/effekt/typer/PreTyper.scala
```diff
@@ -145,7 +159,7 @@ class BoxUnboxInference {
       VarDef(id, annot, region, rewrite(binding))
 
     case DefDef(id, annot, binding) =>
-      DefDef(id, annot, rewriteAsBlock(binding))
+      DefDef(id, annot, rewriteAsBlock(binding, Some(t)))
```
b-studios, line 162: Maybe you can run `rewriteAsBlock` first, then check whether an unbox was inserted, instead of passing the context parameter? The added parameter adds a bit confusion to the signature ("why is it necessary for the transformation to know about the `parentDef`?")

Not sure, whether analyzing and annotating at the point of `DefDef` is much clearer, though. What do you think?

effekt/shared/src/main/scala/effekt/Typer.scala
b-studios, line 411: Maybe we shouldn't talk about `unbox` anymore in this case, since it has been synthesized. What about:

"The right-hand side of a `def` binding should be a block, but got a value of type $vtpe.\nMaybe try `val` if you're defining a value."

b-studios (commented):
Thanks, that looks really good. I left a few comments.

effekt/shared/src/main/scala/effekt/Typer.scala
jiribenes: I used your suggestion and specified that the right-hand side of `def` is _expected_ to be a block.

PR #134, opened 2022-09-08
Jit

b-studios:
This PR is only for discussion and keeping track of the changes in JIT.


effekt/shared/src/main/scala/effekt/jit/PrettyPrinter.scala
```diff
@@ -54,15 +57,13 @@ object PrettyPrinter extends ParenPrettyPrinter {
...
   }
 
   def toDoc(frameDescriptor: FrameDescriptor): Doc = {
-    jsonObjectSmall(ListMap( // TODO: calculate befor pretty-printing
-      "regs_int" -> frameDescriptor.locals.applyOrElse(RegisterType.Integer, x => 0).toString,
-      "regs_cont" -> frameDescriptor.locals.applyOrElse(RegisterType.Continuation, x => 0).toString,
-      "regs_adt" -> frameDescriptor.locals.applyOrElse(RegisterType.Datatype, x => 0).toString
-    ))
+    jsonObjectSmall(RegisterType.valuesNonErased.map(ty => ("regs_%s".format(ty.name)) ->
+      (frameDescriptor.locals.getOrElse(ty, 0).toString: Doc)
```
b-studios, line 65: Should this really be `getOrElse` 0? Or should it rather be apply and crash if not present?

b-studios: I retract. Sorry. This is for each register type not a single register.

effekt/shared/src/main/scala/effekt/jit/Transformer.scala
```diff
@@ -89,7 +89,7 @@ object Transformer {
           emitInlined(block)
         }
         case Type.Continuation() => {
-          ???
+          sys error "Fatal error: Trying to match on continuation"
```
b-studios, line 92: Can this case ever occur?

marzipankaiser: It shouldn't be possible.

marzipankaiser:
@marzipankaiser Make `install` work without jit (warning or similar)

PR #137, opened 2022-09-09
JIT: Find the JIT binary

marzipankaiser:
Search for the JIT binary in the following places:

1. specified as part of the settings arg `--jit-binary`
2. specified in environment variable `EFFEKT_JIT_BIN`
3. in `./bin/${platform}/rpyeffect-jit` relative to PWD
4. in `./bin/${platform}/rpyeffect-jit` relative to the jar


marzipankaiser:
The order is analogous to the one for finding the standard library.

marzipankaiser:
Maybe we should also look at the `PATH`, since it is an executable after all?

PR #141, opened 2022-09-16
LLVM `Double`s

jfrech:
Adding `Double`s and corresponding operators to the LLVM backend.


libraries/llvm/effekt.effekt
```diff
@@ -1,38 +1,57 @@
...
 extern pure def infixAdd(x: Int, y: Int): Int =
-  "%z = add %Int %x, %y ret %Int %z"
+  "%z = call %Int @c_add_Int_Int(%Int %x, %Int %y)   ret %Int %z"
+
+extern pure def infixSub(x: Int, y: Int): Int =
+  "%z = call %Int @c_sub_Int_Int(%Int %x, %Int %y)   ret %Int %z"
 
 extern pure def infixMul(x: Int, y: Int): Int =
-  "%z = mul %Int %x, %y ret %Int %z"
+  "%z = call %Int @c_mul_Int_Int(%Int %x, %Int %y)   ret %Int %z"
```
b-studios, line 30: Does this have any implications on performance?

b-studios (approved):
Looks good to me! I just don't understand why we need `c_mul_int_int`, when instruction `mul` worked?

libraries/llvm/primitives.c
```diff
@@ -1,21 +1,62 @@
...
 
+struct Pos c_println_Double(const Double x) {
+    printf("%g\n", x);
+    return Unit;
+}
+
+
+// arithmetic
+
+#define C_OP(T, OP_NAME, OP) \
```
phischu, line 40: Nice idea of sharing code, but please hard-code the operations in `llvm/effekt.effekt` instead. This will lead to less code and less indirection.

libraries/llvm/primitives.c
```diff
@@ -1,21 +1,62 @@
...
+C_OP(Int, div, /)
+C_OP(Int, mod, %)
+
+// floating-point arithmetic
+C_OP(Double, add, +)
+C_OP(Double, sub, -)
+C_OP(Double, mul, *)
+C_OP(Double, div, /)
+// NOTE: requires linking against `-lm` and `#include <math.h>`
+Double c_mod_Double_Double(Double x, Double y) { return fmod(x, y); }
```
phischu, line 57: I didn't even know this exists. Remove it and the dependency on `libm`.

effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala
```diff
@@ -66,8 +66,12 @@ ${indentedLines(instructions.map(show).mkString("\n"))}
     case BitCast(result, operand, tpe) =>
       s"${localName(result)} = bitcast ${show(operand)} to ${show(tpe)}"
 
+    // TODO: Require `operand0` to be of integer type. It is currently hackily typed dependent on its right operand's type which ought to be a constant.
     case Add(result, operand0, ConstantInt(n)) =>
       s"${localName(result)} = add ${show(operand0)}, $n"
+    // TODO: Require `operand0` to be of floating-point type. It is currently hackily typed dependent on its right operand's type which ought to be a constant.
+    case Add(result, operand0, ConstantDouble(x)) =>
```
phischu, line 73: Add a new instruction `FAdd` instead of reusing `Add` for floating point.

phischu (changes requested):

libraries/llvm/effekt.effekt
jfrech: It very well may, since the linker only has a vague understanding on inlineability; I reverted in https://github.com/effekt-lang/effekt/pull/141/commits/6f1ed0204e7086de8607d66b046108ac2e024340, though.

effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala
jfrech: I surmise https://github.com/effekt-lang/effekt/pull/141/commits/17c4807a0e308f4ecd31fcfe7bfed99d2457c195 is sufficient.

phischu (approved):

PR #148, opened 2022-09-23
Parse return type annotation of operations

jiribenes:
### Motivation

When a user mistakenly specifies a type annotation of an operation handler, they get an unfriendly error message from the parser at the wrong place:

Screenshot 2022-09-23 at 12 45 34

This also happens when specifying a return type of a object operation:

Screenshot 2022-09-23 at 15 09 58

### Solution

* Allow the `OpHandler` AST to include an optional return type.
* Modify the parser to optionally parse the return type of an operation.
* In the `Typer` phase -- if a return type is specified for an operation, throw an error.
* Add negative test for annotated object operation and another negative test for annotated handler operation

Screenshot 2022-09-23 at 12 44 02

#### Alternative solution

We could also just somehow forbid a return type annotation in the parser, but I haven't been able to do it in a nice way. However, I believe that we want to write down this information from the user even if we decide to ignore it.

### Discussion

Is `Typer` the best phase to report this error? We could also report it earlier while in a `Namer` or `PreTyper` phase, but it seems like this is a typing-related problem that we might want to improve in the future.

The solution is not optimal. Ideally, we should take the user-specified return type annotation and _somehow_ take it into consideration for better error messages.
Effect does in fact allow the annotation of the _arguments_ of the handler and even checks them:

Screenshot 2022-09-23 at 12 53 33

Unfortunately, there is an ambiguity about what the return type of an operation handler actually _is_ -- is it the resulting type of the body, or the type of the argument we put into the `resume` continuation?

We also don't support annotations on _object_ operations where we should just check the type and compared to the previously defined type in the interface. There are some subtleties to get right -- see comments below for more details.


effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -311,7 +311,10 @@ object Typer extends Phase[NameResolved, Typechecked] {
         Context.error("Duplicate definitions of operations")
 
       clauses foreach Context.withFocus {
-        case d @ source.OpClause(op, tparams, params, body, resume) =>
+        case _ @ source.OpClause(op, tparams, params, Some(ret), body, resumeId) =>
+          // TODO: Can we somehow use the return type provided by the user?
+          Context.abort(pretty"Unexpected type annotation on operation ${op}.")
```
b-studios, line 316: Does "use" here mean using it for error messages, or for type checking?

b-studios: To be precise, it is ok to have an annotated return type on object definitions. It is just ambiguous for handler definitions. In this method, you can see whether something is an object or handler definition by inspecting `continuationDetails`:

`None` -> object definition
`Some(ctx)` -> handler definition

We could report the error only in the latter case since only there the ambiguity arises.

b-studios:
Maybe you could add a neg-test to show the error message? 

After adding the effekt file to `examples/neg` you can simply run 

```
project effektJVM
test:console
scala> effekt.TestUtils.generateCheckFiles()
```
to generate the check files. Passing `true` to `generateCheckFiles` will regenerate ALL check files.

effekt/shared/src/main/scala/effekt/Typer.scala
jiribenes: > We could report the error only in the latter case since only there the ambiguity arises.

Done in a18277b8b37d176e91ed4a268fabe921441e8277.

However, when done naively, this now introduces a problem since we don't check the user-supplied type.
Screenshot 2022-09-23 at 15 01 57

Ideally, I'd want to use something like `matchDeclared`:
https://github.com/effekt-lang/effekt/blob/b9b56fa12a0d4b2552818e9180182e38367aec43/effekt/shared/src/main/scala/effekt/Typer.scala#L1124

but right now it's overly specialised for `source.Param`s.

Moreover, I'm not actually sure that `matchDeclared` is the right thing to use here -- in a system without subtyping, this would just be an equality constraint:
https://github.com/effekt-lang/effekt/blob/a18277b8b37d176e91ed4a268fabe921441e8277/effekt/shared/src/main/scala/effekt/Typer.scala#L357-L369

jiribenes:
I've updated the PR text and added a few details about what we (plan to?) do with object operations.

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -358,11 +355,26 @@ object Typer extends Phase[NameResolved, Typechecked] {
           }
 
           val Result(_, effs) = continuationDetails match {
-            // no answer type, just check body
-            case None => body checkAgainst tpe
+            case None => retAnnotation match
+
+              case Some(ret) => // TODO: what about the effects?
```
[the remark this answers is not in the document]
b-studios: That's *very* subtle...

b-studios:
Whatever we do, it should behave similarly to how we deal with annotated effects on functions:

https://github.com/effekt-lang/effekt/blob/5fa89a2cfd03f4a9f899a8f242df33ad51e4e6b5/effekt/shared/src/main/scala/effekt/Typer.scala#L594-L614

jiribenes:
I think it might be wise to change the scope of this PR.

Let's first merge this more-or-less as-is, specifically with parsing return type annotations on _all_ operations and reporting them as errors.
Then afterwards, let us add proper support for (type checking) return type annotations on object operations.

effekt/shared/src/main/scala/effekt/Typer.scala
jiribenes: The important parts of this conversation are repeated and clarified below so I'm resolving this thread.

PR #132, opened 2022-09-08
Sketch a typed version of core

b-studios:
I started working on a typed version of our core representation. It would be great to gather feedback on it.

In particular, I am interested in 

- learning whether it impedes or helps other backends
- It impedes or helps potential optimization passes
- how to represent includes in core
- how to model top-level / module-level declarations
- how to represent mutually recursive bindings (blocks / data defs / interface defs)

@marzipankaiser @jiribenes @phischu @jfrech  Please chip in, if you have any comments.

PR #154, opened 2022-09-26
Shallow types in machine so we can compile recursive data types.

phischu:


effekt/shared/src/main/scala/effekt/machine/Tree.scala
```diff
@@ -171,11 +169,11 @@ object builtins {
   val Apply: Tag = 0
 
   val Unit: Tag = 0
-  val UnitType = Positive(List(List()))
+  val UnitType = Positive()
```
marzipankaiser, line 172: This makes `UnitType` and `BooleanType` indistinguishable, doesn't it? (since `Positive` is a case class and thus compared structurally IIRC).

In the JIT backend, I erase Unit (and the primitives depend on that), so I kind of need that information.

effekt/shared/src/main/scala/effekt/machine/Tree.scala
```diff
@@ -171,11 +169,11 @@ object builtins {
...
 
   val True: Tag = 1
   val False: Tag = 0
-  val BooleanType = Positive(List(List(), List()))
+  val BooleanType = Positive()
```
marzipankaiser, line 176: This makes `UnitType` and `BooleanType` indistinguishable, doesn't it? (since `Positive` is a `case class` and thus compared structurally IIRC).

In the JIT backend, I erase `Unit` (and the primitives depend on that), so I kind of need that information.

marzipankaiser: The same holds for `BooleanType` and other data, so I can't implement `Bool` using int registers.

marzipankaiser (approved):
This would work for me. Seems OK for now.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -212,7 +212,7 @@ object Transformer {
     case lifted.BlockLit(params, body) =>
       // TODO deal with evidence
       val parameters = params.map(transform);
-      val variable = Variable(freshName("g"), Negative(List(parameters.map(_.tpe))))
+      val variable = Variable(freshName("g"), Negative("<function>"))
```
b-studios, line 215: Functions don't need to be structural?

b-studios (approved):

b-studios:
I will go ahead and merge, so that you can proceed. We can still redesign it when we merge #132

PR #156, opened 2022-09-27
MLton backend

b-studios:
This is a draft PR for discussion and tracking the changes on this branch :)

TODO
- [ ] Non-sequential function scope (and mutual recursion) (requires analysis to maximize let-generalization)
- [ ] Std lib things
- [x] mlton on ci
- [ ] Suppress warnings

Limitations #185


b-studios:
Hey, the `while` desugaring is now merged into master. You could benefit from it :)

effekt/shared/src/main/scala/effekt/generator/ml/ML.scala
```diff
@@ -93,47 +93,68 @@ object ML extends Backend {
...
-      ml.Type.Alias(name(tc))
-    case TypeConstructor.ExternType(_, _) =>
-      ml.Type.Builtin(name(tc))
+    case symbols.builtins.UnitSymbol       => ml.Type.Unit
+    case symbols.builtins.IntSymbol        => ml.Type.Integer
+    case symbols.builtins.DoubleSymbol     => ml.Type.Real
+    case symbols.builtins.BooleanSymbol    => ml.Type.Bool
+    case symbols.builtins.StringSymbol     => ml.Type.String
+    case TypeConstructor.DataType(_, _, _) => ml.Type.Data(name(tc))
+    case TypeConstructor.Record(_, _, _)   => ml.Type.Alias(name(tc))
```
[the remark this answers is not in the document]
JonathanStarup: Good, point, fixed

effekt/jvm/src/main/scala/effekt/EffektConfig.scala
```diff
@@ -47,7 +47,8 @@ class EffektConfig(args: Seq[String]) extends REPLConfig(args) {
 
   val llvmVersion: ScallopOption[String] = opt[String](
     "llvm-version",
-    descr = "the llvm version that should be used to compile the generated programs (only necessary if backend is llvm, defaults to 12)",
+    descr = "the llvm version that shou" +
```
b-studios, line 50: I guess this is by accident and should be reverted

b-studios:
@JonathanStarup Is there anything that you planned to do before we can merge this PR?

JonathanStarup:
Nothing except some cleaup, the changes to effekt.sh and the build should be reverted right?

b-studios:
> the changes to effekt.sh and the build should be reverted right?

I have to look into this, maybe for now. We still have the open ticket on how to support, unix, and macos at the same time with the same convenience.

examples/ml/raytracer.effekt
```diff
@@ -0,0 +1,275 @@
...
+    val pointAtTime = pos + dir.scale(bestDist);
+    val normal = sphereNormal(bestObj, pointAtTime);
+    Some(Intersection(bestDist, pointAtTime, normal, bestObj))
+  } else { None() }
+}
+
+
+def trace(pos: Point, dir: Vector, scene: Scene, depth: Int): ColorV = {
+
+  def surface(pos: Point, dir: Vector, scene: Scene, depth: Int, intersection: Intersection): ColorV = {
```
b-studios, line 142: Compared to the original raytracer example in `examples/pos` I had to manually reorder (and partially re-nest) the function definitions in order to be sorted topologically. @JonathanStarup Could you maybe open up a ticket so that we remember implementing this as a preprocessing step before emitting ML?

b-studios: I also rewrite the List stdlib to use tailrecursive functions instead of local mutable state since we do not know how fast the latter is in ML at the moment.

b-studios:
@JonathanStarup Can you tell me whether you can think of something that blocks the merge? I would like to click merge tomorrow, so that we can maintain the backend in all subsequent PRs :) It is not necessary that everything works; it is more important that all tests pass (which is the case) and no impediments for other backends or contributors are created.

PR #162, opened 2022-09-29
Machine: `interfaces` and multiple operations

marzipankaiser:
This adds a basic implementation for interfaces and new into Machine
and adds a example for usage.

This also supports multiple methods, but llvm currently does not (and JIT is buggy with the passed parameters).

@b-studios and I implemented the basis for this earlier today during our meeting.


marzipankaiser:
I also now added correctly ordering the methods/operations (in the order they are declared), and multiple operations for effects onto this PR.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -125,6 +125,20 @@ object Transformer {
         noteBlockParams(id, allParams)
         Def(Label(transform(id), allParams), transform(body), transform(rest))
 
+      case lifted.Def(id, tpe, block @ lifted.New(impl), rest) =>
+        // TODO freeParams?
+        // TODO deal with evidenve?
+        val interface = tpe.asInstanceOf[symbols.Interface]
+        val implTransformed = impl.clauses.sortBy[Int]({
+          case (operationName, _) =>
+            interface.ops.indexOf(operationName)
```
b-studios, line 134: While this works, in other places in the compiler we would iterate over `interface.ops` like
```
interface.ops map {
  case op => transform ... impl(op) ...
}
```
This has the advantage that we maintain the invariant that all operations are actually there and all others would be dropped in the worst case.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -125,6 +125,20 @@ object Transformer {
...
+        }).map({
+          case (_, lifted.BlockLit(params, body)) =>
+            // TODO we assume that there are no block params in methods
```
b-studios, line 137: This assumption may soon be invalidated.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
marzipankaiser: Changed that (although `impl` is not a map, so it's `impl.find( ... ).get` now).

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
marzipankaiser: Yes. Will we have to change it at https://github.com/effekt-lang/effekt/pull/162/files#diff-2bc35700065c09eb82c800736c2625a2516ddb669bb20acbfce3d9af8dbca683R329, too?

marzipankaiser:
(rebased to current `master`)

phischu (approved):

PR #163, opened 2022-09-30
Machine: Fix arity mismatch with effect handlers due to evidence

marzipankaiser:
For this code (https://github.com/effekt-lang/effekt/blob/master/examples/llvm/capabilities.effekt):
```
effect Emit(n: Int) : Unit


def loop(n: Int): Unit / Emit = {
  if (n < 0) {
    ()
  } else {
    do Emit(n);
    loop(n - 1)
  }
}

def main() = {
  try {
    loop(10)
  } with Emit { (n: Int) =>
    println(n); resume(())
  }
}
```
the following lifted core is generated:
```
module capabilities

import effekt

record Emit(Emit)

def loop(ev581, ev582, n, Emit$capability) =
  if (infixLt(n, 0)) {
    ()
  } else {
    Emit$capability.Emit(<ev582>, n);
    loop(<ev581>, <ev582>, infixSub(n, 1), Emit$capability)
  }

def main(ev583) =
  try { (ev584, Emit$capability) => 
    loop(<ev584, ev583>, <>, 10, Emit$capability)
  } with Emit {
    def Emit(n, resume) = 
      let _ = println(n);
      resume(<>, ())
  }

()
```
This, with the current machine transformation, leads to an arity mismatch in both the resume and the call of the effect operation.
A hotfix for this part of the issue is to:
- Always ignore the first parameter to `resume` (which is evidence).
- Ignore the first parameter passed into a handler (which is "its" evidence).

This PR implements this Hotfix and un-ignores the tests fixed by this.

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/LLVMTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/LLVMTests.scala
@@ -13,13 +13,7 @@ class LLVMTests extends EffektTests {
 
   override lazy val ignored: List[File] = List(
     // computes the wrong results
-    examplesDir / "llvm" / "generator.effekt",
-    examplesDir / "llvm" / "capabilities.effekt",
     examplesDir / "llvm" / "nested.effekt",
-    examplesDir / "llvm" / "stored.effekt",
-
-    // crashes
-    examplesDir / "llvm" / "forking.effekt",
 
     // polymorphic effect operations not supported, yet
     examplesDir / "llvm" / "choice.effekt",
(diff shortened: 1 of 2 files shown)


marzipankaiser:
@b-studios Is there always exactly one leading evidence parameter for calls to effect operations and `resume`?

b-studios:
At the moment, yes. :)

b-studios (commented):
Looks good to me, one one small nitpick.

b-studios (approved):

PR #164, opened 2022-10-04
Machine: Handle evidence

marzipankaiser:
Handle evidence and use it to execute effect operations in the correct context.

Is not yet correctly implemented in the llvm backend (number of stacks to pop is ignored),
so widely untested.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
@@ -60,7 +60,7 @@ object Transformer {
         val transformedParams = params.map {
           case lifted.ValueParam(id, tpe) => Variable(id.name.name, transform(tpe))
           case lifted.BlockParam(id, tpe) => Context.abort("Foreign functions currently cannot take block arguments.")
-          case lifted.EvidenceParam(_) => ???
+          case lifted.EvidenceParam(id) => Variable(id.name.name, builtins.Evidence)
         }
         emitDeclaration(Extern(transform(name), transformedParams, transform(functionType.result), body))
         transformToplevel(rest, entryPoint)
--- a/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
@@ -187,7 +187,7 @@ object Transformer {
         val returnClause = Clause(List(variable), Return(List(variable)))
         val delimiter = Variable(freshName("returnClause"), Type.Stack())
 
-        LiteralInt(transform(ev), 1,
+        LiteralEvidence(transform(ev), builtins.There(builtins.Here),
           NewStack(delimiter, returnClause,
             PushStack(delimiter,
               New(transform(id), transform(handler),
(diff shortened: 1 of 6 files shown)


marzipankaiser:
For some reason, `fibonacci.effekt` works on my machine but fails in the CI.

marzipankaiser:
Also, while I wanted to use `infixAdd` for my first draft, resolving that Symbol in the `Context` sometimes failed, so I added `EviAdd`.

b-studios:
> For some reason, fibonacci.effekt works on my machine but fails in the CI.

Maybe stack size?

b-studios:
If we absolutely have to add `EviAdd` then I would call it something like `ComposeEvidence`.

marzipankaiser:
> If we absolutely have to add `EviAdd` then I would call it something like `ComposeEvidence`.

Ah, yes, that's a better name.

marzipankaiser:
https://github.com/effekt-lang/effekt/pull/164/commits/f85fb62a1d8a3d6657680bffab0cc3a98248403a uses `"infixAdd"` instead of `ComposeEvidence`. I'm not sure if that is preferable?

marzipankaiser:
@phischu tested it on his machine and it was indeed the stack limit being reached.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -201,7 +200,21 @@ object Transformer {
   def transform(arg: lifted.Argument)(using BlocksParamsContext, Context): Binding[Variable] = arg match {
     case expr: lifted.Expr => transform(expr)
     case block: lifted.Block => transform(block)
-    case lifted.Evidence(_) => transform(lifted.IntLit(0)) // TODO implement
+    case lifted.Evidence(scopes) => {
+      val addSymbol = Context.module.findPrelude.terms("infixAdd").find({
```
phischu, line 204: In my opinion, we should have a separate type of evidence in `machine.Type` and two runtime operations in `machine.Statement`: `LiteralEvidence(n: Int)` and `ComposeEvidence(name: Variable, arguments: Environment, rest: Statement)`.

b-studios: At first we didn't want to add evidence to reduce the number of cases. What are the advantages you hope from having evidence being built into machine?

marzipankaiser:
I have now semi-splitted out evidence. It is still synonymous int in most places (i.e. `machine.builtins.Evidence` is `machine.Type.Evidence` etc), but can be used as if it weren't (there is a type alias `Evidence`, and `builtins.Here` (=`0`) and `builtins.There(_)` (=`_+1`) work for both construction and patterns).

b-studios:
Why does `There` take another evidence? It should just alias the number `1` IIUC.

marzipankaiser:
I think both are reasonable. I understand `Here` and `There` as the constructors of the evidence type, so `There` just says "outside of the current region" and still needs specification/proof as to where exactly (like with e.g. `Data.List.Relation.Unary.Any` etc in Agda).

I am not strongly opposed to changing it to `There : Evidence = 1`, although I think it's nicer this way, as it allows to construct all possible evidence values without using `+` (on `Int`).

marzipankaiser:
After rethinking it, I changed `There` to be `1` now.

marzipankaiser:
(and rebased to current `master`)

phischu (approved):

Issue #16, opened 2020-04-15
Binary on Ubuntu 16.04


b-studios:
On Ubuntu 16.04 it has been reported that Effekt cannot be exeucted since
`java -jar` is treated as *one command* instead of a command and an argument.

---
The Effekt "binary" is created by prefixing:

```
#! /usr/bin/env java -jar
```
to the jar file itself. Installing it with npm makes the binary available (almost) platform independently. In particular on Windows, npm analyses the shebang and automatically generates scripts that correctly start Effekt by invoking `java -jar`.

We could change the shebang to
```
#! java -jar
```
which seems to work on MacOS X, Windows 10 and Ubuntu 16.04 -- however, now we cannot start Effekt as a subprocess from node. This is necessary to start the language server in the VSCode extension.

b-studios:
I could also reproduce the bug on a fresh 18.04 VM:

```
$ effekt
/usr/bin/env: 'java -jar': No such file or directory
```

b-studios:
As a workaround, one can of course run `java -jar /usr/local/bin/effekt`, but this is not satisfactory.

b-studios:
A more reliable workaround is the following:

Create a script `effekt.sh` with the following contents:

```
#!/bin/bash
java -jar /usr/bin/local/effekt $@
```
and make it executable. Put this script in your `PATH` and use `effekt.sh` instead of `effekt`. This also works for VSCode:

- Enter settings
- navigate to Effekt / path to executable
- enter `effekt.sh` (or the absolute path to the script) there

b-studios:
As a workaround I added the script to the distribution (87a4b0) and link it as `effekt.sh`. So linux users for now need to call `effekt.sh` instead of `effekt`. Also VSCode can now be configured to point to `effekt.sh` if `effekt` does not work.

b-studios:
This needs to be documented on the website.

waterlens:
So why not just pass `-S` to `env` to make it recognize the multiple arguments?

JonathanStarup:
> So why not just pass `-S` to `env` to make it recognize the multiple arguments?

This `-S` flag breaks on windows :(

PR #166, opened 2022-10-24
Avoid having two effekt binaries (deprecate effekt.sh)

b-studios:
Previously we had to have two commands:

- `effekt` which worked on MacOs
- `effekt.sh` which worked on Ubuntu systems

in #16 @waterlens proposes to use `-S` (which is short for [`--split-string`](https://man7.org/linux/man-pages/man1/env.1.html)).

I am not sure anymore, why I didn't do that in the first place. Maybe `-S` is not available on some implementations?

Could someone with Ubuntu please try whether this works? Simply run

```
sbt assembleBinary
```
and then there should be `bin/effekt` which you probably need to `chmod +x` and then execute. Can you also try executing it with some arguments (like `--help`)?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,3 +33,9 @@ jobs:
 
     - name: Assemble fully optimized js file
       run: sbt effektJS/fullOptJS
+
+    - name: Try installing effekt binary
+      run: sbt install
+
+    - name: Run effekt binary
+      run: effekt --help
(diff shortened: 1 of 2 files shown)


phischu (approved):
Just works on Ubuntu 20.04

b-studios:
Looks like this has introduced a couple of regressions... We just got reports that it does break on Ubuntu 22.04.

b-studios:
Maybe someone is curious to look into https://github.com/sbt/sbt-native-packager ?

PR #170, opened 2022-11-02
Feature/mutable_state

phischu:


b-studios:
As soon as this branch is in a somewhat stable state, I would be happy to merge it. You can of course continue to work on it, but this way it is easier to deal with upstream changes (then you don't need to rebase so many commits every time master changes).

serkm:
This is ready to merge now.

effekt/shared/src/main/scala/effekt/generator/chez/ChezScheme.scala
```diff
@@ -132,7 +132,7 @@ trait ChezScheme {
       }
       chez.Handle(handlers, toChez(body))
 
-    case Region(body) => chez.Builtin("with-region", toChez(body))
+    case Region(body, _) => chez.Builtin("with-region", toChez(body))
```
b-studios, line 135: Ah I see that we also need the answer type here, yes!

b-studios (approved):

b-studios:
Thanks, good job!

PR #172, opened 2022-11-04
Move from Scalatest to MUnit

### LINKED ISSUE #168: `check` files are sensitive to CRLF / LF lineendings
I edited my `.check` files on windows so they had CRLF line endings which caused the test to declare failure.

b-studios:
Moving to munit has a few benefits:

- the library itself is a lot simpler (and might be easier to get into for students)
- the reported diffs for strings are much better (this is especially helpful, since we do a lot of golden-testing)
- it comes with support for ANSI/windows normalized string comparison, which hopefully fixes #168 
- it allows to filter easily based on the file name: 
image

There are also a few downsides:
- tests cannot be grouped nicely anymore, like they was with scala tests `describe`. Instead I am using many different test suites, now.
- do to the missing grouping, test output is not hierarchical anymore and might be harder to read.

For example, here is a run that shows munit:
image

previously, the tests were grouped by folders.

I am happy to accept these downsides. WDYT?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -45,8 +45,8 @@ lazy val lspDependencies = Seq(
 )
 
 lazy val testingDependencies = Seq(
-  "org.scala-sbt" %% "io" % "1.6.0" % "test",
-  "org.scalatest" %% "scalatest" % "3.2.9" % "test"
+  "org.scala-sbt" %% "io" % "1.6.0" % Test,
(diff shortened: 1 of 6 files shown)


JonathanStarup (approved):

jiribenes (approved):
Looks good to me! :shipit:

PR #183, opened 2022-11-08
Refactor/syntax into enums

b-studios:
In this PR I refactor the source syntax into enums. This makes grouping a lot easier. Also factoring `Named` and `Resolvable` to separate objects separates concerns, which could make the source tree easier to understand for newcomers.

I plan to move those two objects to `Namer`, later.

Advantages:
- Potentially easier to read.
- `Named` serves as documentation (https://github.com/effekt-lang/effekt/pull/183/files#diff-90c294a2587a165cc230975b6fc6787f865eee7a9bc4987cd6cbfffb8a776434R443-R492)

Disadvantages:
- Enum support in Intellij is broken: https://youtrack.jetbrains.com/issue/SCL-20658/Jump-to-definition-does-not-work-for-exported-constructors
- Widening of enums is noisy https://github.com/lampepfl/dotty/issues/16299
- Instead of a complicated feature (existential type members) `Named` uses another complicated (and pretty new) feature "match types".

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Namer.scala
+++ b/effekt/shared/src/main/scala/effekt/Namer.scala
@@ -540,11 +540,6 @@ object Namer extends Phase[Parsed, NameResolved] {
         Substitutions.types(tparams, targs).substitute(tpe)
       case other => Context.abort(pretty"Expected a value type, but got ${other}")
     }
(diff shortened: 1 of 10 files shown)


b-studios:
Sorry @marzipankaiser & @phischu & @JonathanStarup, this is probably yet another annoying-to-rebase PR.

PR #187, opened 2022-11-14
Feature/compile pattern matching

b-studios:
Over the weekend, I implemented a first draft of a match compiler. This should make life easier for all backend developers,
including LLVM (@phischu), JIT (@marzipankaiser), and MLTon (@JonathanStarup).

The match compiler is based on these notes by Jules Jacobs:

  https://julesjacobs.com/notes/patternmatching/patternmatching.pdf

Right now, the only implemented backend is JS (so LLVM and Chez are still missing for the tests to pass).


effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -139,7 +139,7 @@ object Transformer {
       case lifted.App(lifted.BlockVar(id), List(), args) =>
         // TODO deal with BlockLit
         id match {
-          case symbols.UserFunction(_, _, _, _, _, _, _) =>
+          case symbols.UserFunction(_, _, _, _, _, _, _)  | symbols.TmpBlock(_) =>
```
b-studios, line 142: The matching here on symbols here is a smell. What is it that you really need here?

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -179,18 +179,34 @@ object Transformer {
...
+            case (symbols.Constructor(_, _, _, tpe), _) :: _ => tpe match {
+              case symbols.DataType(_, _, constructors) => constructors
+              case symbols.Record(_, _, constructor) => List(constructor)
+              case symbols.ExternType(_, _) => ??? // TODO: can this happen?
+            }
+         case _ => ??? // TODO: can there be an empty set of clauses? How do we get the type then?
+        };
+        transform(scrutinee).run { value =>
+          Switch(value, constructors.map {
+                   constructor => clauses.find { _._1 == constructor } match {
```
b-studios, line 192: This should not be necessary anymore, no?

b-studios: https://llvm.org/docs/LangRef.html#id39

```
switch i32 %val, label %otherwise [ i32 0, label %onzero
                                    i32 1, label %onone
                                    i32 2, label %ontwo ]
```

b-studios: The machine Switch just also needs a default label:

Currently:
```scala
case Switch(scrutinee: Variable, clauses: List[Clause])
```

Proposed:
```scala
case Switch(scrutinee: Variable, clauses: List[(Int, Clause)], default: Clause)
```
Note, that each clause also has an integer value.

b-studios: BTW, unsurprisingly, nested matching now also works in LLVM:

image

PR #191, opened 2022-11-19
Refactor/deprecate local declarations

b-studios:


b-studios:
FYI, the traversal in core.Transformer is very fragile. Looks like at some point I made a mistake in the exercise of using a function as an accumulator to foldRight :)

PR #194, opened 2022-11-22
Drop intrinsic lists in core

b-studios:
Please do not merge.

Using the intrinsic list encoding on core is really annoying on the module level, since a special return value usally unit has to be made up to terminate the list. This is a technical debt that we carry around for quite some time now,

As a step forward, here we

- [X] introduce `core.Definition`
- [X] introduce `core.Scope` to locally use `core.Definition` 
- [x] reenable simple optimizations on Core
- [ ] add additional optimizations to simplify `Scope(Scope...))`
- [ ] introduce `core.Defintion.Mutual` to separate out mutually recursive block definitions 
- [x] refactor the toplevel module declarate to use `List[Definition]` instead of `Stmt`
- [x] Also refactor `lifted`.

Completely unrelated to this PR, [7ff2ce2](https://github.com/effekt-lang/effekt/pull/194/commits/7ff2ce2c07c0f3f27b5185ed37f99a8e06a0cbc2) also desugars while into a recursive function to further simplify core and lifted.


effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -41,14 +41,19 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
     ModuleDecl(path, mod.imports.map { _.path }, declarations, externals, optimized, exports)
   }
 
+  def addToScope(definition: Definition, body: Stmt): Stmt = body match {
+    case Scope(definitions, body) => Scope(definition :: definitions, body)
+    case other => Scope(List(definition), other)
+  }
```
b-studios, line 47: We use this helper function (for now) to not introduce too many `Scope`s in the translation.

effekt/shared/src/main/scala/effekt/generator/js/JavaScript.scala
```diff
@@ -256,28 +256,30 @@ trait JavaScript extends Backend {
...
+
+    case Definition.Def(id, tpe, block) =>
+      js.Const(nameDef(id), toJS(block))
+
+    case Definition.Let(Wildcard(_), tpe, binding) =>
+      js.ExprStmt(toJS(binding))
+
+    case Definition.Let(id, tpe, binding) =>
+      js.Const(nameDef(id), toJS(binding))
+  }
```
b-studios, line 272: The refactoring bares first fruit.

effekt/shared/src/main/scala/effekt/lifted/LiftInference.scala
```diff
@@ -115,16 +116,21 @@ object LiftInference extends Phase[CoreTransformed, CoreLifted] {
...
+      val env = pretransform(definitions)
+      val body = transform(rest)(using env, Context)
+
+      // TODO once we also port lifted to non-intrinsic lists, this can be simplified
+      definitions.foldRight(body) {
+        case (core.Definition.Def(id, tpe, block), rest) =>
+          Def(id, tpe, transform(block)(using env, Context), rest)
+        case (core.Definition.Let(id, tpe, binding), rest) =>
+          Let(id, tpe, transform(binding)(using env, Context), rest)
+      }
```
b-studios, line 129: Here we go back to the horrible intrinsic lists, to refactor one bit at a time. :)

b-studios:
@phischu @marzipankaiser there is probably a better way to implement the translation to Machine. For now all LLVM tests go through. Maybe you could revise it? I propose the following:

1. we merge this PR if it looks at all ok to you
2. we implement `Mutual` blocks in `core` and `lifted` (like sketched in core)
3. we revisit the computation of free variables, both in lifted and in Machine

WDYT?

phischu (approved):

PR #196, opened 2022-11-25
Whole program js

b-studios:
WIP.

Idea: also use whole program compilation in JS. We *do* need to support some form of separate compilation for the website (for performance reasons and to avoid calling `eval` over and over again -- which results in a quadratic explosion). 

However, I first want to make the JS backend conform with all the other backends and THEN think about how to support the website again. Maybe this results in a design that works better in the common case.

For example, it would be fine to not run optimizations on the code generated for the website (so optimizations can be whole-program).


b-studios:
Update: I am done homogenizing the different backends. This should make it slightly easier for Machine (LLVM, JIT) and mlton to generate code. More importantly. Aggregating all dependencies into a single file is now performed in ONE location. This is important since this way all backends can rely on this fact and do not have to deal with imports anymore. 

Even more importantly, we can perform whole program optimizations once for all backends (@fritzladwig).

PR #199, opened 2022-11-30
Polymorphism in machine-based backends (through boxing)

marzipankaiser:
*[EDIT: Old (outdated) version below)]*

This adds a phase `PolymorphismBoxing`, replacing primitive types `T` by `BoxedT` in type arguments and inserting coercions where necessary. `BoxedT` should be defined in the standard library for each type `T`.

#### Old version:
This is WIP to bring polymorphism to machine-based backends.
It currently works using an additional pass over lifted, auto-boxing and unboxing values.

It also is more of a proof-of-concept/prototype in it's current stage.

Basic polymorphism works, but not with block parameters (e.g. a function `Int => Int` as argument for a parameter of type `A => B`) - This will give a readable error message, however.


marzipankaiser:
Some remarks/things I would do differently:
1. Wait for #132 if possible. Since auto-boxing/-unboxing is basically implicit conversion, we need types all the time and in some places they're not easily available. (e.g. type of `BlockLit`s)
     - Using a BlockLit of type `Int => Int` as `A => B` needs wrapping and for this we need the type of the `BlockLit`.
2. Maybe do the transformation earlier (e.g. in Core?, so we do not need to consider Evidence parameters)
3. Parameters are somewhat inconsistent between type and term level: `FunctionType` has different kinds of parameters separated, `BlockLit` just has one List. this makes it harder to find the corresponding ones.

marzipankaiser:
Adding the new Phase crashes with alll backends tried, known bugs so far:
- for all machine backends: ...App(BlockLit(...)...) is not supported in machine. This was easiest to generate, maybe we could support this? (alternatively: inline it)
- for all lift backends: LiftInference does not work correctly (always passes here to the wrapped function instead of forwarding the evidence)

The generated core looks fine for me, though (on simple examples).

b-studios:
Wow cool, so machine-based backends get boxing "for free" if they implement (polymorphic) datatypes?

marzipankaiser:
It should also be possible to make specialized versions of this phase that box only some, or more types (or use different boxes) when used polymorphically, by overriding the partial function `box: ValueType -> Boxer`.

marzipankaiser:
I started adding tests using #205 now, will continue doing so as I find the time.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -423,7 +433,14 @@ object Transformer {
 
     case symbols.InterfaceType(interface, List()) => Negative(interface.name.name)
 
-    case symbols.ValueTypeApp(typeConstructor, List()) => Positive(typeConstructor.name.name)
+    case symbols.ValueTypeApp(typeConstructor, targs) =>
+      // TODO: Add something like:
+      // if(!targs.forall(isSimpleDataType)) {
+      //   Context.abort(s"Unsupported type ${tpe} polymorphic over non-data type")
+      // }
```
b-studios, line 440: Can you try to just translate the `targs` here and automatically fail?

b-studios: BTW if you don't add this check now, then we probably never will ;)

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -432,6 +449,15 @@ object Transformer {
   def transform(id: Symbol): String =
     s"${id.name}_${id.id}"
 
+  def isSimpleDataType(tpe: core.ValueType): Boolean = {
```
b-studios, line 452: What do you mean by simple data type? That it is not boxed?

effekt/shared/src/main/scala/effekt/core/Parser.scala
```diff
@@ -168,13 +174,19 @@ class CoreParsers(positions: Positions, names: Names) extends EffektLexers(posit
...
   // Blocks
   // ------
-  lazy val block: P[Block] =
+  lazy val block: P[Block] = {
+    ( blockNonMember ~ many((`.` ~> id) ~ (`:` ~> blockType)) ^^ {
+      case firstReceiver ~ fields => fields.foldLeft(firstReceiver) {
+        case (receiver, field ~ tpe) => Block.Member(receiver, field, tpe)
+      }
+    }) | blockNonMember
+  }
```
b-studios, line 183: typically we would drop the braces around the body and put `| blockNonMember` on a new line to read like a grammar in a paper :)

effekt/shared/src/main/scala/effekt/core/Parser.scala
```diff
@@ -136,12 +136,18 @@ class CoreParsers(positions: Positions, names: Names) extends EffektLexers(posit
...
   // Pure Expressions
   // ----------------
-  lazy val pure: P[Pure] =
+  lazy val pure: P[Pure] = {
+    pureNonAccess ~ many((`.` ~> id) ~ (`:` ~> valueType)) ^^ {
+      case firstTarget ~ accesses => accesses.foldLeft(firstTarget){
+        case (target, field ~ tpe) => Pure.Select(target, field, tpe)
+      }
+    }
+  }
```
b-studios, line 145: same as below with dropping `{}`

effekt/shared/src/main/scala/effekt/Compiler.scala
```diff
@@ -231,15 +231,27 @@ trait Compiler {
...
+      (Frontend andThen Middleend andThen CoreDependencies andThen Aggregate andThen Backend.whole).apply(source)
+  }
 
-  def compileAll(source: Source)(using Context): Option[CoreTransformed] =
-    (Frontend andThen Middleend andThen CoreDependencies andThen Aggregate).apply(source)
+  def compileAll(source: Source)(using C: Context): Option[CoreTransformed] = C.config.backend() match {
+    case "llvm" =>
+      (Frontend andThen Middleend andThen core.PolymorphismBoxing andThen CoreDependencies andThen Aggregate).apply (source)
+    case _ =>
+      (Frontend andThen Middleend andThen CoreDependencies andThen Aggregate).apply(source)
```
b-studios, line 255: This is a good step forward, thanks!

effekt/jvm/src/test/scala/effekt/PolymorphismBoxingTests.scala
```diff
@@ -0,0 +1,286 @@
...
+        |extern {} def bar() {g: MyInterface}: Foo at {g} = "my decl"
+        |
+        |def prim = { () => return 42 }
+        |def union = { () => if (true) (f: () => Int @ {f})() else (g: () => Unit @ {g})() }
+        |def main = { () => return () }
+        |
+        |export prim
+        |""".stripMargin
+    assertTransformsTo(code, code)
+  }
```
b-studios, line 149: I am very happy about these test, thanks!

effekt/jvm/src/test/scala/effekt/PolymorphismBoxingTests.scala
```diff
@@ -0,0 +1,286 @@
...
+
+    def findSource(path: String): Option[kiama.util.Source] = None
+  }
+
+  val names = new core.Names(boxtpes ++
+    symbols.builtins.rootTypes ++ Map(
+    // TODO maybe add used names
+  ))
+
+  class ConsistentRenamer(names: core.Names, prefix: String = "l") extends core.Tree.Rewrite {
```
b-studios, line 57: I think `Consistent` should be the default. Just call it `Renamer` :)

effekt/jvm/src/test/scala/effekt/PolymorphismBoxingTests.scala
```diff
@@ -0,0 +1,286 @@
...
+      val oldBound = bound
+      bound = ids ++ bound
+      val res = f
+      bound = oldBound
+      res
+    }
+    def withBinding[R](id: symbols.Symbol)( f: => R ): R = withBindings(List(id))(f)
+
+    override def pure: PartialFunction[Pure, Pure] = {
+      case Pure.ValueVar(id, tpe) => Pure.ValueVar(rewrite(id), tpe)
```
b-studios, line 69: Hmmmmm visiting the names seems like a very common case... Maybe we can change rewrite to also have a case for this?

effekt/jvm/src/test/scala/effekt/PolymorphismBoxingTests.scala
```diff
@@ -0,0 +1,286 @@
...
+  def assertTransformsTo(input: String, expected: String): Unit = {
+    val pInput = core.CoreParsers.module(input, names) match {
+      case Success(result, next) => result
+      case nosuccess: NoSuccess => fail(nosuccess.toMessage)
+    }
+    val pExpected = core.CoreParsers.module(expected, names) match {
+      case Success(result, next) => result
+      case nosuccess: NoSuccess => fail(nosuccess.toMessage)
+    }
+    implicit val pctx: core.PolymorphismBoxing.PContext = new PolymorphismBoxing.PContext(List())(using context)
```
b-studios, line 127: In Scala 3 this would look a bit different. I can't remember myself and always have to look up the syntax, though :(

effekt/jvm/src/test/scala/effekt/PolymorphismBoxingTests.scala
```diff
@@ -0,0 +1,286 @@
...
+import effekt.util.messages
+import effekt.util.messages.DebugMessaging
+import kiama.parsing.{Failure, NoSuccess, Success}
+import kiama.util.Severities
+
+abstract class AbstractPolymorphismBoxingTests extends munit.FunSuite {
+
+  def boxDef(tpe: ValueType.ValueTypeApp): List[symbols.Symbol] = {
+    val tpeCns: symbols.TypeConstructor.Record =
+      symbols.TypeConstructor.Record(symbols.LocalName("Boxed" ++ tpe.constructor.name.name), List(), null)
```
b-studios, line 19: `null`??? really?

b-studios: I know why you do this, but ideally we wouldn't need to cook up some symbols, but only add declarations to the core module.

b-studios: Maybe you can add the symbols to `builtins` for now? WDYT?

effekt/jvm/src/test/scala/effekt/PolymorphismBoxingTests.scala
```diff
@@ -0,0 +1,286 @@
...
+    tpeCns.constructor.fields = List(
+      symbols.Field(symbols.LocalName("unbox" ++ tpe.constructor.name.name),
+        symbols.ValueParam(symbols.LocalName("value"), Some(tpe)), tpeCns.constructor))
+    List(tpeCns, tpeCns.constructor, tpeCns.constructor.fields.head)
+  }
+
+  val boxtpes = symbols.builtins.rootTypes.values.flatMap {
+    case t: TypeConstructor.ExternType => boxDef(ValueType.ValueTypeApp(t, List()))
+    case _ => Nil
+  }.map { c => (c.name.name, c) }.toMap
```
b-studios, line 31: This triggers an interesting question: how do you represent extern types?

effekt/jvm/src/test/scala/effekt/PolymorphismBoxingTests.scala
```diff
@@ -0,0 +1,286 @@
...
+        }
+      }
+    }
+
+    object messaging extends DebugMessaging
+
+    def contentsOf(path: String): Option[String] = None
+
+    def findSource(path: String): Option[kiama.util.Source] = None
+  }
```
b-studios, line 50: This might seem akward, but is actually good ;)

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
marzipankaiser: Yes, it's probably not a big problem (PolymorphismBoxing should have removed all of them). Actually, IIUC, the whole function should ideally be dropped (it is `transform(symbols.Type)` which we don't want to use anymore, but it is still called from https://github.com/effekt-lang/effekt/pull/199/files/f50fb68cac88083e264fd21b96cfbaf0a61c89d4#diff-2bc35700065c09eb82c800736c2625a2516ddb669bb20acbfce3d9af8dbca683R343 for the types of fields.
I just noticed that this is now annotated with a type anyway, so we can use that and drop this function.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
marzipankaiser: Yes, exactly. I am open to suggestions for a good name here..

effekt/jvm/src/test/scala/effekt/PolymorphismBoxingTests.scala
marzipankaiser: Agreed.
Thinking about it, having something like this for types, too, might even make the `PolymorphismBoxing` phase simpler 🤔

effekt/jvm/src/test/scala/effekt/PolymorphismBoxingTests.scala
marzipankaiser: We could do that, yes (We would still temporarily use something like null to be able to tie the knot, though).

The main reason, why I used symbols here is, that the `Boxed*` types should be part of the stdlib module, and this way, we can run this phase separately on the modules (before `Aggregate`).

marzipankaiser: Currently, they are defined here: https://github.com/effekt-lang/effekt/pull/199/files/f50fb68cac88083e264fd21b96cfbaf0a61c89d4#diff-feddea404f7474369f1a93d26ee304085c61ce17003f0ea38c076fc30cd6e118R162-R166
and will then be searched in the prelude module (by name); this is generating them for testing.

marzipankaiser: That raises a general question: Not relying on symbols, where do we get the declarations for imported types/... when working on core? Or should the declarations contain _all_ declarations and not only the ones from this module? (In that case: How can we differentiate between imported and other declarations?)

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
marzipankaiser: That function is now gone.

effekt/jvm/src/test/scala/effekt/PolymorphismBoxingTests.scala
marzipankaiser: It's `given T = <some expression of type T>`. Fixed.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -350,9 +360,9 @@ object Transformer {
 
   def getTagFor(constructor: symbols.Constructor)(using Context): Int = constructor.tpe match {
```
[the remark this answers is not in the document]
marzipankaiser: This is used
- for `Match` here: https://github.com/effekt-lang/effekt/blob/f50fb68cac88083e264fd21b96cfbaf0a61c89d4/effekt/shared/src/main/scala/effekt/machine/Transformer.scala#L162 
- for `PureApp` here: https://github.com/effekt-lang/effekt/blob/f50fb68cac88083e264fd21b96cfbaf0a61c89d4/effekt/shared/src/main/scala/effekt/machine/Transformer.scala#L331

which should be easy enough to fix using the context introduced in #209 (https://github.com/effekt-lang/effekt/pull/209/files#diff-2bc35700065c09eb82c800736c2625a2516ddb669bb20acbfce3d9af8dbca683R470 ).
For the latter, we would still match on the symbol type, though (this happens in multiple places).

marzipankaiser:
Rebased onto current master and cleaned up the commit history slightly in the process (moving bugfix commits to the point where the bug was introduced, mostly).

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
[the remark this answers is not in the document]
marzipankaiser: Fixed.

marzipankaiser:
With something like #213, this could be changed to generate a `BoxedT` exactly for each `extern type T`, allowing this to be backend-specific without explicitly exchanging/passing parameters to this Phase.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
marzipankaiser: @me `requiresBoxing`

effekt/jvm/src/test/scala/effekt/PolymorphismBoxingTests.scala
marzipankaiser: Done.

marzipankaiser:
@b-studios I think I did all the changes we talked about now.

PR #203, opened 2022-12-02
Feature/typed core draft

b-studios:
This is very much WIP. Baby steps towards a typed version of core.

TODO
- [x] properly substitute types for patterns in the match compiler.
- [ ] fuse elaboration and transformer to avoid inconsistent intermediate states.
- [x] Maybe drop `cargs` again (since they are always bargs_i.capt).


effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -147,53 +142,61 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
...
+      Member(transformAsBlock(receiver), s.definition, transform(Context.inferredBlockTypeOf(tree)))
 
     case s @ source.New(h @ source.Implementation(tpe, members)) =>
       // we need to establish a canonical ordering of methods
       // we use the order in which they are declared in the signature (like with handlers)
       val clauses = members.map { cl => (cl.definition, cl) }.toMap
       val sig = h.definition
 
-      New(Implementation(sig, sig.ops.map(clauses.apply).map {
+      val coreType = transform(Context.inferredBlockTypeOf(s)) match {
```
phischu, line 161: Should `s` be `h`?

b-studios:
I think this PR is now ready for review. A review should mostly focus on the design (also wrt to how we can also port lifted at some point). 

Symbols should be treated as opaque starting from core (in the future). Some backends still use symbol information instead of looking up the information in the declarations section of the core module.

b-studios:
@marzipankaiser I added you as a reviewer since you are most affected by it, since you want to implement type-boxing/unboxing based on typed core.

b-studios:
@fritzladwig this PR could also really help you, since the new representation might be easier to test. We could easily write a parser for this core representation. The parsed trees then could be used in your tests for optimization.

effekt/shared/src/main/scala/effekt/core/Type.scala
```diff
@@ -61,7 +61,26 @@ object Type {
...
   val TRegion = BlockType.Extern(builtins.RegionSymbol, Nil)
 
-  def join(tpe1: ValueType, tpe2: ValueType): ValueType = tpe1 // TODO implement properly
+  /**
+   * Function types are the only type constructor that we have subtyping on.
+   */
+  def merge(tpe1: ValueType, tpe2: ValueType, covariant: Boolean): ValueType = (tpe1, tpe2) match {
+    case (ValueType.Boxed(btpe1, capt1), ValueType.Boxed(btpe2, capt2)) =>
+      ValueType.Boxed(merge(btpe1, btpe2, covariant), merge(capt1, capt2, covariant))
+    case (tpe1, tpe2) => tpe1 // TODO we could also fail here.
```
phischu, line 70: And `TBottom`?

effekt/shared/src/main/scala/effekt/core/Tree.scala
```diff
@@ -59,39 +74,44 @@ case class ModuleDecl(
...
  */
 enum Declaration extends Tree {
-  def id: Symbol
+  def id: Id
 
-  case Data(id: Symbol, ctors: List[Symbol])
-  case Record(id: Symbol, fields: List[Symbol])
-  case Interface(id: Symbol, operations: List[Symbol])
+  case Data(id: Id, tparams: List[Id], constructors: List[Constructor])
+  case Record(id: Id, tparams: List[Id], constructor: Constructor)
```
phischu, line 87: Consider absorbing `Record` into `Data` (and likewise on the object language type level). Keep `Select` (it happens to work on all data types with a single constructor).

b-studios: I agree, let's do this. However, we should do this after this PR since we also need to push this change through lifted and I don't want to touch lifted too much in this PR.

b-studios:
Thanks @phischu for the comments. We should address them in a separate PR

PR #204, opened 2022-12-05
Common Context objects for phases on core and lifted

b-studios:


marzipankaiser:
Rebased and added a first draft of a `CoreContext` that tracks the `effekt.context.Context` and some `core.Declaration`s.

marzipankaiser:
Additional note on `CoreContext`: If the `Decl`s in lifted stay like they are (a different type), we would probably want something like this for `lifted`, too.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -449,7 +431,7 @@ object Transformer {
    * Extra info in context
    */
 
-  def abort(message: String)(using C: Context) =
+  def abort(message: String)(using C: ErrorReporter) =
     C.abort(message)
```
b-studios, line 435: Is this even called somewhere?

marzipankaiser:
If we want anything after core.Transformer not to inspect symbols, we will have to also encode `builtins` into the declarations/externs.

marzipankaiser:
> If we want anything after core.Transformer not to inspect symbols, we will have to also encode `builtins` into the declarations/externs.

I think this will be a bigger (-ish) and somewhat separate change, and thus make a separate PR for it after this one.

effekt/shared/src/main/scala/effekt/generator/chez/ChezScheme.scala
```diff
@@ -59,16 +59,16 @@ trait ChezScheme {
   /**
    * Entrypoint used by the LSP server to show the compiled output
    */
-  def compileSeparate(input: AllTransformed)(using C: Context) =
-    C.using(module = input.main.mod) { Some(chez.PrettyPrinter.format(compile(input.main))) }
```
b-studios, line 63: This module is also used for error reporting positions.

marzipankaiser: Ah, TIL. Then we have to keep it, of course.

b-studios: But if we do not use error reporting anymore, it can of course be dropped.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -347,27 +357,22 @@ object Transformer {
...
 
-  def getTagFor(constructor: symbols.Constructor)(using Context): Int = constructor.tpe match {
-    case symbols.DataType(name, Nil, constructors) => constructors.indexOf(constructor)
-    case symbols.DataType(name, tparams, constructors) => Context.abort("Not yet supported: (data) type polymorphism")
-    case symbols.Record(name, Nil, constructor) => builtins.SingletonRecord
-    case symbols.Record(name, tparams, constructor) => Context.abort("Not yet supported: (data) type polymorphism")
-    case t @ symbols.ExternType(name, tparams) => Context.abort(s"Application to an unknown symbol: $t")
-  }
+  def getTagFor(constructor: core.Id)(using DeclarationContext, ErrorReporter): Int =
+    DeclarationContext.getConstructor(constructor).data.constructors.indexWhere(_.id == constructor)
```
b-studios, line 364: Maybe this can be part of the `DeclarationContext`?

b-studios:
Otherwise it LGTM

effekt/shared/src/main/scala/effekt/generator/js/JavaScript.scala
```diff
@@ -259,13 +248,8 @@ object JavaScript extends Backend {
...
 
-  def tagFor(constructor: Id)(using C: GeneratorContext): js.Expr = {
-    val position = C.declarations.collectFirst {
-      case Declaration.Data(_, _, cs) if cs.exists(c => c.id == constructor) =>
-        cs.indexWhere(c => c.id == constructor)
-    }.getOrElse {
-      C.context.panic(s"Cannot find constructor ${constructor} in data type declarations.")
-    }
+  def tagFor(constructor: Id)(using C: DeclarationContext, E: Context): js.Expr = {
+    val position = C.getConstructor(constructor).data.constructors.indexWhere(_.id == constructor)
```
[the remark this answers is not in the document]
marzipankaiser: Ideally, we'd throw an error. This can be part of the method in the DeclarationContext.

PR #205, opened 2022-12-06
Core parser

b-studios:
Still missing:

- [x] Match

This is an example of how to use the parser:

```scala
object ParserExample extends scala.App {

  val parserPositions = new Positions
  val names = new Names(Map("Int" -> symbols.builtins.IntSymbol))
  object messages extends DebugMessaging

  val parsers = CoreParsers(parserPositions, names)

  given ErrorReporter with { var focus = NoSource; val messaging = messages; val positions = parserPositions }

  val result = parsers.parse(StringSource(
    s"""module foo
       |
       |import list/foo
       |
       |type List['A] {
       |  Nil()
       |  Cons(head: 'A, tail: List['A])
       |}
       |
       |interface Exc['E] { raise: ['A]('E) => 'A }
       |
       |extern "my extern"
       |
       |extern {io} def foo['A2](): 'A2 = "my decl"
       |extern {} def bar() {f@f: MyInterface}: Foo at {f} = "my decl"
       |extern {} def bar() {g: MyInterface}: Foo at {g} = "my decl"
       |
       |def hof = { ['b](){ f@f: MyInterface } => (f: (){g:MyIDC} => (Sig at {g}) @ {})(){ f:NyInterface@{hello} } }
       |def prim = { () => return 42 }
       |def union = { () => if (true) (f: () => Int @ {f})() else (g: () => Unit @ {g})() }
       |def baz = { ['b](){ f@f: MyInterface } => (f: () => Unit @ {})() }
       |def bar = { ['b](){ f@f: MyInterface } => return box {f} (f : MyInterface @ {f}) }
       |
       |let myList = !(Cons : (Int, List[Int]) => List[Int] @ {io})(1, (Cons : (Int, List[Int]) => List[Int] @ {io})(2, (Nil : () => List[Int] @ {io})()))
       |
       |export bam
       |""".stripMargin, "input.effekt"))

  def showType(tpe: ValueType) =
    PrettyPrinter.pretty(PrettyPrinter.toDoc(tpe), 30).layout

  def showType(tpe: BlockType) =
    PrettyPrinter.pretty(PrettyPrinter.toDoc(tpe), 30).layout

  println(result)
  println("----")
  result.foreach { module =>
    println(core.PrettyPrinter.format(module).layout)
    module.definitions(0) match {
      case Definition.Def(id, block) =>
        println(showType(block.tpe))
        println(block.capt)
      case Definition.Let(id, binding) => ()
    }
  }
  messages.buffer.foreach {
    msg => println(msg)
  }
}
```

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Driver.scala
+++ b/effekt/jvm/src/main/scala/effekt/Driver.scala
@@ -61,7 +61,7 @@ trait Driver extends kiama.util.Compiler[Tree, ModuleDecl, EffektConfig, EffektE
     C.setup(config)
 
     def compile(): Option[String] = C.compileWhole(src) map {
-      case Compiled(main, outputFiles) =>
+      case Compiled(_, main, outputFiles) =>
         outputFiles.foreach {
           case (filename, doc) =>
             saveOutput(filename, doc.layout)
--- a/effekt/shared/src/main/scala/effekt/Compiler.scala
+++ b/effekt/shared/src/main/scala/effekt/Compiler.scala
@@ -19,48 +19,51 @@ import kiama.util.{ Positions, Source }
  *
  * All phases have a source field, which is mostly used to invalidate caches based on the timestamp.
  */
-sealed trait PhaseResult { val source: Source }
+enum PhaseResult {
 
-/**
- * The result of [[Parser]] parsing a single file into a [[effekt.source.Tree]].
- */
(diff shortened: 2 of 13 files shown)


b-studios:
I am merging this, since this can be of help for @marzipankaiser and @fritzladwig.
However, it would be great to discuss the design-aspect of the compiler pipeline in #204 and maybe revert / change that part, later.

PR #206, opened 2022-12-07
Optimizations on Core

fritzladwig:


fritzladwig:
During review check optimizations in regard to work duplication (e.g. beta reduction)

b-studios:
See #267

PR #208, opened 2022-12-08
Remove deprecated test

JonathanStarup:


b-studios:
I was not sure whether at some point we will check this again, but I think you are right and we can delete it.

PR #209, opened 2022-12-09
Annotate more core types in lifted

### LINKED ISSUE #207: LLVM: crashes on `foldr`
The following code:
```effekt
type ListInt {
    Nil(); Cons(hd: Int, tl: ListInt)
}

def foldr(l: ListInt, init: Int){ f: (Int, Int) => Int }: Int = {
    l match {
        case Nil() => init
        case Cons(hd, tl) => f(hd, foldr(tl, init){f})
    }
}

def main() = {
   val lst = [1,2,3,4,5,6,7,8,9,10]
   val sum = foldr(lst, 0){ (x,y) => x + y}
   println(sum)
}
```
([source code](https://gist.github.com/marzipankaiser/d083bd6df71818354a05d6b6abddb4b4))
causes the LLVM backend to crash with the following error:
```
opt: ./out/issue_b.ll:1112:20: error: '%init_102' defined with type '%Neg = type { ptr, ptr }' but expected 'i64'
        store %Int %init_102, ptr %init_102p.120
                   ^
[error] java.lang.RuntimeException: Nonzero exit value: 1
```
[stack trace omitted]

marzipankaiser:
This annotates core types at some more positions in lifted,
and uses this to also compute types in `freeVariables`.

As a necessity for this, it also threads the `lifted.Decl`s through the `machine.Transformation` (so it does not inspect the symbol anymore in at least one case).

Fixes #207.


effekt/shared/src/main/scala/effekt/generator/chez/ChezSchemeLift.scala
```diff
@@ -99,17 +99,17 @@ object ChezSchemeLift extends Backend {
...
       chez.Let(List(Binding(nameDef(id), chez.Builtin("box", toChez(init)))), toChez(body))
 
-    case State(id, init, region, body) =>
+    case State(id, init, stateTpe, region, body) =>
       chez.Let(List(Binding(nameDef(id), chez.Builtin("fresh", Variable(nameRef(region)), toChez(init)))), toChez(body))
 
     case Try(body, tpe, handler) =>
       val handlers: List[chez.Handler] = handler.map { h =>
-        val names = RecordNames(h.id)
+        val names = RecordNames(h.id.symbol)
```
b-studios, line 110: I previously tried to get rid of the dependency on symbols in all phases after core. Why is this necessary here?

effekt/shared/src/main/scala/effekt/lifted/Tree.scala
```diff
@@ -102,7 +100,7 @@ case object Hole extends Stmt
  *
  * Used to represent handlers / capabilities, and objects / modules.
  */
-case class Implementation(id: symbols.Interface, operations: List[Operation]) extends Tree
+case class Implementation(id: core.BlockType.Interface, operations: List[Operation]) extends Tree
```
b-studios, line 103: Ah, this is why :)

effekt/shared/src/main/scala/effekt/lifted/Tree.scala
```diff
@@ -120,72 +118,70 @@ def Here() = Evidence(Nil)
 
 class EvidenceSymbol() extends Symbol { val name = Name.local(s"ev${id}") }
 
-def freeVariables(d: Definition): Set[Symbol] = d match {
+def freeVariables(d: Definition): Set[Param] = d match {
```
b-studios, line 121: Not sure collecting params here is the right thing to do. Eventually, we might drop param and replace it by separate lists of value and block bindings (like in core). Is it because you want to collect `(Set[ValueSymbol], Set[BlockSymbol])`?

b-studios: On the other hand, also having the types does make sense.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -54,18 +55,19 @@ object Transformer {
...
-          case Definition.Def(id, tpe, block @ lifted.BlockLit(params, body)) =>
+          case Definition.Def(id, tpe, block @ lifted.BlockLit(tparams, params, body)) =>
             // TODO does not work for mutually recursive local definitions
             val freeParams = lifted.freeVariables(block).toList.collect {
-              case id: symbols.ValueSymbol => Variable(transform(id), transform(tpe))
-              case id: symbols.BlockParam => Variable(transform(id), transform(tpe))
-              case id: symbols.ResumeParam => Variable(transform(id), transform(tpe))
-              case id: lifted.EvidenceSymbol => Variable(transform(id), builtins.Evidence)
+              case lifted.ValueParam(id, tpe) => Variable(transform(id), transform(tpe))
+              case lifted.BlockParam(id: (symbols.BlockParam | symbols.ResumeParam), tpe) =>
```
b-studios, line 67: Beautiful, much better!

b-studios: Why do you need the union type here? Can't you just drop it?

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -453,6 +455,20 @@ object Transformer {
...
+        case i: lifted.Decl.Interface if i.id == id => i
+      }.getOrElse{ Context.abort(s"Expected an interface, got ${id}") }
+    }
+    def findData(id: Symbol)(using Context): lifted.Decl.Data = {
+      declarations.collectFirst {
+        case i: lifted.Decl.Data if i.id == id => i
+      }.getOrElse{ Context.abort(s"Expected a data type, got ${id}") }
+    }
+  }
+  def ModuleContext(using MC: ModuleContext): ModuleContext = MC
```
b-studios, line 470: Great, at some point we should start sharing this.

b-studios (approved):
LGTM. I have a few comments and questions, but they can be adressed later.

b-studios:
In a separate commit, we could start homogenizing lifted to core. I am out of this game ;) This is the job of you (@marzipankaiser), @mm0821 , and @phischu who are the owners of `lifted`.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
marzipankaiser: This is not the nicest solution. IIUC, we should collect only those `freeVariables` that are bound somewhere on the way to the toplevel, minus local functions (which will become toplevel definitions, too).
See also https://github.com/effekt-lang/effekt/pull/209/files#diff-2bc35700065c09eb82c800736c2625a2516ddb669bb20acbfce3d9af8dbca683R71

effekt/shared/src/main/scala/effekt/lifted/Tree.scala
marzipankaiser: Yes, having the types is the main reason, and that there is no type for evidence. An alternative would be a `List[(Id, Type)]` (or similar), but that would require us to add a type for Evidence (which would then ideally be used in other places, too.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
marzipankaiser: I think this code here might be the reason that `pos/defdef.effekt` is failing for LLVM 🤔

PR #210, opened 2022-12-15
Wrap pure/direct blocks in core to use the correct calling convention

marzipankaiser:


effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -170,10 +170,45 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
     Box(transformAsBlock(tree), transform(Context.inferredCapture(tree)))
 
   def transformAsBlock(tree: source.Term)(using Context): Block = tree match {
```
b-studios, line 172: I think this will translate too many cases. For example in the call `println(42)` it might eta expand `println`, no?

b-studios: Maybe we can have a separate `transformAsControlBlock` or similar (ideally with a better name / comment what that means)?

marzipankaiser: I had a `transformAsBlockArgument` originally, but then noticed that there were no relevant usages of the original function left.
In function calls, the callee is always an Id: https://github.com/effekt-lang/effekt/blob/d4a12916ae5b38326b4dbc30aa72ae90e0d88e36/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L407
Called from here, directly with source symbol: https://github.com/effekt-lang/effekt/blob/d4a12916ae5b38326b4dbc30aa72ae90e0d88e36/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L336-L338
Receivers of method calls get fed through this, but are not wrapped, so this should be fine IIUC.

marzipankaiser: Usages of `transformAsBlock` are:
- https://github.com/effekt-lang/effekt/blob/d4a12916ae5b38326b4dbc30aa72ae90e0d88e36/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L66
  https://github.com/effekt-lang/effekt/blob/d4a12916ae5b38326b4dbc30aa72ae90e0d88e36/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L149
  => must be wrapped, since any symbol referring to a definition is always called using `App`
- https://github.com/effekt-lang/effekt/blob/d4a12916ae5b38326b4dbc30aa72ae90e0d88e36/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L170
  => I thought boxed functions must be wrapped. Apparently, this is not necessarily the case. However I suspect this might be a problem:
  https://github.com/effekt-lang/effekt/blob/d4a12916ae5b38326b4dbc30aa72ae90e0d88e36/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L330-L334
  If we `box` a toplevel definition this will lead to the original issue IIUC :thinking:
- https://github.com/effekt-lang/effekt/blob/d4a12916ae5b38326b4dbc30aa72ae90e0d88e36/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L178
  https://github.com/effekt-lang/effekt/blob/d4a12916ae5b38326b4dbc30aa72ae90e0d88e36/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L295-L296
  => here, the block is an interface, so no wrapping will occur
- https://github.com/effekt-lang/effekt/blob/d4a12916ae5b38326b4dbc30aa72ae90e0d88e36/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L299
  https://github.com/effekt-lang/effekt/blob/d4a12916ae5b38326b4dbc30aa72ae90e0d88e36/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L327
  https://github.com/effekt-lang/effekt/blob/d4a12916ae5b38326b4dbc30aa72ae90e0d88e36/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L413
  => These are block arguments, must be wrapped.

marzipankaiser: I could still rename the function to make this more obvious.

marzipankaiser: Actually, for the "is an interface" cases, we might want to have a separate `transformAsInterfaceBlock`? This also adds an additional sanity check, that we don't call methods on a function or try to call an interface (instead of one of it's methods). 🤔

b-studios: > we might want to have a separate transformAsInterfaceBlock?
I think this is fine not to have :)

marzipankaiser: After having done that split, I am not entirely sure it is a good idea: There is no case where we call `transformAsControlBlock` from the rest of the transformation (as there are only symbols in function position), and it adds some duplication between the two (esp. Select).
Also, dispatching is more convoluted than it should be IMHO.

effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -169,25 +169,114 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
   def transformBox(tree: source.Term)(implicit C: Context): Pure =
     Box(transformAsBlock(tree), transform(Context.inferredCapture(tree)))
 
+  /**
+   * Transforms the soruce to a function (expecting to be called using [[core.Stmt.App]] or an interface.
```
b-studios, line 173:
```suggestion
   * Transforms the source to a function (expecting to be called using [[core.Stmt.App]] or an interface.
```

effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -169,25 +169,114 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
...
+   */
+  def transformAsObject(tree: source.Term)(using Context): Block = tree match {
+    case v: source.Var =>
+      val sym = v.definition
+      val tpe = Context.blockTypeOf(sym)
+      tpe match {
+        case _: BlockType.FunctionType =>
+          Context.panic(s"Using function symbol ${sym} but an object was expected.")
+        case _: BlockType.InterfaceType => BlockVar(sym.asInstanceOf)
+      }
```
b-studios, line 205: Do we need this check here?

effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -169,25 +169,114 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
...
+          Context.abort(s"Expected a function but got an object of type ${t}")
+      }
+
     case source.BlockLiteral(tps, vps, bps, body) =>
       val tparams = tps.map(t => t.symbol)
       val cparams = bps.map { b => b.symbol.capture }
       BlockLit(tparams, cparams, vps map transform, bps map transform, transform(body))
 
+    case s @ source.Select(receiver, selector) =>
+      Member(transformAsObject(receiver), s.definition, transform(Context.inferredBlockTypeOf(tree)))
```
b-studios, line 272: I might be confusing things, but I hoped this could've been shared.

effekt/shared/src/main/scala/effekt/core/Transformer.scala
b-studios: I see you concerns. At the same time, I like this because we don't pretend to automagically eta-expand objects :) Not sure... hmm...

marzipankaiser: That's true. I'll leave it as it is.

effekt/shared/src/main/scala/effekt/core/Transformer.scala
marzipankaiser: No, this should already be catched by Typer (I think?). It is mostly there because I re-used the structure of the earlier definitions...^^'

effekt/shared/src/main/scala/effekt/core/Transformer.scala
marzipankaiser: It could, using yet another function like `transformAsSelectOrUnbox` (ideally with a better name). Since it's not too much/complicated I decided not to and just copied it, though I could just extract it.

effekt/shared/src/main/scala/effekt/core/Transformer.scala
marzipankaiser: I removed it for simplicity.

b-studios:
Looks like thinks are getting out of hand with additional methods ;) I think I have to stare at it after we merged it.

marzipankaiser:
A little, yes. Some kind of kind system for this would be better, but is much more work to implement (and this at least makes it not crash).

b-studios:
LGTM, as soon as the tests pass I'll merge.

PR #211, opened 2022-12-18
Refactor/visitor macro

b-studios:
Start implementing macro to generate tree traversals by simple structural recursion.

@marzipankaiser we talked about this. I am not yet sure the gains are worth the additional complexity. It was fun to implement, though. 

Gains:
- much more friendly with respect to refactorings
- much easier to also add rewrites for types
- impossible to forget cases (it is easily possible to forget whole non-terminals, though)
- very easy to add a new IR (and rewrites).
- saved a few lines (though, this is not the main motivation)

Problems:
- impossible to understand
- very difficult to debug (I added a couple of comments and commands to help with that, though)


b-studios:
I think I prefer manually writing recursive functions and then sprinkling in the recursive macro for boring cases. Like in:

```scala
enum Exp { case Add(l: Exp, r: Exp); case Sub(l: Exp, r: Exp); case Lit(n: Int) }
export Exp.*

object Optimizations extends Structural {
  def dropAddZero(e: Exp): Exp = e match {
    case Add(l, Lit(0)) => dropAddZero(l)
    case Add(Lit(0), r) => dropAddZero(r)
    case e => structuralRewrite(e)
  }
}

@main def demo() = println(Optimizations.dropAddZero(Add(Lit(0), Add(Lit(1), Lit(0)))))
```

b-studios:
If there are no other voices about this PR, I would go ahead and merge it. You may speak now or forever hold your peace.

b-studios:
I take this as a "OK" :)

Issue #30, opened 2020-10-20
Module System


phischu:
We want a module system for Effekt. At first we should focus on a simple and clear basis on which we then build interesting extensions.

# Properties

Some features we might want are:

1. Separate compilation should be possible
2. It should be possible to define modules locally, in a file, and then gradually move parts to different files.

# Idea

We look at [namespaces in Flix](https://doc.flix.dev/namespaces/) and objects as modules in Scala for inspiration.

```
module A {
  def f() = ...
}
def main() = {
  A.f()
}
```

## Opening modules

# Corner cases

Here we gather examples where the behavior is not so clear

## Nested folders versus nested modules in one file

Consider the following file layout
```
|- A.effekt
|- A
  |- B.effekt
```
and the following content
```
// A.effekt
module B { module C {} }
// A/B.effekt
module C {}
```
If we now use `A.B.C.f()` what should happen?

anfark:
WIP see branch [feature/modules](https://github.com/effekt-lang/effekt/tree/feature/modules)

b-studios:
Most use cases are covered by #388, but we still need to reconsider the design objects / classes etc.

PR #213, opened 2022-12-19
WIP: Make more types extern instead of builtin

marzipankaiser:
This PR is a draft to move types from being builtins (in `symbols.builtins`) to `extern type` declarations in the prelude.
To do so, it annotates the operations using those types (`If`, `While` for `Boolean`, `Assign` for `Unit` (and `Ref`)) with `Id`s referring to them, which are then resolved by Namer and used by Typer.

For phases after core, it will also add an `ExternContext` (sim. to `DeclarationContext` in #204).

Implications:
- Using e.g. `If` in a backend that does not support `Boolean`s yet will lead to an error in Namer
- All of the builtin types are identified by name (only) and can potentially be overwritten, possibly leading to strange behaviour
    - This might be solved by having a #30.
    - On the plus side, this also allows individual backends to decide that, e.g., `Boolean` is a data type (not builtin)

### DIFF
--- a/effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
@@ -1,6 +1,6 @@
 package effekt.llvm
 
-import effekt.llvm.Operand.LocalReference
+import effekt.llvm.Operand.{ConstantGlobal, LocalReference}
 import effekt.machine
 import effekt.machine.analysis.*
 
@@ -361,8 +361,33 @@ object Transformer {
   def stkType = NamedType("Stk");
   def regionType = NamedType("Region");
 
+  case class ExternTypeDecl(
+     tpe: Type,
+     size: Int,
+     share: Option[Operand.ConstantGlobal],
+     erase: Option[Operand.ConstantGlobal])
+  /**
+   * Parses the body of a `extern type` declaration.
+   *
+   * For LLVM, this has the form:
+   * `llvm-type-name bit-size [refcount-increase refcount-decrease]`
+   * (`[]` marking the last two parameters as optional)
+   */
+  def parseExternTypeDecl(machineTpe: machine.Type.Extern): ExternTypeDecl = {
+      machineTpe.name.split(" ").toList match {
+        case List(tpeName, size, share, erase) =>
+          ExternTypeDecl(RawType(tpeName), size.toInt,
+            Some(ConstantGlobal(PointerType(), share)),
+            Some(ConstantGlobal(PointerType(), erase)))
+        case List(tpeName, size) =>
+          ExternTypeDecl(RawType(tpeName), size.toInt, None, None)
+        case _ =>
+          ??? // TODO thread error reporter through to here
+      }
+  }
+
   def transform(tpe: machine.Type): Type = tpe match {
-    case machine.Type.Extern(name)   => RawType(name.split(" ")(0))
+    case e: machine.Type.Extern   => parseExternTypeDecl(e).tpe
     case machine.Positive(_)         => positiveType
     case machine.Negative(_)         => negativeType
     case machine.Type.Stack()        => stkType
@@ -375,7 +400,7 @@ object Transformer {
 
   def typeSize(tpe: machine.Type): Int =
     tpe match {
-      case machine.Type.Extern(name) => name.split(" ")(1).toInt
+      case e: machine.Type.Extern => parseExternTypeDecl(e).size
       case machine.Positive(_)       => 16
       case machine.Negative(_)       => 16
       case machine.Type.Stack()      => 8 // TODO Make fat?
@@ -544,11 +569,8 @@ object Transformer {
 
   def shareValue(value: machine.Variable)(using FunctionContext, BlockContext): Unit = {
     value.tpe match {
-      case machine.Type.Extern(name) => {
-        val decl = name.split(" ")
-        if (decl.length > 2) {
-          emit(Call("_", VoidType(), ConstantGlobal(PointerType(), decl(2)), List(transform(value))))
-        } else ()
+      case e: machine.Type.Extern => parseExternTypeDecl(e).share.foreach{ share =>
+          emit(Call("_", VoidType(), share, List(transform(value))))
       }
       case machine.Positive(_)       => emit(Call("_", VoidType(), sharePositive, List(transform(value))))
       case machine.Negative(_)       => emit(Call("_", VoidType(), shareNegative, List(transform(value))))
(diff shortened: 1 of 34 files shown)


marzipankaiser:
`Unit` is harder than I thought, since State (`Ref`) depends on it as the return type of `put`.
The symbols regarding State are used extensively in multiple backends, so this will be some work to move.

marzipankaiser:
For `Ref` (and also maybe for the other changes here, but to a lesser degree), an interesting question comes up:
We can
1. Add the relevant Ids to the source Tree and annotate them (e.g. `put` to `Assign`, `Ref` to `State`, ...), or
2. use new kinds of annotation to annotate the nodes with the Getter/Setter/...

1 is what I am doing so far, but there is a difficulty for `Var`: We do not actually know if `get` is used until we resolve it in `Namer`. So we can either use 2 here, or we will have a spurious reference to `get` for each reference to a `val`.

b-studios:
> On the plus side, this also allows individual backends to decide that, e.g., Boolean is a data type (not builtin)


But wouldn't this also require overloading literals `true` and `false`?

b-studios:
> For `Ref` [...]

TBH I don't understand the discussion. Are you talking about "how can we figure out the correct symbols for state in Typer"?

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -548,25 +544,31 @@ object Transformer {
 
   def shareValue(value: machine.Variable)(using FunctionContext, BlockContext): Unit = {
     value.tpe match {
+      case machine.Type.Extern(name) => {
+        val decl = name.split(" ")
+        if (decl.length > 2) {
```
b-studios, line 549: This looks very ad hoc :) What is happening here?

effekt/shared/src/main/scala/effekt/machine/Tree.scala
```diff
@@ -203,12 +203,10 @@ export Statement.*
  * Types
  */
 enum Type {
+  case Extern(name: Predef.String)
   case Positive(name: Predef.String)
   case Negative(name: Predef.String)
   case Stack()
```
b-studios, line 209: Maybe define `Int` as `Type.Extern(%Int 8)` some where to be reused?

effekt/shared/src/main/scala/effekt/symbols/builtins.scala
```diff
@@ -67,10 +55,6 @@ object builtins {
 
   val rootTypes: Map[String, TypeSymbol] = Map(
     "Unit" -> UnitSymbol,
-    "Boolean" -> BooleanSymbol,
-    "Int" -> IntSymbol,
-    "Double" -> DoubleSymbol,
-    "String" -> StringSymbol,
```
b-studios, line 73: Fantastic!

effekt/shared/src/main/scala/effekt/machine/Tree.scala
b-studios: Nevermind, it is defined in the prelude, of course.

effekt/shared/src/main/scala/effekt/machine/Tree.scala
```diff
@@ -217,7 +215,7 @@ export Type.{ Positive, Negative }
 type Evidence = Int
 object builtins {
 
-  val Evidence = Type.Int()
+  val Evidence = Type.Extern("%Int 8")
```
b-studios, line 218: Does it make sense to define this in the prelude as well?

```
extern type Evidence ="..."
```

marzipankaiser:
> > On the plus side, this also allows individual backends to decide that, e.g., Boolean is a data type (not builtin)
> 
> But wouldn't this also require overloading literals `true` and `false`?

Yes, it would (in some way)

marzipankaiser:
> > For `Ref` [...]
> 
> TBH I don't understand the discussion. Are you talking about "how can we figure out the correct symbols for state in Typer"?

Yes, exactly. Currently, this adds the symbols to the source tree (like for `Literal`). Though now, I am leaning more towards introducing them in `Namer` as a new kind of annotation (e.g. `Getter` on `Var` to annotate the symbol of the `get` function, `ReferenceTypeConstructor` on `State` etc).

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
marzipankaiser: It kind of is. An `extern type` gets a `String` body, which can (will) be backend-specific. IIUC, for LLVM, there are 4 kinds of information we need: 
1. The LLVM 'type' (e.g. `%Int`)
2. The Number of bits it occupies
3. ~ How to increase refcounts
4. ~ How to decrease refcounts

3 and 4 are optional (only needed for types that are partially on the heap).

Yes, ideally we would use a more structured format here.

effekt/shared/src/main/scala/effekt/machine/Tree.scala
marzipankaiser: I am unsure about that one, too.
On one hand, we do not currently have a mechanism to hide it from the user in that case. On the other hand, that might be introduced via #30 someday, and this is very ad hoc.

Is the representation of evidence sufficiently uniform? If so, we could in principle also maybe have `LiftInference` introduce it (as some kind of type alias or something like that).

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
b-studios: I think it would be good to have a function called "parseExternSomething" (or variations thereof) that explains this in a comment.

effekt/shared/src/main/scala/effekt/machine/Tree.scala
b-studios: I think the latter is a great idea! We also need to document somewhere which builtin types are required to exist when.

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
marzipankaiser: Agreed. I have changed that now.

marzipankaiser:
For `Ref` (mutable state) the possible strange behaviour without a #30 occurs. Some tests (at least `examples/pos/probabilistic.effekt`) define `Ref` themselves.

marzipankaiser:
While I still would support the direction this tries to move in, and this did create some insights, I don't think that it's worthwhile to update this to the current version (in contrast to similar changes in the current one), so closing.

Issue #218, opened 2023-01-16
Machine transformation: passing functions that close over local state directly fails


marzipankaiser:
In a program like the following:
```effekt
def hof(x: Int){ f: Int => Int }: Int = {
    f(x)
}

def foo(a: Int) = {
    def bar(b: Int): Int = b*a
    hof(5){bar}
}

def main() = { 
    foo(2)
}
```
after the machine translation, the function `bar` will have an additonal parameter (for :
```effekt-machine
  def bar_196((ev257_257 : Int, b_195 : Int, a_192 : Int)) = {
    let x_275 = infixMul_30(b_195, a_192);
    return x_275
  };
```
but will still be passed directly to `hof`:
```effekt-machine
  subst [ev254_254 !-> ev_acc_272, ev255_255 !-> ev_zero_273, x_190 !-> x_274, f_189 !-> bar_196];
  jump hof_19
```
and there be called without this parameter:
```effekt-machine
def hof_191((ev254_254 : Int, ev255_255 : Int, x_190 : Int, f_189 : <function>)) = {
  let ev_zero_269 = Here;
  let ev_acc_270 = ev255_255 + ev_zero_269;
  invoke f_189.0(ev_acc_270, x_190)
}
```

This leads to a ll-error in the llvm backend and a crash in the jit backend.

marzipankaiser:
This is related to #217.

marzipankaiser:
Note: As a workaround, one can always eta-expand the block parameter.

PR #219, opened 2023-01-16
machine: Fix passing of free variables to local functions

### LINKED ISSUE #217: Machine transformation does not correctly close over local state
When using a machine-based backend (llvm, jit), the following code:
```effekt
def main() = { 
    var x = 0
    def next() = {
        x = x + 1;
        x
    }
    println(next())
    println(next())
}
```
(https://gist.github.com/marzipankaiser/4dbeafc1365d0630d8b22e84ce33f592)
will crash (jit: produce a wrong result),

as the label for `next` will not have `x` as a parameter due to the transformation here:
https://github.com/effekt-lang/effekt/blob/c10eaaaf3869a6839b10df61ab9340714cf80ab2/effekt/shared/src/main/scala/effekt/machine/Transformer.scala#L70-L80

(Also, we should not really rely on the symbols types here, anymore..)

marzipankaiser:
Now this uses all free Ids minus block labels, the name of this definition and constructor names,
and also adds the free Ids of called functions.

This also eta-expands functions passed directly as block parameters (since they are labels and thus, this will never work).

Additionally, adds a minor bugfix to `transform(BlockType)`.

Resolves #217.


marzipankaiser:
Marked as draft because this still does not work for:
```effekt
def sumPaths(): Int = {
  var state = 0

  def explore(): Int = state
  def explore2(): Int = explore()

  explore2()
}

def main() =  println(sumPaths())
```

marzipankaiser:
Since both the problem and the solution for #218 are closely related, I will fix that in this PR, too.

marzipankaiser:
Resolves #218

marzipankaiser:
@phischu This is what we talked about back then. Since it is now rebased and working, I will just merge it.

PR #220, opened 2023-01-17
Add support for Holes to machine and llvm

marzipankaiser:
When executed, a hole will print a generic error message and exit with exit code 1.


b-studios:
Can we also desugar holes into a primitive call instead of a separate syntax object?

marzipankaiser:
You mean in machine already? Or even earlier?

b-studios:
Right now I was thinking of Machine; but maybe even as early as core?

marzipankaiser:
Then, we would have to somehow get at the symbol of the builtin. 🤔
Also, this might make it harder to implement some backend-specific behaviours for holes (e.g. if we would ever compile to a target that has support for holes, it would be nice if we could just use that).

I don't know what's the better solution, I think that - for now - this is more consistent with the other parts (core, lifted), though.

b-studios:
Ok, I don't have a strong preference since it is easily changed.

marzipankaiser:
Requested review mostly fyi. Will now merge.

PR #221, opened 2023-01-27
Convert filename of location of node to URI

mm0821:
LSP-Locations should be provided with a URI as source, however in the current implementation of `locationOfNode` the source name of the node is not necessarily a URI but can also be a (relative) path without scheme. This causes the "textDocument/definition"-action to fail in neovim if the definition is not in the current file, as the file the definition is in cannot be opened then.

This PR contains simple solution that correctly yields a URI and works for all my test cases. But someone with a VSCode installation should probably check whether the VSCode-plugin still works fine with this change.

Moreover, I'm not perfectly sure if this is the correct place (maybe it should also be available somewhere else?) for the definition of `toURI` and whether this case analysis is sufficient. For me only these cases occurred but I'm not sure if an incomplete relative path like "folder/filename.effekt" without a "./" could occur as well. I hope not, as it does not seem to be clear whether prepending `cwd()` would always work in that case.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Server.scala
+++ b/effekt/jvm/src/main/scala/effekt/Server.scala
@@ -7,7 +7,7 @@ import effekt.source.{ FunDef, Hole, ModuleDecl, Tree }
 import effekt.util.{ PlainMessaging, getOrElseAborting }
 import effekt.util.messages.EffektError
 
-import kiama.util.{ Position, Services, Source }
+import kiama.util.{ Filenames, Position, Services, Source }
 import kiama.output.PrettyPrinterTypes.Document
 
 import org.eclipse.lsp4j.{ Diagnostic, DocumentSymbol, SymbolKind, ExecuteCommandParams }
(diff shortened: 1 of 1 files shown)


effekt/jvm/src/main/scala/effekt/Server.scala
```diff
@@ -107,12 +107,22 @@ trait LSPServer extends kiama.util.Server[Tree, ModuleDecl, EffektConfig, Effekt
   } yield info
 
   // The implementation in kiama.Server does not support file sources
+  def toURI(filename: String): String = {
+    if (filename startsWith "file://")
+      filename
+    else
+      if (filename startsWith "./")
+        "file://" ++ Filenames.cwd() ++ "/" ++ Filenames.dropPrefix(filename, ".")
```
b-studios, line 115: I am not sure using `/` directly here is a good idea. It might be problematic with Windows users.

b-studios: Probably it would also be better to patch directly here: 

https://github.com/effekt-lang/kiama/blob/main/jvm/src/main/scala/kiama/util/Server.scala#L349-L363

b-studios: Hmm.. maybe not, looking at the code of override `locationOfNode` below I don't understand why we have the override.

b-studios: Regarding windows: it shouldn't be an issue if the implementation of the windows client adheres to the LSP:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uri

which sends it according to

https://tools.ietf.org/html/rfc3986

that is:
```
 foo://example.com:8042/over/there?name=ferret#nose
  \_/   \______________/\_________/ \_________/ \__/
   |           |            |            |        |
scheme     authority       path        query   fragment
   |   _____________________|__
  / \ /                        \
  urn:example:animal:ferret:nose
```

mm0821: Yes, I think `/` is used as path separator in URIs independent of any OS, so it should not be an issue here.

Wrt where to patch you certainly know better than me. But the implementation of `locationOfNode` in https://github.com/effekt-lang/kiama/blob/main/jvm/src/main/scala/kiama/util/Server.scala#L349-L363
does not deal with source filenames in nodes directly, hence the override. (At least, that's what I gathered and what the comment says.)

PR #222, opened 2023-01-30
JIT: New calling convention

marzipankaiser:
Prepending parameters/returned values to the stored environment (instead of appending), and requiring the parameters to be in the correct registers already.

@b-studios This is a breaking change in the output format of the jit backend.


marzipankaiser:
LLVM install fails in CI (apparently due to missing keys)

PR #223, opened 2023-02-01
More extensive stdlib for LLVM

marzipankaiser:
Most of the structure and effekt code is reused from the JIT stdlib.


b-studios:
Ideally at some point we start sharing stdlibs... :)

b-studios:
Is there any important reason we cannot merge this? I often need the `io/args` part of it :)

b-studios:
@marzipankaiser could you ping @phischu next week and try to get this merged? That would be great!

marzipankaiser:
It does not break tests and I just resolved the conflicts with master.
This is a draft mostly because some of the implementations are highly non-optimal and/or hacky.

phischu:
I've added tests but they fail with
```
[error] No unique declaration for Tuple2. Options:
```
and I can't find the reason right now.

marzipankaiser:
The problem seems to be the caused by the destructuring `val` in `sortBy` (though rewriting it to a `match` doesn't help):
```effekt
// this implementation is not stacksafe!
def sortBy[A](l: List[A]) { compare: (A, A) => Boolean }: List[A] =
  l match {
    case Nil() => Nil()
    case Cons(pivot, rest) =>
      val (lt, gt) = rest.partition { el => compare(el, pivot) };
      val leftSorted = sortBy(lt) { (a, b) => compare(a, b) }
      val rightSorted = sortBy(gt) { (a, b) => compare(a, b) }
      leftSorted.append(Cons(pivot, rightSorted))
  }
```
(in the `Cons` case, on the lhs of `partition`)

which generates the following core:
```effekt-core
def sortBy[A](l: List[A]){compare} =
  let tmp853 = l;
  def tmp854() =
    Nil[A]();
  def tmp876(pivot: A, rest: List[A]) =
    val tmp857 = partition[A](rest, { (el: A) => 
      compare(el, pivot)
    });
    let tmp858 = tmp857;
    def tmp869(lt: List[A], gt: List[A]) =
      val leftSorted = sortBy[A](lt, { (a: A, b: A) => 
        compare(a, b)
      });
      val rightSorted = sortBy[A](gt, { (a: A, b: A) => 
        compare(a, b)
      });
      append[A](leftSorted, Cons[A](pivot, rightSorted))
    
    tmp858 match {
      case Tuple2 { (tmp870: List[A], tmp871: List[A]) => 
        tmp869(tmp870, tmp871)
      }
    }
  
  tmp853 match {
    case Nil { () => 
      tmp854()
    }
    case Cons { (tmp877: A, tmp878: List[A]) => 
      tmp876(tmp877, tmp878)
    }
  }
```

marzipankaiser:
Ah, found the problem:
`PolymorphismBoxing` assumes that all data/record types are declared (so it does not depend on the symbols). However, in the current LLVM backend it is run *before* aggregating the modules, and errors out since it can't find `Tuple2` (and thus, does not know its structure).

marzipankaiser:
For now, run `PolymorphismBoxing` after `Aggregate` (it took me way too long to notice that `Aggregate` still exists... 😓).

Issue #225, opened 2023-02-01
`lifted.Tree.freeVariables` distinguishes Variables by type


marzipankaiser:
The function `freeVariables` distinguishes Variables that have the same symbol but different type:
https://github.com/effekt-lang/effekt/blob/8365cbd7f24cd1a59f7c9b5a7a21a1c4dc2c7826/effekt/shared/src/main/scala/effekt/lifted/Tree.scala#L112-L177

This is a problem:
1. When types are syntactically different (e.g. alpha-equivalence, capabilities changing along the enclosing scopes)
   - If we want to prevent this, we'd have to rewrite the types in appropriate positions
2. When types are computed incorrectly

marzipankaiser:
2 occurs in `PolymorphismBoxing` e.g. in the following program:
```effekt
effect Increment {
  def increment(): Unit
}
def main() = {
  try {

    def step(): Int = {
        def looper(): Int / Increment = 0
        looper()
    }

    step()
  } with Increment {
      def increment() = resume(())
  }
}
```
because the types in argument position `BlockVar`s are not transformed correctly.

b-studios:
Yes, we should fix both, then.

PR #226, opened 2023-02-05
Refactor/typed lifted

b-studios:


b-studios:
@marzipankaiser it would be very valuable to get a review from you. In particular, I simply duplicated `DeclarationContext`, which definitively needs a better solution.

b-studios:
It looks like my last commit broke the build. However, I think it is only a symptom of something broken in machine. Function types should take one evidence parameter for the function itself -- this is how our translation currently work. Maybe machine assumed that it does not?

marzipankaiser:
> @marzipankaiser it would be very valuable to get a review from you. In particular, I simply duplicated `DeclarationContext`, which definitively needs a better solution.

I don't really see how we can avoid that duplication as long as we have it in the Tree. The Scala types used by `DeclarationContext` in core/lifted are only related by name and structure IIUC. So either we'd need to change that somehow or use reflection/macros.

marzipankaiser:
> It looks like my last commit broke the build. However, I think it is only a symptom of something broken in machine. Function types should take one evidence parameter for the function itself -- this is how our translation currently work. Maybe machine assumed that it does not?
> 
> We get the following error (only introduced by changing the translation of types to add one evidence parameter for the function itself):
> 
> ```
> [info] running effekt.Server --backend llvm examples/llvm/recursive.effekt
> opt: ./out/examples_llvm_recursive.ll:1323:9: error: multiple definition of local value named 'func_121'
>         %func_121 = load %Neg, ptr %func_121p.129
>         ^
> ```
> 
> @marzipankaiser Do you have an idea how this can happen?

Machine probably assumes the types it gets are core types, so the evidence parameters are added twice - once in lift inference and once by machine. At least that would be my first guess.

b-studios:
> Machine probably assumes the types it gets are core types, so the evidence parameters are added twice - once in lift inference and once by machine. At least that would be my first guess.

That's what I guessed, but I also touched all places where `core.ValueType` etc. was used and turned it into `lifted.ValueType`?

b-studios:
@marzipankaiser do you think you could look into the machine/LLVM issue or do you think this is something @phischu needs to do? Is it difficult to check whether this also affects the JIT backend?

marzipankaiser:
> > Machine probably assumes the types it gets are core types, so the evidence parameters are added twice - once in lift inference and once by machine. At least that would be my first guess.
> 
> That's what I guessed, but I also touched all places where `core.ValueType` etc. was used and turned it into `lifted.ValueType`?

Yes, but there are some places where we create evidence parameters from thin air in the machine translation IIRC.

marzipankaiser:
> @marzipankaiser do you think you could look into the machine/LLVM issue or do you think this is something @phischu needs to do? Is it difficult to check whether this also affects the JIT backend?

I am looking into it. If it affects LLVM, it will most likely affect JIT, too - I could try to merge them locally later.

b-studios:
> Yes, but there are some places where we create evidence parameters from thin air in the machine translation IIRC.

Do you have a particular place where you expect this to happen; I could not find it.

marzipankaiser:
Turns out I was wrong: The problem seems to be the calculation of `freeVariables` (maybe related to #225) - e.g. in `closures.effekt`, the parameter `g` of `f` is returned by `freeVariables` on the block, and thus gets passed twice:

https://github.com/effekt-lang/effekt/blob/1c1b4a01377e41c15472934d036886b5fea073a7/effekt/shared/src/main/scala/effekt/machine/Transformer.scala#L70-L82

Here, (the same) `g` occurs both in `params` and in `freeParams`.

marzipankaiser:
I think I found the issue:
Block parameters are not removed correctly from the free variables, as the types don't (never) match up between the use site and the parameter. They differ in the `Id` of `self` (which prints the same but has a different `.id`)

b-studios:
Thanks for the analysis @marzipankaiser. I turned the list of names for evidence parameters into a list of "Unit" (a natural number... / arity) to make sure structural equality still holds.

b-studios:
As discussed in person, I am going ahead and merge this. We need to

1. rebase / merge into the mlton implementation
2. add explicit shift construct to lifted

PR #227, opened 2023-02-06
lifted.freeVariables: Don't overload by type, assert equality

marzipankaiser:
This changes `lifted.freeVariables` to compare variables by `id` only. It, however, asserts equality for the params, too (They should be consistent).

Note: This makes the following example from #225 crash:
```effekt
effect Increment {
  def increment(): Unit
}
def main() = {
  try {

    def step(): Int = {
        def looper(): Int / Increment = 0
        looper()
    }

    step()
  } with Increment {
      def increment() = resume(())
  }
}
```
This is a good thing, as this makes a bug in `PolymorphismBoxing` obvious.


b-studios:
Could you add the example as a test? This way we can avoid reintroducing the bug.

Issue #233, opened 2023-02-16
Proposal: Backend interface


b-studios:
Like every 6 month or so, we redefine how backends are defined :) Here is a new proposal with the following idea:

Instead of abstracting and reusing specific parts of the compilation pipeline, each backend copy & pastes the full pipeline and has full control. This leads to duplication, but also makes it much easier for newcomers to understand the pipeline since it is in one place (and they only work on one backend). Backends would still reuse `Frontend` and `Middleend`, but can add arbitrary phases.

```scala
// TODO, could be backend specific in the future
//  not all of them need to be supported by each backend
enum Stage {
  case Core
  case Lifted
  case Machine
  case Target
}

case class Compilation[E](outputFiles: Map[String, Document], executable: E)

/**
 * ...
 *
 * Every single method can use the context to report errors or throw
 * exceptions.
 *
 * @tparam Executable information of this compilation run, which is passed to
 *                 the corresponding backend runner (e.g. the name of the main file)
 */
trait Backend[Executable] extends Compiler {

  /**
   * Called after running the frontend from editor services.
   *
   * Can be overridden to implement backend specific checks (exclude certain
   * syntactic constructs, etc.)
   */
  def validate(source: Source, mod: Module)(using Context): Unit = ()

  /**
   * Show the IR of this backend for [[source]] after [[stage]].
   * Backends can return [[None]] if they do not support this.
   *
   * Used to show the IR in VSCode. For the JS Backend, also used for
   * separate compilation in the web.
   */
  def prettyIR(source: Source, stage: Stage)(using Context): Option[Document]

  /**
   * Return the backend specific AST for [[source]] after [[stage]].
   * Backends can return [[None]] if they do not support this.
   *
   * The AST will be pretty printed with a generic [[effekt.util.PrettyPrinter]].
   */
  def treeIR(source: Source, stage: Stage)(using Context): Option[Any]

  /**
   * Should compile [[source]] with this backend. Each backend can
   * choose the representation of the executable.
   */
  def compile(source: Source)(using Context): Compilation[Executable]

  /**
   * Should compile [[source]] with this backend, the compilation result should only include
   * the contents of this file, not its dependencies. Only used by the website and implemented
   * by the JS backend. All other backends can return `None`.
   */
  def compileSeparate(source: Source)(using Context): Option[Document] = None
}


/**
 * Interface used by [[Driver]] and [[EffektTests]] to run a compiled program.
 *
[...]

PR #234, opened 2023-02-17
Refactor backend interfaces

b-studios:
Adresses #233


b-studios:
There is still a few things to do:

1. Merge Compiler and BackendCompiler into one trait
2. Rename BackendRunner into Runner
3. Drop generator.Backend
4. Inline the pipelines of the backends into the respective compilers

b-studios:
It is starting to look better. One thing that I still don't like is the `Either[String, Unit]`... Maybe we should throw an exception instead?

PR #235, opened 2023-02-22
Naive state

b-studios:
Here I am experimenting with a different representation of state that is easier to monomorphize.


b-studios:
The PR is starting to look solid. I implemented the new local state for all backends, except LLVM (and machine for that matter).

It would be very valuable to get feedback from @marzipankaiser, @serkm, and @phischu on how to proceed with machine. You don't have to look at the whole PR, the essential change is this in `lifted.Tree`:

```diff
-  case State(id: Id, init: Expr, region: Symbol, ev: Evidence, body: Stmt)
+  // allocates into a (type-monomorphic?) region.
+  // e.g. var x in r = init; body
+  case Alloc(id: Id, init: Expr, region: Id, ev: Evidence, body: Stmt)
+
+  // creates a fresh state handler to model local (backtrackable) state.
+  // e.g. state(init) { (ev){x: Ref} => ... }
+  case Var(init: Expr, body: Block.BlockLit)
```

https://github.com/effekt-lang/effekt/pull/235/files#diff-0555635c10275b9368035de048ce1782d57df07d499fc1671fb799d45586aa0bR125-R132

For now, I changed the test files to use an explicit region instead of local mutable state. 

```diff
-def main() = {
-    var n = 0
+def main() = region r {
+    var n in r = 0
```
https://github.com/effekt-lang/effekt/pull/235/files#diff-52c3c58f95cb33c9fbda0911bef989780e255f229c04410a7f5e94d98320429cL18-L19

but this is only a temporary fix. Any feedback on how to proceed with machine is very welcome!

phischu:
@serkm I have yet another idea for the implementation of local mutable state.

It crucially relies on going back to the "copy-on-share" strategy. This means that `shareStack` immediately deep-copies the stack instead of bumping the reference count. This way every reference is to a stack indeed is always unique. I don't remember why we changed that but there was a reason.

Now when we want to create a local mutable reference we push a frame that contains the value. References are offsets from the bottom of the stack. Getting the content of a reference means finding the correct stack using evidence and then reading the value at the offset. Setting the content of a reference finds the location in the same way and directly writes to it on the stack.

For arenas we could do the same except that the frame contains a reference to the arena and `erase` and `share` and `return` do the correct thing with it.

A corner case is allocating into the global region which would create a reference on the heap.

PR #237, opened 2023-02-23
PolymorphismBoxing: Fix transformation of Operations

marzipankaiser:


marzipankaiser:
This is a minor bugfix, so merging directly.

PR #241, opened 2023-03-08
machine: Val(x,s,Return(x)) -> s

marzipankaiser:
Currently, `machine.Transformer` generates an unnecessary `PushFrame` when given lifted code like
```lifted
val x = s; return x
```

In particular, PolymorphismBoxing generates code of this form sometimes. Other phases might, too (someday).
This is analogous to the special case for Run(Return(...)), and avoids pushing the unnecessary stack frame.


marzipankaiser:
cc @phischu

Issue #244, opened 2023-03-15
Multiple copies of a region on the stack


serkm:
When a resumption is called within itself, it leads to a region being a subregion of its copy.

Here is an example where a resumption is captured in `f` and called within the resumption itself. `resume(10)` calls `f()` which then calls `resume(4)` (and resets `f` to avoid infinite recursion). 

```
effect E {
    def e() : Int
}


def main() : Unit = region r1 {
    var f = fun() { () }
    try region r2 {
      val i = do e()
      var x = 0
      f()
      println(x)
      x = x + i
      println(x)
    } with E {
        def e() = {
          f = fun() {
            f = fun() { () }
            resume(4)
          }
          resume(10)
        }
    }
}
```
After `resume(4)` is called, the stack has the shape `r2 : r2 : r1 : []`.
This code prints
```
0
4
0
10
```
which is what I'd expect, but should it be possible have multiple copies of a region on the stack?

serkm:
Update: I moved the allocation of x before the effect usage and the output is NOT what I expected.
```

effect E {
    def e() : Int
}
def main() : Unit = region r1 {
    var f = fun() { () }
    try region r2 {
      var x = 0
      val i = do e()
      f()
      println(x)
      x = x + i
      println(x)
    } with E {
        def e() = {
          f = fun() {
            f = fun() { () }
            resume(4)
          }
          resume(10)
        }
    }
}
```
Output:
```
0
4
4
14
```
Shouldn't the value of x be 0 when the resumption is called, regardless of other calls of that resumption?
Tested on the js backend and all chez backends.

b-studios:
I tried running the example and on llvm it outputs:

```
[info] running effekt.Server --noexit-on-error --backend llvm issue244.effekt
[error] scala.NotImplementedError: an implementation is missing
[error] 	at scala.Predef$.$qmark$qmark$qmark(Predef.scala:344)
[error] 	at effekt.machine.Transformer$.transform(Transformer.scala:470)
[error] 	at effekt.machine.Transformer$.transform(Transformer.scala:461)
[error] 	at effekt.machine.Transformer$.transform(Transformer.scala:133)
[error] 	at effekt.machine.Transformer$.transform$$anonfun$22(Transformer.scala:276)
[error] 	at effekt.machine.Transformer$.pure$$anonfun$1(Transformer.scala:533)
[error] 	at effekt.machine.Transformer$.transform(Transformer.scala:276)
[...]

PR #248, opened 2023-03-16
Mutable state using evidence

serkm:
Here is an alternative implementation for mutable state in the LLVM backend that should work with #244. It doesn't refer to regions by name and instead uses evidence for allocation and access.

This comes with the cost of not having direct pointers as references anymore. On the other hand it makes memory management a lot simpler.


serkm:
@marzipankaiser I encountered an issue with boxed types while implementing this.
```
var x = 0
x = 1
```
The generated core tree for above statements looks like this:
```
List(Let(tmp124_139, Literal(0, Data(Int_98, Nil)))),
State(
  x_116,
  ValueVar(tmp124_139, Data(Int_98, Nil)),
  this_127,
  App(
    Member(
      BlockVar(
        x_116,
        Interface(Ref_118, List(Data(BoxedInt_82, Nil))),
        Set(this)
      ),
      put_140,
      Function(
        Nil,
        Nil,
        List(Data(Int_98, Nil)),
        Nil,
        Data(Unit_119, Nil)
      )
    ),
    Nil,
    List(Literal(1, Data(Int_98, Nil))),
    Nil
  )
)
```
The var declaration uses `Int_98` and the access uses `BoxedInt_82`. This lead to usage of the wrong arena, because Ints and Objects are allocated in separate arenas.
As a workaround I hardcoded the boxed types for now (see 9bb9c1ac5d0f307e47dff852dc0538bc260227f0)

marzipankaiser:
> @marzipankaiser I encountered an issue with boxed types while implementing this.
> 
> ```
> var x = 0
> x = 1
> ```
> 
> The generated core tree for above statements looks like this:
> 
> ```
> List(Let(tmp124_139, Literal(0, Data(Int_98, Nil)))),
> State(
>   x_116,
>   ValueVar(tmp124_139, Data(Int_98, Nil)),
>   this_127,
>   App(
>     Member(
>       BlockVar(
>         x_116,
>         Interface(Ref_118, List(Data(BoxedInt_82, Nil))),
>         Set(this)
>       ),
>       put_140,
>       Function(
>         Nil,
>         Nil,
>         List(Data(Int_98, Nil)),
>         Nil,
>         Data(Unit_119, Nil)
>       )
>     ),
>     Nil,
>     List(Literal(1, Data(Int_98, Nil))),
>     Nil
>   )
> )
> ```
> 
> The var declaration uses `Int_98` and the access uses `BoxedInt_82`. This lead to usage of the wrong arena, because Ints and Objects are allocated in separate arenas. As a workaround I hardcoded the boxed types for now (see [9bb9c1a](https://github.com/effekt-lang/effekt/commit/9bb9c1ac5d0f307e47dff852dc0538bc260227f0))

Ah, sorry, I completely forgot about answering here. In principle, it should be easy enough to fix this in `PolymorphismBoxing`, depending on what's the correct behavior, i.e.: should Ref[Int] be special-cased and allowed? Or should state always be boxed?

serkm:
I special-cased references for now.
@phischu I also added support for nested handlers here.

effekt/shared/src/main/scala/effekt/machine/PrettyPrinter.scala
```diff
@@ -65,23 +64,23 @@ object PrettyPrinter extends ParenPrettyPrinter {
...
 
-    case Load(name, ref, rest) =>
-      name <+> "=" <+> "*" <> ref <> ";" <> line <> toDoc(rest)
+    case Load(name, ref, ev, rest) =>
+      name <+> "=" <+> "*" <> ref <> "<" <> ev <> ">" <> ";" <> line <> toDoc(rest)
 
-    case Store(ref, value, rest) =>
-      "*" <> ref <+> "=" <+> value <> ";" <> line <> toDoc(rest)
+    case Store(ref, value, ev, rest) =>
+      "*" <> ref <> "<" <> ev <> ">" <+> "=" <+> value <> ";" <> line <> toDoc(rest)
```
phischu, line 74: I would have used keywords like `alloc` `load` `store` but this is fine.

libraries/llvm/rts.ll
```diff
@@ -56,37 +56,21 @@
...
+; - Values (Integer, Double)
+; - Objects (Positive, Negative)
+; - Strings
+; For each group we have an arena where mutable state is allocated.
 ;
-;   +[ StateHeader ]+--------------+
-;   | Next  | Size  | Payload ...  |
-;   +---------------+--------------+
-%State = type ptr
+%Region = type [ 3 x %Mem ]
```
phischu, line 65: I don't quite like this solution and would prefer storing a sharer and an eraser with each object. But ok.

libraries/llvm/rts.ll
```diff
@@ -233,29 +217,89 @@ define void @eraseNegative(%Neg %val) alwaysinline {
...
-    %statenext = getelementptr %StateHeader, %State %state, i64 0, i32 0
-    %statesize = getelementptr %StateHeader, %State %state, i64 0, i32 1
+    %ret.0 = insertvalue { ptr, %Ref } undef, %Sp %sp, 0
+    %ret.1 = insertvalue { ptr, %Ref } %ret.0, %Ref %offset, 1
+    ret { ptr, %Ref } %ret.1
 
-    %regionval = load %RegionVal, %Region %region
-    %oldstate = extractvalue %RegionVal %regionval, 0
-    %oldsize = extractvalue %RegionVal %regionval, 1
+realloc:
```
phischu, line 273: Later we could perhaps move this reallocation into its own function.

libraries/llvm/rts.ll
```diff
@@ -233,29 +217,89 @@ define void @eraseNegative(%Neg %val) alwaysinline {
...
+
+    %ret..0 = insertvalue { ptr, %Ref } undef, %Sp %newsp, 0
+    %ret..1 = insertvalue { ptr, %Ref } %ret..0, %Ref %arenasize, 1
+    ret { ptr, %Ref } %ret..1
 
-    %ptr = getelementptr i8, ptr %state, i64 16
-    ret %Ref %ptr
 }
 
+define ptr @getPtr(%Ref %ref, i64 %idx, i64 %evidence) alwaysinline {
```
phischu, line 296: The `%idx` is for the 3 different kinds of values, right?

libraries/llvm/rts.ll
```diff
@@ -440,55 +410,95 @@ copy:
...
-    %regionsize = extractvalue %RegionVal %regionval, 1
-    %backupsize = add i64 %regionsize, 16
-    %backupmem = call ptr @malloc(i64 %backupsize)
-    store %RegionVal %regionval, %RegionBackup %backupmem
-    %backupdata = getelementptr i8, ptr %backupmem, i64 16
-    call void @backupRegion(%State %regionstate, ptr %backupdata)
-    ret %Stk %newstk
+    %objectsbase = extractvalue %Region %region, 1, 1
+    %objectssp = extractvalue %Region %region, 1, 0
+    call void @forEachObject(%Base %objectsbase, %Sp %objectssp, %Eraser @sharePositive)
```
phischu, line 428: Nice! Higher-order functions in LLVM.

libraries/llvm/rts.ll
```diff
@@ -440,55 +410,95 @@ copy:
...
+
+    %last = icmp eq %Stk %rest, %stk
+    br i1 %last, label %closecycle, label %next
+
+next:
+    %nextnew = call ptr @malloc(i64 112)
+    store %Stk %nextnew, ptr %newstkrest
+    br label %loop
+
+closecycle:
```
phischu, line 498: I forgot why the meta stack is cyclic...

phischu (approved):
Very good job!

libraries/llvm/rts.ll
serkm: The meta-stack isn't cyclic but first-class stacks are.
This representation saves us a few operations on `PushStack`/`PopStack`, because we only have to swap the heads of the meta-stack and the first-class stack.

libraries/llvm/rts.ll
serkm: Yes, perhaps it should be renamed to `%kind` or something similar.

libraries/llvm/rts.ll
serkm: This would become obsolete if we switch to monomorphic regions.

PR #250, opened 2023-03-28
Fast Integer Maps

DaCrizzle:


b-studios:
Thanks a lot for your contribution, but we cannot merge it in the current state.

PR #253, opened 2023-04-03
Deprecate multiple value argument sections

b-studios:
AFAIK we never use this feature anymore and it just leads to confusion, since by flattening `foo()()()`==`foo()`.


jiribenes (approved):
LGTM

b-studios:
This change would be more general as it would allow `foo()()`:

```
  lazy val callTarget: P[CallTarget] =
    ( funCall ^^ ExprTarget.apply // allows f()() as shorthand for (unbox f())()
    | `(` ~> expr <~ `)` ^^ ExprTarget.apply
    | idTarget
    )

  lazy val idTarget: P[CallTarget] =
    idRef ^^ IdTarget.apply

  lazy val unboxExpr: P[Term] = `unbox` ~/> expr ^^ Unbox.apply

  lazy val newExpr: P[Term] = `new` ~/> implementation ^^ New.apply

  lazy val funCall: P[Term] =
    ( callTarget ~ maybeTypeArgs ~ valueArgs ~ blockArgs ^^ Call.apply
    | callTarget ~ maybeTypeArgs ~ valueArgs ~ success(List.empty[Term]) ^^ Call.apply
    | callTarget ~ maybeTypeArgs ~ success(List.empty[Term]) ~ blockArgs ^^ Call.apply
    )
```

However, we still need to address a problem with left-recursion here.

Issue #239, opened 2023-02-27
Unhelpful error messages when misusing block arguments


jiribenes:
## Motivation

I wanted to define a function `foo` which takes a function `f` and handles its exceptions:
<img width="410" alt="Screenshot 2023-02-27 at 14 17 42" src="https://user-images.githubusercontent.com/11269173/221573932-66847744-57ce-4323-995b-b69322251138.png">

So I tried removing the `/ Exc` part and got an even weirder error message:
<img width="393" alt="Screenshot 2023-02-27 at 14 17 57" src="https://user-images.githubusercontent.com/11269173/221573930-e22eec24-cb63-4337-8ac8-2362c506a38f.png">

Then I tried only keeping the `Exc` as an argument which worked fine but I didn't understand why:
<img width="261" alt="Screenshot 2023-02-27 at 14 20 51" src="https://user-images.githubusercontent.com/11269173/221574349-3bee4587-83eb-41a9-9db4-e9cb8b826ccf.png">

Only afterwards I realised that the block argument needs to be a function (block) type... 🤦
<img width="393" alt="Screenshot 2023-02-27 at 14 21 54" src="https://user-images.githubusercontent.com/11269173/221574576-9e04a2f2-31ed-4b06-b0d4-5752194115c6.png">

### Relevant code

```
effect Exc(msg: String): Unit

def foo[R] { f : R / Exc }: R = <>
def bar[R] { g : R }: R = <>
def baz[R] { h : Exc }: R = <>

def whatIWanted[R] { f : () => R / Exc }: R = <>
```

## Summary

In the case of `foo` and `bar` above, provide a clearer error message to signal that the compiler (parser/typer) expects block types to be actual function types.

Issue #161, opened 2022-09-28
Resuming an object operation produces confusing error message


jiribenes:
I was playing around with objects and managed to write `resume( ... )` by accident.
This however produced a very confusing error message about not being able to infer the type of the continuation.

I think that we should be able to special-case this and produce a more relevant message to the user.

<img width="859" alt="Screenshot 2022-09-27 at 17 11 01" src="https://user-images.githubusercontent.com/11269173/192715558-e01e2f95-8441-43cc-b8cc-dd57a030cac8.png">

I believe this is the relevant line of code throwing the error message -- it might be helpful to check `continuationDetails` 
https://github.com/effekt-lang/effekt/blob/471f9da2380dac9cd2d057d12cad60496cc20bb3/effekt/shared/src/main/scala/effekt/Typer.scala#L317

b-studios:
+1 This is a wellformedness check that should report something useful like "resuming is only supported in effect handlers, not when defining objects."

Issue #146, opened 2022-09-22
Explicitly handling the same effect twice should be fine


b-studios:
```
effect Exc(): Unit
def main() =
  try { e1.Exc() } 
  with e1: Exc { () }
  with e2: Exc { () }
```
triggers:

![image](https://user-images.githubusercontent.com/408265/191777536-cf412e6e-fd17-42e4-8f40-289b34f42ba8.png)

However, it should be ok since they are bound explicitly. I do not know what the implicit semantics should be, though.

Issue #145, opened 2022-09-22
Explicitly binding capabilities triggers spurious warning 


b-studios:
```effekt
effect Exc(): Unit
def main() =
  try { e.Exc() } with e: Exc { () }
```
![image](https://user-images.githubusercontent.com/408265/191777056-43964eda-471d-4ef0-beb0-2bb11507e6e4.png)

Expected behavior: No warning.

jiribenes:
Today I've independently ran into this issue as well.
<img width="363" alt="Screenshot 2022-09-23 at 10 36 51" src="https://user-images.githubusercontent.com/11269173/191922502-6d43d0f3-58c4-4320-8087-6af3beb013f1.png">

b-studios:
Yes, this is guaranteed to occur every single time, since we do not consider bound capabilities as "effect usage" at the moment. :)

PR #255, opened 2023-04-07
Improve errors

b-studios:
In this PR I try to address a few issues that are all concerned with error messages and warnings issued by the Effekt compiler. In particular, these are:

- #239 
- #161
- #256
- #145
- #146

A few of them should be solved with this PR, a few others are solved "for now" (while there are better solutions).


b-studios:
Commit 3da0288 also address #145

b-studios:
@jiribenes Since many of the above issues are authored by you, it would be valuable if you could review this PR.

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -745,6 +746,14 @@ class EffektLexers(positions: Positions) extends Parsers(positions) {
...
     rep1sep(p, sep) ^^ { _.toList }
 
+  extension [T] (p: Parser[T]) {
+    def !!(errorMessage: T => String): Parser[Nothing] =
+      p.flatMap(t => error(errorMessage(t)))
+
+    def !!!(errorMessage: String): Parser[Nothing] =
+      p.flatMap(_ => error(errorMessage))
+  }
+
```
jiribenes, line 756: Thanks for these utility functions, they will be really helpful for improving error messages from the parser in the future! :)

jiribenes (approved):
Thanks a lot for these improvements! 👍

PR #267, opened 2023-05-14
Optimize Core

b-studios:
This PR builds on the work by @fritzladwig and supersedes #206.

It performs (in this order)

1. Deadcode elimination
2. Inlines uniquely used blocks (functions and objects).
3. Normalizes a bit (`return(run(e)) = e`, and so on)

It does not implement a static argument transformation. We need to reimplement that later at some point. Ideally on a CPS representation. We should also look into the standard shrinking techniques (https://www.cs.princeton.edu/~appel/papers/shrink.pdf) at that point.

It turned out that the simplifications were "too good" and loops became fully direct style (test: raytracer.effekt). In order to work around this issue in the JS backend, I introduced a new microphase that patches up recursive (non-guarded) function calls by thunking them under a trivial `val _ = return (); recursive()`. In the JS implementation this will cause a frame to be pushed on the monadic trampoline, which effectively makes recursive calls stack safe.


b-studios:
Thanks @fritzladwig for your work again! It might look like I re-implemented everything, but that was much easier after you having implemented it already once.

b-studios:
@CodiumAI-Agent /review

PR #272, opened 2023-06-20
Multiple Return Values

denhie:


effekt/jvm/src/main/scala/effekt/EffektConfig.scala
```diff
@@ -45,6 +45,14 @@ class EffektConfig(args: Seq[String]) extends REPLConfig(args) {
     noshort = true
   ).map(Backend.backend)
 
+  val no_optimize: ScallopOption[String] = choice(
+    choices = List("none", "all", "inlining", "sat", "beta", "dead"),
+    name = "no_optimize",
+    descr = "Optimizations that should not be used",
+    default = Some("none"),
+    noshort = true
+  )
```
b-studios, line 54: This looks like a left over from the optimization PR

effekt/jvm/src/main/scala/effekt/Repl.scala
```diff
@@ -108,7 +108,7 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
       runFrontend(StringSource(""), module.make(UnitLit()), config) { cu =>
         module.definitions.foreach {
           case u: Def =>
-            outputCode(DeclPrinter(context.symbolOf(u)), config)
+            outputCode(DeclPrinter(context.symbolsOf(u)), config)
```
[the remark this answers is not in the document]
b-studios: My intuition tells me that there are invariants which are not represented here. This will lead to *a lot* of code changes that might not be necessary.

effekt/jvm/src/main/scala/effekt/Repl.scala
```diff
@@ -199,14 +199,14 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
...
-            outputCode(DeclPrinter(sym), config)
+            outputCode(DeclPrinter(List(sym)), config)
         }
         output.emitln(s"\nImported Functions\n==================")
         cu.terms.toList.sortBy { case (n, _) => n }.foreach {
           case (name, syms) =>
             syms.collect {
               case sym if !sym.isSynthetic =>
-                outputCode(DeclPrinter(sym), config)
+                outputCode(DeclPrinter(List(sym)), config)
```
b-studios, line 209: I don't like that DeclPrinter now takes a list... This feels wrong.

effekt/jvm/src/main/scala/effekt/Repl.scala
```diff
@@ -215,21 +215,22 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
     case d: Def =>
       val extendedDefs = module + d
       val decl = extendedDefs.make(UnitLit())
+      val output = config.output()
```
b-studios, line 218: This is unused

effekt/jvm/src/main/scala/effekt/Repl.scala
```diff
@@ -215,21 +215,22 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
...
 
       runFrontend(source, decl, config) { cu =>
         module = extendedDefs
 
         // try to find the symbol for the def to print the type
-        (context.symbolOption(d.id) match {
+        d.ids.foreach(id => (context.symbolOption(id) match {
```
b-studios, line 224: I believe this is not the correct implementation. It should print
```
x:Int, y:String, z:Bool
```
not
```
x: Int
y: String
z: Bool
```

or what do you think?

effekt/jvm/src/main/scala/effekt/Server.scala
```diff
@@ -3,14 +3,13 @@ package effekt
 import effekt.context.Context
 import effekt.core.PrettyPrinter
 import effekt.lifted.LiftInference
-import effekt.source.{ FunDef, Hole, ModuleDecl, Tree }
-import effekt.util.{ PlainMessaging, getOrElseAborting }
+import effekt.source.{FunDef, Hole, ModuleDecl, Tree}
+import effekt.util.{PlainMessaging, getOrElseAborting}
```
b-studios, line 7: Please try to keep this changes out of the PR or change them upstream, so that the diff is smaller :)

effekt/jvm/src/main/scala/effekt/Server.scala
```diff
@@ -231,10 +234,11 @@ trait LSPServer extends kiama.util.Server[Tree, EffektConfig, EffektError] with
     }
   } yield res
 
-  def needsUpdate(annotated: (ValueType, Effects), inferred: (ValueType, Effects))(using Context): Boolean = {
+  def needsUpdate(annotated: (List[ValueType], Effects), inferred: (List[ValueType], Effects))(using Context): Boolean = {
     val (tpe1, effs1) = annotated
```
b-studios, line 238: This should be called `tpes1` since they are now multiple.

effekt/jvm/src/main/scala/effekt/Server.scala
```diff
@@ -231,10 +234,11 @@ trait LSPServer extends kiama.util.Server[Tree, EffektConfig, EffektError] with
...
     val (tpe2, effs2) = inferred
-    tpe1 != tpe2 || effs1 != effs2
+
+    tpe1.size != tpe2.size || tpe1.zip(tpe2).forall { case (t1, t2) => t1 != t2 } || effs1 != effs2
```
b-studios, line 241: please don't write huge boolean expressions like this.

```suggestion
    val differentArity = tpes1.size != tpes2.size 
    val differentEffects = effs1 != effs2 
    def differentTypes = (tpes1 zip tpes2).forall { case (tpe1, tpe2) => tpe1 != tpe2 }
    
    differentArity || differentEffects || differentTypes
```

effekt/jvm/src/main/scala/effekt/Repl.scala
phischu: No, this is the correct behavior in this case. It can still be refactored, though.

effekt/jvm/src/main/scala/effekt/Repl.scala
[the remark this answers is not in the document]
denhie: Yes, a definition now has multiple symbols.

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -520,13 +528,13 @@ object Typer extends Phase[NameResolved, Typechecked] {
...
       bindings
-  } match { case res => Context.annotateInferredType(pattern, sc); res }
+  } match { case res => Context.annotateInferredType(pattern, List(sc)); res } // TODO MRV: List(sc)
 
   //</editor-fold>
 
   //<editor-fold desc="statements and definitions">
 
-  def checkStmt(stmt: Stmt, expected: Option[ValueType])(using Context, Captures): Result[ValueType] =
+  def checkStmt(stmt: Stmt, expected: Option[List[ValueType]])(using Context, Captures): Result[List[ValueType]] =
```
b-studios, line 537: I think this should be a list of option since each individual component might be known or not.

PR #275, opened 2023-07-18
Multiple operations

serkm:
This adds support for effects and interfaces with more than one operation in the LLVM backend.


phischu (approved):
Beautiful!

PR #277, opened 2023-08-10
Bugfix at toInt

serkm:


b-studios:
Probably the repeated calls to `unsafeCharAt(s, i)` should be "ANFed" as well.

Issue #269, opened 2023-05-18
Numbers higher than INT32_MAX crash the parser


marvinborner:
The parser crashes with a `NumberFormatException` [here](https://github.com/effekt-lang/effekt/blob/c9dd22879f7818f14584a538b27adf6811a979bf/effekt/shared/src/main/scala/effekt/Parser.scala#L476) when numbers higher than $2^{31}-1$ get used:

```
def main() =
  println(2147483648)
```

Interestingly enough, the same error occurs for doubles (e.g. 2147483648.42) -- which is weird because they should be able to store higher numbers and get tried by the parser first(?).

I'm unsure what the correct action should be, probably a "Not implemented" error or something for now?

b-studios:
Thanks! At the very least we could catch this exception and report a parse error like: "Not a 32bit integer literal". 
We should think about changing the representation of `Int` to `Long` internally, for the very least.
Doubles is indeed very confusing.

b-studios:
I keep this issue open to remind us to change the representation from `Int` to `Long`. But then, we also need to change the error message again :)

PR #278, opened 2023-08-16
Fix/32bit integer literals

b-studios:
Adresses #269 by throwing a better error.


b-studios:
Eventually, we should use Long internally. Almost all backends use 64bit integers anyway.

PR #279, opened 2023-08-17
Support 64bit integer literals

### LINKED ISSUE #269: Numbers higher than INT32_MAX crash the parser

b-studios:
This PR adds support for 64bit integer literals by using `Long` instead of `Int` (in Scala) to represent literals.

This closes #269.


b-studios:
Note that the JavaScript backend will (as before) use the generic number type, so it will not have enough precision to accurately store full 64bit integers.

Issue #202, opened 2022-12-01
Lexer doesn't recognise escaped double quotes in strings


jiribenes:
I was just trying to use double quotes in a string 😄 

<img width="690" alt="Screenshot 2022-12-01 at 11 04 18" src="https://user-images.githubusercontent.com/11269173/205024944-0a9c5a41-ee2a-4bd6-8bb6-5e9cd33debf8.png">

This is not only a VSCode issue, since the message originates in the compiler:
<img width="626" alt="Screenshot 2022-12-01 at 11 06 55" src="https://user-images.githubusercontent.com/11269173/205025206-95fac05d-8755-4c4b-8117-d7c493e2302b.png">

Here's the regex in question:
https://github.com/effekt-lang/effekt/blob/18c94138ec3829b793095865c562d1ba74992994/effekt/shared/src/main/scala/effekt/Parser.scala#L208

jiribenes:
Proposed regular expression:
```perl
\"(\\.|[^\"])*\"
```
- First, match `"`
- Then, zero or more times:
    - **EITHER** match `\` followed by _anything_
    - **OR** match something else than `"`
- Finally, match `"`

b-studios:
I would be fine with merging this simple fix. Eventually, we should actually design strings and escape sequences at some point.

PR #290, opened 2023-10-13
Lexer: match escape sequences in strings

dvdvgt:
This PR addresses and fixes issue #202.

Of course, I am open to suggestions on how to properly integrate escape sequences into the lexing process as proposed by @b-studios (https://github.com/effekt-lang/effekt/issues/202#issuecomment-1500515704):

> [...] we should actually design strings and escape sequences at some point. 

What changes do you think this entails?

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Parser.scala
+++ b/effekt/shared/src/main/scala/effekt/Parser.scala
@@ -730,7 +730,7 @@ class EffektLexers(positions: Positions) extends Parsers(positions) {
    */
   lazy val integerLiteral = "([-+])?(0|[1-9][0-9]*)".r
   lazy val doubleLiteral = "([-+])?(0|[1-9][0-9]*)[.]([0-9]+)".r
-  lazy val stringLiteral = """\"([^\"]*)\"""".r
+  lazy val stringLiteral = """\"(\\.|[^\"])*\"""".r
 
   // Delimiter for multiline strings
   val multi = "\"\"\""


b-studios:
> What changes do you think this entails?

Most languages say in their spec what a String is and what an escape sequence is (for instance here for JS: https://tc39.es/ecma262/multipage/text-processing.html#prod-CharacterEscape).

dvdvgt:
So we would need to decide on a grammar for strings and change the lexer accordingly? For example
```
<string> ::= " <string2>* "
<string2> ::= <escaped> | <char>
<escaped> ::= \" | \n | \t | \r | \v | \f
```

Currently, a string literal is tokenized as a whole. If we want to have better errors regarding invalid control characters/escape sequences, each of these will be probably need to be its own token. However, the lexer is currently skipping all whitespaces, making it impossible to correctly tokenize a string containing such whitespaces (`"a b"` would be tokenized as `"`, `a`, `b`). Therefore, we either need to accept the updated regex for string literals or change lexing such that whitespaces are not ignored. Or am I perhaps missing something?

What are your thoughts on this?

b-studios:
I know too little about string escapes, their use cases, and how they would be transported in the compiler to the different backends; so I don't have an informed opinion. The grammar you suggest makes somewhat sense but I don;t know what the `v` and `f` options are supposed to be.

dvdvgt:
> I know too little about string escapes, their use cases, and how they would be transported in the compiler to the different backends; so I don't have an informed opinion. The grammar you suggest makes somewhat sense but I don;t know what the `v` and `f` options are supposed to be.

They are just taken from the JS spec sheet you posted. You can read more about them [here](https://en.wikipedia.org/wiki/Control_character#In_ASCII).

jiribenes:
On the meeting, we talked about possible designs of escape sequences in strings and settled on similar design as Zig:
https://ziglang.org/documentation/master/#Escape-Sequences

image

dvdvgt:
What do you mean by align? 

The given grammar is not compatible across all backends. For example, `"\u039e"` is displayed as `Ξ` in node, yet is rejected by the MLton compiler: `String constant with character too large for type: #"\u039E"`.

b-studios:
No, I just meant that ideally strings would potentially work the same on all backends. But it is fine for now (TM)

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -730,7 +730,7 @@ class EffektLexers(positions: Positions) extends Parsers(positions) {
    */
   lazy val integerLiteral = "([-+])?(0|[1-9][0-9]*)".r
   lazy val doubleLiteral = "([-+])?(0|[1-9][0-9]*)[.]([0-9]+)".r
-  lazy val stringLiteral = """\"([^\"]*)\"""".r
+  lazy val stringLiteral = """\"(\\.|[^\"])*\"""".r
```
b-studios, line 733:
```suggestion
  lazy val stringLiteral = """\"(\\[ntr'\"\\]|[^\"])*\"""".r
```

b-studios (approved):

b-studios:
Eventually we can write a custom parser like `;` to actually lex it.

PR #292, opened 2023-10-15
early for loop implementation

IR0NSIGHT:


b-studios (commented):
Hey, thank  you for your PR. Two questions:

1. how is `forN` not covered by `repeat` in the JS standard library? (https://github.com/effekt-lang/effekt/blob/master/libraries/js/effekt.effekt#L232)
2. what is the motivation for adding this test? Does it currently fail?

IR0NSIGHT:
> Hey, thank you for your PR. Two questions:
> 
>     1. how is `forN` not covered by `repeat` in the JS standard library? (https://github.com/effekt-lang/effekt/blob/master/libraries/js/effekt.effekt#L232)
> 
>     2. what is the motivation for adding this test? Does it currently fail?
> 
> 
> Right now it cannot succeed since a check file is missing.

actually, i wasnt aware of repeat, im not sure yet if repeat (or forN) do what i want, ill know more on wednesday.
The PR is still a draft, as im still figuring out how to work with effekt :)

Ill set it to ready-for-review once its worked out 👍

IR0NSIGHT:
each(0, array.size()) already does what forN(array.size()) was intended to do.
PR is obsolete.

Issue #273, opened 2023-07-11
There should be an option to produce an executable without running it


phischu:
Currently invoking `effekt main.effekt` will compile and run the given file and `effekt --compile main.effekt` will just generate code and then stop.

For the llvm backend (and also the mlton backend) it would be great if it was possible to also invoke the llvm compiler (respectively mlton compiler) to produce an executable but then not run it.

b-studios:
I agree, `--compile` should do exactly that.

b-studios:
Duplicate of #254

Issue #254, opened 2023-04-05
`--compile` should generate executable


marzipankaiser:
Currently `--compile` just does not execute the generated code.
Instead, it should also generate some kind of executable at a known location
(e.g. by adding a shebang to the output, wrapping it, ...).

dvdvgt:
I am afraid I don't fully grasp this issue.

- What kind of executable do you have in mind? How does this differ between the different backends, that is, does this only amount to changing the shebang, if we would choose to go into this direction or is there something else to take into consideration?
- Compiling it to a "known location" is already available via the `--out` option, no?
- What platforms do we want to target with this? Only UNIX systems or also Windows (shebangs won't work there)?

Can you perhaps eleborate a bit?

b-studios:
Each Backend will need to decide what it constitutes to compile. Right now we already have abstractions in place where each backend can say how it compiles Effekt code. This should be extended to configure what `--compile` means.

dvdvgt:
I think it might be good to have both options

1. `--compile`: Compile an Effekt source file to an executable.
2. `--source` (not exactly sure about the name yet): Transform an Effekt source file into the source language of the respective backend. This might be nice to have to manually inspect the generated code.

This means we would keep the current functionality of `--compile` and only rename it to `--source`. For the "new" `--compile` we take the output of `--source` and pass through some backend specific commands. For that, it would be helpful if you could let me know what this usually involves. That is, which backend specific commands do you typically manually invoke?

Something else to take into consideration is the IO involved here to call the backend specific compiler as a `Process`. Since the compiler seems to be prohibited from using IO, I would value your input here to which degree calling the backend specific compiler would break the invariant mentioned here:

https://github.com/effekt-lang/effekt/blob/523e11bea6c862935c86e84f8660a4d88574b95d/effekt/shared/src/main/scala/effekt/Compiler.scala#L84-L90

b-studios:
Maybe we can keep `--compile` and call the new one `--build` ? @marzipankaiser @phischu any opinions?

dvdvgt:
Resolved by PR #294.

PR #294, opened 2023-10-18
Rework `--compile` and add a seperate `--build` option

dvdvgt:
This issue addresses #254 (and #273). 

# Changes

In the `Runner` we now have two methods, `build` just for building an executable and `eval` for then evaluating (running) said executable. Furthermore, we now have one new flag in addition to the `--compile` flag, which also has been reworked.
- The `--build` flag builds executables when selecting certain backends (LLVM and MLton) instead of only outputting the transpiled code. This option overwrites the `--compile` option, as the transpiled code is needed for building.
- The `--compile` flag only outputs (saves) the transpiled code.

# Discussion

I am not quite sure how to handle the ChezScheme and JS backend in terms of "building" an executable. Prepending a shebang (`#!/usr/bin/env node ...`) only works on UNIX systems. Furthermore, this would require more drastic changes to the current implementation, as I would need to edit the pretty-printed output of `Compiler.compile` in the respective runner, which is currently not available in the `Runner`.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Runner.scala
+++ b/effekt/jvm/src/main/scala/effekt/Runner.scala
@@ -52,9 +52,10 @@ trait Runner[Executable] {
   def build(executable: Executable)(using Context): String
 
   /**
-   * Runs the executable (e.g. the main file).
+   * Runs the executable (e.g. the main file) by calling the build function.
    */
-  def eval(executable: Executable)(using Context): Unit
+  def eval(executable: Executable)(using Context): Unit =
+    exec(build(executable))
 
   def canRunExecutable(command: String*): Boolean =
     try {
@@ -84,25 +85,6 @@ trait Runner[Executable] {
     }
     go(progs0)
   }
-
-  /**
-   * Create a executable file with at the given path with the given content. The file will have
-   * the permissions UNIX permissions 744.
-   */
-  def createExecutableFile(path: String, content: String)(using C: Context): Unit = {
-    import java.nio.file.{Files, Path}
-    import java.nio.file.attribute.PosixFilePermission.*
-    import scala.jdk.CollectionConverters.SetHasAsJava
-
-    val path1 = Path.of(path)
-    val perms = Set(
-      OWNER_READ, OWNER_WRITE, OWNER_EXECUTE,
-      GROUP_READ,
-      OTHERS_READ
-    )
-    IO.createFile(path, content)
-    Files.setPosixFilePermissions(path1, SetHasAsJava(perms).asJava)
-  }
 }
 
 object JSRunner extends Runner[String] {
@@ -118,18 +100,19 @@ object JSRunner extends Runner[String] {
     if canRunExecutable("node", "--version") then Right(())
     else Left("Cannot find nodejs. This is required to use the JavaScript backend.")
 
+  /**
+   * Creates an executable `.js` file besides the given `.js` file ([[path]])
+   * and then returns the absolute path of the created executable.
+   */
   def build(path: String)(using C: Context): String =
-    val out = C.config.outputPath().getAbsolutePath.stripSuffix(".")
-    val jsFilePath = (out / path.stripPrefix(".")).unixPath
+    val out = C.config.outputPath().getAbsolutePath
+    val jsFilePath = (out / path).unixPath
     // create "executable" using shebang besides the .js file
     val jsScriptFilePath = jsFilePath.stripSuffix(s".$extension")
     val jsScript = s"require('${jsFilePath}').main().run()"
     val shebang = "#!/usr/bin/env node"
-    createExecutableFile(jsScriptFilePath, s"$shebang\n$jsScript")
+    IO.createFile(jsScriptFilePath, s"$shebang\n$jsScript", true)
     jsScriptFilePath
-
-  def eval(path: String)(using C: Context): Unit =
-    exec(build(path))
 }
 
 trait ChezRunner extends Runner[String] {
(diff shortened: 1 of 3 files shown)


b-studios:
Thanks, as for the discussion: I am not sure whether you would need to edit the pretty printed output. It should suffice to prefix it. I think for the JS backend, it would be enough to output a script next to the compiled result (maybe called like a binary would) that contains a shebang and the `require(...).main().run()` boilerplate.

For Chez, there is an option to create a binary that does WPO and bundles the interpreter. Maybe this one helps: https://github.com/gwatt/chez-exe/blob/master/compile-chez-program.ss

dvdvgt:
> However, I think for the JS backend, it would be enough to output a script next to the compiled result (maybe called like a binary would) that contains a shebang and the require(...).main().run() boilerplate.

Good idea! I will do just that.

> For Chez, there is an option to create a binary that does WPO and bundles the interpreter. Maybe this one helps: https://github.com/gwatt/chez-exe/blob/master/compile-chez-program.ss

It seems to be using the `compile-whole-program` function which does not output a binary but some object file which can be executed by the scheme interpreter just like an `.ss` file. The actual process of turning that into a standalone executable is a bit more elaborate and goes beyond just executing a reasonable amount of commands in the `build` function but rather embedding in it [custom C code](https://github.com/gwatt/chez-exe/blob/937629ad6cdcec698f0b25d996bb9b01a3b7cce7/utils.ss#L27).

b-studios:
I haven't tested it, but I thought the idea of the `chez-exe` project is that we could just reuse their implementation and run one chez command?

b-studios:
I thought there was a way to automatically generate a binary that also bundles `petit`, see https://cisco.github.io/ChezScheme/csug9.5/use.html#./use:h8

dvdvgt:
> I thought there was a way to automatically generate a binary that also bundles `petit`, see https://cisco.github.io/ChezScheme/csug9.5/use.html#./use:h8

Unless I am missing something, it is not possible to create a single standalone binary. See this part of the link you posted:

> Here is how we might create and run a simple "echo" application from a Linux shell:
> echo '(suppress-greeting #t)' > myecho.ss
> echo '(scheme-start (lambda fns (printf "~{~a~^ ~}\n" fns)))' >> myecho.ss
> echo '(compile-file "myecho.ss") \
>       (make-boot-file "myecho.boot" (quote ("petite")) "myecho.so")' \
>      | scheme -q
> scheme -b myecho.boot hello world
> If we take the extra step of installing a copy of the Petite Chez Scheme executable as myecho and copying myecho.boot into the same directory as petite.boot (or set SCHEMEHEAPDIRS to include the directory containing myecho.boot), we can simply invoke myecho to run our echo application

Tldr: Using this approach, you can only compile a `.boot` file from a `.ss` file and then have a copy of the `petite` named like the `.boot` file besides it.

Therefore, it is only up to discussion if
1. We simply want to output a simple bash script (with a shebang) calling the scheme interpreter with the transpiled `.ss` file.
2. Use the aforementioned `chez-exe` project to create a real standalone binary. With this approach, however, the user has to have yet another program installed in their environment.

b-studios:
I think it would also suffice to do something similar to the JS solution: compile the ss scheme as much as possible using the documented process. Then put a bash script next to it that just starts it.

b-studios (approved):
Looks good to me! Feel free to merge (if you have the permission, otherwise I'll do it).

Ideally, we would have some CI to check whether a binary can be created for each backend using `build`. However, running the backend specific tests probably already covers this?

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -78,6 +84,25 @@ trait Runner[Executable] {
...
+    import scala.jdk.CollectionConverters.SetHasAsJava
+
+    val path1 = Path.of(path)
+    val perms = Set(
+      OWNER_READ, OWNER_WRITE, OWNER_EXECUTE,
+      GROUP_READ,
+      OTHERS_READ
+    )
+    IO.createFile(path, content)
+    Files.setPosixFilePermissions(path1, SetHasAsJava(perms).asJava)
```
b-studios, line 104: Maybe this could be used to our kiama fork as a helper function `makeExecutable` or similar?

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -186,18 +226,21 @@ object MLRunner extends Runner[String] {
...
+    executable
+
+  /**
+   * Compile the MLton source file (`<...>.sml`) to an executable.
+   *
+   * Requires the MLton compiler to be installed on the machine.
+   * Assumes [[path]] has the format "SOMEPATH.sml".
+   */
+  def eval(path: String)(using C: Context): Unit =
+    exec(build(path))
```
b-studios, line 245: This is now actually always the same implementation and could be moved to the parent trait.

b-studios: The documentation should be moved to another reasonable place (like `compile`, or `build`)

dvdvgt:
> Ideally, we would have some CI to check whether a binary can be created for each backend using `build`. However, running the backend specific tests probably already covers this?

Yes, it does. However, only for the platform we run the CI on. So only Linux (Ubuntu) if I am not mistaken. It also "runs on my system" (OSX). Windows will probably be problematic, though.

> Looks good to me! Feel free to merge (if you have the permission, otherwise I'll do it).

Hm looks like I do not have permission. I'll just message you when I have implemented your proposed changes and it's ready.

b-studios:
> Windows will probably be problematic, though

Yeah, that's annoying to setup and maintain.

b-studios:
Thanks @dvdvgt!

PR #295, opened 2023-10-19
Draft: Make Case studies work in JIT

marzipankaiser:
Only for discussion. The case studies are broken for other backends by this, due to stdlib incompatibilities.


marzipankaiser:
Problem in ML:
Screenshot 2023-10-19 at 19 23 32

marzipankaiser:
IIUC, this refers to:
```
def parseCalls(): Int / Parser =
  or { number(); 1 } {
    var count = 1;
    ident();
    punct("(");
    count = count + parseCalls();
    many {
        punct(",");
        count = count + parseCalls()
    };
    punct(")");
    count
  }
  ```

**Comment by b-studios** [comment]
yes that makes sense, the scope of mutable variable extends over the recursive call. You need to rewrite the example, potentially not using mutable variables

**Comment by marzipankaiser** [comment]
Ah, thanks - I will try to do so. Afterwards (i.e. when commenting out the above), we will still have this:
Screenshot 2023-10-19 at 21 18 23

Is there a way to make mutually recursive definitions for the ML backend?

**Comment by marzipankaiser** [comment]
So, Effekt now compiles it to ML when commenting out the `parseExpr` calls (that still have a "polymorphic recursion" error in `or`), but then mlton fails with:
```
sbt:effekt> run --backend ml examples/casestudies/parser.md
[info] running effekt.Server --backend ml examples/casestudies/parser.md
Error: /Users/gaisseml/dev/effekt/effekt/out/examples_casestudies_parser.sml 630.77-637.17.
  Function applied to incorrect argument.
    expects: [(TokenKind_799, string, (int, int, int) Position_794) Token_805]
             -> _
    but got: [unit] -> _
    in: ((member2of2 (NondetDollarcapabil  ...  58 tok_1228))) k2965
[error] java.lang.RuntimeException: Nonzero exit value: 1
```

PR #298, opened 2023-10-30
Integrate higher-rank polymorphism on block parameters

### LINKED ISSUE #246: Higher-rank polymorphism
It would be great if we had higher-rank polymorphism on block parameters

```
def within { [V]() => Unit / State[V] }: Unit = ...
```

dvdvgt:
This PR resolves #246 and integrates higher-rank polymorphism on block parameters. This only required a few changes:
- The Parser had to be changed slightly to optionally accept type parameters for block types and type arguments for function arguments.
- The pretty printer for types was changed such that value parameters are in parentheses if type parameters are present, e.g., `[A]List[A] => Int` is instead pretty printed as `[A](List[A]) => Int`.
- The namer required a slight change but for here onward, no changes seem to be required.


b-studios (approved):

b-studios:
Approved by the Effekt committee in person :)

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -524,13 +526,15 @@ class EffektParsers(positions: Positions) extends EffektLexers(positions) {
   lazy val captureSet: P[CaptureSet] = `{` ~> manySep(idRef, `,`) <~ `}` ^^ CaptureSet.apply
 
   lazy val blockType: P[BlockType] =
-    ( (`(` ~> manySep(valueType, `,`) <~ `)`) ~ many(blockTypeParam) ~ (`=>` ~/> primValueType) ~ maybeEffects ^^ FunctionType.apply
-    |  some(blockTypeParam) ~ (`=>` ~/> primValueType) ~ maybeEffects ^^ { case tpes ~ ret ~ eff => FunctionType(Nil, tpes, ret, eff) }
-    | primValueType ~ (`=>` ~/> primValueType) ~ maybeEffects ^^ { case t ~ ret ~ eff => FunctionType(List(t), Nil, ret, eff) }
+    ( (`[` ~> manySep(idDef, `,`) <~ `]`).? ~ (`(` ~> manySep(valueType, `,`) <~ `)`).? ~ many(blockTypeParam) ~ (`=>` ~/> primValueType) ~ maybeEffects ^^ {
```
b-studios, line 529: @dvdvgt maybe have auxiliary definition for `typeParam` to make the role of Id explicit?

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -228,7 +228,9 @@ class EffektParsers(positions: Positions) extends EffektLexers(positions) {
     )
 
   lazy val functionArg: P[BlockLiteral] =
-    ( `{` ~> lambdaParams ~ (`=>` ~/> stmts <~ `}`) ^^ { case (vps, bps) ~ body => BlockLiteral(Nil, vps, bps, body) : BlockLiteral }
+    ( `{` ~> (`[` ~> manySep(idDef, `,`) <~ `]`).? ~ lambdaParams ~ (`=>` ~/> stmts <~ `}`) ^^ {
```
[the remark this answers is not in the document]
b-studios: Or at the very least use existing `maybeTypeParams` :)

effekt/shared/src/main/scala/effekt/Parser.scala
b-studios: I just noticed, we already have `maybeTypeParams` for this.

PR #312, opened 2023-11-16
Feature/multiple return values invariants

phischu:


effekt/jvm/src/main/scala/effekt/Repl.scala
```diff
@@ -108,7 +108,7 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
       runFrontend(StringSource(""), module.make(UnitLit()), config) { cu =>
         module.definitions.foreach {
           case u: Def =>
-            outputCode(DeclPrinter(context.symbolsOf(u)), config)
+            outputCode(DeclPrinter(List(u.id.symbol)), config)
```
b-studios, line 111: This is confusing to me, doesn't u.id.symbol already sometimes return a list?

effekt/jvm/src/main/scala/effekt/Repl.scala
```diff
@@ -220,16 +220,16 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
...
         // try to find the symbol for the def to print the type
-        d.ids.foreach(id => (context.symbolOption(id) match {
+        context.symbolOption(d.id) match {
           case Some(v: ValueSymbol) =>
-            Some(context.valueTypeOf(v))
+            Some(v, context.valueTypeOf(v))
           case Some(b: BlockSymbol) =>
-            Some(context.blockTypeOf(b))
+            Some(b, context.blockTypeOf(b))
           case t =>
```
b-studios, line 228: As said before, did you check this works with the REPL? I would guess it does not print anything. Also it should probably be foreach and not map.

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -225,11 +225,10 @@ object Namer extends Phase[Parsed, NameResolved] {
...
       Context.bind(p.capture)
 
-    case d @ source.ValDef(ids, annots, binding) =>
-      val tpes = annots.map(a => a.map(resolve)) // TODO MRV: remove Option[]
+    case d @ source.ValDef(binders, binding, _) =>
       resolveGeneric(binding)
-      (ids zip tpes) foreach { (id, tpe) =>
-        Context.define(id, ValBinder(Context.nameFor(id), tpe, d))
+      binders.foreach { case source.ValueParam(id, tpe) =>
+        Context.define(id, ValBinder(Name.local(id), tpe.map(resolve), d))
```
b-studios, line 231: Is this change from `nameFor` to `local` intentional?

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -296,16 +296,10 @@ class EffektParsers(positions: Positions) extends EffektLexers(positions) {
...
-          val ids = defs.map(_._1)
-          val tpes = defs.map(_._2)
-          ValDef(ids, tpes, binding)
-     }
-
-  lazy val valDef: P[(IdDef, Option[ValueType])] =
-    idDef ~ (`:` ~/> valueType).? ^^ { case id ~ tpe => (id, tpe)}
+  lazy val valDef: P[Def] =
+    `val` ~> someSep(valueParamOpt, `,`) ~ (`=` ~/> stmt) ^^  {
+      case binders ~ binding => ValDef(binders, binding)
```
b-studios, line 301: thanks, much better

b-studios (commented):
Looks good to me for now, but can you briefly answer the questions (if you haven't done so already)?

b-studios:
Thanks for merging, but please let us answer the questions, still

effekt/shared/src/main/scala/effekt/context/Annotations.scala
```diff
@@ -504,15 +504,6 @@ trait AnnotationsDB { self: Context =>
...
 
-  /**
-   * Searching the symbol for a definition
-   *
-   * These lookups should not fail (except there is a bug in the compiler)
-   */
-
-  def symbolsOf(tree: source.Definition): List[Symbol] =
-    tree.ids.map(symbolOf)
-
```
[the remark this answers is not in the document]
denhie: This function was written by me to get the list of symbols. It was only used in a few places.

effekt/jvm/src/main/scala/effekt/Repl.scala
denhie: Prints nothing in the repl because d.id is empty. In case of ValDef we need d.binders.map(_.id)

b-studios: You could in principle match on d and special case if it is a ValDef.

PR #313, opened 2023-11-16
Refactor/mrv valdef

b-studios:
Please don't merge.

This is a PR that tries to reorganize how symbols are looked up. It is orthogonal to the thesis, but will help to eventually merge.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/source/Tree.scala
+++ b/effekt/shared/src/main/scala/effekt/source/Tree.scala
@@ -507,7 +507,7 @@ object Named {
 
   type Params = ValueParam | BlockParam
   type Externs = ExternDef | ExternResource | ExternInterface | ExternType
-  type Defs = FunDef | ValDef | VarDef | DefDef | InterfaceDef | DataDef | RecordDef | TypeDef | EffectDef
+  type Defs = FunDef | VarDef | DefDef | InterfaceDef | DataDef | RecordDef | TypeDef | EffectDef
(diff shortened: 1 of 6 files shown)


effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -544,7 +538,40 @@ object Named {
...
+      C.annotationOption(Annotations.Symbol, t.id).map(_.asInstanceOf[SymbolFor[T]])
+  }
+
+  /**
+   * `ValDef`s resolve to potentially multiple symbols.
+   */
+  extension (t: ValDef) {
+    def boundSymbols(using C: Context): List[effekt.symbols.Binder.ValBinder] =
+      // the symbol of a ValueParam is a symbols.ValueParam, not a ValBinder!
+      t.binders.map { (b: ValueParam) => b.symbol; ??? }
```
b-studios, line 562: @phischu I purposefully crash here. Can we look into this together? We need to decide how to represent them on the symbols level.

b-studios: We could conflate the two. I just want to mention that having separate symbols for different constructs makes generating better error messages in Intelligence easier, because we can simply match on the type and specialize the error.

b-studios: I also don't like the name `boundSymbols`, but since `symbols` is a package name, calling it like this will crash everything....

b-studios:
@phischu

Issue #309, opened 2023-11-10
Share standard library code between backends


phischu:
In folder `./libraries` we have multiple standard libraries. One for each backend. These overlap partially. The goal is to share code between them.

We identify 5 _categories_ of definitions:

1. Backend-specific extern definitions (e.g. DOM manipulation)
2. General extern definitions (e.g. `Int`, `infixAdd`)
3. Definitions that are extern for performance reasons or for quicker implementation (e.g. `Regex`)
4. Definitions using only general extern definitions (e.g. `repeat`)
5. Definitions using no extern definitions (e.g. `List`)

Category (1) can never be shared. Category (5) can always be shared. Category (4) can be shared as long as the used extern definitions have roughly the same semantics. Therefore we would like some kind of informal specification for those in category (2). Category (3) I would like to forbid.

We should come up with a concrete organizational scheme that separates these categories. We should also come up with concrete rules for extending the standard libraries and write these down.

b-studios:
That's a very important issue. However, we cannot forget that some backends might implement stdlib functions differently (either for performance reasons, or for lack of missing features).

For instance in JS we would often use `while` and `var` -- where in other backends we would write a local tail-recursive function. Ideally, in this concrete example the two things mean the same, though.

jiribenes:
Is this fixed by #449, @phischu?

phischu:
Yes.

PR #315, opened 2023-11-21
Port break & continue to all backends

### LINKED ISSUE #314: loop with early return
Feature request:
a for loop with ability to return early/break

Right now there is no such feature (afaik): 
each(n,m) executes from start to finish, no option to break early.

In order to use a return early functionality (as is dictated in my case by the existing are-we-fast-yet benchmarks) i have to fall back to "while (run)" ... and mutate run in the loop:
```
var run = true
while (run) {
   if (...) { run = false}
}
```
This is much harder to read/understand/verify and invites bugs (mutable state and unnecessary while loop)

so: 
each() with break (and continue) would be nice to have.
(any kind of early return would suffice, it doesnt have to be "break")

jiribenes:
## Motivation

We can port the new user-defined `break` and `continue` into all backends.
Resolves #314 

## Questions

- [x] Can this break some benchmarks we use? => It would be nice to benchmark this before merging. 
- [x] Relates to #309: can we share this and similar code among backends without needing to copy it into every backend separately?
- [x] Do we even want this change? Is this a design we're happy with?


b-studios:
I am a big fan of this, since it shows that effects cannot only be used the high-level but also for very low-level things like jumps.

> Can this break some benchmarks we use? => It would be nice to benchmark this before merging.

Breaks running the benchmarks => That's a severe problem (we should have CI for this!)
Breaks performance => I don't care (then we should fix the performance to support these kinds of programs)

> can we share this

Yes! We need a solution for this... ASAP

> Do we even want this change? Is this a design we're happy with?

I'd say just merge and redesign later ;) Things to concider:

- break with a value (requires `Control[T]`)
- how to break out of nested loops

b-studios:
@marzipankaiser @phischu could you briefly check whether this breaks the benchmarks?

phischu:
Breaks neither build nor performance for the benchmarks using the ml backend

PR #316, opened 2023-11-24
DS-JS Backend

b-studios:
This is the work-in-progress PR to add a direct-style JS backend.  It builds on our work of bytecode instrumentation from [2018](https://dl.acm.org/doi/abs/10.1145/3276481). Although I am implementing a variant here that is closer to [generalized stack inspection](https://dl.acm.org/doi/abs/10.1145/1090189.1086393).


b-studios:
@phischu observed that the generated continuations always are only used exactly once. We could inline them in the future. However, I am not sure how this could affect performance.

b-studios:
@marzipankaiser could you help me debug the breakage in PolymorphismBoxing? 
I added captures to BlockParams and guess the breakage is caused by the following change:

https://github.com/effekt-lang/effekt/pull/316/files#diff-e3c105bf7369dea9df7e42bc4aaa7ec444a2d5d472068ca01688ca38c5fd269cR343

marzipankaiser:
Quick benchmark results of this(`>`) in comparison to old js results (`<`) that I had lying around.
Columns are Benchmark, mean [s], stddev.
```diff
2c2
< countdown,171015.9,486.7
---
> countdown,33.9,0.7
5,10c5,10
< iterator,19537.8,64.2
< nqueens,34128.0,107.4
< product early,,
< resume nontail,,
< tree explore,,
< triples,3963.5,30.0
---
> iterator,33.2,0.3
> nqueens,34.1,0.3
> product early,39.6,0.3
> resume nontail,47.2,0.4
> tree explore,37.6,0.3
> triples,10473.5,38.1

```

b-studios:
@marzipankaiser I fixed a few things in `Renamer` and moved it from `test/core` to `main/core` since it is useful also to write transformations.

Note that the implementation is still not correct. Local definitions are not mutually recursive (at the moment). Since renamer currently allows this, there might be some accidental capture.

b-studios:
BTW the "Text file busy" CI bug is really getting annoying... #322

effekt/jvm/src/test/scala/effekt/core/RenamerTests.scala
```diff
@@ -148,4 +150,28 @@ class RenamerTests extends CoreTests {
...
+  //        |
+  //        | def main = { () =>
+  //        |   let renamed1 = 1
+  //        |   let renamed2 = 2
+  //        |   return renamed2:Int
+  //        | }
+  //        |""".stripMargin
+  //
+  //    assertRenamedTo(input, expected)
+  //  }
```
b-studios, line 176: @marzipankaiser this test shows that `Renamer` right now is not correct.

b-studios:
There are still few things to be done, like making sure that the monadic version still works and adapting the homepage to the new calling convention.

However, I think this now starts to look good enough to become the new standard backend.

I know it is a huge PR, but maybe someone is interested to look into it?

PR #320, opened 2023-11-28
JS backend: Do not drop first element for Array.toList (fixes #319)

### LINKED ISSUE #319: JS backend: Array.toList drops first element
Executing `.toList` on an array drops the first element.

## Example
For the program:
```effekt
import mutable/array

def main() = {
  val array = arrayFromList([1,2,3]);
  println(array.toList)
}
```
we get the output:
```
Cons(2, Cons(3, Nil()))
```

marzipankaiser:
Fixes #319.


jiribenes (approved):
LGTM :shipit:

PR #321, opened 2023-11-29
Testing infrastructure for core transformations

marzipankaiser:
This is a WIP to factor out some of the infrastructure written for testing `PolymorphismBoxing`,
so we can use it to test other transformations on core, too.


marzipankaiser:
(Note for future @marzipankaiser: When rebasing after #316 was merged, 17c23c6 should become empty and 36ae18d should be changed to fit those changes)

PR #332, opened 2023-12-04
Fix Renamer to conform to contract given in docs

marzipankaiser:
Fixes the problem noted in https://github.com/effekt-lang/effekt/pull/316#discussion_r1413046445 ,
and should make the output Barendregt unless the generated names (prefix + number / original name + number) are already in use.

@b-studios .

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/Renamer.scala
+++ b/effekt/shared/src/main/scala/effekt/core/Renamer.scala
@@ -9,17 +9,14 @@ import effekt.context.Context
  * @param names used to look up a reference by name to resolve to the same symbols.
  *              This is only used by tests to deterministically rename terms and check for
  *              alpha-equivalence.
- * @param prefix if the prefix is empty, the original name will be used
+ * @param prefix if the prefix is empty, the original name will be used as a prefix
  *
  * @param C the context is used to copy annotations from old symbols to fresh symbols
  */
 class Renamer(names: Names = Names(Map.empty), prefix: String = "") extends core.Tree.Rewrite {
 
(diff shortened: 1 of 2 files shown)


effekt/shared/src/main/scala/effekt/core/Renamer.scala
```diff
@@ -9,17 +9,14 @@ import effekt.context.Context
...
  * @param C the context is used to copy annotations from old symbols to fresh symbols
  */
 class Renamer(names: Names = Names(Map.empty), prefix: String = "") extends core.Tree.Rewrite {
 
-  // TODO PROBLEM: if the input is not Barendregt, then the output will also not be...
-  //   withBindings currently is invoked multiple times for the same id...
-
   // maps symbols bound in the current scope to their renamed variants.
-  private var bound: Map[Id, Id] = Map.empty
+  private var bound: List[Map[Id, Id]] = List.empty
```
b-studios, line 19: Do we really need this list of maps?

marzipankaiser: *Need* is a bit strong. We can get away without, but then we can't have `unbind` and have to do something like:
```
val tBinding = rewrite(binding)
withBinding(x) { Let(rewrite(x), tBinding, rewrite(body)) }
```
for non-recursive bindings.

Which is fine, too, admittedly.

marzipankaiser: Oh, and we could - of course - replace this with a proper `Environment` type with a `lookup`.

Issue #293, opened 2023-10-17
Make effekt have a non-0 exit code when compilation fails


marzipankaiser:
When using `effekt ... --compile` to compile a program, the exit code is always 0, even when there is an error.
This makes writing scripts (shell scripts, build/test scripts etc) using it harder.

Example:
<img width="932" alt="Screenshot 2023-10-17 at 14 30 50" src="https://github.com/effekt-lang/effekt/assets/2551444/430b17d6-c104-4c74-bdbb-554bcf599cdf">

Intended output:
Terminate with exit code non-0 (e.g. 1), if there are errors.

b-studios:
Fixed by #338

PR #338, opened 2023-12-04
Fix non-zero exit code on error

marvinborner:
This PR fixes issue #293.

If a message with error severity is logged in compilation/interpret/build mode, effekt now fails with code 1. If run as REPL or server, errors won't cause exits.


effekt/jvm/src/main/scala/effekt/Driver.scala
```diff
@@ -100,5 +100,9 @@ trait Driver extends kiama.util.Compiler[EffektConfig, EffektError] { outer =>
   def afterCompilation(source: Source, config: EffektConfig)(implicit C: Context): Unit = {
     // report messages
     report(source, C.messaging.buffer, config)
+    
+    // exit with non-zero code if not in repl/server mode and messaging buffer contains errors
+    if (!(config.repl() || config.server()) && C.messaging.hasErrors)
```
jiribenes, line 105: Nitpick: I think it might be a bit easier to read this if it were 
```suggestion
    if (config.shouldExitOnError() && C.messaging.hasErrors)
```

with 
```scala
def shouldExitOnError(): Boolean = !repl() && !server()
```
in `EffektConfig.scala` (perhaps with a better name 🤔)

(Similarly to `requiresCompilation()` in `EffektConfig.scala`)

effekt/jvm/src/main/scala/effekt/EffektConfig.scala
```diff
@@ -121,6 +121,10 @@ class EffektConfig(args: Seq[String]) extends REPLConfig(args) {
 
   def interpret(): Boolean = !server() && !compile() && !build()
 
+  def repl(): Boolean = filenames().isEmpty && !server() && !compile()
+
+  def exitOnError(): Boolean = !repl() && !server()
```
b-studios, line 126: Do we have a semantic name for this? What characterizes these options? Being in batch mode?

b-studios:
In our meeting @marzipankaiser mentioned that this might kill `sbt` when we use `sbt run myfile.effekt` and an error occurs (and it actually does).

marzipankaiser (approved):
Tried it locally on my machine.
It seems to do (close enough to) the right thing for me:
- running `effekt ...` terminates with useful error code (1 on error)
- running `sbt`, then `project effektJVM`, then `run ...` does not terminate shell on error, generates compiled output correctly
- running `sbt "effektJVM/run ..."` terminates with useful error code (1 on error)
- running `sbt`, then `effektJVM/run ...` *will* terminate the sbt shell on error. While this is somewhat unexpected behaviour, I would not run into this in normal use, so I think this is a good compromise for now.

In Summary: LGTM.

Issue #66, opened 2022-04-15
Typeclasses


b-studios:
We should consider adding typeclasses or similar to prevent repeating code such as sorting datastructures, etc.

serkm:
This seems to me like a straightforward transformation to codata:
- class definition -> interface definition
- instance -> top level object
- type constraints -> implicit parameters

I guess the tricky part would be inferring the correct instance.

serkm:
A suggestion for syntax:
```
class C[A] {
    // function declarations
}

instance C[T] {
    // function definitions
}

// type constraints
def f [A] <C[A]> (a: A) = {...}
```

serkm:
Another idea would be to add implicit parameters to Effekt and use regular codata to express type classes:

```
// implicit parameter of type A
def f () <a: A> = {...}

def main() = {
   f() // passes implicit value iff there is exactly one value of type A in scope
}
```

PR #339, opened 2023-12-04
Expand standard library

jiribenes:
## Motivation

I'm doing Advent of Code 2023 and wanted to upstream a few useful functions :)

## Changes

In general, I tried to reconcile the different versions, conventions, and limitations of all of the different backends... 🤯
It would be really nice to have a solution for #309 -- some of the functions would work on any backend and I wouldn't have to copy them over 5 times.

Newly added functions by module:

### effekt

_all backends_

- `min(n: Int, m: Int): Int`, no specific tests, upstreamed from LLVM backend
- `max(n: Int, m: Int): Int`, no specific tests, upstreamed from LLVM backend

### immutable/list

_all backends_

- `foreachIndex[A](l: List[A]) { f: (Int, A) => Unit } : Unit`, no specific tests
- `collect[A, B](l: List[A]) { f : A => Option[B] }: List[B]` with new tests in `examples/pos/list/collect`
- `flatMap[A, B](l: List[A]) { f : A => List[B] }: List[B]` with new tests in `examples/pos/list/flatMap`
- `sum(l: List[Int]): Int` with new tests in `examples/pos/list/sum`

### text/string

_not in LLVM backend_

- `indexOf(str: String, sub: String): Option[Int]` with new tests in `examples/pos/string/indexOf`; not in the Chez backend
- `lastIndexOf(str: String, sub: String): Option[Int]` with new tests in `examples/pos/string/indexOf`; not in the Chez backend
- `startsWith(str: String, prefix: String): Boolean` and `endsWith(str: String, suffix: String): Boolean`, no specific tests; not in the Chez backend
- `substring(str: String, from: Int, to: Int): String` with new tests in `examples/pos/string/substring`
  - most backends had `substring(str: String, from: Int): String` with many different semantics
  - currently claps `from` and `to` to be between `0` and `str.length`
  - ... and makes sure that if `from >= to`, then the result is `""`
- ANSI escape codes (`ANSI_GREEN`, `ANSI_RED`, `ANSI_RESET`)

### mutable/array

_only in the JS backend_

- reformatted the whole module to be more consistent with the rest of the codebase
- `foreach[T](arr: Array[T]){ action: T => Unit / Control }: Unit`
- `foreachIndex[T](arr: Array[T]){ action: (Int, T) => Unit / Control }: Unit`
- `sum(list: Array[Int]): Int` with new tests in  `examples/pos/array/sum`
- there's also new big set of tests for going between `Array[T]` and `List[T]` in `examples/pos/array/list_conversion` to prevent problems like in #319 
- **BREAKING** renamed: `arrayFromValue→fill`, `arrayFromLoop→build`, `arrayFromList→fromList`

### mutable/map

_only in the JS backend_

- reformatted the whole module to be more consistent with the rest of the codebase
- added `keys`, `values` which return a JS array from `mutable/array` -- there are no tests for this

### test

- ported from ML backend to JS and Chez with its tests
- could/should be extracted out to shared stdlib (#309)
- added `assertEqual` -- this could possibly fail on backends without general equality, but it's still very useful to me
  - uses an explicit `show` block argument because of backends that don't have a `show` function for all types (related: #66)


libraries/ml/text/string.effekt
```diff
@@ -18,19 +18,28 @@ extern pure def length(str: String): Int =
...
 
+def substring(str: String, from: Int, to: Int): String = {
+    def clamp(lower: Int, x: Int, upper: Int) = max(lower, min(x, upper))
+
+    val len = str.length
+    val clampedFrom = clamp(0, from, len)
+    val clampedTo = clamp(0, to, len)
+
+    val substringLength = max(0, clampedTo - clampedFrom)
+    str.unsafeSubstring(clampedFrom, clampedTo)
```
jiribenes, line 35:
```suggestion
    str.unsafeSubstring(clampedFrom, substringLength)
```

😭

libraries/js/mutable/map.effekt
```diff
@@ -1,18 +1,25 @@
...
 
 extern io def update[K, V](m: Map[K, V], key: K, value: V): Unit =
-    "(function() { m.set(key, value); return $effekt.unit })()"
+  "(function() { m.set(key, value); return $effekt.unit })()"
+
+extern pure def values[K, V](map: Map[K, V]): Array[V] =
+  "Array.from(map.values())"
+
+extern pure def keys[K, V](map: Map[K, V]): Array[K] =
+  "Array.from(map.keys())"
```
jiribenes, line 25: I also wanted to add `.entries` which returns a _JavaScript_ tuple of `(key, value)`, but then the conversion is awful since JS uses arrays instead of tuples and the JS array is a heterogeneous array. I would just call `map.entries.map(([k, v]) => { Tuple2(k, v) }`, but `Tuple2` is always renamed (to something like `Tuple2_201`).

jiribenes:
Squashed the 33 commits into a single large commit with a huge description :)

jiribenes:
Please note the **breaking changes** in `mutable/array`: renamed:
- arrayFromValue → fill
- arrayFromLoop → build
- arrayFromList → fromList

A [untested!] command like `sed -i '' -e 's/arrayFromValue/fill/g' -e 's/arrayFromLoop/build/g' -e 's/arrayFromList/fromList/g' some-effekt-file.effekt` should be enough for migration.

cc @IR0NSIGHT

jiribenes:
Rebased on current `master` (be63dbb8).

libraries/chez/common/immutable/list.effekt
```diff
@@ -22,6 +22,45 @@ def foreach[A](l: List[A]) { f: A => Unit } : Unit =
       foreach(rest) { a => f(a) }
   }
 
+def foreachIndex[A](l: List[A]) { f: (Int, A) => Unit } : Unit = {
```
b-studios, line 25: Most other languages that I am aware of first list the element, then the index. WDYT?

b-studios (approved):
LGTM the only bikeshed I have is the order of arguments on foreachIndex

libraries/chez/common/immutable/list.effekt
jiribenes: No preference from my point of view, I just followed the order which was already present in `immutable/list` on the JS backend 😇 

https://github.com/effekt-lang/effekt/blob/master/libraries/js/immutable/list.effekt#L23

b-studios: well... what did I think back then? Let's merge first, then share, then reorganize.

jiribenes: > Most other languages that I am aware of first list the element, then the index. WDYT?

I took a quick look at other languages:

Python with `enumerate`:
```py
for index, elem in enumerate(myIterator):
```

Go with `range`:
```
for index, elem := range mySlice { ... }
```

C++20 with `enumerate`:
```
for (auto const [index, elem] : std::views::enumerate(v))
```

Kotlin:
```
fun <T> Iterable<T>.forEachIndexed(
    action: (index: Int, T) -> Unit)
)
```

Haskell (`vector` library):
```
iforM :: Monad m => Vector a -> (Int -> a -> m b) -> m (Vector b)
```

... but Scala uses `zipWithIndex` and has the `(T, Int)` order instead

b-studios: Funny, I looked at JS, Scala, Ruby...

PR #347, opened 2023-12-08
Add sbt task for building the JS file

dvdvgt:
This PR adds a sbt task for building the JS file conveniently. The resulting file is saved under `out/effekt.js` and is to be used by the [website's](https://github.com/effekt-lang/effekt-website) "update compiler" workflow so that we can get rid of this hacky part [here](https://github.com/effekt-lang/effekt-website/blob/26493b1a1063dea79c33df9f3e45be789118c37c/.github/workflows/update-compiler.yml#L30).


build.sbt
```diff
@@ -167,6 +168,15 @@ lazy val effekt: CrossProject = crossProject(JSPlatform, JVMPlatform).in(file("e
   )
   .jsSettings(
 
+    assembleJS := {
+      (Compile / clean).value
+      (Compile / compile).value
+      val resultDir = (Compile / fastLinkJSOutput).value
```
b-studios, line 174: Why fastLinkJSOutput? I think we should use fullOpt, no?

PR #348, opened 2023-12-08
Fix "Annotate Captures" feature of the VSCode extension

dvdvgt:
## Description

The VSCode extension has the feature to annotate the captures of boxes and definitions. Currently, however, the VSCode extension does not show any captures since the relevant phase of the compiler ([AnnotateCaptures](https://github.com/effekt-lang/effekt/blob/c918d0d68e758169be0b2be1bc44971ca90c5167/effekt/shared/src/main/scala/effekt/source/AnnotateCaptures.scala#L17)) is not run. Thus, this PR introduces the following changes:

- Extend the phases run by the language server to also include wellformedness checks and the annotation of captures
- A subtle bug caused old captures information to be shown on previously seen files. Therefore, each time the `AnnotateCaptures` phase is run, the information about the collected captures has to be reset (`allCaptures`).

## Discussion

We may need to keep an eye on the responsiveness of the language server due to increased number of phases run. However, so far I have not noticed any slowdowns.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Compiler.scala
+++ b/effekt/shared/src/main/scala/effekt/Compiler.scala
@@ -125,13 +125,14 @@ trait Compiler[Executable] {
       mod
     }
 
-  def runMiddleend(source: Source)(using Context): Option[Typechecked] =
-    (Parser andThen
-    Namer andThen
-    BoxUnboxInference andThen
-    Typer andThen
-    Wellformedness andThen
-    AnnotateCaptures)(source)
+  /**
+   * Used by the server to typecheck, report type errors and show
+   * captures at boxes and definitions 
+   */
+  def runMiddleend(source: Source)(using Context): Option[Module] =
+    val typechecked = Frontend(source)
+    typechecked.foreach(res => validate(source, res.mod))
+    typechecked.flatMap(Middleend.run).map(_.mod)
 
   /**
    * Called after running the frontend from editor services.
--- a/effekt/shared/src/main/scala/effekt/Compiler.scala
+++ b/effekt/shared/src/main/scala/effekt/Compiler.scala
@@ -130,9 +130,10 @@ trait Compiler[Executable] {
    * captures at boxes and definitions 
    */
   def runMiddleend(source: Source)(using Context): Option[Module] =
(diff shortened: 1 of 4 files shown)


effekt/shared/src/main/scala/effekt/Compiler.scala
```diff
@@ -125,6 +125,14 @@ trait Compiler[Executable] {
       mod
     }
 
+  def runMiddleend(source: Source)(using Context): Option[Typechecked] =
+    (Parser andThen
+    Namer andThen
+    BoxUnboxInference andThen
```
b-studios, line 131: This duplicates Frontend, except parser is not cashed.

b-studios: Is it possible to reuse Frontend or is the non caching of parser essential?

dvdvgt:
Apparently, it is possible to use the cached versions of the middle and frontend.

effekt/shared/src/main/scala/effekt/Compiler.scala
```diff
@@ -125,6 +125,15 @@ trait Compiler[Executable] {
...
     }
 
+  /**
+   * Used by the server to typecheck, report type errors and show
+   * captures at boxes and definitions 
+   */
+  def runMiddleend(source: Source)(using Context): Option[Module] =
+    val typechecked = Frontend(source)
+    typechecked.foreach(res => validate(source, res.mod))
+    typechecked.flatMap(Middleend.run).map(_.mod)
```
b-studios, line 135: Sorry for nitpicking so much, but isn't this roughly equivalent to:
```
(Frontend andThen Middleend).map { res => validate(res.source, res.mod); res.mod }
 ```

dvdvgt: No worries. If it doesn't matter when the `validate` function is called, then this should equivalent, true. I thought it might be important to validate right after the frontend.

b-studios: Maybe I am missing something, but the most important thing is that validate is called as part of the LSP service

PR #349, opened 2023-12-11
Feature/core constructor

b-studios:
This PR special cases constructors in core. This way, it is trivial for backends to detect whether a `PureApp` is a constructor call or not. In consequence, all remaining `PureApp`s are FFI calls, which nicely stratifies things.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/Tree.scala
+++ b/effekt/shared/src/main/scala/effekt/core/Tree.scala
@@ -169,12 +169,24 @@ case class Run(s: Stmt) extends Expr
  * -------------------------------------------
  */
 enum Pure extends Expr {
+
   case ValueVar(id: Id, annotatedType: ValueType)
 
   case Literal(value: Any, annotatedType: ValueType)
 
-  // invariant, block b is pure.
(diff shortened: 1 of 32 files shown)


effekt/shared/src/main/scala/effekt/core/Tree.scala
```diff
@@ -347,9 +369,16 @@ object normal {
...
 
   // "match" is a keyword in Scala
-  // TODO perform matching here, if scrutinee statically known
   def patternMatch(scrutinee: Pure, clauses: List[(Id, BlockLit)], default: Option[Stmt]): Stmt =
-    Match(scrutinee, clauses, default)
+    scrutinee match {
+      case Pure.Make(dataType, ctorTag, vargs) =>
+        clauses.collectFirst { case (tag, lit) if tag == ctorTag => lit }
+          .map(body => app(body, Nil, vargs, Nil))
+          .orElse { default }.getOrElse { sys error "Should not happen" }
```
b-studios, line 377: This is a case-of-known-case analysis that was dear to my heart :)

PR #351, opened 2023-12-18
Add testing suite for LSP server

marvinborner:
This PR aims to add a test suite for Effekt's LSP server.

Apart from the obvious reasons, LSP testing could also be useful for finding reflection bugs (PR #329).

The implementation relies on ScalaJS and Microsoft's `vscode-languageserver-protocol` node module. The resulting facade bindings could also be useful if we wanted to replace LSP4J with a ScalaJS implementation (see [here](https://www.scala-js.org/doc/interoperability/facade-types.html) for more information on writing facade types).

For now, this PR just consists of a basic proof-of-concept bindings/connection/initialization (I had a lot of trouble with ScalaJS and facade types).


jiribenes:
Hey, looks great so far! :)

Just wondering, have you tried [ScalablyTyped](https://scalablytyped.org/docs/readme.html)?
It could save you lots of time spent on writing wrappers (it uses some advanced magic to get the types via TypeScript).

marvinborner:
Ah yes, I wanted to mention this in the PR. I actually tried it and it didn't work properly (or maybe I used it wrong). It gave a lot of warnings about unresolvable/etc. types and the generated Scala files missed several relevant classes. Maybe I'll try again, it would really save a lot of time.

marvinborner:
Can anyone reproduce the [CI failure](https://github.com/effekt-lang/effekt/actions/runs/7806995999/job/21294472921#step:13:35)? It seems to think that `hashTriple` and `main` end at a different line/character and I can't find the reason.

Aside from that, this PR should be mergable soon, so I'm looking for some feedback.

jiribenes:
Deprecated by #885.

PR #324, opened 2023-12-01
Change async API to monadic reflection

b-studios:

PR #355, opened 2023-12-25
Expand immutable/list library in every backend

jiribenes:
### Motivation

🎶 _You better watch out, you better not cry, better not pout, I'm telling you why, a \*non-denominational gift giver\* is coming to town_.
_He's making an **immutable list**, and he's doing it twice; gonna find out, which backend's naughty or nice, a \*non-denominational gift giver\* is coming to town_. 🎶 

### Changes

For the JS & Chez & LLVM & ML (= all) backends:
- Added the following functions:
    - `fill[A](size: Int, default: A): List[A]`
    - `build[A](size: Int) { index: Int => A }: List[A]`
    - `get[A](list: List[A], index: Int): A / Exception[OutOfBounds]`
    - `all[A](list: List[A]) { predicate: A => Boolean }: Boolean`
    - `any[A](list: List[A]) { predicate: A => Boolean }: Boolean`
    - `join[A](lists: List[List[A]]): List[A]`
    - `join[A](lists: List[List[A]], between: List[A]): List[A]` (Haskell's `intercalate`)
    - `slice[A](list: List[A], start: Int, stopExclusive: Int): List[A]`
    - `splitAt[A](list: List[A], index: Int): (List[A], List[A])`
    - `updateAt[A](list: List[A], index: Int) { update: A => A }: List[A]`
    - `modifyAt[A](list: List[A], index: Int) { update: A => A }: List[A] / Exception[OutOfBounds]`
    - `deleteAt[A](list: List[A], index: Int): List[A]`
    - `insert[A](list: List[A], index: Int, x: A): List[A]`
    - `replace[A](list: List[A], index: Int, x: A): List[A]`
    - `zip[A, B](left: List[A], right: List[B]): List[(A, B)]`
    - `zipWith[A, B, C](left: List[A], right: List[B]) { combine : (A, B) => C }: List[C]`
    - `unzip[A, B](pairs: List[(A, B)]): (List[A], List[B])`
    - `isSortedBy[A](list: List[A]) { compare: (A, A) => Boolean }: Boolean`
- Modified the semantics of `.take` and `.drop` so that for all integers `n` and lists `list`: `val l = list.take(i); val r = list.drop(i); l.append(r) == list`.
- Tested all functions, wrote doc comments with examples and complexities, added a plenty of automatic tests to a vast majority of the functions.
    - The tests try not to use wild `show` or `println` overloads in order to also work on the JS/ML backends.
- After this change, the `immutable/list` library is the exact same for the JS backend and for all of the Chez backends, thus paving the way towards a library shared by all backends. :)
- LLVM and ML are also included (except for a few problems)!
- The actions in `foreach` and `foreachIndex` now may use `Control`
- The Exception interface now returns `Nothing`
```scala
interface Exception[E] {
  def raise(exception: E, msg: String): Nothing
}
```

### Future work

- ~~port `immutable/list` in its entirety to the LLVM backend~~
- ~~port `immutable/list` in its entirety to the ML backend~~
- make a `libraries/common` folder and move the file `immutable/list` there
- move the rest of stdlib tests into `examples/stdlib`

### Unresolved questions

- I'm open to all possible nitpicking and bikeshedding. :)
- ~~the tests now seem to take about a minute more, but I'm a bit reluctant to cut down on tests~~
    - We made a new stdlib test runner to address this. 
- See new questions below.


examples/pos/arrays.effekt
```diff
@@ -18,12 +18,12 @@ def main() = {
...
+    val arrWithVals: Array[String] = fill(4, "Hello World");
     println(arrWithVals)
 
-    val arrByF = build(5){ i => i };
+    val arrByF: Array[Int] = build(5){ i => i };
     println(arrByF);
 
-    var arrByList = fromList([1, 5, 17]);
+    var arrByList: Array[Int] = fromList([1, 5, 17]);
     println(arrByList)
```
jiribenes, line 28: These annotations are necessary since we don't have qualified imports yet.
Thus the compiler cannot distinguish between:
```
fill[T](Int, T): Array[T]
```
and
```
fill[T](Int, T): List[T]
```

b-studios:
Wow, thanks, that's awesome! 

Regarding the tests, maybe we can have a separate Testrunner to run tests that test the stdlib and not the compiler itself?

This way we can run both on CI but only the compiler tests locally when developing

libraries/chez/common/immutable/list.effekt
```diff
@@ -3,124 +3,599 @@ module immutable/list
...
-    case Cons(a, rest) => rest.reverseWith(Cons(a, acc))
-  }
-  l.reverseWith(Nil())
+  var res = Nil[A]()
+  l.foreach { el => res = Cons(el, res) }
+  res
 }
 
-// [1,2,3].reverseOnto([4,5,6])
+// Reverse a list `l` and append it to `other`.
```
marvinborner, line 231: Wording: `append` should be `prepend`

jiribenes: Thanks! :)

jiribenes:
@b-studios:
> Regarding the tests, maybe we can have a separate Testrunner to run tests that test the stdlib and not the compiler itself?

Sure. Do you want me to do that in this PR or in a separate PR?
Moreover, should it be a test runner for stdlib per backend or a test runner for stdlib for all backends together?
(`JavaScriptStdTests, ..., MLStdTests` vs. `StdTests`)

libraries/chez/common/immutable/list.effekt
jiribenes: I decided to use:
```
// Reverse a list `l` and append `other` to it.
```
to avoid using the word `prepend` which is not used anywhere else. :)

However, let me know if you have a better description.
To be honest, I'm not sure that a function like this needs to be in the stdlib... 🤔 (it's only used in the definition of append, IIRC)

jiribenes:
Added two more useful functions:
- `join[A](lists: List[List[A]]): List[A]`
- `join[A](lists: List[List[A]], between: List[A]): List[A]` (Haskell's `intercalate` under an overloaded name, not sure I like it too much...)

libraries/chez/common/immutable/list.effekt
marvinborner: We might also want to consider a similar approach with swapped arguments used in the [standard library of Idris 2](https://github.com/idris-lang/Idris2/blob/0e831ed5ef6670d01e025197f9d1b3178586be9b/libs/base/Data/List1.idr#L113)

libraries/chez/common/immutable/list.effekt
```diff
@@ -3,124 +3,643 @@ module immutable/list
...
+def nonEmpty[A](l: List[A]): Boolean = l match {
+  case Nil() => false
+  case Cons(a, rest) => true
+}
+
+// Return the first element of a given list.
+// Throws a `EmptyList` exception if it's empty.
+//
+// O(1)
+def head[A](l: List[A]): A / Exception[EmptyList] = l match {
```
marvinborner, line 63: Wouldn't the name `unsafeHead` be more appropriate? I guess this is a philosophical question but I would prefer totality by default. The `headOption` could then be named `head`.
Same for `tail`/`unsafeTail`

b-studios: We sometimes use `unsafe` to really mean unsafe (as in "unchecked ffi"). Here, it throws an exception which still can be caught.

I think this is one important point in the design of a stdlib in presence of effects and handlers. Do we want to force users to pattern match or to handle?

jiribenes: That's an important point, yes.
In this PR, I didn't change the `head` and `headOption` functions at all in order to side-step the issue :)

However, I added `get` and `unsafeGet` where the former returns an option and the latter panics at runtime (so it's _truly unsafe_, mostly intended for library writers...)
I'm not convinced that we really need the unsafe version though. Having `get` and `getOption` might be more consistent.

I'm very much open to debate here — now that there's a sketch of a library with a few more functions, we get a small view into library design in the language.

jiribenes:
At the Ξ Effekt Working Group meeting, we talked about the following changes:
- using `Exception[OutOfBounds]` for functions that can fail (like `get` and `head`)
- in order to facilitate nicer working with `Exception`s, port [the reification of the `Exception` effect into the `Result` type](https://github.com/effekt-lang/effekt/pull/324/files#diff-d98f9fe70c2dd70a63691e244089dce836f03f0d46f1fe10ba8f04708ef520c2R7) from #324 into the standard library
- use the `Control` effect in functions like `foreach`, `foreachIndex` and similar, possibly providing a "simplified type" in the documentation like in the `lens` library in Haskell

jiribenes:
`effekt.MLTests.examples/ml/lib/seq.effekt` is currently broken as the MLTon backend (probably?) crashes [?]

effekt/jvm/src/test/scala/effekt/StdlibTests.scala
```diff
@@ -0,0 +1,48 @@
...
+}
+class StdlibLLVMTests extends StdlibTests {
+  def backendName: String = "llvm"
+
+  override lazy val ignored: List[File] = List(
+    // For every function tested using `immutable/result`:
+    // [error] Unsupported coercion from Exception47234[E48288] to Exception47234[OutOfBounds47515]
+    examplesDir / "stdlib" / "list" / "get.effekt",
+    examplesDir / "stdlib" / "list" / "modifyat.effekt",
+  )
```
jiribenes, line 47: cc @marzipankaiser Polymorphism boxing? 🤔

effekt/jvm/src/test/scala/effekt/StdlibTests.scala
```diff
@@ -0,0 +1,48 @@
...
+  def backendName: String = "chez-lift"
+}
+class StdlibMLTests extends StdlibTests {
+  def backendName: String = "ml"
+
+  override lazy val ignored: List[File] = List(
+    // For every function using `foreach`:
+    // [error] Effect polymorphic recursion is not allowed. The following definition is effect polymorphic since unification variable X occurs in instantiation ...
+    examplesDir / "stdlib" / "list" / "join.effekt",
+    examplesDir / "stdlib" / "list" / "flatmap.effekt",
```
jiribenes, line 36: cc @b-studios 
This is because we use
```
var reminder = l
loop { // recursive from `Control`
  // ...
}
```
but we can't do that in the MLTon backend.

What's the best way to go about this? Should we:
1. not use `Control` at all [or at least not in the MLTon backend]
2. write a better version
3. ignore the issue for now

libraries/ml/effekt.effekt
```diff
@@ -196,24 +196,23 @@ extern io def panic[R](msg: String): R =
   "raise Fail msg"
 
 effect Exception[E] {
-  def raise(exception: E, msg: String): Nothing
+  def raise[A](exception: E, msg: String): A
 }
```
jiribenes, line 200: This is a notable change since all of the other backends used `def raise[A](exception: E, msg: String): A`, but the ML backend used `def raise(exception: E, msg: String): Nothing`.
What's the better choice?

effekt/jvm/src/test/scala/effekt/StdlibTests.scala
jiribenes: Here's the relevant code I'm talking about:
https://github.com/effekt-lang/effekt/pull/355/files#diff-f8dcc2131b31d921d0b18b09a86024e013b5b6902d7a0dc065a3f71eacf941deR111-R122

libraries/ml/effekt.effekt
b-studios: That's a very good question. The only reason we do this in MLton is since it cannot monomorphize rank-2. This will not be the last rank-2 operation, so we need a general strategy for dealing with this.

effekt/jvm/src/test/scala/effekt/StdlibTests.scala
b-studios: To be honest, long term I would like to reimplement MLTon's optimizations in our backend and then either move to OCaml or replace it by LLVM / RiscV / Wasm. MLton was a great way of showing what is possible with WPO, but was only ever a means not the end.

libraries/ml/effekt.effekt
jiribenes: Although thinking about the possible backends (and compilation strategies) is important, I feel like the question here is a semantic one: what is the _ideal_ version of `raise`? Returning A for all A or returning Nothing? Which variant is nicer/easier to work with?

b-studios: Of course in terms of parametricity, both are the same. For this particular case I might actually prefer `Nothing` since it is more explicit and less indirect. We need to make sure that the subtyping has the right ergonomics in the concrete implementation, though.

effekt/jvm/src/test/scala/effekt/LLVMTests.scala
```diff
@@ -102,5 +101,8 @@ class LLVMTests extends EffektTests {
   override lazy val ignored: List[File] = List(
     // Issue #207
     examplesDir / "llvm" / "polymorphism_blockparams.effekt",
+
+    // See PR #355
+    examplesDir / "llvm" / "string_toint.effekt",
```
jiribenes, line 106:
```
$ effekt -b --backend llvm examples/llvm/string_toint.effekt
opt: ./out/string_toint.ll:895:9: error: multiple definition of local value named 'tmp223018_2298'
        %tmp223018_2298 = load %Pos, ptr %tmp223018_2298p_189
        ^
```
No idea why this happens in the LLVM backend on this branch, cc @phischu

jiribenes:
Rebased to current `master`.

jiribenes:
Should be ready for a final review: let's finally unwrap this gift and enjoy the contents :)

libraries/js/effekt.effekt
```diff
@@ -203,24 +203,25 @@ extern io def panic[R](msg: String): R =
...
 
 effect Exception[E] {
-  def raise[A](exception: E, msg: String): A
+  def raise(exception: E, msg: String): Nothing
 }
 record RuntimeError()
 
-def raise[A](msg: String): A / Exception[RuntimeError] = do raise(RuntimeError(), msg)
+def raise[A](msg: String): A / Exception[RuntimeError] = do raise(RuntimeError(), msg) match {}
+def raise[A, E](exception: E, msg: String): A / Exception[E] = do raise(exception, msg) match {}
```
jiribenes, line 211: The ML backend had these, I just wanted to add them for compatibility reasons.

libraries/llvm/text/string.effekt
```diff
@@ -57,16 +57,16 @@ def indexOf(str: String, start: Int, substr: String): Option[Int] = {
...
    if(c == "\x22") {
     "\\\x22"
    } else if (c == "\\") {
     "\\\\"
    } else c
  } ++ "\x22"
-def show(l: List[String]): String = showList(l){e => show(e)}
-def show(o: Option[String]): String = showOption(o){ e => show(e) }
+def showQuoted(l: List[String]): String = showList(l){e => showQuoted(e)}
+def showQuoted(o: Option[String]): String = showOption(o){ e => showQuoted(e) }
```
jiribenes, line 69: No idea if somebody depends on this, but without this change, I get errors as we define `show` for lists in `immutable/list` and `show` for options in `immutable/option`.
cc @phischu

effekt/jvm/src/test/scala/effekt/StdlibTests.scala
jiribenes: With #383, we could now make the `Control` code a little simpler, potentially.

effekt/jvm/src/test/scala/effekt/StdlibTests.scala
```diff
@@ -0,0 +1,48 @@
...
+  )
+}
+class StdlibLLVMTests extends StdlibTests {
+  def backendName: String = "llvm"
+
+  override def ignored: List[File] = List(
+    // For every function tested using `immutable/result`:
+    // [error] Unsupported coercion from Exception47234[E48288] to Exception47234[OutOfBounds47515]
+    examplesDir / "stdlib" / "list" / "get.effekt",
+    examplesDir / "stdlib" / "list" / "modifyat.effekt",
```
b-studios, line 46: This is somethin @marzipankaiser might be able to figure out.

libraries/js/immutable/list.effekt
```diff
@@ -58,92 +179,440 @@ def flatMap[A, B](l: List[A]) { f : A => List[B] }: List[B] = {
...
   var rights: List[A] = Nil()
   l.foreach { el =>
-    if (pred(el)) { lefts = Cons(el, lefts) }
-    else { rights = Cons(el, rights) }
-  };
+    if (pred(el)) {
+      lefts = Cons(el, lefts)
+    } else {
+      rights = Cons(el, rights)
+    }
```
b-studios, line 605: This could (in principle) now be written using our new pattern matching syntax.

b-studios:
In general this looks good to me and I will merge. However, there are a few things that are strange (poly boxing, LLVM tmps, etc.)

Also, the one thing that worries me the most are the immense build times. We need to start looking into speeding up the compiler itself, now.

Issue #157, opened 2022-09-27
Check return type annotation of object operations


jiribenes:
Following up on #148, we should type check the return type annotation of object operations.

As of right now, we reject all return type annotations given by the user on an object operation.
Instead, we should properly check both the return type and the given set of effects.
There is some subtlety required to do this properly -- capability passing should be informed by the declared effects,
not by the user-annotated effects. See the discussion in #148 for more details.

Relevant code to be changed:
https://github.com/effekt-lang/effekt/blob/a6201f56e7c33d7674708d192e9ee39b41a2fc09/effekt/shared/src/main/scala/effekt/Typer.scala#L357-L368

b-studios:
I just want to mention that one intermediate step might be to implement block parameters on interfaces and objects (but not handlers, yet). This would make it much easier to implement tests that involve capture sets.

PR #361, opened 2024-01-16
Fix: Object typing for bidirectional effects

### LINKED ISSUE #152: Effectful signatures are not implemented for objects
It looks like we forgot to implement capability passing for the new object feature:

```
effect Exc(): Unit

interface MyClass {
  def get(): Int / { Exc }
}

def impl = new MyClass {
  def get() = { do Exc(); 42 }
}
```

The same mechanism does already work for bidirectional effects:

```
effect Exc(): Unit

interface MyClass {
  def get(): Int / { Exc }
}

def impl() = 
  try { o.get() } 
  with o: MyClass {
    def get() = { resume { do Exc(); 42 } }
  }
  with Exc { 0 }
```
we simply forgot to implement it for objects.

@jiribenes this is related to what we discussed before (#148).

dvdvgt:
This issue is to close #152.

## Description

Check #152 for details.

## Tasks

- [X] Get simple example to compile
- [X] Add block parameters to operations
- [X] Write tests and find issues with current implementation
- [ ] Discuss solution for #157 

## Changelog

- **Breaking**: Deprecate `effect ... { ... }` notation in favour of `interface ... { ... }`. `effect ...` is now only allowed for singleton interfaces.
-  Add support for block parameters for operations


effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -4,10 +4,10 @@ package typer
 /**
  * In this file we fully qualify source types, but use symbols directly
  */
-import effekt.context.{ Annotation, Annotations, Context, ContextOps }
+import effekt.context.{Annotation, Annotations, Context, ContextOps}
 import effekt.context.assertions.*
-import effekt.source.{ AnyPattern, Def, IgnorePattern, MatchPattern, ModuleDecl, Stmt, TagPattern, Term, Tree, resolve, symbol }
-import effekt.symbols.*
+import effekt.source.{AnyPattern, Def, Effectful, IgnorePattern, MatchPattern, ModuleDecl, Stmt, TagPattern, Term, Tree, resolve, symbol}
+import effekt.symbols.{BlockType, Capture, *}
```
b-studios, line 10: This was probably added by accident in IntelliJ

b-studios:
I minimized one of the failing tests to

```
interface Compose {
    def compose[A, B] {g: A => B}: A => B at {g}
}

def handler {prog: () => Unit / Compose} =
  try { prog() } with Compose {
    def compose[F, G]() = ()
  }

def main() = handler { do compose[Int,  Int] { x => 1 }; () }
```

In fact, bidirectional handlers are not yet implemented in backends other than JS (I don't know about LLVM, TBH; @phischu ?).

b-studios:
I think it is fine to ignore the failing test for the ML backend for now and document it as "mutual recursion not supported".

PR #370, opened 2024-01-24
Directly forward execution output to stdout

### LINKED ISSUE #336: println output should not be buffered during execution
using panic("hello world") will swallow ALL other print statements of the execution, making debugging it impossible
having an infiinite loop or recursion will swallow ALL print statements, making debugging impossible

both issues are caused by console output not being directly passed to the console but being buffered (or something) until the execution finished.
with the absence of a stepper, println is my only way to debug other than "wir lösen das problem durch anstarren" :smile:

marzipankaiser:
Fixes #336 (since I was looking into the Process API r.n. anyway).

Issue #336 is caused by the Process being started using `.!!` which:
```
Starts the process represented by this builder, blocks until it exits, and returns the output as a String. Standard error is sent to the console. If the exit code is non-zero, an exception is thrown.
```

We now use a custom `ProcessLogger` and also connect stdin.
This makes this mostly work, apart from potential race conditions between input and output. (Note: `.!<` would work well in that regard but breaks the tests.)

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Runner.scala
+++ b/effekt/jvm/src/main/scala/effekt/Runner.scala
@@ -54,11 +54,10 @@ trait Runner[Executable] {
   /**
    * Runs the executable (e.g. the main file) by calling the build function.
    */
-  def eval(executable: Executable)(using Context): Unit = {
+  def eval(executable: Executable)(using C: Context): Unit = {
     val execFile = build(executable)
-    val exitCode = Process(execFile).!<
-    if(exitCode != 0) {
-      Context.error(s"Process exited with exit code ${exitCode}.")
+    Process(execFile).lazyLines.foreach{ line =>
+      C.config.output().emitln(line)
     }
   }
 
--- a/effekt/jvm/src/main/scala/effekt/Runner.scala
+++ b/effekt/jvm/src/main/scala/effekt/Runner.scala
@@ -54,8 +54,13 @@ trait Runner[Executable] {
   /**
    * Runs the executable (e.g. the main file) by calling the build function.
    */
-  def eval(executable: Executable)(using Context): Unit =
-    exec(build(executable))
+  def eval(executable: Executable)(using Context): Unit = {
+    val execFile = build(executable)
+    val exitCode = Process(execFile).!<
+    if(exitCode != 0) {
+      Context.error(s"Process exited with exit code ${exitCode}.")
+    }
+  }
 
   def canRunExecutable(command: String*): Boolean =
     try {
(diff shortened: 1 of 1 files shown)


marzipankaiser:
Apparently this is used in some way during testing. Am investigating.

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -54,8 +54,26 @@ trait Runner[Executable] {
...
+        System.out.flush()
+      }
+
+      override def err(s: => String): Unit = System.err.println(s)
+
+      override def buffer[T](f: => T): T = f
+
+    }, connectInput = true).exitValue()
+
+    if(exitCode != 0) {
```
b-studios, line 73: nitpick
```suggestion
    if (exitCode != 0) {
```

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -54,8 +54,26 @@ trait Runner[Executable] {
...
-  def eval(executable: Executable)(using Context): Unit =
-    exec(build(executable))
+  def eval(executable: Executable)(using C: Context): Unit = {
+    val execFile = build(executable)
+
+    val exitCode = Process(execFile).run(new ProcessLogger {
+
+      override def out(s: => String): Unit = {
+        C.config.output().emitln(s)
+        System.out.flush()
```
b-studios, line 64: Do we need to flush each time?

b-studios:
It's great that you made progress here. But I would have hoped to just have a thin wrapper that acts exactly like programming against the host native interface.

effekt/jvm/src/main/scala/effekt/Runner.scala
marzipankaiser: We can probably drop this; it is left over from trying to get the ordering right when printing not-full-lines.
`emitln` should flush anyway.

marzipankaiser:
This was mostly a product of me reading into the process API docs anyway.

If I understand that correctly, then we'd need to change how tests are run (because they currently check the output in `C.config`). IMHO, we might want to run tests differently anyway (esp. w.r.t. stdin and stdout), but I'm not sure how big of a change this would be. - Then, using `.!<` should do exactly what we'd want here (just use stdin/out/err for in/out/err), no?

effekt/jvm/src/main/scala/effekt/Runner.scala
b-studios, line 64:
```suggestion
```

PR #350, opened 2023-12-12
Feature/extern unquotes

b-studios:

PR #374, opened 2024-01-26
Website breaks to missing extern definitions

b-studios:
All examples on the website output

> println$impl is not defined

The reason is that recently introduced inlining of extern definitions #350 will inline definitions like `println`

https://github.com/effekt-lang/effekt/blob/5dffce2c7f61b03c5038ddfe9d3c1f0fc0ae2ae1/libraries/js/effekt.effekt#L14-L15

which in turn rely on extern includes or literals, such as

https://github.com/effekt-lang/effekt/blob/5dffce2c7f61b03c5038ddfe9d3c1f0fc0ae2ae1/libraries/js/effekt.effekt#L8

The problem now is that the website uses separate compiling, where the extern includes are only visible to the one module including them. Previously, `println` would simply **close over** the extern includes. But now, since we inline it, it cannot anymore.

## Solution implemented in this hotfix

Inlining externs now differs (in JS) between separate and whole program compilation.
- When compiling whole-program, we inline extern defs.
- When compiling separate, we do **not** inline extern defs, since:
       
## Drawback

- it adds complexity in the implementation of the javascript Transformer
- the hotfix looses some code-sharing with the direct-style backend, which currently is not used (neither in tests, nor on the website), since I moved some code into the "Monadic..." variants to easier control inlining.

### DIFF
--- a/effekt/js/src/main/scala/effekt/LanguageServer.scala
+++ b/effekt/js/src/main/scala/effekt/LanguageServer.scala
@@ -1,7 +1,7 @@
 package effekt
 
 import effekt.context.{ Context, VirtualFileSource, VirtualModuleDB }
-import effekt.generator.js.TransformerMonadic
+import effekt.generator.js.TransformerMonadicSeparate
 import effekt.lifted.LiftInference
 import effekt.util.{ PlainMessaging, getOrElseAborting }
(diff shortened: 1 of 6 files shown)


b-studios:
@dvdvgt once the PR is merged, could you update the website again and see whether this resolves the problem? The test I added should actually reproduce this error, though.

PR #376, opened 2024-01-27
Do not ignore type parameters of receiver in method calls

### LINKED ISSUE #373: Function call type checking for operations
## Description

When calling operations, the argument types are of course checked against the declared type. But which actually is the declared type if there are multiple definition of an operation given by the following?

```
interface Eff[A] {
  def op(x: A): A
}
```

Consider this object:

```
def impl: Eff[Int] = new Eff[Int] {
  def op(x: Int) = x / 1
}
```

When calling `impl.op("hello")`, the function type for the symbol of `op` is looked up. However, the function type is annotated as `[A](x: A): A` _once_. Thus, when checking the function call, the call is only ever checked with respect to the original declared type and not all potentially existing types of concrete implementations.

One solution could be to annotate each concrete type during typechecking the implementation of `Eff`, but since we only have exactly one symbol for all possible implementations of `op`, updating the type of `op` overwrites the previous annotation.

Note that this is not an issue exclusive to objects, but to also for handlers.

```
try { eff.op("hello") }
with Eff[Int] { def op(x) = resume(x / 1) }
```

Both of these examples compile and yield `NaN`.

## Idea

We need to have unique symbols for each operation implementation and then annotate the specific type for each implementation to this unique symbol.

## Relevant

- https://github.com/effekt-lang/effekt/blob/5dffce2c7f61b03c5038ddfe9d3c1f0fc0ae2ae1/effekt/shared/src/main/scala/effekt/Typer.scala#L364
  We need to bind this type to the operation here: `Context.bind(op.symbol.asOperation, <>)`. This overwrites previous annotations, though.

- https://github.com/effekt-lang/effekt/blob/5dffce2c7f61b03c5038ddfe9d3c1f0fc0ae2ae1/effekt/shared/src/main/scala/effekt/Typer.scala#L176-L177
- https://github.com/effekt-lang/effekt/blob/5dffce2c7f61b03c5038ddfe9d3c1f0fc0ae2ae1/effekt/shared/src/main/scala/effekt/Typer.scala#L910-L912

b-studios:
For now I am special casing the treatment of method calls to fix #373. What I do not like about this solution:

- previously there was no `freshTypeVar` in `Typer`--it all was encapsulated behind `instantiate`.
- now the treatment of method and function calls is not entirely uniform and slightly differs.

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/ConstraintTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/ConstraintTests.scala
@@ -33,7 +33,7 @@ abstract class ConstraintTests extends munit.FunSuite {
 
 
   def freshTypeVar(name: String) =
-    scope.fresh(TypeParam(Name.local(name)), NoSource)
+    scope.freshTypeVar(TypeParam(Name.local(name)), NoSource)
 
   def freshCaptVar(name: String) =
     scope.freshCaptVar(CaptUnificationVar.VariableInstantiation(CaptureParam(Name.local(name)), NoSource))
--- a/effekt/shared/src/main/scala/effekt/Typer.scala
+++ b/effekt/shared/src/main/scala/effekt/Typer.scala
@@ -950,13 +950,41 @@ object Typer extends Phase[NameResolved, Typechecked] {
 
     val Result(recvTpe, recvEffs) = checkExprAsBlock(receiver, None)
 
-    val interface = recvTpe.asInterfaceType.typeConstructor
+    val interface = recvTpe.asInterfaceType
     // filter out operations that do not fit the receiver
-    val candidates = methods.filter(op => op.interface == interface)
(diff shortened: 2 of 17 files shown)


b-studios:
To test this fix I added a few variations of #373. @dvdvgt could you double-check that this fix is working?

dvdvgt:
Looks good to me and also seems to reject the "crazy" examples. One note, though, you only tested this using handlers, however I think it would be more robust if we had at least one negative test for objects as well. While they should do the same, the implementation is different and we should also for this issue there. Besides, adding the more elaborate examples as negatives test too shouldn't hurt, right?

b-studios:
Yeah I thought about adding them as well, but then they are really only instances of the tests already there. Each test, at least, adds time to CI :) (which is an excuse for me not to add it)

Issue #375, opened 2024-01-27
Make writing neg-tests easier and more robust


b-studios:
If I understand correctly, the Scala compiler uses the following strategy to specify neg-tests:

- either there is a check file with expected output
- or the source file has one or multiple annotations as comments

We could do similar:
```scala
def main() = {
   1 + "test" // ERROR
}
```

That is, we check whether there is an error, reported on that line.

This is more robust since it does not look into the message and a lot easier to specify then creating a separate `.check` file.

b-studios:
Later, we could also support different kinds of comments that allow checking for more (e.g., is this a parse error, typer, etc.)

jiribenes:
Related work is LLVM's `lit` tool: here's an example test: https://github.com/llvm/llvm-project/blob/main/clang/test/Lexer/header.cpp

b-studios:
Closing for now. Of course we can keep discussing more advanced implementations :)

PR #377, opened 2024-01-27
Improve Test Runner

b-studios:
This PR addresses  #375 and supports the following formats:

Running tests in `examples/neg` depends on whether there is a check file or not:

1. `.check` file exists: `.effekt` program will be compiled and run; output is compared against `.check`
2. `.check` file does not exist: `.effekt` program will be compiled. The following checks will be performed
  - if no line has an error annotation, then it is checked whether an error exists at all.
  - for each line that ends in either `// ERROR` or `// WARN` a corresponding error or warning is searched for.

Additionally each `// ERROR` and `// WARN` comment can be followed by an excerpt of the error message that is expected. 

https://github.com/effekt-lang/effekt/blob/a103be8e5fcfafebf1c55e95a3648329ae211a25/examples/neg/blocks_wrong_arguments.effekt#L1-L18

The trailing message will be searched **verbatim** in the error messages.

_Additional_ errors that are not marked with a comment are currently ignored.

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
@@ -9,7 +9,7 @@ import scala.language.implicitConversions
 
 abstract class ChezSchemeTests extends EffektTests {
 
-  override def included: List[File] = List(
+  override def positives: List[File] = List(
     examplesDir / "pos",
     examplesDir / "casestudies",
     examplesDir / "chez",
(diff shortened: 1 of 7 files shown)


b-studios:
@dvdvgt This might be helpful for you to test typer more easily.

b-studios:
@jiribenes You mentioned the LLVM test runner. Is there anything you think would be super helpful in addition to what I implemented?

b-studios:
Once the tests go through I will merge this since I already want to use it on another PR :)

PR #378, opened 2024-01-27
Warn if a type parameter shadows a non-parameter

### LINKED ISSUE #368: Add warning when shadowing types
Have a warning when using a name as a type parameter which is already in scope (as a type).

## More detailed
More than once we and others stumbled into programs like:
```
effect Foo[A](x: A): A
//...
def bar(): Int = 
try {
  //...
} with Foo[Int] { 
  (x: Int) => 12
}
```
Which errors out with something like:
```
Expected Int but got Int.
```
which is hard to understand (and due to `Int` being bound at `with Foo[Int]`, shadowing `Int`).

We should output a warning here.

b-studios:
This tries to fix #368 by emitting a warning in case a type parameter shadows an outer type, which is _not_ a type parameter (since this sometimes is desired and could lead to too many false negatives).

An example can be seen here:

https://github.com/effekt-lang/effekt/blob/2c43c68d1e45f60573515bb21005d9251c47bb3c/examples/neg/namer/shadowing-types.effekt#L1-L17

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Namer.scala
+++ b/effekt/shared/src/main/scala/effekt/Namer.scala
@@ -293,7 +293,7 @@ object Namer extends Phase[Parsed, NameResolved] {
     case source.InterfaceDef(id, tparams, operations, isEffect) =>
       val effectSym = Context.resolveType(id).asControlEffect
       effectSym.operations = operations.map {
-        case source.Operation(id, tparams, params, ret) =>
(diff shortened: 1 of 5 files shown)


b-studios:
TBH while going through this code I was tempted to refactor `Scope` to be an `enum`. This code is older than Scala 3 and writing it in a functional style might be clearer.

PR #379, opened 2024-01-28
Fix #354 by not exporting extern resources

### LINKED ISSUE #354: Online evaluator crashes on extern resources
When I go to https://effekt-lang.org/quickstart.html and enter the following program

```
interface Greet { }

extern resource my_resource : Greet

def helloWorld() = ""
```

and in the REPL field I enter

```
helloWorld()
```

then the output is

```
ReferenceError: my_resource_1396 is not defined
```

instead of the result.

b-studios:
The problem underlying the error 

> ReferenceError: my_resource_1396 is not defined

observed by @phischu in #354 was that extern resources have been "exported" in the separate-compilation-based Javascript backends. That is, JS code like

```javascript
module.export = {
  // ...
  my_resource_1396: my_resource_1396
}
```
was generated, but extern resources don't have a term-level representation so the right-hand side of `my_resource_1396` was unbound.

#### Solution Attempt

First attempt to solve #354 the problem was
```diff
  def shouldExport(sym: Symbol)(using D: DeclarationContext): Boolean = sym match {
    // do not export fields, since they are no defined functions
    case fld if D.findField(fld).isDefined => false
    // do not export effect operations, since they are translated to field selection as well.
    case op if D.findProperty(op).isDefined => false
+   // do not export extern resources, since they do not exist on the term-level
+   case r: symbols.ExternResource => false
    // all others are fine
    case _ => true
  }
```
However, after `core` we should not match on symbols anymore; so this is not a good solution, since it is not stable under renaming.

In particular, we currently do not include extern resources in the core representation:
https://github.com/effekt-lang/effekt/blob/3aa78e8795b8adb1d347827a3f2b004c2b783be2/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L111-L112

Same holds for `ExternType`, `ExternResource`, `ExternInterface`

#### Chosen Solution

Instead, now I do not use `module.exports` on the symbol, anymore. We should try to avoid this after going to `core`, anyways.

This also helped simplifying the separate compilation aspect of JS a bit (3c26361) since now we do not need to filter twice. I am a little bit afraid that this might introduce a regression, but checked that the tests actually caught this before performing the refactoring.

### DIFF
--- a/effekt/js/src/test/scala/effekt/WebTests.scala
+++ b/effekt/js/src/test/scala/effekt/WebTests.scala
@@ -37,9 +37,13 @@ object WebTests extends TestSuite {
         mod.module
       }
 
-    def evaluate[A](imports: List[String], content: String) = {
-      server.writeFile("interactive.effekt", imports.map(i => s"import $i").mkString("\n") + s"\n\ndef main() = ${content}")
+    def evaluate[A](imports: List[String], content: String) =
+      run[A](imports, s"\n\ndef main() = { ${content} }\n")
+
+    def run[A](imports: List[String], content: String) = {
+      server.writeFile("interactive.effekt", imports.map(i => s"import $i").mkString("\n") + s"\n\n${content}")
       val mainFile = server.compileFile("interactive.effekt")
+
       load(mainFile.replace("out/", "")).main().run().asInstanceOf[A]
     }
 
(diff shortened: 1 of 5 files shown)


b-studios:
Now I do not use `module.exports` on the symbol, anymore. We should try to avoid this after going to `core`, anyways.

It also helped simplifying the separate compilation aspect of JS a bit since now we do not need to filter twice. I am a little bit afraid that this might introduce a regression, but checked that the tests actually caught this before performing the refactoring.

PR #381, opened 2024-01-28
Also check wellformedness of effects in namer

### LINKED ISSUE #281: Handling of type parameters for effects
If we declare an effect with type parameters, e.g.
```effekt
effect Eff[A] {
  def use(): A
}
```

Then the number of type parameters is not checked at the usage site, so all of those compile:
```effekt
def foo(): Int / Eff = <>
def foo2(): Int / Eff[Int, Unit] = <>
```

Only if we ever use `do use`, it will complain that `Eff[Int]` is not allowed in that context, or that the number of type parameters does not match up *for the use call*. 
EDIT: If we ever try to define a handler with `try ... with ...`, then we get the appropriate error message there.

### LINKED ISSUE #352: Misleading error message when using parametric effects
The following code:
```scala
effect Emit[A](x: A): Unit
def emit[A](x: A): Unit / Emit[A] = do Emit[A](x)
```

returns the error message:
```
Effect Emit is not allowed in this context.
```
pointing to the RHS of the second declaration:
Screenshot 2023-12-19 at 11 58 01

---

The correctly to write the program is to not let the `Emit` in the return type take a type parameter:
```
effect Emit[A](x: A): Unit
def emit[A](x: A): Unit / Emit = do Emit[A](x)
```

It would therefore be nice to provide a better error message (and to report it in the return type where it stems from).

b-studios:
Fixes  #281 and also fixes #352 by inserting a single wellformedness check.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Namer.scala
+++ b/effekt/shared/src/main/scala/effekt/Namer.scala
@@ -643,7 +643,7 @@ object Namer extends Phase[Parsed, NameResolved] {
    * Resolves an interface type, potentially with effect aliases on the top level
    */
   def resolveWithAliases(tpe: source.BlockTypeRef)(using Context): List[InterfaceType] = Context.at(tpe) {
-    tpe match {
+    val resolved: List[InterfaceType] = tpe match {
       case source.BlockTypeRef(id, args) => Context.resolveType(id) match {
         case EffectAlias(name, tparams, effs) =>
           if (tparams.size != args.size) {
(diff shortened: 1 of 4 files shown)


b-studios:
This resulted in a CI failure on master:

```
java.lang.AssertionError: assertion failed: Cannot add signature for interface twice: Exception
    at effekt.lifted.mono.FlowAnalysis.addInterface(Analyzer.scala:38)
    at effekt.lifted.mono.Analyzer$package$.analyze(Analyzer.scala:341)
    at effekt.lifted.mono.Analyzer$package$.analyze$$anonfun$1(Analyzer.scala:93)
```

Is this yet another flaky CI?

Issue #383, opened 2024-01-28
Proposal: Pattern Matching


b-studios:
_this is still work in progress_

Here I propose quite a lot of changes to pattern matching and control flow structures (`if` and `while`). All of the changes will happen in the front end. In particular, the representation of core and all backends should stay largely unaffected.

## Current Implementation
We currently implement pattern matching in the style of [Jules Jacobs](https://julesjacobs.com/notes/patternmatching/patternmatching.pdf).

From an implementors perspective, this is particularly attractive since compiling pattern matching automatically gives us exhaustiveness and redundancy checks. Pattern matching is compiled in the translation form [source to core](https://github.com/effekt-lang/effekt/blob/3634fe2cda7962f7e3910a7400cdec282036df75/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L612).

https://github.com/effekt-lang/effekt/blob/3634fe2cda7962f7e3910a7400cdec282036df75/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L579-L590

The internal representation of a single clause is the _conjunction_ of potentially several scrutinees  with the expected pattern, followed by the clause to be executed in case the patterns are all satisfied.

## Current Limitations

User-facing, we currently only support matching on data type constructors, such as

```scala
e match {
  case Some(v) => ...
  case None => ...
}
```

In particular, we do not implement
- guards (that is, `case Some(v) if v > 0 => ...`)
- as-patterns (that is `case Some(v) as x => ...` or `case x @ Some(v) => ...`
- literal patterns (that is `case 1 => ...`).

As @phischu noted in #363, we can easily support literal patterns if we support guards.

Patterns, currently, can only occur in a match expression. Some languages (like Rust's [while let](https://doc.rust-lang.org/rust-by-example/flow_control/while_let.html)) support matching as part of conditionals such as:

```scala
while (find(x) is Some(v)) {
  ...
}
```

This is sometimes more efficient, since it prevents us from having to call `find` twice
```scala
while (find(x).isDefined) {
 ...  find(x).get ....
}
```
Of course we can always write this explicitly as a recursive function, which is more noisy:
```scala
def loop(): Unit = find(x) match {
  case Some(v) =>  ...; loop()
  case None() => ()
}
loop()
```

## Proposal
This proposal aims to address the following issues:

- Support literal patterns
- Support guards
- Support matching in conditionals (while and if)

while retaining most of the simple implementation style of Jules.

[...]

PR #385, opened 2024-01-29
Implement pattern matching with literals and guards

b-studios:
This PR implements proposal #383. Notes about the implementation can be found at the top of

- [`PatternMatchingCompiler.scala`](https://github.com/effekt-lang/effekt/blob/551053f5786c0943c07a93ea8361389ebce52cef/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala#L12)
- [`ExhaustivityChecker.scala`](https://github.com/effekt-lang/effekt/blob/551053f5786c0943c07a93ea8361389ebce52cef/effekt/shared/src/main/scala/effekt/typer/ExhaustivityChecker.scala)

The implementation of the compiler was rather modest; exhaustivity checking was surprisingly tricky since it is difficult to maintain the necessary information (like types to check exhaustivity of nested scrutinees).

Both implementations (compiler and checker) are structure to 

1. preprocess clauses into an internal representation (transformation requires `Context`)
2. perform the actual work (transformation does NOT require `Context` and can thus be tested more easily)

Note: for now, I decided that Or-Patterns are not worth it.

---
Implemented in frontend
- [x] pattern guards
- [x] boolean guards
~- [x] or patterns (properly, which requires significant changes to always bind the SAME symbol)~
- [x] literal patterns
- [x] guards on if and while
- [x] exhaustivity check

pattern compiler
- [x] guards
~- [ ] or patterns~
- [x] literal patterns
- [x] heuristic
- [x] compilation of if and while
~- [ ] fuse subsequent predicates using `&&` but also make `&&` by-name in its arguments.~

translation
- [x] default clause in matches
- [x] non trivial guards in while and if


b-studios:
Reminder to future me: if you want to add Or-Patterns again, start by reverting [d4452a3](https://github.com/effekt-lang/effekt/pull/385/commits/d4452a326e4974f504936cbb5a738f24c76816bc).

b-studios:
To fix in the future: Using `Map[ValueVar, Pattern]` as a semantic domain for patterns in the compiler is wrong. It should be `List[(ValueVar, Pattern)]` since we might have multiple matches on the same scrutinee.

PR #386, opened 2024-02-02
Add examples from OOPSLA-2022 artifact as tests

### LINKED ISSUE #63: Integrate System C into Effekt
As part of our work on the OOPSLA'22 paper

https://se-tuebingen.github.io/oopsla-2022-artifact/

, we developed System C, a language similar to Effekt, but with explicit capability passing.

As an important next step, we need to integrate System C with Effekt master.

This means:

- coming from System C, we need to incorporate effect inference and capability passing transformation
- coming from Effekt, we need to add explicit capability passing and named capabilities
- the region checker in Effekt currently has a much better error reporting than the unification based implementation in System C

The effect system of System C is already largely implemented in Effekt (as a region checker). However, lessons learnt from our work on the paper need to be backported and incorporated into Effekt.

It is not just a matter of merging the two implementations. Instead, we need to find a design that integrates explicit and implicit capability passing nicely. Ideally it also should be integrated with a module system and explicit name spacing.

The System C code base currently lives in the branch: https://github.com/effekt-lang/effekt/tree/system-c-steps

In PR #59 I started cherry-picking some commits -- but there is much more to do.

dvdvgt:
## Description

Closes #63 by porting the examples from the [artifact](https://se-tuebingen.github.io/oopsla-2022-artifact/) and adding them as tests. I also had a look at the case studies listed there, but it seems there are already tests that closely resemble them.

Since [some examples](https://se-tuebingen.github.io/oopsla-2022-artifact/paper/section3.html) were using the [`timestamp` function](https://github.com/effekt-lang/effekt/blob/3634fe2cda7962f7e3910a7400cdec282036df75/libraries/ml/effekt.effekt#L23-L24) and this is of course inherently impossible to test, I instead used a dummy counter instead.

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
@@ -77,7 +77,7 @@ class ChezSchemeLiftTests extends ChezSchemeTests {
     examplesDir / "benchmarks" / "generator.effekt",
     examplesDir / "pos" / "capture" / "regions.effekt",
     examplesDir / "pos" / "capture" / "selfregion.effekt",
-    // boxing is not (yet) supported for list-based backends
+    // boxing is not (yet) supported for lift-based backends
     examplesDir / "pos" / "file.effekt",
 
     // known issues:


dvdvgt:
Are these tests failing because of the usage of unsupported features of the respective backend, or rather are actually bugs?

b-studios:
Rank-2 not supported in ML
- `println[A]()` in [this test](https://github.com/effekt-lang/effekt/pull/386/files#diff-0a0026735f94abdb113543b83ac530b599039cb0dd13386406ef963e8e33b399R2) leads to [this error](https://github.com/effekt-lang/effekt/actions/runs/7756432120/job/21153715064#step:9:681)
   
- Boxing (especially non-empty captures) not supported in lift-based implementations (Chez lift, ML, and LLVM)
  boxing in [this test](https://github.com/effekt-lang/effekt/pull/386/files#diff-209fde35e0a61b84402ba7d1b20592b1956f9ca3761f03d174b58872eed64399R10) is not expected to work. 
  
This is why [this error](https://github.com/effekt-lang/effekt/actions/runs/7756432120/job/21153715064#step:9:708) shows up:
  
```diff
-[error] examples/pos/global_capabilities.effekt:74:35: Cannot typecheck call.
-There are multiple overloads, which all fail to check:
-
-Possible overload: effekt.println$4 of type String => Unit
-  Wrong number of type arguments 1
-
-Possible overload: effekt.println of type Int => Unit
-  Wrong number of type arguments 1
-
-Possible overload: effekt.println$2 of type Double => Unit
-  Wrong number of type arguments 1
-
-Possible overload: effekt.println$1 of type Unit => Unit
-  Wrong number of type arguments 1
-
-Possible overload: effekt.println$3 of type Boolean => Unit
-  Wrong number of type arguments 1
-
-        def myprintln[A](msg) = { println[A](msg); resume(()) }
-                                  ^^^^^^^^^^^^^^^
+Hello!
+Hello!
...
```
The ML backend needs to have monomorphic definitions of println. You could define it at `String` and then use the specific `show` function (on int for example).

examples/neg/global_caps_captures.effekt
```diff
@@ -0,0 +1,23 @@
+interface Exc { def throw(msg: String): Unit }
+interface Console { def println[A](a: A): Unit }
```
b-studios, line 2:
```suggestion
interface Console { def println(a: String): Unit }
```
would make it work on MLton

effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
```diff
@@ -77,6 +77,8 @@ class ChezSchemeLiftTests extends ChezSchemeTests {
     examplesDir / "benchmarks" / "generator.effekt",
     examplesDir / "pos" / "capture" / "regions.effekt",
     examplesDir / "pos" / "capture" / "selfregion.effekt",
+    // boxing is not (yet) supported for list-based backends
```
b-studios, line 80:
```suggestion
    // boxing is not (yet) supported for lift-based backends
```

PR #394, opened 2024-02-12
Add a generic structural comparison primitive to JS backend

jiribenes:
### Summary

In the JavaScript backend, we can use reflection to create a _generic_ structural comparison (function `genericCompare: ∀A. (A, A) => -1 | 0 | 1`). We can use such function for building more advanced data structures that require comparisons e.g., binary search trees.

### Why the strange name?

The name of the resulting function is `genericCompare` _on purpose_ to convey that it's just _some_ structural comparison, not necessarily a great one or a stable-across-versions one. Unison calls the same function [`Universal.ordering`](https://share.unison-lang.org/@unison/base/code/releases/2.14.1/latest/terms/Universal/ordering).

### Future

In the future, we might be able to add a `__compare` method to the JavaScript classes we generate, similarly to `__equals`, for example:
https://github.com/effekt-lang/effekt/blob/1227d5cd32b31554614ddc544ca662c010ba01fc/effekt/shared/src/main/scala/effekt/generator/js/Transformer.scala#L52-L62

This function might become deprecated if we add type classes in the future: #66.
Nevertheless, I believe this PR is still pretty good as a first step towards a richer standard library.


jiribenes:
Currently, this results in a fascinating error somewhere in templating code:
```
java.util.NoSuchElementException: None.get
    at scala.None$.get(Option.scala:627)
    at scala.None$.get(Option.scala:626)
    at effekt.symbols.Module.findPrelude(symbols.scala:66)
    at effekt.core.Transformer$.equalsFor$1(Transformer.scala:546)
    at effekt.core.Transformer$.transformPattern$1(Transformer.scala:570)
```

Here's the relevant code in `core.Transformer.equalsFor`:
https://github.com/effekt-lang/effekt/blob/1227d5cd32b31554614ddc544ca662c010ba01fc/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L545-L556

There's some weird interplay between the `infixEq` code and the newly added code, namely this part where we reuse `equals$impl`:
https://github.com/effekt-lang/effekt/blob/e1a3f4ca6b5126013c723d2697b57cf28d2c1aa1/libraries/js/effekt_builtins.js#L27-L29

b-studios:
Strangely it fails already in finding the prelude, not only infixEq. The `None.get` is the one in

```
dependencies.find(_.name.name == "effekt").get
```
that @marzipankaiser added at one point to look up things in the stdlib.

b-studios:
One more step towards the root cause: the module that we try to find the standard prelude in, IS the prelude ;)

libraries/js/effekt.effekt
```diff
@@ -127,6 +127,30 @@ extern pure def infixEq[R](x: R, y: R): Boolean =
...
+   Greater()
+ }
+
+ extern pure def genericCompareImpl[R](x: R, y: R): Int =
+   "compare$impl(${x}, ${y})"
+
+ // Compares two values of the same type, returning an `Ordering`.
+ def genericCompare[R](x: R, y: R): Ordering = {
+   genericCompareImpl(x, y) match {
+    case -1 => Less()
```
b-studios, line 147: Indeed, this will fail since we search the prelude in the *dependencies* not in the module itself.

b-studios:
Changing `findPrelude` to

```
def findPrelude: Module = {
    // we are the prelude itself
    if (name.name == "effekt") this
    else dependencies.find(_.name.name == "effekt").getOrElse {
      sys error "Cannot find prelude, this should not happen"
    }
  }
```

fixes the bug.

jiribenes:
Seems to work great, thanks @b-studios! :)

b-studios (approved):
LGTM

PR #398, opened 2024-02-27
Add types to symbol outline

### LINKED ISSUE #396: Add types to symbols outline in LSP
We infer types but then don't show them, which is sad.

The symbols outline could show them making it a bit more useful

image

This could for starters look very similar to my LLM experiments:

image

Ideally, structured a bit more clearly (that would need to be done on the VSCode end, though).

marvinborner:
Closes #396. I also added collapsible items for imports and namespaces.

image

I'm not sure how to get the types of sub-level symbols yet, so I just put a '?' when the type is not available. We also need to add some kind of caching so the outline doesn't flicker when the file can't be processed (ie. while actively typing).

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Server.scala
+++ b/effekt/jvm/src/main/scala/effekt/Server.scala
@@ -12,6 +12,8 @@ import kiama.output.PrettyPrinterTypes.Document
 
 import org.eclipse.lsp4j.{ Diagnostic, DocumentSymbol, SymbolKind, SymbolTag, ExecuteCommandParams }
 
+import scala.jdk.CollectionConverters._
+
 /**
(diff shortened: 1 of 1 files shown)


b-studios:
See also https://github.com/effekt-lang/effekt/tree/feature/llm

b-studios:
Hey @marvinborner what are the next steps here?

marvinborner:
I'm really not sure. As discussed, annotating the scopes adds a **lot** of allocations.
Also, I don't think I understand the `scope` data type fully, mapping every scope to LLM's `allBindings` as I did here was definitely the wrong approach as it recurses infinitely.

I think we need to find a better (and cached) solution, and also cache other stages first. Otherwise, merging (something like) this will decrease LSP's performance a lot.

marvinborner:
@timsueberkrueb

PR #416, opened 2024-03-19
Fix highlights for consoles with white text color

marvinborner:
On systems with white text color, highlights make the text invisible. This is the default for several of my machines. The issue can be easily fixed by also forcing the text/foreground color to be black using ANSI escape codes.

Before:
image

After:
image


jiribenes (approved):
LGTM, looks great! :shipit:

PR #417, opened 2024-03-19
Type omission for operation implementations and block arguments

### LINKED ISSUE #345: Type omission for computation literals with higher-rank-poly
## Description

Currently, it is possible to omit parameter types for computation literals like in this (minimal) example

```scala
def f() { g: Int => Int }: Int = g(42)
// ...
f { a => a + 1 }
// equals
f { (a: Int) => a + 1 }
```

However, with the proper implementation of higher-rank-polymorphism in #246, it should also be possible to omit the type parameter and parameter type:

```scala
def h() { g: [A](A) => A }: Unit = { g(42); () }
// ...
h { a => a } // error: Wrong number of type arguments, given 0, but function expects 1.
h { [A](a: A) => a } // checks out
```

We should investigate, whether this can be made possible.

### LINKED ISSUE #414: Omission of block/computation types in objects
Currently, the user has to repeat the type of all block/computation parameters when implementing an interface as an object. However, this should not be necessary, as the type is already defined by the interface's definition. It should be sufficient to look up the type given by the definition and check it against the definition. This would also be more inline with the omission of types for values in implementations.

## Example

```scala
interface Defer {
  def defer[A, B] { prog: () => A } { finally: A => B }: B
}

// --- currently ---
def deferImpl = new Defer {
  def defer[A, B] { prog: () => A } { finally: A => B } = {
    val res = prog()
    finally(res)
  }
}

// --- proposed ---
def deferImpl = new Defer {
  def defer[A, B] { prog } { finally } = { // no types here
    val res = prog()
    finally(res)
  }
}

def main() = {
  deferImpl.defer { 40 + 1 } { x => x + 1 }
}
```

## Relevant

https://github.com/effekt-lang/effekt/blob/9828d900754901c98a84163e987a7aea69545264/effekt/shared/src/main/scala/effekt/Typer.scala#L371

https://github.com/effekt-lang/effekt/blob/9828d900754901c98a84163e987a7aea69545264/effekt/shared/src/main/scala/effekt/Typer.scala#L402-L403

https://github.com/effekt-lang/effekt/blob/9828d900754901c98a84163e987a7aea69545264/effekt/shared/src/main/scala/effekt/Typer.scala#L413-L419

dvdvgt:
This closes #414 and closes #345 by allowing the omission of block/computation types when implementing objects. See the issue for further information.

## TODO

- [x] Currently, the implementation is still a bit "hacky". I just unsafely `get` all `BlockParam` types, which may lead to crashes. While the parser, in its current form, does only permit the omission of block types when parsing the definition of an object, this would be yet another implicit invariant we would need to uphold throughout the compiler.
- [x] Add tests
- [x] Make type parameters optional as well (?)


dvdvgt:
## Discussion

In the example

```scala
interface Defer {
  def defer[A, B] { prog: () => A } { finally: A => B }: B
}

def deferImpl = new Defer {
  def defer[A, B] { p } { f } = {
    val res = p()
    f(res)
  }
}
```

we still have to explicitly name the type parameters even though, in this case, they are not used since we omitted the block types. It might be nice to allow the omission of these as well. Then, it would be permitted to specify an object like this

```scala
interface It {
  def op[A_1, ..., A_N](x_1: A_1, ..., xn: A_N) { f_1: A_1 => A_2 } { f_2: A_2 => A_3 } ... { f_n : A_N-1 => A_N }: A_N
}

def interfaceImpl = new It {
  def op(x, ..., z) { f } ... { g } = s
}
```

b-studios:
Not that I recall, but you can always just "recreate" them:
```
TypeVar.TypeParam(tp.name)
```
It might be that they then lack the position information and other LSP relevant things, though.

dvdvgt:
The omission of the types of blocks in handlers turns out to be a bit more difficult. As far as I can see, there are two options:

1. Either we special-case the `resume` block in the compiler such that it is typed independently of other blocks and block types can only be omitted within the resume block's block argument.
2. or we allow the omission of block types for all block parameters but then have to sprinkle `getOrElse(Context.abort("expected block type here"))` everywhere.

What do you think?

b-studios:
Definitely number 2: if the type of a called higher-order function is known, we should be able to omit parameter types when calling it.

dvdvgt:
This should be ready now. The way this works now is that the parser only allow the omission of block types for operation implementations (naturally, this only affects objects) and function/block arguments (which includes `resume`). The reason for this is simple: we already know the type, either due to the operation declaration or the type signature of the call target. Thus, no type is needed and if one is given, we try to match these two. Since all other blocks are required to have types, we only need to get the aforementioned two cases right in the `Typer` and unsafe `get`s should be permittable.

Type parameters can also be omitted, however, then no types referring to these are allowed.

```scala
def Id { def id[A] { prog: () => A }: A }
def impl = new Id { def id { prog: () => A } = prog() } // error:  Could not resolve type A
def impl = new Id { def id[A] { prog: () => A } = prog() } // checks out!
```

examples/pos/type_omission_op.effekt
```diff
@@ -0,0 +1,52 @@
+module examples/pos/polymorphic/type_omission_op
+
+interface Eff[A] {
+	def eff[B, C](x: B, y: C) { f: B => A } { g: C => A }: A
```
b-studios, line 4:
```suggestion
    def eff[B, C](x: B, y: C) { f: B => A } { g: C => A }: A
```

effekt/shared/src/main/scala/effekt/source/ExplicitCapabilities.scala
```diff
@@ -137,6 +137,6 @@ object ExplicitCapabilities extends Phase[Typechecked, Typechecked], Rewrite {
   def definitionFor(s: symbols.BlockParam)(using C: Context): source.BlockParam =
     val id = IdDef(s.name.name)
     C.assignSymbol(id, s)
-    val tree: source.BlockParam = source.BlockParam(id, source.BlockTypeTree(s.tpe))
+    val tree: source.BlockParam = source.BlockParam(id, Some(source.BlockTypeTree(s.tpe.get)))
```
b-studios, line 140: This could be a map?

examples/neg/unresolved_type.effekt
```diff
@@ -0,0 +1,11 @@
+interface Defer {
+	def defer[A, B] { prog: () => A } { finally: A => B }: B
+}
+
+def impl = new Defer {
+	def defer { prog: () => A } { finally: A => B }: B = { // ERROR Could not resolve type A		
+		val res = prog()
```
b-studios, line 7: Taaaaabs :)

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -57,7 +57,7 @@ object Typer extends Phase[NameResolved, Typechecked] {
               // bring builtins into scope
               builtins.rootTerms.values.foreach {
                 case term: BlockParam =>
-                  Context.bind(term, term.tpe)
+                  Context.bind(term, term.tpe.get)
```
b-studios, line 60: I guess this is safe because we assume builtins are always annotated with their types, right?

Maybe document like:
```suggestion
                  Context.bind(term, term.tpe.getOrElse { INTERNAL_ERROR("Builtins should always be annotated with their types.") })
```

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -246,7 +247,7 @@ object Typer extends Phase[NameResolved, Typechecked] {
         // (4) Wellformedness checks
         val handlerFor = providedCapabilities.map { cap =>
           // all effects have to be concrete at this point in time
-          val concreteEffect = Context.unification(cap.tpe.asInterfaceType)
+          val concreteEffect = Context.unification(cap.tpe.get.asInterfaceType)
```
b-studios, line 250: How do we know that capabilities have a type here?

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -940,7 +953,7 @@ object Typer extends Phase[NameResolved, Typechecked] {
       }
       val bps = bparams.map { p =>
         val param = p.symbol
-        val tpe = param.tpe
+        val tpe = param.tpe.get
```
b-studios, line 956: What if this is not annotated? Can this happen?

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -1454,15 +1467,15 @@ trait TyperOps extends ContextOps { self: Context =>
 
   private [typer] def bindingCapabilities[R](binder: source.Tree, caps: List[symbols.BlockParam])(f: => R): R = {
     bindCapabilities(binder, caps)
-    capabilityScope = BindSome(binder, caps.map { c => c.tpe.asInterfaceType -> c }.toMap, capabilityScope)
+    capabilityScope = BindSome(binder, caps.map { c => c.tpe.get.asInterfaceType -> c }.toMap, capabilityScope)
```
b-studios, line 1470: Do we know by construction that there always will be a type?

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -1454,15 +1467,15 @@ trait TyperOps extends ContextOps { self: Context =>
...
     val result = f
     capabilityScope = capabilityScope.parent
     result
   }
 
   private [typer] def bindCapabilities[R](binder: source.Tree, caps: List[symbols.BlockParam]): Unit =
     val capabilities = caps map { cap =>
-      assertConcrete(cap.tpe.asInterfaceType)
+      assertConcrete(cap.tpe.get.asInterfaceType)
```
b-studios, line 1478: Sorry for all those comments, but how do we know that the type exists here?

effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -87,7 +87,7 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
           // like in asSeenFrom we need to make up cparams, they cannot occur free in the result type
           val capabilities = effects.canonical
           val tparams = tps.drop(tparamsInterface.size)
-          val bparamsBlocks = bps.map(b => transform(b.tpe))
+          val bparamsBlocks = bps.map(b => transform(b.tpe.get))
```
b-studios, line 90: This should be there, because interface declarations are annotated, right? Maybe we can document this here like above

effekt/shared/src/main/scala/effekt/Typer.scala
dvdvgt: Yes, we do either get the `caps` argument from `Context.freshCapabilityFor` where, by construction, a type is always given or from line 219 where we also ensure a block type is given (or abort if not). So we do not get a runtime error further down the line when making changes, I added the usual `Context.abort("type required...")`. Is this sufficient?

effekt/shared/src/main/scala/effekt/Typer.scala
dvdvgt: This function is only ever called by `bindingCapabilities` where we already ensured that a type is given. However, I'll also add `.getOrElse { Context.abort ... }` here, just in case later modifications change these guarantees.

dvdvgt:
Should be mergeable after the CI passes, @b-studios.

PR #418, opened 2024-03-20
Add Existentials

### LINKED ISSUE #245: Existential types
It would be great if we had existential types

```
type Packet {
   Wrap[E](x: E, E => E at {})
}
```

b-studios:
On popular demand, this PR brings existentials back (fixes #245)

However, there are still a few things that need to be done. 

### Before Merging
 - [ ] More extensive testing. Maybe someone can help with this?

### After Merging
In follow up PRs we could
- [ ] check scope extrusion (see the neg test for how this looks like at the moment) and report better errors
- [ ] allow annotating type parameters on matches to bind them in the pattern (and actually in guards). This is somewhat related to what @dvdvgt is working on right now.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/symbols/symbols.scala
+++ b/effekt/shared/src/main/scala/effekt/symbols/symbols.scala
@@ -264,15 +264,15 @@ case class Constructor(name: Name, tparams: List[TypeParam], var fields: List[Fi
   lazy val vparams: List[ValueParam] = fields.map { f => f.param }
   val bparams: List[BlockParam] = Nil
 
-  val returnType: ValueType = ValueTypeApp(tpe, tparams map ValueTypeRef.apply)
-  def annotatedResult: Option[ValueType] = Some(returnType)
+  val appliedDatatype: ValueType = ValueTypeApp(tpe, tpe.tparams map ValueTypeRef.apply)
+  def annotatedResult: Option[ValueType] = Some(appliedDatatype)
   def annotatedEffects: Option[Effects] = Some(Effects.Pure)
 }
 
 // TODO maybe split into Field (the symbol) and Selector (the synthetic function)
 case class Field(name: Name, param: ValueParam, constructor: Constructor) extends Callable {
   val tparams: List[TypeParam] = constructor.tparams
-  val vparams = List(ValueParam(constructor.name, Some(constructor.returnType)))
+  val vparams = List(ValueParam(constructor.name, Some(constructor.appliedDatatype)))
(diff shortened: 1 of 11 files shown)


b-studios:
@jiribenes I am marking you as reviewer since you requested this feature last :)

b-studios:
We should be aware that there is no way at the moment to support this featur ein MLton due to the restrictions of SML and Hindley Milner.

b-studios:
@phischu you also asked for this.

examples/neg/exists_fresh.effekt
```diff
@@ -0,0 +1,16 @@
+type Exists {
+  Op[Z](el: Z, combine: (Z, Z) => Z at {})
+}
+
+def test2fresh() = {
+  val intBox = Op(1, box { (x, y) => x + y });
+
+  def twoBoxes(b1: Exists, b2: Exists) = (b1, b2) match {
+    case (Op(el1, combine1), Op(el2, combine2)) =>
+      combine2(el2, el1) // ERROR Expected Z but got Z
```
b-studios, line 10: Arguably this is not the best error message :)

examples/neg/exists_extrudes.effekt
```diff
@@ -0,0 +1,15 @@
...
+  Op[Z](el: Z, combine: (Z, Z) => Z at {})
+}
+
+def test2extruding() = {
+  val intBox = Op(1, box { (x, y) => x + y });
+
+  val x = intBox match {
+    case Op(el, combine) => el
+  }
+  x + 2 // ERROR Cannot typecheck
```
b-studios, line 11: We do not check for extrusion, but since the existentials are fresh it shouldn't be possible to do anything with them. Inferred types will just mention strange, not bound, type parameters.

effekt/shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -264,15 +264,15 @@ case class Constructor(name: Name, tparams: List[TypeParam], var fields: List[Fi
   lazy val vparams: List[ValueParam] = fields.map { f => f.param }
   val bparams: List[BlockParam] = Nil
 
-  val returnType: ValueType = ValueTypeApp(tpe, tparams map ValueTypeRef.apply)
-  def annotatedResult: Option[ValueType] = Some(returnType)
+  val appliedDatatype: ValueType = ValueTypeApp(tpe, tpe.tparams map ValueTypeRef.apply)
```
b-studios, line 267: I renamed this to be consistent with `Operation`s

PR #419, opened 2024-03-20
Add support for names ending in ? and !

b-studios:
Finally we can name predicates with ? :)

```
list.empty?
result.defined?
...
```


jiribenes (approved):
`pr.LGTM!` :shipit:

It would be nice to write down a style guide to indicate how/where `?` and `!` are supposed to be used, then rename all of the relevant identifiers in the standard library.

PR #421, opened 2024-03-22
Explicit interface boxing

### LINKED ISSUE #288: Explicitly box interfaces
Currently the following is accepted:

```
def create(): Fresh at {} =
  new Fresh { def fresh() = "" }
```

However, the following is not and I think it should be:

```
def create(): Fresh at {} =
  box new Fresh { def fresh() = "" }
```

jiribenes:
Resolves #288

TODO:
- [ ] convert the example from #288 into a positive test


effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -531,7 +531,9 @@ class EffektParsers(positions: Positions) extends EffektLexers(positions) {
   lazy val string = stringLiteral ^^ { s => StringLit(s.substring(1, s.size - 1)) }
 
   lazy val boxedExpr: P[Term] =
-    `box` ~> captureSet.? ~ (idRef ^^ Var.apply | functionArg) ^^ { case capt ~ block => Box(capt, block) }
+    `box` ~> captureSet.? ~ (idRef ^^ Var.apply | functionArg | newExpr) ^^ {
+      case capt ~ block => Box(capt, block)
+    }
```
jiribenes, line 536: Let's minimise the diff:
```suggestion
    `box` ~> captureSet.? ~ (idRef ^^ Var.apply | functionArg | newExpr) ^^ { case capt ~ block => Box(capt, block) }
```

phischu (approved):

PR #424, opened 2024-03-22
Allow using `get` and `put` with UFCS

### LINKED ISSUE #335: Builtin get and set operations cannot be shadowed when using dot notation
linter bug: Unbox requires a boxed type, but got Array[Int].effekt
minified example:
```
def test(arr: Array[Int]) = {
	arr.put(3,12)
}
```

jiribenes:
Fixes #335 by removing the `get` and `put` operations from the list of root terms.
We might be able to garbage-collect even a larger portion of the `builtins.scala` file. :)

~~I haven't tested this locally, but I believe the test coverage here should be sufficient (we do use mutable variables a lot :))~~
EDIT: _I have tested this locally, I modified existing tests which tried to avoid `array.get` and `list.get` and added a new test._


jiribenes:
Still need to add a test for `array.get(...)` and similar.

effekt/shared/src/main/scala/effekt/symbols/builtins.scala
```diff
@@ -82,9 +82,7 @@ object builtins {
   lazy val globalRegion = ExternResource(name("global"), TRegion)
 
   val rootTerms: Map[String, TermSymbol] = Map(
```
jiribenes, line 84: I'm keeping the rest of this file (so we're still creating `TState.get` and `TState.put`) in case we want to do revive this part of the codebase later.

examples/pos/arrays.effekt
```diff
@@ -10,20 +10,21 @@ def show(o: Option[String]) = o match {
...
-    val arrWithVals: Array[String] = fill(4, "Hello World");
+    arr.put(0, "hello");
+    arr.put(2, "world");
+    println(show(arr.get(0)));
+    println(show(arr.get(2)));
+    println(show(arr.get(4)));
+    arr.put(4, "foo");
+    println(show(arr.get(4)))
+
+    val arrWithVals: Array[String] = mutable::array::fill(4, "Hello World");
```
jiribenes, line 22: I would have liked to say just `array::fill`, but that doesn't seem to work. Is that a bug in namespace-handling code?

jiribenes: Made this into an issue ~> https://github.com/effekt-lang/effekt/issues/426

b-studios (commented):
LGTM :)

PR #425, opened 2024-03-22
Add shortcutting for boolean operations

### LINKED ISSUE #387: Make && and || by name
We can make these two logical operators by name, by simply desugaring them to `if`.

This could either happen before Typer (that is in Parser), or after Typer (that is in the translation to core).

```
EXPR1 || EXPR2 = if (EXPR1) true else EXPR2
EXPR1 && EXPR2 = if (EXPR1) EXPR2 else false
```

Doing it in Parser has the advantage that it is really just a few lines of code to implement.
Doing it in core.Transformer has the advantage that we can selectively perform this translation on `effekt.infixOr` and `effekt.InfixAnd` and the operators can be overloaded (for instance for logic programming).

We could also do the translation selectively and **only** translate it if EXPR2 is a Stmt or a side effecting Expr (EXPR1 does not matter, since it is forced anyway).

marvinborner:
This PR closes #387. For now I've implemented it directly in the parser.
There might be bugs left, so please try it out if the CI passes.

---

CI did not pass. LLVM has obvious errors that could be fixed by improving the inliner.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Parser.scala
+++ b/effekt/shared/src/main/scala/effekt/Parser.scala
@@ -617,6 +617,7 @@ class EffektParsers(positions: Positions) extends EffektLexers(positions) {
   private def binaryOp(lhs: Term, op: String, rhs: Term): Term =
      Call(IdTarget(IdRef(Nil, opName(op))), Nil, List(lhs, rhs), Nil)
 
+  // thunkedBinaryOp(lhs, op, rhs) = op { lhs } { rhs }
   private def thunkedBinaryOp(lhs: Term, op: String, rhs: Term): Term =
      Call(IdTarget(IdRef(Nil, thunkedOpName(op))), Nil, Nil, List(BlockLiteral(Nil, Nil, Nil, Return(lhs)), BlockLiteral(Nil, Nil, Nil, Return(rhs))))
 
--- a/effekt/shared/src/main/scala/effekt/Parser.scala
+++ b/effekt/shared/src/main/scala/effekt/Parser.scala
@@ -375,8 +375,8 @@ class EffektParsers(positions: Positions) extends EffektLexers(positions) {
    * Expressions
    */
   lazy val expr:    P[Term] = matchExpr | assignExpr | orExpr | failure("Expected an expression")
-  lazy val orExpr:  P[Term] = orExpr  ~ "||" ~/ andExpr ^^ liftedBinaryOp | andExpr
-  lazy val andExpr: P[Term] = andExpr ~ "&&" ~/ eqExpr ^^ liftedBinaryOp | eqExpr
+  lazy val orExpr:  P[Term] = orExpr  ~ "||" ~/ andExpr ^^ thunkedBinaryOp | andExpr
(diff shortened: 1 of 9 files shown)


b-studios:
It seems the inliner is turned off?

https://github.com/effekt-lang/effekt/blob/62a551e0c8b5090a8e376a3549abde5f9781e40d/effekt/shared/src/main/scala/effekt/core/Optimizer.scala#L24

b-studios:
Looks like @dvdvgt disabled it here: fc29e654

dvdvgt:
This was disabled for debugging during the development of the accompanying PR. This can, of course, be turned back on!

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -617,9 +617,15 @@ class EffektParsers(positions: Positions) extends EffektLexers(positions) {
   private def binaryOp(lhs: Term, op: String, rhs: Term): Term =
      Call(IdTarget(IdRef(Nil, opName(op))), Nil, List(lhs, rhs), Nil)
 
-  private def opName(op: String): String = op match {
+  private def liftedBinaryOp(lhs: Term, op: String, rhs: Term): Term =
```
b-studios, line 620: I would call this `delayedBinaryOp` or `thunkedBinaryOp`. We already use "lifted" for other things a lot :)

b-studios: Also, could you give an example in pseudocode like

```
delayedBinaryOp(lhs, op, rhs) = op { lhs } { rhs }
```
?

libraries/chez/callcc/effekt.effekt
```diff
@@ -141,11 +141,11 @@ extern pure def infixGte(x: Double, y: Double): Bool =
 extern pure def not(b: Bool): Bool =
   "(not ${b})"
 
-extern pure def infixOr(x: Bool, y: Bool): Bool =
-  "(or ${x} ${y})"
+def infixOr { first:  () => Bool } { second: () => Bool }: Bool =
```
b-studios, line 144: You could use the shorthand here

```suggestion
def infixOr { first:  => Bool } { second: => Bool }: Bool = 
```

which emphasizes the thunkiness

b-studios:
Just out of curiosity. Could you post the generated code (js, chez, or llvm) of a simple example (after rebasing on the master with the inliner fix)?

It would be interesting to see whether the "overhead" of the function call and all traces of the "monadic" encoding are completely removed.

marvinborner:
I can't reproduce the [CI failure](https://github.com/effekt-lang/effekt/actions/runs/8571765961/job/23492781723) on `stdlib/list/insert.effekt` of `chez-callcc`. The [previous CI run](https://github.com/effekt-lang/effekt/actions/runs/8521298424/job/23339157347) passed (although with a slightly different, yet equivalent, git history).
Could this be related to our other flakiness issues? Otherwise I wouldn't know which of my changes causes this.

b-studios (approved):
Beautiful, thanks!

PR #427, opened 2024-03-22
Use feature flags for extern definitions

marzipankaiser:
This implements feature flags for externs.

# Syntax
```effekt
extern def infixAdd(a: Int, b: Int): Int =
     js "${a} + ${b}"
     llvm { someEffektStmt }
     default "${a} + ${b}"
```
(Note: You might not want to use `default` like this... Just an example.)

The old syntax should still work and is equivalent to having just a `default` clause.

This also exists for includes like this:
```effekt
extern js """
...
"""
extern include js " ... "
```

# Notes about implemented behaviour
Current behaviour is:
- Extern includes with non-matching flags are ignored
- Extern defs without matching flag (for current backend) give out a warning during naming (for now; ideally: Check if it is actually used!)
- Extern defs with an unannotated/`default` **string** will give out a warning

# Future work
- Output warnings only if we (transitively) use unsupported externs from main
- Move generation of warnings from `Namer` into a separate phase, potentially with some other smaller checks. Alternatively: Move to `ResolveExternDefs`.
- Allow using interfaces/functions passed as block parameters (currently fails with a type error due to captures).


effekt/shared/src/main/scala/effekt/FeatureFlag.scala
```diff
@@ -1,4 +1,4 @@
-package effekt.symbols
+package effekt
```
jiribenes, line 1: _Why does this help?_ 🤔

This is very strange.

marzipankaiser:
Design decision reg. Effekt code in extern definitions:
Currently, the syntax is just `(expr)` (forced parentheses for parsing) instead of a string.
They (all of them) are passed through to the backends and transformed on the way, the backend can then choose.

Upside: The backends can do whatever they want with it.

Downside: This has to be handled in the backends, which is non-trivial at least in LLVM.

Alternative: Choose the "correct" extern definition in `core`. Then make the pseudo-extern definitions normal definitions.
(Pro/con as above but reversed)

b-studios:
I don't think I understand your proposal. Do you have an example of how this looks like?

marzipankaiser:
Note to self: Add a simple phase on core to select & desugar extern defs (the backends can decide to use it or not)

marzipankaiser:
Changed:
- Now uses `{ stmt }` instead of `( expr )` for effekt-defined externs; this makes the second point easier and seems more natural to me
- Added a Core -> Core phase that selects the correct definitions and makes them "normal" definitions if they are defined using Effekt code

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -288,13 +290,20 @@ object Namer extends Phase[Parsed, NameResolved] {
         resolveGeneric(body)
       }
 
-    case f @ source.ExternDef(capture, id, tparams, vparams, bparams, ret, body) =>
+    case f @ source.ExternDef(capture, id, tparams, vparams, bparams, ret, bodies) =>
+      if (!bodies.supportedByFeatureFlags(Context.compiler.supportedFeatureFlags)) {
+        Context.warning(s"Extern definition ${id} is not supported by this backend.")
+      }
```
marzipankaiser, line 296: This is where the warnings are currently generated.

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -829,11 +830,98 @@ object Typer extends Phase[NameResolved, Typechecked] {
...
+              case x: CaptUnificationVar => List(x)
+              case _ => Nil
+            }
+            val Result(funTpe, unhandledEffects) = Context.withUnificationScope(captVars) {
+
+              sym.vparams foreach Context.bind
+              sym.bparams foreach Context.bind
+
+              // to subtract the capabilities, which are only inferred bottom up, we need a **second** unification variable
+              val inferredCapture = Context.freshCaptVar(CaptUnificationVar.AnonymousFunctionRegion(BlockLiteral(tps, vps, bps, body)))
```
marzipankaiser, line 858: Apart from the role here, all of this code is just copy-pasted from the case for `FunDef`.

effekt/shared/src/main/scala/effekt/core/ResolveExternDefs.scala
```diff
@@ -0,0 +1,50 @@
...
+import effekt.context.Context
+import effekt.source.FeatureFlag
+
+/**
+ * [[Phase]] that resolves feature-flag annotated externs to unique extern (string-based) defs or new definitions.
+ *
+ * @param supportedFeatureFlags are a *prioritized* list (i.e., earlier ones always match).
+ * Unsupported externs are replaced with a definition according to [[defaultExternBody]].
+ */
+case class ResolveExternDefs(supportedFeatureFlags: List[String]) extends Phase[CoreTransformed, CoreTransformed] {
```
marzipankaiser, line 12: This is the phase resolving extern defs

effekt/shared/src/main/scala/effekt/Namer.scala
jiribenes: Looks great! 👍
For clarity, would it be possible to also dump why exactly it's not supported? (something like `Extern definition ${id} is not supported as it requires feature flags ${featureFlags}, but the current backend only supports ${backendFeatureFlags}`)

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -163,20 +173,27 @@ class EffektParsers(positions: Positions) extends EffektLexers(positions) {
...
-    | failure(s"Expected an extern definition, which can either be a single-line string (e.g., \"x + y\") or a multi-line string (e.g., $multi...$multi)")
+  lazy val externBody: P[ExternBody] =
+    (maybeFeatureFlag ~ ("""(?=[\"])""".r ~> commit(
+              multilineString ^^ { s => Template(List(s), Nil) }
+            | guard(regex(s"(?!${multi})".r)) ~> templateString(expr)
+            | failure(s"Expected an extern definition, which can either be a single-line string (e.g., \"x + y\") or a multi-line string (e.g., $multi...$multi), potentially qualified")
+            )) ^^ {
+                case ff ~ body => ExternBody.StringExternBody(ff, body)
+            }
+    | featureFlag ~ ("{" ~> stmt <~ "}") ^^ { case ff ~ body => ExternBody.EffektExternBody(ff, body) }
```
b-studios, line 192:
```suggestion
    | featureFlag ~ ("{" ~> stmts <~ "}") ^^ { case ff ~ body => ExternBody.EffektExternBody(ff, body) }
```

effekt/shared/src/main/scala/effekt/Typer.scala
b-studios: That's a lot of code. Can we somehow share?

marzipankaiser: We probably could. Thinking about it, frankly, I'm not sure this is the correct way to check those.
We probably want at least "internal" type checking, but I'm not quite certain how much we want at the boundary.

Will look at it again once I find the time.

marzipankaiser:
At some point we might want to move the warnings at
https://github.com/effekt-lang/effekt/blob/5646f6731f7983dcead702a6e60321f9cc9186c2/effekt/shared/src/main/scala/effekt/Namer.scala#L294-L304
and
https://github.com/effekt-lang/effekt/blob/5646f6731f7983dcead702a6e60321f9cc9186c2/effekt/shared/src/main/scala/effekt/Namer.scala#L385-L389
from `Namer` to a separate phase with more of those smaller checks.
Maybe there are also other checks that would fit there?

marzipankaiser:
Tests are failing because of tests that use definitions like
https://github.com/effekt-lang/effekt/blob/5646f6731f7983dcead702a6e60321f9cc9186c2/examples/pos/capture/resources.effekt#L3
which will potentially never work on some backends. 
Potentially some can be rewritten to use Effekt definitions instead (i.e., `extern def ... = else { ... }`).

effekt/shared/src/main/scala/effekt/Compiler.scala
```diff
@@ -66,6 +66,17 @@ export PhaseResult.*
...
 
+/**
+ * Constants for common feature flags (not specific to one generator)
+ */
+object FeatureFlag {
+  /** For backends that translate to a programming language supporting at least:
+   * - directly splicing in variables ("${...}" works)
+   * - "simple" literals like "42"
+   */
+  val simpleAtom: String = "simpleAtom"
```
marzipankaiser, line 77: Here, I now define common feature flags (currently one).
`simpleAtom` is meant to be supported by backends that translate such that both simple literals (ints like "42") and variable names (mostly for "${x}") are allowed.

examples/pos/capture/ffi_blocks.effekt
```diff
@@ -1,6 +1,6 @@
 interface Foo {}
-extern io def something {f: Foo}: Int = "${box f}"
-extern io def createFoo(): Foo at {global} = "42"
+extern io def something {f: Foo}: Int = simpleAtom "${box f}"
+extern io def createFoo(): Foo at {global} = simpleAtom "42"
```
b-studios, line 3: This is just a bad test (already has been), not your fault

examples/pos/capture/mbed.effekt
```diff
@@ -9,10 +9,10 @@ extern resource deep: Builtin
 
 // Some Builtins pretending to use the resources
 // ---------------------------------------------
-extern {in}    def useIn(): Unit = "0"
-extern {out}   def useOut(): Unit = "0"
-extern {block} def performBlocking(): Unit = "0"
-extern {deep}  def blockDeepSleep(): Unit = "0"
+extern {in}    def useIn(): Unit = simpleAtom "0"
```
b-studios, line 12:
```suggestion
extern {in}    def useIn(): Unit = else { 0 }
```

marzipankaiser:
`else` -> `default`

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -829,11 +830,22 @@ object Typer extends Phase[NameResolved, Typechecked] {
           Result((), effBinding)
         }
 
-      case d @ source.ExternDef(pure, id, tps, vps, bps, tpe, body) => Context.withUnificationScope {
+      case d @ source.ExternDef(captures, id, tps, vps, bps, tpe, bodies) => Context.withUnificationScope {
         d.symbol.vparams foreach Context.bind
         d.symbol.bparams foreach Context.bind
 
-        body.args.foreach { arg => checkExpr(arg, None) }
+        flowingInto(Context.lookupCapture(id.symbol.asBlockSymbol)) {
```
b-studios, line 837: would the following work?
```suggestion
        flowingInto(Context.lookupCapture(d.symbol)) {
```

effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
```diff
@@ -59,9 +60,17 @@ object PrettyPrinter extends ParenPrettyPrinter {
...
+      "extern" <+> toDoc(capt) <+> "def" <+> toDoc(id) <+> "=" <+> paramsToDoc(tps, vps, bps) <> ":" <+> toDoc(ret) <+> "=" <+> vcat(bodies map {
+        case ExternBody.StringExternBody(ff, body) => toDoc(ff) <+> toDoc(body)
+        case ExternBody.EffektExternBody(ff, body) => toDoc(ff) <+> "(" <+> toDoc(body) <+> ")"
+      })
+    case Extern.Include(ff, contents) => emptyDoc // right now, do not print includes.
+  }
+
+  def toDoc(ff: FeatureFlag): Doc = ff match {
+    case FeatureFlag.NamedFeatureFlag(name) => name
+    case FeatureFlag.Default => "else"
```
b-studios, line 73: Nice, here it is already called `Default` :)

effekt/shared/src/main/scala/effekt/core/Tree.scala
```diff
@@ -125,8 +126,22 @@ case class Property(id: Id, tpe: BlockType) extends Tree
...
-  case Include(contents: String)
+  case Def(id: Id, tparams: List[Id], cparams: List[Id], vparams: List[Param.ValueParam], bparams: List[Param.BlockParam], ret: ValueType, annotatedCapture: Captures, bodies: List[ExternBody])
+  case Include(featureFlag: FeatureFlag, contents: String)
+}
+sealed trait ExternBody extends Tree {
+  def featureFlag: FeatureFlag
+}
+object ExternBody {
+  case class StringExternBody(featureFlag: FeatureFlag, contents: Template[Pure]) extends ExternBody
+  case class EffektExternBody(featureFlag: FeatureFlag, body: Stmt) extends ExternBody
```
b-studios, line 137: Potentially risking conflicts with `String` this could be shortened to
```scala
enum ExternBody extends Tree {
  def featureFlag: FeatureFlag
  
  case String(featureFlag: FeatureFlag, contents: Template[Pure])
  case EffektExternBody(featureFlag: FeatureFlag, body: Stmt)
}
```

effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
marzipankaiser: Just noticed so too when renaming 😅

effekt/shared/src/main/scala/effekt/core/Tree.scala
marzipankaiser: I'm not sure it's worth it given the potential conflicts. In particular, adding any helper methods on `ExternBody` that use `String`s as parameters/return values would become awkward (and might not be that unlikely).
Maybe we can also come up with a better name (I couldn't, hence the not entirely helpful one... -s, actually).

effekt/shared/src/main/scala/effekt/Typer.scala
marzipankaiser: Should work, will change.

examples/pos/liftinference.effekt
```diff
@@ -5,7 +5,7 @@ interface Foo {
   def bar(): String
 }
 
-extern control def fancy { f: Unit => Int }: Int = "foo"
+extern control def fancy { f: Unit => Int }: Int = simpleAtom "foo"
```
marzipankaiser, line 8: Here, using `default { ... }` instead would potentially make the test less useful - after all, there probably was a reason for using an extern here, and this will never reach `LiftInference` then (will be desugared in core already).

b-studios: Oh that is true. But we could also run it only in one of the lift backends.

marzipankaiser: I'll probably just duplicate it. 🤔

examples/pos/capture/ffi_blocks.effekt
marzipankaiser: I tried to fix it up, and it found a bug: We cannot actually call any operations on `f` since the capture is not allowed by our current implementation of Typer here.
Just allowing this in typer, however, leads to crashes in the backends (more precisely: undefined variables). I think this might be related to not passing the captures to the extern 🤔.

effekt/shared/src/main/scala/effekt/Compiler.scala
marzipankaiser: (Removed)

marzipankaiser:
Maybe it would be better to move the Resolving further to the front after all.
I would still prefer it to be after Typing, but we could do it in `Middleend` on `Typechecked`.

marzipankaiser:
The approach of adding it after Typer does not work properly either: The calling-convention issue hinges on the symbol type used in `core.Transformer`:
https://github.com/effekt-lang/effekt/blob/ee905df3182f203730cec00d6cffd3472b84cbdb/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L635-L638

effekt/shared/src/main/scala/effekt/core/Parser.scala
```diff
@@ -69,13 +69,11 @@ class CoreParsers(positions: Positions, names: Names) extends EffektLexers(posit
...
-      case captures ~ (id, tparams, cparams, vparams, bparams, result) ~ bodies =>
-        Extern.Def(id, tparams, cparams, vparams, bparams, result, captures, bodies.map {
-          case ff ~ (body: Stmt) =>
-            ExternBody.EffektExternBody(ff, body)
+    | `extern` ~> (captures <~ `def`) ~ signature ~ (`=` ~> (featureFlag ~ externBody)) ^^ {
+      case captures ~ (id, tparams, cparams, vparams, bparams, result) ~ body =>
+        Extern.Def(id, tparams, cparams, vparams, bparams, result, captures, body match {
           case ff ~ (body: String) =>
-            ExternBody.StringExternBody(ff, Template(List(body), Nil))
+            ExternBody(ff, Template(List(body), Nil))
```
b-studios, line 76: Beautiful! Thanks, that's much better

effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -117,18 +117,19 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
     case f @ source.ExternDef(pure, id, _, vps, bps, _, bodies) =>
       val sym@ExternFunction(name, tps, _, _, ret, effects, capt, _) = f.symbol
       assert(effects.isEmpty)
+      assert(bodies.length == 1, "Externs are resolved before core")
```
b-studios, line 120: The `.head` below is a bit hidden. Maybe refactor to

```
val tBody = bodies match {
  case source.ExternBody.StringExternBody(ff, body) :: Nil => ...
  case _ => Context.panic("Externs should be resolved and desugared before core.Transformer")
}
```
?

effekt/shared/src/main/scala/effekt/core/Tree.scala
```diff
@@ -126,16 +126,11 @@ case class Property(id: Id, tpe: BlockType) extends Tree
  * FFI external definitions
  */
 enum Extern extends Tree {
-  case Def(id: Id, tparams: List[Id], cparams: List[Id], vparams: List[Param.ValueParam], bparams: List[Param.BlockParam], ret: ValueType, annotatedCapture: Captures, bodies: List[ExternBody])
+  case Def(id: Id, tparams: List[Id], cparams: List[Id], vparams: List[Param.ValueParam], bparams: List[Param.BlockParam], ret: ValueType, annotatedCapture: Captures, body: ExternBody)
```
b-studios, line 129: nice!

effekt/shared/src/main/scala/effekt/generator/js/TransformerMonadic.scala
```diff
@@ -183,22 +179,14 @@ trait TransformerMonadic extends Transformer {
...
       js.Call(toJS(f), vargs.map(toJS) ++ bargs.map(toJS))
 
     case PureApp(f: core.Block.BlockVar, targs, vargs) if canInline(f) =>
       val extern = D.getExternDef(f.id)
-      extern.bodies.forFeatureFlags(jsFeatureFlags).getOrElse { ??? /* TOOD insert hole */ } match {
-        case ExternBody.StringExternBody(_, body) =>
-          inlineExtern(vargs, extern.vparams, body)
-        case ExternBody.EffektExternBody(_, body) => sys error "Effekt extern body should have been removed"
-      }
+      inlineExtern(vargs, extern.vparams, extern.body.contents)
```
b-studios, line 189: Much, much better!

effekt/shared/src/main/scala/effekt/Compiler.scala
```diff
@@ -66,6 +66,12 @@ export PhaseResult.*
 
 enum Stage { case Core; case Lifted; case Machine; case Target; }
 
+/**
+ * Constants for common feature flags (not specific to one generator)
+ */
+object FeatureFlag {
+}
```
b-studios, line 73: Do we still need this now it is empty?

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -288,13 +290,28 @@ object Namer extends Phase[Parsed, NameResolved] {
         resolveGeneric(body)
       }
 
-    case f @ source.ExternDef(capture, id, tparams, vparams, bparams, ret, body) =>
+    case f @ source.ExternDef(capture, id, tparams, vparams, bparams, ret, bodies) =>
+      if (!bodies.supportedByFeatureFlags(Context.compiler.supportedFeatureFlags)) {
+        val featureFlags = bodies.map(_.featureFlag)
+        Context.warning(pp"Extern definition ${id} is not supported as it is only is defined for feature flags ${featureFlags.mkString(", ")}," +
```
b-studios, line 296:
```suggestion
        Context.warning(pp"Extern definition ${id} is not supported as it is only defined for feature flags ${featureFlags.mkString(", ")}," +
```

effekt/shared/src/main/scala/effekt/generator/js/TransformerMonadic.scala
```diff
@@ -35,7 +37,8 @@ object TransformerMonadicWhole extends TransformerMonadic {
       val hasBlockParameters = bparams.nonEmpty
       val isControlEffecting = annotatedCapture contains symbols.builtins.ControlCapability.capture
       !hasBlockParameters && !isControlEffecting
-    case Extern.Include(contents) => false
+    case Extern.Include(ff, contents) if ff.matches(jsFeatureFlags) => false
+    case Extern.Include(ff, contents) => true // it's nothing anyway
```
b-studios, line 41: This I don't understand. Before we didn't inline those at all.

effekt/shared/src/main/scala/effekt/generator/js/TransformerMonadic.scala
```diff
@@ -110,11 +113,13 @@ trait TransformerMonadic extends Transformer {
...
-    case Extern.Def(id, tps, cps, vps, bps, ret, capt, body) =>
-      js.Function(nameDef(id), (vps ++ bps) map toJS, List(js.Return(toJS(body))))
+    case Extern.Def(id, tps, cps, vps, bps, ret, capt, ExternBody(featureFlag, contents)) =>
+      js.Function(nameDef(id), (vps ++ bps) map toJS, List(js.Return(toJS(contents))))
 
-    case Extern.Include(contents) =>
+    case Extern.Include(ff, contents) if ff.matches(jsFeatureFlags) =>
       js.RawStmt(contents)
+
+    case Extern.Include(_, _) => js.RawStmt("") // ignore, not meant for us
```
b-studios, line 122: Can this still happen? I thought selection happens earlier and only leaves those that we care about.

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -40,12 +42,14 @@ object Transformer {
...
-      case machine.Extern(functionName, parameters, returnType, body) =>
+      case machine.Extern(functionName, parameters, returnType, machine.ExternBody(_, body)) =>
         VerbatimFunction(transform(returnType), functionName, parameters.map {
           case machine.Variable(name, tpe) => Parameter(transform(tpe), name)
         }, transform(body))
-      case machine.Include(content) =>
+      case machine.Include(ff, content) if ff.matches(llvmFeatureFlags) =>
         Verbatim(content)
+      case machine.Include(ff, content) =>
+        Verbatim("") // ignore, not meant for us
```
b-studios, line 52: Same as above, is this necessary?

b-studios:
I reviewed the Scala source changes and they look mostly good to me. I only have a few remarks:

1. https://github.com/effekt-lang/effekt/pull/427#discussion_r1572385267
2. https://github.com/effekt-lang/effekt/pull/427#discussion_r1572605230 (and for other backends)
3. https://github.com/effekt-lang/effekt/pull/427#discussion_r1572603819

effekt/shared/src/main/scala/effekt/core/Transformer.scala
marzipankaiser: Good idea.

effekt/shared/src/main/scala/effekt/Compiler.scala
marzipankaiser: Not really, removing it.

effekt/shared/src/main/scala/effekt/generator/js/TransformerMonadic.scala
marzipankaiser: Or just drop the check altogether. I'll do the latter.

marzipankaiser:
(interpreted that as approval)

PR #432, opened 2024-03-22
Concurrent IO

b-studios:
This eventually replaces #324.

There is still a lot to do, but some mechanisms are already in place.


b-studios:
This is how you can use the various libraries:

```
import io
import io/network
import io/time
import immutable/bytes
import text/string

// creates a simple echo server.
def main() = eventloop(box {
  var x in global = 0;
  server("localhost", 8080, box {
    loop {
      val msg = do receive();
      x = x + 1
      //println("Received a request: " ++ msg.toString)
      if (msg.toString.startsWith("GET /")) {
        if (x <= 1) {
          wait(1000)
        }
        do send(toBytes("HTTP/1.1 200 OK\r\n\r\nHello from Effekt!"))
      } else {
        do send("HTTP/1.1 400 Bad Request\r\n\r\n".toBytes)
      }
      do end()
    }
  })
})
```

However, it will currently crash, since `global` is not bound and the temporary hack only worked for a single file (and not across several other files).

b-studios:
@jiribenes With #434 being merged, is the library now usable for you?

jiribenes:
> @jiribenes With #434 being merged, is the library now usable for you?

Yep -- except that `immutable/bytes` is not a part of this PR so nobody else can actually use this. It would be nice to have a test or two to detect such problems if at all possible.

Here's what I tried:
```scala
import io
import io/network
import immutable/bytes

import http

def serve(): Unit / Socket = {
  val _ = do receive();
  
  val responseBody = http::Body("Hello from Effekt!".toBytes)
  do send(http::ok(responseBody).encode)
  do end()
}

def main() = 
  eventloop(box {
    server("localhost", 8080, box {
      serve()
    })
  })
```

b-studios:
Agreed, there should be tests. I simply forgot to add the file :(

b-studios:
Superseded by #469

Issue #412, opened 2024-03-12
Parser performance: rewrite parser using different combinator library


dvdvgt:
By investigating the time spent per phase using PR #411, it became apparent that the parser is a major bottleneck of the compiler. The parser regularly occupies more than **50% of the total time** spent compiling (for more complex/realistic programs with imports). This offers a massive opportunity for speeding up the compiler.

In my opinion, we have three options here. Use a parser combinator, parser generator library or a completely handwritten parser. 

Using a *handwritten parser* can be a lot of work, and given the experimental stage Effekt is in, also in terms of syntax, it may be hard to make changes to a handwritten parser down the road. After a quick search, I did not find any actively maintained, reasonably popular *parser generators* for Scala. This leaves us with *parser combinators*, and there are a few candidates:

1. https://github.com/j-mie6/Parsley
2. https://github.com/com-lihaoyi/fastparse

`fastparse` promises some nice [speed-ups](https://com-lihaoyi.github.io/fastparse/#Performance).

b-studios:
Minor comment: we do not use `scala-parser-combinator`s but the Kiama library.

dvdvgt:
Thanks, I changed it.

Maybe we should first add some more tests for the parser, so that we can more easily detect regressions.

b-studios:
The problem is the positions which mostly die up in error messages and with LSP. Adding those tests could be a lot of work.

PR #435, opened 2024-03-26
Adding a separate lexer

dvdvgt:
This PR aims to speed up the parsing phase of the compiler, as discussed in #412. Since we would want to have a separate lexer regardless of the possibility of switching to a different parsing library or tuning the current implementation for speed, this should be useful.

The lexer is handwritten, and I took care to make it reasonably fast (of course, I am happy to hear your feedback). The reason for a handwritten lexer are of course speed, however, at a loss of ease of making changes. Though, since I would not expect the keywords, syntax of string and number literals and those for comments to change anytime soon, this should be negligible. Also, we can offer better error messages for invalid syntax than with current implementation. 

<details>
<summary> unterminated string example </summary>

```scala
def main() = {
  val x = "hello
  ()
} 
>>> Expected a statement
```

</details>

## TODO

- [X] Have a working lexer for Effekt's syntax
- [X] Cover basics functionality with tests
- [X] Add support for quoted strings / string interpolation: `"result: ${ xs.map { x => x + 1 } }"`
- [ ] Integrate lexer into current parser by making the current parser implementation input agnostic -- first merge changes into kiama?


effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -0,0 +1,514 @@
...
+            val token = makeToken(nextToken())
+            token.kind match {
+              case TokenKind.`}` => {
+                delimiters.pop() match {
+                  // check if quote has been terminated
+                  case `${` => {
+                    terminated = true
+                    stringTokens += Token(token.start, token.end, TokenKind.`}$`)
+                  }
+                  case _ =>
```
b-studios, line 371:
```suggestion
                  case `{{` =>
```

PR #438, opened 2024-03-27
Enable inlining of small block definitions

b-studios:
This PR reenables the inliner (see #425) and also enables inlining of small definitions.

Right now the inlining threshold is set arbitrary to `20`:

https://github.com/effekt-lang/effekt/pull/438/files#diff-bc7040e8ee95a824eba8f56e3bfc6a8c47fe30cf6e0422baeaf52e7521496717R26

Also, higher-order functions are currently inlined, regardless of the size. 

Neither of these two decisions is founded on actual benchmarks or other good sources, which would be important to investigate in the future.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/Renamer.scala
+++ b/effekt/shared/src/main/scala/effekt/core/Renamer.scala
@@ -81,6 +81,19 @@ class Renamer(names: Names = Names(Map.empty), prefix: String = "") extends core
       }
   }
 
+  override def rewrite(o: Operation): Operation = o match {
+    case Operation(name, tparams, cparams, vparams, bparams, resume, body) =>
+      withBindings(tparams ++ cparams ++ vparams.map(_.id) ++ bparams.map(_.id) ++ resume.map(_.id).toList) {
+        Operation(name,
+          tparams map rewrite,
+          cparams map rewrite,
+          vparams map rewrite,
+          bparams map rewrite,
+          resume map rewrite,
+          rewrite(body))
+      }
+  }
+
   def apply(m: core.ModuleDecl): core.ModuleDecl =
     suffix = 0
     m match {
--- a/effekt/shared/src/main/scala/effekt/core/Tree.scala
+++ b/effekt/shared/src/main/scala/effekt/core/Tree.scala
@@ -64,7 +64,25 @@ import effekt.util.messages.INTERNAL_ERROR
  *
  * -------------------------------------------
  */
-sealed trait Tree
+sealed trait Tree extends Product {
+  /**
+   * The number of nodes of this tree (potentially used by inlining heuristics)
+   */
+  lazy val size: Int = {
+    var nodeCount = 1
+
+    def all(t: IterableOnce[_]): Unit = t.iterator.foreach(one)
+    def one(obj: Any): Unit = obj match {
+      case t: Tree => nodeCount += t.size
+      case s: effekt.symbols.Symbol => ()
+      case p: Product => all(p.productIterator)
+      case t: Iterable[t] => all(t)
+      case leaf           => ()
+    }
+    this.productIterator.foreach(one)
+    nodeCount
+  }
+}
 
 /**
  * In core, all symbols are supposed to be "just" names.
--- a/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
@@ -71,7 +71,7 @@ object Transformer {
         definitions.foreach {
           case Definition.Def(id,  block @ lifted.BlockLit(tparams, params, body)) =>
 
-            params.collect { case lifted.BlockParam(id, tpe) => noteParameter(id) }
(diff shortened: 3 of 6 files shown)


effekt/shared/src/main/scala/effekt/core/Renamer.scala
```diff
@@ -81,6 +81,19 @@ class Renamer(names: Names = Names(Map.empty), prefix: String = "") extends core
...
+      withBindings(tparams ++ cparams ++ vparams.map(_.id) ++ bparams.map(_.id) ++ resume.map(_.id).toList) {
+        Operation(name,
+          tparams map rewrite,
+          cparams map rewrite,
+          vparams map rewrite,
+          bparams map rewrite,
+          resume map rewrite,
+          rewrite(body))
+      }
+  }
```
b-studios, line 95: I needed to fix renamer since we forgot to rename parameters bound by operations.

effekt/shared/src/main/scala/effekt/core/Tree.scala
```diff
@@ -64,7 +64,25 @@ import effekt.util.messages.INTERNAL_ERROR
...
+      case t: Tree => nodeCount += t.size
+      case s: effekt.symbols.Symbol => ()
+      case p: Product => all(p.productIterator)
+      case t: Iterable[t] => all(t)
+      case leaf           => ()
+    }
+    this.productIterator.foreach(one)
+    nodeCount
+  }
+}
```
b-studios, line 85: I added this to `Tree` in order for it to be cached. Otherwise, we would be recomputing it over and over again.

b-studios:
There seems to be an issue with LLVM.

Running `examples/llvm/multihandler.effekt` and printing the lifted core for the failing operation shows that `resume15` should be bound

<details>

```
List(Operation(Put,BlockLit(List(),List(EvidenceParam(ev492), ValueParam(n14,Data(Int,List()))),
  Shift(Evidence(List(Var(ev492))),
    BlockLit(List(),List(BlockParam(resume15,Function(List(),List(EvidenceType()),List(Data(Unit,List())),List(),Data(Unit,List())))),
      Val(_16,App(Member(BlockVar(s7,Interface(Ref,List(Data(Int,List())))),put,Function(List(),List(EvidenceType()),List(Data(Int,List())),List(),Data(Unit,List()))),List(),List(Evidence(List()), ValueVar(n14,Data(Int,List())))),App(BlockVar(resume15,Function(List(),List(EvidenceType()),List(Data(Unit,List())),List(),Data(Unit,List()))),List(),List(Evidence(List()), Literal((),Data(Unit,List()))))))))))
 ```

</details>

**Comment by b-studios** [comment]
@phischu In the last commit I tried to get rid of the matching on symbols in machine transformer by remembering the "kind" of a block symbol in the translation context.

Could you please look into this and fix it? Right now I get

```
opt: ./out/multihandler.ll:789:9: error: multiple definition of local value named 'resume12_465'
        %resume12_465 = extractvalue {%Stk, %Sp} %tmp_48, 0
        ^
```

but I don't know how to debug this.

File: effekt/shared/src/main/scala/effekt/machine/Transformer.scala [hunk 3]
```diff
@@ -508,22 +530,60 @@ object Transformer {
...
+      case BlockInfo.Parameter(tpe) => sys error s"Expected a function definition, but got a block parameter: ${id}"
+      case BlockInfo.Resumption => sys error s"Expected a function definition, but got a continuation: ${id}"
+    }
+    def blockParams(id: Symbol): Environment = definition(id).blockParams
+    def freeParams(id: Symbol): Environment = definition(id).freeParams
+  }
+  enum BlockInfo {
+    case Definition(freeParams: Environment, blockParams: Environment)
+    case Parameter(tpe: lifted.BlockType)
+    case Resumption
```
b-studios, line 546: The idea of this type is that it reflects things we know about block variables in the program.

- `Definition` is a known function (also object? @phischu ?)
- `Parameter` is an unknown function or object
- `Resumption` is a continuation

The calling convention differs between these three, which is why we need to know.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -133,24 +140,34 @@ object Transformer {
...
         )
+
+      // TODO deal with BlockLit
       case lifted.App(lifted.BlockVar(id, tpe), targs, args) =>
         if(targs.exists(requiresBoxing)){ ErrorReporter.abort(s"Types ${targs} are used as type parameters but would require boxing.") }
-        // TODO deal with BlockLit
-        id match {
-          case symbols.BlockParam(_, _) =>
+
+        BPC.info(id) match {
```
b-studios, line 148: Instead of inspecting symbols, we now dispatch on what we know (that is, recorded) about the binder of the variable.

b-studios:
I am merging this now so that @marvinborner can rebase his PR #425 on top of it. 

@phischu if you find the time, please look into the semantics of the machine transformer.

PR #158, opened 2022-09-27
Homogenize paths for windows

b-studios:

Issue #423, opened 2024-03-22
Escaping newlines in single-line strings


jiribenes:
Continuing the discussion from #410, do we actually want to be able to escape newlines like this:
```scala
val myString = "let's say I want to \
continue on the other line"      // ^ escapes a newline

// ~> myString = "let's say I want to continue on the other line"
```

Personally, I'm completely fine to forbid this, especially since we have multiple-line strings as of #410.

marzipankaiser:
Additional comment: If we want this, this is broken on Windows.

jiribenes:
This was resolved by the new lexer #435, such strings are no longer supported.

PR #442, opened 2024-04-09
Fix running Effekt on Windows

### LINKED ISSUE #401: Effekt v0.2.2 crashes on Windows
### Steps to reproduce:
1. Use Windows
2. Install Effekt v0.2.2
3. Open Effekt REPL
4. Write `42` and press Enter.
5. See the error message below.

### Error message

The error message points to the fact that we added a `setPosixFilePermissions` to our Kiama fork a few months ago in commit https://github.com/effekt-lang/kiama/commit/89515c8d17c2772d1caba6f1ef7a9ff5d1d93022:
```
Exception in thread "main" java.lang.UnsupportedOperationException
        at java.base/java.nio.file.Files.setPosixFilePermissions(Files.java:2168)
        at kiama.util.IO$.createFile(IO.scala:81)
        at effekt.JSRunner$.build(Runner.scala:131)
        at effekt.JSRunner$.build(Runner.scala:124)
        at effekt.Runner.eval(Runner.scala:58)
        at effekt.Runner.eval$(Runner.scala:14)
```

### How to resolve

We should at least catch the `UnsupportedOperationException`.
If we want, we might be able to use [`AclFileAttributeView`](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/nio/file/attribute/AclFileAttributeView.html) on Windows or just fallback to `java.io.File.setExecutable()`.

Note that although it's technically an issue in our Kiama fork, I'm filing it here for visibility (and in order to expedite it). :)

marzipankaiser:
Fix #401 by, if running on Windows, generating a `.bat` file instead of the shebang'ed script.
Note: Only tested with JavaScript-backend.

This also changes the CI to:
- In principle run on windows-latest (if enabled, and without Chez, ML, and LLVM - since I don't know how to install stuff on Windows)
- Have an additional task execute *just one example file* (not checking the output) on Windows using JS (to detect that this works).


marzipankaiser:
Potentially, we could factor out common `createScript` / `createScriptOrShebang` functions.

.github/workflows/ci.yml
```diff
@@ -7,10 +7,35 @@ on:
...
+        node-version: '12.x'
+
+    - name: Run tests
+      run: sbt "effektJVM/clean; effektJVM/run examples/pos/sideeffects.effekt"
+
   build-jar:
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latests]
```
jiribenes, line 38: This might be a typo ;) 
(the tests are waiting for almost 2 hours for a machine of this name -- I'll just cancel those)

marzipankaiser: It is. ****.
Thanks for cancelling!

.github/workflows/ci.yml
```diff
@@ -7,6 +7,31 @@ on:
...
+      with:
+        java-version: 11
+
+    - name: Set up NodeJS
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
+
+    - name: Run tests
+      run: sbt "effektJVM/clean; effektJVM/testOnly effekt.JavaScriptTests -- --tests=.*examples/pos/sideeffects.*"
```
b-studios, line 33: Can we also try a neg-test for the path?

marzipankaiser:
Apparently (https://github.com/effekt-lang/effekt/actions/runs/8722693317/job/23929134561?pr=442#step:5:297) bat prints what it does. Have to fix that.

marzipankaiser:
And of course paths are a problem: https://github.com/effekt-lang/effekt/actions/runs/8722798996/job/23929494648?pr=442#step:5:512

effekt/jvm/src/test/scala/effekt/EffektTests.scala
```diff
@@ -59,6 +61,27 @@ trait EffektTests extends munit.FunSuite {
...
+      val windowsPathRegex = """(?:([a-zA-Z]):\\{1,2})?((?:[^\\/:*?"<>|\r\n]+\\)*[^\\/:*?"<>|\r\n]*)""".r
+      windowsPathRegex.replaceAllIn(s, m => {
+        val drive = Option(m.group(1)) // Retrieve the drive letter or an empty string if not present
+        val path = m.group(2).replaceAll("\\\\", "/") // Replace backslashes with forward slashes in directories
+        drive match {
+          case Some(d) => s"$d/$path" // Combine the drive letter (if present) and the Unixified path
+          case None => path
+        }
+      })
+  }
```
marzipankaiser, line 84: While `munit` does some normalization/ignores some things in `assertNoDiff`, I could not find a way to hook into this.
So now, depending on current OS, I normalize things that look like paths.

b-studios: Doesn't look this this could be customized in munit: 

https://github.com/scalameta/munit/blob/cd435cb2cb0f8784f5de65c64355040e852cef27/munit/shared/src/main/scala/munit/internal/difflib/Diffs.scala#L34-L58

b-studios:
BTW this is related to #158. Maybe you can close #158 if the current one subsumes it?

marzipankaiser:
I actually like the solution in #158 more (if it works) - there, we do not normalize the test outputs, but instead always print Unix-Style paths in error messages IIUC. This prevents us from potentially (if unlikely) making test pass by accident on Windows.

Maybe, once we have a Windows machine for testing, we can check if this minus https://github.com/effekt-lang/effekt/pull/442/commits/535215d15ac4da2879d6faaf7bd8702dc9b743a9 plus #158 works, too.

marzipankaiser:
This now uses the solution from #158, which I like better and also seems to work.
The CI seems to run, still has to be tested on an actual Windows machine.

jiribenes (approved):
I tested this out today on an older Windows machine at 6ff1a33 (current `feature/windows-fix`) & everything seems to work except for 
- the fact that in tests in my setup, the filenames are like `examples\\pos\\exists.effekt
- `examples/pos/escaped_linebreaks.effekt`, which doesn't quite escape correctly (Windows line-endings!), returning:
```patch
==> X effekt.JavaScriptTests.examples\\pos\\escaped_linebreaks.effekt (js)  0.022s munit.ComparisonFailException: F:\Development\effekt\effekt\jvm\src\test\scala\effekt\EffektTests.scala:97
96:        test(s"${f.getPath} (${backendName})") {
97:          assertNoDiff(run(f), expected)
98:        }
diff assertion failed
=> Obtained
    """|[error] examples/pos/escaped_linebreaks.effekt:1:22: ')' expected but '"' found
       |def main() = println("a\
       |                     ^
       |""".stripMargin
-[error] examples/pos/escaped_linebreaks.effekt:1:22: ')' expected but '"' found
-def main() = println("a\
-                     ^
+a    "b"       cd
```
All the more reason to forbid this with the new lexer #435 (see #423 for more details about escaping newlines).

As a disclaimer, I needed to swap `node` for the full path everywhere in `Runner`, but I _think_ that it's just a weird quirk of my set-up. :)

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -125,13 +139,23 @@ object JSRunner extends Runner[String] {
...
+      case OS.POSIX =>
+        val shebang = "#!/usr/bin/env node"
+        val jsScriptFilePath = jsFilePath.stripSuffix(s".$extension")
+        IO.createFile(jsScriptFilePath, s"$shebang\n$jsScript", true)
+        jsScriptFilePath
+
+      case OS.Windows =>
+        val jsMainFilePath = jsFilePath.stripSuffix(s".$extension") + "__main.js"
+        val exePath = jsFilePath.stripSuffix(s".$extension")
+        val batScript = s"node $jsMainFilePath"
```
jiribenes, line 155: This variable is unused, right?

marzipankaiser: Ah, yes, this is left-over from an older version.
```suggestion
```

marzipankaiser:
@jiribenes I think I'd delay the fix of escaped newlines to another issue/PR (out of scope for this). So: Merging :tada:

PR #445, opened 2024-04-21
Added case study on probabilistic programming and inference

sinaschaefer:
Added the case study `Inference` that contains the contents of my bachelor's thesis `Probabilistic Programming and Programmable Inference in Effekt`.

The content includes Slice Sampling and Metropolis-Hastings algorithms, as well as some examples to show how these algorithms can be applied.


b-studios:
The tests are failing since a check file is missing. You just need to create a file with the same name that ends with `.check` and contains the expected outputs. You can look into the other case studies to see how this works.

b-studios:
The expected output might be a bit difficult to state since it could rely on `random`. Is it possible to use a different (deterministic) source of randomness for the tests?

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,534 @@
...
+    do Weight(prob1 / prob0)
+  }
+  ((result1, trace1), prob1)
+}
+```
+`metropolisStep` is then called in the algorithm implementation
+```
+def metropolisHastingsAlgo[A]{program: () => A / {Sample, Weight}} = {
+  val ((result0, trace0), prob0) = handleWeight{
+    handleRandom{
```
b-studios, line 278: Since `handleRandom` is implemented in terms of "global" randomness, the order of handlers here shouldn't matter, right? So we could remove `handleRandom` here and leave randomness uninterpreted.

sinaschaefer: I think it would be possible to move `handleRandom` from this position to the wrapper `metropolisHastings`, but I'm not sure if it is possible to leave the randomness uninterpreted since `Random` is used to generate samples.
In terms of the order of the handlers, since `handleSample` (indirectly) uses the effect `Random`, `handleRandom` has to be applied after applying `handleSample`.

b-studios: I am proposing to move it completely to the outside, that is `main`. This way everything could in principle use randomness, but we can implement it in a deterministic manner.

b-studios: I am proposing to move it completely to the outside, that is `main`. This way everything could in principle use randomness, but we can implement it in a deterministic manner for testing.

dvdvgt:
Thanks for your contribution, @sinaschaefer. In order for the tests to pass, we need to take care of two more things:

1. As you can see [here](https://github.com/effekt-lang/effekt/actions/runs/8860282023/job/24382841865#step:9:802), you need to annotate your extern definition of `pow` with our new ["feature-flag"](https://github.com/effekt-lang/effekt/pull/427). Since we have multiple different backends, one extern definition is ambiguous and compiler can not safely figure out for which backend your extern definition works.
  
    This may all sound horribly complicated, but in the end, you just need to prepend `js` to the right-hand side of the definition:
    ```
    extern pure def pow(x: Double, y: Double): Double = js "Math.pow(x, y)"
    ```
2. Now, we can finally get the CI to run. Since your examples only target the JavaScript backend, we need to ignore your tests when testing the other backends (or add definitions for other backends to the extern definition). For that, you need to add the line `examplesDir / "casestudies" / "inference.effekt.md"` to the files 
    https://github.com/effekt-lang/effekt/blob/master/effekt/jvm/src/test/scala/effekt/LLVMTests.scala
    https://github.com/effekt-lang/effekt/blob/master/effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
    https://github.com/effekt-lang/effekt/blob/master/effekt/jvm/src/test/scala/effekt/MLTests.scala
    that is, to all backends for which your extern definitions do not work.

I hope this is helpful, and if not, do not hesitate to shoot us a message.

dvdvgt:
@sinaschaefer, if you'd give me [write permission](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-access-to-your-personal-repositories/inviting-collaborators-to-a-personal-repository) to your fork, I can happily make these changes for you.

dvdvgt:
@sinaschaefer, it appears you gave me permissions for the wrong repository. I need them for this one: https://github.com/sinaschaefer/effekt. Thanks in advance!

sinaschaefer:
@dvdvgt Oh sorry, now it should be right. Thanks!

b-studios:
@dvdvgt could you round to make test go through?

dvdvgt:
It seems there's no simple solution for this. An generic `Emit[R]` effect is used for printing the results. Therefore, it is not know wether `R` is something that can be rounded before printing. Rounding where `do emit` is envoked does not work either because, there as well, the type of the thing being emited is not know and cannot be assumed to be a double.

I am enclinded to just comment out the parts that cause the failing of the tests due to varying precision of doubles depending on the machine it is run on.

b-studios:
We could handle into a structure like `List[T]` instead of inspecting. This would make testing much easier anyways. 

What do you think @phischu ?

phischu (approved):

phischu:
Merge it and revise it later.

b-studios:
I agree that we need to merge it soon -- but you know that we will never revise it, if we do not fix it now. Since @dvdvgt is already working on it, we could "quickly" fix it. Otherwise somebody in the future has to wrap their heads around it again and again just to enable these tests.

examples/casestudies/inference.effekt.md
```diff
@@ -309,90 +310,83 @@ def metropolisStep[A](prob0: Probability, trace0: Trace){program: Trace => (A, P
...
+    program()
   }
-
   def loop(result: A, trace: Trace, prob: Probability): Unit / Emit[A] = {
     do emit(result)
-    val ((result1, trace1), prob1)=
-    handleSample{
-      handleRejection{
-        metropolisStepSingleSite(prob, trace){ trace => handleWeight { handleReusingTrace(trace) { program() } } }
+    val ((result1, trace1), prob1)= handleSample {
```
b-studios, line 380: Here also the spacing seems off

examples/casestudies/inference.effekt.md
```diff
@@ -295,12 +297,11 @@ def propose(trace: Trace): Trace / Sample = {
...
 
 ```
-def metropolisStep[A](prob0: Probability, trace0: Trace){program: Trace => (A, Probability)} = {
+def metropolisStep[A](prob0: Probability, trace0: Trace) { program: Trace => (A, Probability) } = {
   val trace1 = propose(trace0)
   val (result1, prob1) = program(trace1)
-
-  if(prob1 < prob0){
-    do Weight(prob1 / prob0)
+  if(prob1 < prob0) {
```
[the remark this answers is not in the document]
dvdvgt: I tried to. I guess I missed some

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+  with Emit[A] {
+    def emit(element) = {
+      do emit(element);
+      steps = steps - 1;
+      if (steps > 0) { resume(()) }
+    }
+  }
+}
+
+def collect[A,R](n: Int){ program: () => R / Emit[A] }: List[A] = {
```
b-studios, line 188: You can delete this, if we didn't use it

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+      steps = steps - 1;
+      if (0 < steps) { resume(()) }
+    }
+  }
+  result.reverse
+}
+```
+
+## Rejection Sampling
+The effect `Weight` is also the basis of the rejection handling algorithm
```
b-studios, line 204: Here weight should also be lower case

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+
+```
+def handleRejection[R] { program: () => R / weight }: R / random = {
+  try { program() }
+  with weight { (prob) =>
+    if (do random() < prob) { resume(()) }
+    else { handleRejection{ program } }
```
b-studios, line 211:
```suggestion
    else { handleRejection {program} }
```
Actually in this case I was briefly confused. Maybe calling `program()` here is clearer.

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+    with handleObserve
+    do emit(program())
+    loop()
+  }
+  loop()
+}
+```
+
+## Slice Sampling
+
```
b-studios, line 232: @phischu is there a sentence of explanation that you could/ would add here?

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+}
+```
+
+## Metropolis-Hastings
+
+### Tracing
+A trace records past samples used by an algorithm. The trace also controls where the algorithm draws samples from in the next iterations.
+Constructing traces is possible by handling the effect `Sample`, not with the default handler, but one that draws new samples and records them in a trace.
+
+```
```
b-studios, line 257: I would move the definition of type Trace here, where it is explained.

In general, this is a literate programming file: Explanations and definitions should be close to each other.

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+  }
+  loop(result, prob)
+}
+```
+
+## Metropolis-Hastings
+
+### Tracing
+A trace records past samples used by an algorithm. The trace also controls where the algorithm draws samples from in the next iterations.
+Constructing traces is possible by handling the effect `Sample`, not with the default handler, but one that draws new samples and records them in a trace.
```
b-studios, line 255:
```suggestion
Constructing traces is possible by handling the effect `sample`, not with the default handler, but one that draws new samples and records them in a trace.
```

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+
+```
+def handleTracing[R] { program: () => R / sample }: (R, Trace) / sample = {
+  try { (program(), []) }
+  with sample { (dist) =>
+    val d = do sample(dist);
+    val (r, trace) = resume(d);
+    (r, Cons(d, trace))
+  }
+}
```
b-studios, line 265: Feel free to ignore this, but why not use mirac6le state, like in other handlers?

```suggestion
  var trace = []
  val resul = try { program() }
  with sample { (dist) =>
    val d = do sample(dist);
    trace = Cons(d, trace)
    resume(d)
  }
  (result, trace.reverse)
}
```

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+  def rounded(d: Double) = round(d, 3).show
+  "State(" ++ x.rounded ++ ", " ++ y.rounded ++ ", " ++ vx.rounded ++ ", " ++ vy.rounded ++ ")"
+}
+```
+
+In this example we also define a type alias for `Measurement`
+
+```
+type Measurement = Double
+// movement of the robot in one time intervall
```
b-studios, line 458:
```suggestion
// movement of the robot in one time interval
```

maybe quickly run the case stay through a spell checker / LLM

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+def move(s: State): State / sample = {
+  s match {
+    case State(x, y, vx, vy) =>
+      val xa = do sample(Gaussian(0.0, 1.0))
+      val ya = do sample(Gaussian(0.0, 1.0))
+      return State(x+vx+xa, y+vy+ya, vx+xa, vy+ya)
+  }
+}
+// measures distance to robot and observes how probable position is given current state
+def measure(s: State, m: Measurement): Unit / observe = {
```
b-studios, line 468: This is a literate file, why have a comment here is this can be part of the text?

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+```
+record Population(susceptible: Double, infected: Double, recovered: Double)
+
+def show(s: Population): String = {
+  val Population(susceptible, infected, recovered) = s
+  def rounded(d: Double) = round(d, 3).show
+  "Population(" ++ susceptible.rounded ++ ", " ++ infected.rounded ++ ", " ++ recovered.rounded ++ ")"
+}
+
+//multiply s,i,r parts of population with transition probabilitys to get new state of population
```
b-studios, line 495: Same as above, also

```suggestion
// multiply s,i,r parts of population with transition probabilitys to get new state of population
```

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+  p match {
+    case Population(s, i, r) =>
+      val noise = do sample(Gaussian(0.0, 0.01))
+      val s1 = 0.5 * s - noise
+      val i1 = 0.3 * i + 0.5 * s + 0.1 * r + noise
+      val r1 =  0.9 * r + 0.7 * i
+      return Population(s1, i1, r1)
+  }
+}
+//testing the popultaion(like covid test)
```
b-studios, line 506:
```suggestion
// testing the population (like covid test)
```

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+def test(p: Population, pr: Double): Unit / observe = {
+  p match {
+    case Population(s, i, r) =>
+      val mean = i
+      val sampleSize = 100.0
+      do observe(pr, Beta(mean, sampleSize))
+  }
+}
+//approximate next state of population based on current state
```
b-studios, line 515:
```suggestion
// approximate next state of population based on current state
```

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+def step(p0: Population, pr1: Double): Population / { sample, observe } = {
+  val p1 = progression(p0)
+  test(p1, pr1)
+  return p1
+}
+```
+
+Calling the algorithms on the examples.
```
b-studios, line 523: I think this sentence could be revised to sum up everything better
```suggestion
Finally, we can use the implemented inference methods to...
```

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,617 @@
...
+# Inference
+This case study shows how we can perform inference over probabilistic models with the help of effects.
+
+---
+
+We require some imports
+
+```
+import list
+extern pure def pow(x: Double, y: Double): Double =
```
b-studios, line 16: Why is pow defined here and not in the stdlib?

examples/casestudies/inference.effekt.md
b-studios: thanks, typed on my phone and untested

examples/casestudies/inference.effekt.md
dvdvgt: This cannot be done. I am not quite sure why, though.

```
/Users/david/Develop/effekt/inference-effekt/out/inference.js:1689
    return handleRejection_2899((weight$capability148573665550262348_20704) => {
                                ^

RangeError: Maximum call stack size exceeded
    at /Users/david/Develop/effekt/inference-effekt/out/inference.js:1689:33
    at /Users/david/Develop/effekt/inference-effekt/out/inference.js:1041:22
    at /Users/david/Develop/effekt/inference-effekt/out/inference.js:1041:22
    at /Users/david/Develop/effekt/inference-effekt/out/inference.js:1041:22
    at /Users/david/Develop/effekt/inference-effekt/out/inference.js:1041:22
    at /Users/david/Develop/effekt/inference-effekt/out/inference.js:1041:22
    at /Users/david/Develop/effekt/inference-effekt/out/inference.js:1041:22
    at /Users/david/Develop/effekt/inference-effekt/out/inference.js:1041:22
    at /Users/david/Develop/effekt/inference-effekt/out/inference.js:1041:22
    at /Users/david/Develop/effekt/inference-effekt/out/inference.js:1041:22

Node.js v22.5.1
```

examples/casestudies/inference.effekt.md
dvdvgt: I am not sure what you mean by "mirac6le state"

dvdvgt:
I did some inference myself and tried to explain most of the parts of the case study, as I found it hard to follow without pre-existing knowledge. The "metropolis hastings section" might still need some more explanations, though.

I tried to add some formalisation to the linear regression example. I am not sure if my interpretation is correct, and I will happily discuss it.

libraries/common/effekt.effekt
```diff
@@ -407,6 +407,19 @@ extern pure def exp(x: Double): Double =
   chez "(exp ${x})"
   ml "Math.exp ${x}"
 
+def pow(base: Double, exponent: Int): Double = {
+  def loop(acc: Double, exponent: Int): Double = {
+    if (exponent == 0) acc
+    else loop(acc * base, exponent - 1)
+  }
+  loop(1.0, exponent)
+}
```
jiribenes, line 416: We can also make this slightly faster (logarithmic instead of linear) by using a classic algorithm:
```suggestion
def pow(base: Double, exponent: Int): Double = {
  def loop(base: Double, exponent: Int, acc: Double): Double = {
    if (exponent == 0) acc
    else if (exponent.mod(2) == 0) loop(base * base, exponent / 2, acc)  // even exponent
    else loop(base, exponent - 1, acc * base)  // odd exponent
  }
  loop(base, exponent, 1.0)
}
```

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,693 @@
...
+    -0.13857109526572012,
+    9.9843695780195716 * pow(10.0, 0.0 - 6.0),
+    1.5056327351493116 * pow(10.0, 0.0 - 7.0)]
+  // takes the i-th element of a list
+  def takeith(i: Double, p0: List[Double]): Double = {
+    p0 match {
+      case Nil() => panic("empty list")
+      case Cons(p, ps) =>
+        if(i == 1.0) {p}
+        else {takeith(i - 1.0, ps)}
```
dvdvgt, line 91: spaces

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,693 @@
...
+    676.5203681218851,
+    -1259.1392167224028,
+    771.32342877765313,
+    -176.61502916214059,
+    12.507343278686905,
+    -0.13857109526572012,
+    9.9843695780195716 * pow(10.0, 0.0 - 6.0),
+    1.5056327351493116 * pow(10.0, 0.0 - 7.0)]
+  // takes the i-th element of a list
+  def takeith(i: Double, p0: List[Double]): Double = {
```
dvdvgt, line 86: This is only used with integer. Use the one from the stdlib.

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,693 @@
...
+    p0 match {
+      case Nil() => panic("empty list")
+      case Cons(p, ps) =>
+        if(i == 1.0) {p}
+        else {takeith(i - 1.0, ps)}
+    }
+  }
+  //lanczos approximation
```
dvdvgt, line 94: space

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,693 @@
...
+Currently, we support models with Gaussian, Uniform or Beta distributions. Thus, we need a way to draw from these distributions and to compute the probability density.
+
+```
+def draw(dist: Distribution): Double / random = {
+  dist match {
+    case Gaussian(mean, variance) =>
+      val u1 = do random();
+      val u2 = do random();
+      val sigma = variance
+      //box-muller transform
```
dvdvgt, line 53: space

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,693 @@
...
+    current = current * prob;
+    resume(())
+  }
+}
+```
+
+For generating random numbers, we either use the `random` function from stdlib
+
+```
+def handleRandom[R] { program: () => R / random }: R = {
```
dvdvgt, line 159: Too many surrounding curly braces

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,693 @@
...
+
+With the effect `emit` it is also possible to limit infinite loops. This allows for flexible numbers of steps to be performed with any algorithm that uses the effect `emit`.
+
+```
+def onEmit[A] { handler: A => Unit } { program: => Any / Emit[A] }: Unit = {
+  try { program(); () }
+  with Emit[A] {
+    def emit(element) = { handler(element); resume(()) }
+  }
+}
```
dvdvgt, line 190: Too many curly braces

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,693 @@
...
+  val c = do sample(Gaussian(0.0, 2.0))
+  observations.foreach {
+    case Point(x, y) => do observe(y, Gaussian(m * x + c, 1.0))
+  }
+  return Point(m, c)
+}
+```
+
+Remember from earlier that `handleObserve` uses the `weight` effect operation.
+Therefore, the call `do observe(y, Gaussian(...))` can be understood as approximating $\argmax_{\mathbf{\theta} \in \Theta} P(\mathbf{Y} \mid \mathbf{\theta})$ where $\mathbf{Y}$ are the observations and $\mathbf{\theta}$ are the parameters of a Gaussian given by $m \sim \mathcal{N}(0, 3)$ and $c \sim \mathcal{N}(0, 2)$.
```
dvdvgt, line 473: @phischu Please take a look again at this

examples/casestudies/inference.effekt.md
```diff
@@ -0,0 +1,693 @@
...
+```
+
+For simulating the movement of the robot at a given state, we sample accelerations `xa`, `ya` along the x and y axis from a Gaussian distribution.
+By applying the laws of notion and assuming that the difference between each time step is one, we can derive the update equations given the sampled acceleration for the velocity
+$$\begin{align*}a_t = \frac{d v_t}{d t} \Leftrightarrow v_t - v_{t-1} = a_t \Leftrightarrow v_t = v_{t-1} + a_t\end{align*}$$
+and position
+$$v_t = \frac{d x_t}{d t} \Leftrightarrow d x_t = v_t \Leftrightarrow x_t - x_{t-1} = v_t \Leftrightarrow x_t = v_t + x_{t-1} = v_{t-1} + a_t + x_{t-1}.$$
+
+```
+def move(s: State): State / sample = {
```
dvdvgt, line 505: Too many curly braces

examples/casestudies/inference.effekt.md
phischu, line 473:
```suggestion
Remember from earlier that `handleObserve` uses the `weight` effect operation.
Therefore, the call `do observe(y, Gaussian(...))` can be understood as approximating $\argmax_{\mathbf{\theta} \in \Theta} P(\mathbf{Y} \mid \mathbf{\theta})$ where $\mathbf{Y}$ are the observations and $\mathbf{\theta}$ are the parameters of a Gaussian given by $m \sim \mathcal{N}(0, 3)$ and $c \sim \mathcal{N}(0, 2)$.
```
```suggestion
Remember from earlier that `handleObserve` uses the `weight` effect operation.
Therefore, the call `do observe(y, Gaussian(...))` can be understood as $P(\mathbf{Y} \mid \mathbf{\theta})$ where $\mathbf{Y}$ are the observations and $\mathbf{\theta}$ are the parameters of a Gaussian given by $m \sim \mathcal{N}(0, 3)$ and $c \sim \mathcal{N}(0, 2)$.
```

phischu (approved):

b-studios:
It's great that this was finally merged! 💯 

However... with #590 we now support effects like `emit` better. Maybe you could refactor the emit effect to a singleton operation, still?

PR #449, opened 2024-04-25
Refactor Stdlib

b-studios:
This PR uses #427 to refactor the stdlib and share more parts of it.

Issues / annoyances (re)discovered on the way:

- we really need to allow importing "from-the-back" (I remember @jiribenes reported this somewhere)
- we need to split `import` into `include` and `import` and add a project mode!
- there are still issues with #427 (@marzipankaiser). In `arrays.effekt` I meant to write
  ```
  extern def get[T](arr: Array[T], index: Int): Option[T] =
    js {  js::undefinedToOption(arr.unsafeGet(index)) }
  ```
  but it did not work.


examples/casestudies/anf.effekt.md
```diff
@@ -159,8 +159,8 @@ our input:
 ```
 def pipeline(input: String): String =
   parse(input) { parseExpr() } match {
-    case Success(tree) => pretty(translate(tree))
-    case Failure(msg) => msg
+    case examples::casestudies::parser::Success(tree) => pretty(translate(tree))
```
b-studios, line 162: Here I'd like to write `parser::Success(tree)` to disambiguate.

b-studios:
Working on the PR I understand @phischu's sentiment to keep the stdlib and prelude small. Almost every test will load the stdlib which is not the point of the test. So maybe we can change the prelude for tests?

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -291,6 +303,17 @@ object Namer extends Phase[Parsed, NameResolved] {
...
+    //        val featureFlags = bodies.map(_.featureFlag)
+    //        //        Context.warning(pp"Extern definition ${id} is not supported as it is only defined for feature flags ${featureFlags.mkString(", ")}," +
+    //        //          pp"but the current backend only supports ${Context.compiler.supportedFeatureFlags.mkString(", ")}.")
+    //      }
+    //      bodies.foreach {
+    //        case source.ExternBody.StringExternBody(ff, _) if ff.isDefault =>
+    //          Context.warning(pp"Extern definition ${id} contains extern string without feature flag. This will likely not work in other backends, "
+    //            + pp"please annotate it with a feature flag (Supported by the current backend: ${Context.compiler.supportedFeatureFlags.mkString(", ")})")
+    //        case _ => ()
+    //      }
```
b-studios, line 316: @marzipankaiser I have this commented out here; don't we have your check on master anymore? The diff suggests this

b-studios: I'll go ahead and delete this to minify the diff; please complain if we still need this.

b-studios:
This is finally ready for review. It would be great if someone could help me make this work on the website again since the paths changed from

```
immutable/list
```

to just
```
list
```

Also, the prelude is much larger (maybe we want to use a smaller prelude on the website?). Can someone check out how long loading and compiling things now takes on the website, given the version in this PR?

Maybe @dvdvgt (since you are familiar with the website)?

dvdvgt:
> This is finally ready for review. It would be great if someone could help me make this work on the website again since the paths changed from
> 
> ```
> immutable/list
> ```
> 
> to just
> 
> ```
> list
> ```
> 
> Also, the prelude is much larger (maybe we want to use a smaller prelude on the website?). Can someone check out how long loading and compiling things now takes on the website, given the version in this PR?
> 
> Maybe @dvdvgt (since you are familiar with the website)?

Yes, I'll look into it

b-studios:
I'm gonna click merge; please don't build the website with this, yet, until @dvdvgt finished fixing things up.

PR #450, opened 2024-04-29
Cache compiler after stdlib to speed up tests

b-studios:
This Draft PR experiments with caching the compiler after compiling a trivial file.

On my machine, this way I can run all tests (except ML) within 55 seconds (they run _in parallel_, though). For the ML backend the performance of mlton is still the limiting factor.

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/EffektTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/EffektTests.scala
@@ -1,6 +1,5 @@
 package effekt
 
-import effekt.context.Annotations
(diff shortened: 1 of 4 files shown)


b-studios:
Apparently it did not solve the Text File Busy :)

```
java.io.IOException: Cannot run program "/home/runner/work/effekt/effekt/./out/tests/effekt.chezschemelifttests/patternmatching": error=26, Text file busy
```

b-studios:
Overall the caching is quite successful: it brings down the CI times from about 18min to roughly 8-9min where most of the time is spent in the ML compiler.

PR #451, opened 2024-04-30
Extern warnings only if they reach code generation

### LINKED ISSUE #447: Output "unsupported extern" warnings only if used
When compiling an extern definition like
```
extern def foo(): Unit = js """
    "foo"
"""
```
in a non-matching backend (e.g. chez), we output a warning.
It would be nice to only output this warning if the definition is actually used (reachable transitively from main).

(Future work from #427)

marzipankaiser:
Fixes #447 by (variant A):
- Adding a new variant `Unsupported` for `ExternBody`s in all IRs.
- Generating this in `ResolveExternDefs` if no matching clauses were found.
- In the backends, reporting the warning for `Unsupported` and generating code similar to holes.

Also moves the generation of the other warnings into `ResolveExternDefs`.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
@@ -4,6 +4,7 @@ package llvm
 
 import effekt.machine
 import effekt.util.intercalate
+import effekt.util.messages.ErrorReporter
 import effekt.machine.analysis.*
 
 import scala.collection.mutable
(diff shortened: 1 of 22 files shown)


effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -40,12 +41,23 @@ object Transformer {
...
+            }, transform(contents))
+          case u: machine.ExternBody.Unsupported =>
+            u.report
+            VerbatimFunction(transform(returnType), functionName, parameters.map {
+                case machine.Variable(name, tpe) => Parameter(transform(tpe), name)
+              },
+              """call void @hole()
+                |unreachable
+                |""".stripMargin)
+        }
```
b-studios, line 60: @phischu can you look at this?

phischu: Looks ok

Issue #405, opened 2024-03-04
Emit CommonJS in JS backends


jiribenes:
Currently, the `commonjs` export produces the following export at the end:
```js
module.exports = { main: () => main_1394().run() };
```
Unfortunately, browsers use a completely different module system so this produces a crash at runtime (as `module` doesn't exist).

### Alternative 1

Let's hot-fix that by generating the following export (or a similar one) which works both in a browser and in Node.JS:
```js
(typeof module != 'undefined' && module !== null ? module : {}).exports = $MODULE_NAME = {
  main: () => main_1435().run()
}
```
[where `$MODULE_NAME` is the name of our module]

<details>
<summary> 🪄 Here's the `sed` call needed </summary>

```sh
sed -i'' -e 's/module.exports/(typeof module != "undefined" \&\& module !== null ? module : {}).exports = $MODULE_NAME/g' out/$MYFILE.js
```

</details >

Here's the code that would need to be changed:
https://github.com/effekt-lang/effekt/blob/6f8973ae77e4962b67b3cb626142fab7430a6cd8/effekt/shared/src/main/scala/effekt/generator/js/Tree.scala#L46-L48

### Alternative 2

Let's use [ES modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) instead! They have been stable for about 15 years by now and work both in a browser and in Node*.

This way, we could also get rid of the following hack:
https://github.com/effekt-lang/effekt/blob/6f8973ae77e4962b67b3cb626142fab7430a6cd8/effekt/js/src/test/scala/effekt/WebTests.scala#L29-L34

*The only issue is that for Node.JS, all of our files would have to have the `.mjs` extension...

b-studios:
For context: at some point we generated several different module formats (which is the reason why this is still called `commonjs`

https://github.com/effekt-lang/effekt/blob/6f8973ae77e4962b67b3cb626142fab7430a6cd8/effekt/shared/src/main/scala/effekt/generator/js/Tree.scala#L33

It shouldn't be too hard to first hotfix and then implement Alternative 2. 

It might also be relevant that the standalone compilation mode now is whole program. That is,  we do not generate separate files that need to be linked when compiling to JS. This is only enabled for the website.

b-studios:
Let's please implement Alternative 2 with `.mjs` file extensions.

PR #454, opened 2024-05-07
Add feature flags `jsNode` and `jsWeb`

marzipankaiser:
Currently distinguished by the `Backend` creating the `JavaScript` generator object (backend).
Untested.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Backend.scala
+++ b/effekt/jvm/src/main/scala/effekt/Backend.scala
@@ -27,7 +27,8 @@ case class Backend[E](name: String, compiler: Compiler[E], runner: Runner[E])
 object Backend {
 
   def backend(name: String): Backend[_] = name match {
-    case "js"           => Backend("js", js.JavaScript(List("jsNode")), JSRunner)
+    case "js"           => Backend("js", js.JavaScriptNode(), JSNodeRunner)
+    case "js-web"       => Backend("js-web", js.JavaScriptWeb(), JSWebRunner)
     case "chez-monadic" => Backend("chez-monadic", chez.ChezSchemeMonadic(), ChezMonadicRunner)
     case "chez-callcc"  => Backend("chez-callcc", chez.ChezSchemeCallCC(), ChezCallCCRunner)
     case "chez-lift"    => Backend("chez-lift", chez.ChezSchemeLift(), ChezLiftRunner)
--- a/effekt/jvm/src/main/scala/effekt/Backend.scala
+++ b/effekt/jvm/src/main/scala/effekt/Backend.scala
@@ -27,7 +27,7 @@ case class Backend[E](name: String, compiler: Compiler[E], runner: Runner[E])
 object Backend {
 
   def backend(name: String): Backend[_] = name match {
-    case "js"           => Backend("js", js.JavaScript(), JSRunner)
+    case "js"           => Backend("js", js.JavaScript(List("jsNode")), JSRunner)
     case "chez-monadic" => Backend("chez-monadic", chez.ChezSchemeMonadic(), ChezMonadicRunner)
     case "chez-callcc"  => Backend("chez-callcc", chez.ChezSchemeCallCC(), ChezCallCCRunner)
     case "chez-lift"    => Backend("chez-lift", chez.ChezSchemeLift(), ChezLiftRunner)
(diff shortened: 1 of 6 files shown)


effekt/jvm/src/main/scala/effekt/Backend.scala
```diff
@@ -27,7 +27,7 @@ case class Backend[E](name: String, compiler: Compiler[E], runner: Runner[E])
 object Backend {
 
   def backend(name: String): Backend[_] = name match {
-    case "js"           => Backend("js", js.JavaScript(), JSRunner)
+    case "js"           => Backend("js", js.JavaScript(List("jsNode")), JSRunner)
```
jiribenes, line 30: What if I want to generate code which is designed to run in the browser (like my presentation tool), wouldn't I need `jsWeb` here too? 🤔

marzipankaiser: Hm, currently AFAICT there is no web/node distinction when generating code (except that IIUC the website uses the scala.js-compiled Compiler? Or doesn't it?)

One way to solve this would be to add a separate `js-web` backend. Alternatively, we could drop this PR and instead do a runtime check for node/browser.

marzipankaiser:
@jiribenes I vaguely remember you needed to change some things in the js code for your presentation tool - What was that again? We could integrate those changes here for the `JSWeb` backend.

marzipankaiser:
This is related to #405: When introducing a js-web backend, we should have js code that can run in the browser.

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -131,6 +131,47 @@ object JSRunner extends Runner[String] {
...
     jsScriptFilePath
 }
+object JSWebRunner extends Runner[String] {
+  import scala.sys.process.Process
+
+  val extension = "js"
+
+  def standardLibraryPath(root: File): File = root / "libraries" / "js"
+
+  override def prelude: List[String] = List("effekt", "immutable/option", "immutable/list")
```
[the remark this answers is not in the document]
b-studios: +1

b-studios:
Maybe this also needs to be changed, when merging this PR: http://localhost:4000/docs/getting-started#compiling-programs-with-effekt

marzipankaiser:
> Maybe this also needs to be changed, when merging this PR: http://effekt-lang.org/docs/getting-started#compiling-programs-with-effekt

This should have remained unchanged - apart from some names in code, the js (node) backend is unchanged.

marzipankaiser:
Marking this as ready for review. - While interrelated to #511, the hotfix works for now™️ and this could be merged first. Then, we could also start adjusting the stdlib where appropriate.

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -157,6 +157,43 @@ object JSRunner extends Runner[String] {
...
+  import scala.sys.process.Process
+
+  val extension = "js"
+
+  def standardLibraryPath(root: File): File = root / "libraries" / "common"
+
+  override def prelude: List[String] = List("effekt", "option", "list", "result", "exception", "array", "string", "ref")
+
+  def checkSetup(): Either[String, Unit] =
+    Left("Running js-web code directly is not supported (yet). Use `--compile` to generate a js file / `--build` to generate a html file.") // TODO
```
marzipankaiser, line 170:
```suggestion
    Left("Running js-web code directly is not supported. Use `--compile` to generate a js file / `--build` to generate a html file.")
```
I don't think we want to support this, after all.

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -157,6 +157,43 @@ object JSRunner extends Runner[String] {
...
+  def standardLibraryPath(root: File): File = root / "libraries" / "common"
+
+  override def prelude: List[String] = List("effekt", "option", "list", "result", "exception", "array", "string", "ref")
+
+  def checkSetup(): Either[String, Unit] =
+    Left("Running js-web code directly is not supported. Use `--compile` to generate a js file / `--build` to generate a html file.")
+
+  /**
+   * Creates an openable `.html` file besides the given `.js` file ([[path]])
+   * and then returns the absolute path of a shell script opening it.
```
marzipankaiser, line 174:
```suggestion
   * and then errors out, printing it's path.
```

PR #455, opened 2024-05-07
Import with feature flag

marzipankaiser:
Draft imports conditional on feature flags.
Current syntax:
```effekt
import chez/list on chez
```

- This DOES NOT check that this is ok (i.e. the imported names are actually only used on this backend).
- Untested.


marzipankaiser:
It seems that this does not actually not import, as `Namer` still loads it.
But removing it in/before `Namer` will make it fail if used.

marzipankaiser:
I don't think there is a reasonably simple way to implement this well right now, so closing for now.

We could revisit this if we ever track backends in namer (i.e.: Those definitions are only there on backend x, this code is only for backend y, ...).

PR #459, opened 2024-05-15
More benchmarks

phischu:


examples/benchmarks/arewefastyet/bounce.effekt
```diff
@@ -0,0 +1,106 @@
+import io/args
+import text/string
+import mutable/array
+
+record Ball(
+  x: Double,
+  y: Double,
+  xVel: Double,
+  yVel: Double
+)
```
b-studios, line 10: TBH, I would model this as

```
interface Ball {
  def bounce(): Bool
}
```
to be closer to the JS version of arewefastyet.

Then create balls of  type `Ball in global` everywhere and use global mutable state.

examples/benchmarks/arewefastyet/list.effekt
```diff
@@ -0,0 +1,50 @@
...
+  } else {
+    zs
+  }
+
+def run(n: Int) = {
+  var s in global = 0;
+  each(0, n) { _ =>
+    val result = tail(makeList(15), makeList(10), makeList(6));
+    s = s + length(result)
+  };
```
b-studios, line 42: Please benchmark each individual iteration *within* Effekt. Adding this higher-order function call might be significantly expensive compared to the small program.

examples/benchmarks/arewefastyet/nbody.effekt
```diff
@@ -0,0 +1,179 @@
...
+
+    bodies.put(0, sun);
+    bodies.put(1, jupiter);
+    bodies.put(2, saturn);
+    bodies.put(3, uranus);
+    bodies.put(4, neptune);
+
+    var px = 0.0;
+    var py = 0.0;
+    var pz = 0.0;
```
b-studios, line 84: We have to decide what we want to measure. Those variables do not NEED to backtrack correctly and could be allocated in `global` which is way more efficient in JS.

examples/benchmarks/runner.effekt
```diff
@@ -1,7 +1,7 @@
 import args
 import bench
 
-def benchmark(iterations: Int) { bench: Int => Int } { test: => Unit }: Unit = commandLineArgs() match {
+def benchmark { run: Int => Int } { test: => Unit } = commandLineArgs() match {
```
b-studios, line 4: Why drop iterations? We cannot benchmark JS this way.

phischu:
I am happy, ready for merge.

PR #460, opened 2024-05-17
Unnamed types

phischu:
Make the type system of Machine more structural, fixing some issues with upcasting `Nothing`. I hope this does not break anyone.


b-studios:
Are the 256MB *per* stack segment?

PR #461, opened 2024-05-21
Fix boxing of Nothing / Add explicit type annotation to Val, Let

b-studios:


effekt/shared/src/main/scala/effekt/core/Parser.scala
```diff
@@ -105,7 +105,10 @@ class CoreParsers(positions: Positions, names: Names) extends EffektLexers(posit
   // Definitions
   // -----------
   lazy val definition: P[Definition] =
-  ( `let` ~> id ~ (`=` ~/> expr) ^^ Definition.Let.apply
+  ( `let` ~> id ~ opt(`:` ~/> valueType) ~ (`=` ~/> expr) ^^ {
```
b-studios, line 108: there should already by an `optionalType` or similar.

b-studios: See line 126 case `Val`

effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala
```diff
@@ -139,9 +139,19 @@ object PolymorphismBoxing extends Phase[CoreTransformed, CoreTransformed] {
...
+      val coerce = coercer(binding.tpe, transform(tpe))
+      if (coerce.isIdentity) {
+        List(Definition.Let(id, transform(tpe), transform(binding)))
+      } else {
+        val orig = TmpValue()
+        val origTpe = binding.tpe
+        List(
+          Definition.Let(orig, origTpe, transform(binding)),
+          Definition.Let(id, transform(tpe), coerce(ValueVar(orig, origTpe))))
+      }
```
b-studios, line 154: why do you need to translate one let into two here?

effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -185,8 +191,13 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
...
         Def(f.symbol, BlockLit(tparams, cparams, vparams, bparams, transform(body)), transform(rest))
 
-      case v @ source.ValDef(id, _, binding) if pureOrIO(binding) =>
-        Let(v.symbol, Run(transform(binding)), transform(rest))
+      case v @ source.ValDef(id, tpe, binding) if pureOrIO(binding) =>
+        val transformed = Run(transform(binding))
+        val transformedTpe = v.symbol.tpe match {
+          case Some(tpe) => transform(tpe)
+          case None => transformed.tpe
+        }
```
b-studios, line 199: This pattern shows up quite often, how can we simplify?

effekt/shared/src/main/scala/effekt/core/Parser.scala
marzipankaiser: Thanks, will do.

effekt/shared/src/main/scala/effekt/core/Transformer.scala
marzipankaiser: Something like `foo(v.symbol.tpe, transformed.tpe)` might be possible, equivalent to a
`v.symbol.tpe.map(transform).getOrElse(transformed.tpe)`.
Though I'm not sure it's worth it.

I am unsure if we can abstract over more easily since the type of `binding`/`transformed` is different in most cases (`Stmt` for `val`, `Expr` for `let`). - This also means that the call to transform (on the binding) is a different one.

b-studios:
Sadly, this still doesn't solve the LLVM issue for me on #449 

Even when merging this PR here I still get

```
opt: ./out/strings.ll:13764:9: error: multiple definition of local value named 'tmp31441_4389'
        %tmp31441_4389 = load %Pos, ptr %tmp31441_4389p_4562
        ^
```
on 

```
run --backend llvm examples/stdlib/string/strings.effekt
```

marzipankaiser:
@phischu @b-studios
The next minified example is:
```
interface MyException {
  def myraise(): Nothing
}

def myDefault[T]{d: => T}{body: => T / MyException} = try { body() } with MyException {
  def myraise() = d()
}

def main() = {
  try {
    myDefault { val t = do myraise(); t }{ 0 };
    ()
  } with MyException {
    def myraise() = ()
  }
}
```

marzipankaiser:
I think the problem is here:
https://github.com/effekt-lang/effekt/blob/2739c4791191fc18aab554d443ed3fa81d556f25/effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala#L187
https://github.com/effekt-lang/effekt/blob/2739c4791191fc18aab554d443ed3fa81d556f25/effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala#L218
https://github.com/effekt-lang/effekt/blob/2739c4791191fc18aab554d443ed3fa81d556f25/effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala#L226

Where now, we also need to check the paramaters and potentially box/unbox those.
This is only due to subtyping, which was added recently (for parametric polymorphism, representations of ValueTypes match up after instantiation, and BlockTypes do, too - since we do not instantiate implicitly! - this is however no longer true for subtyping).

marzipankaiser:
@b-studios Now, the `StdlibLLVMTests` pass for me (when merging `fix/boxing-nothing` with `refactor/stdlib`).

b-studios:
Can you add the new minified example to the hotfix?

b-studios:
This suggests to me that using PolymorphismBoxing to also deal with Nothing and Any was maybe not a good idea.

b-studios:
I merged it into #449 and it indeed resolves the problems with the LLVM backend there. Let's keep boxing on our list and try to clean this up in the future.

PR #466, opened 2024-05-23
Experiments generating ocaml instead of sml

b-studios:
Just a draft for now.


b-studios:
This was easier than expected. 

@phischu if you are interested, you could benchmark this to compare it to our MLTon implementation used at https://github.com/effect-handlers/effect-handlers-bench

This way we could estimate how well the (global) optimizations performed by MLton perform compared to OCaml.

Issue #464, opened 2024-05-23
Don't treat the unit value as a literal


phischu:
While we do have literal syntax for the unit value `()`, we should not treat it as a literal in later stages of the compiler. Pattern matching should not generate and equality check. We should delete the `infixEq` function on `Unit`. We should not box and unbox `Unit`.

jiribenes:
What's the status of this issue cc @phischu?

phischu:
This will continue to confuse people, but we can close it.

PR #468, opened 2024-05-24
Fix/dont match on unit

b-studios:
This treats one half of #464. Ideally, we would represent unit as a data type (which is defined in the prelude). 

However, we still need to special case its rendering to `()`.

A data type like

```
type Unit { unit() }
```

would be shown as `unit()` by the generic show methods (in `inspect`), which would be surprising. This is not impossible to work around, however.


phischu (approved):
I don't understand it, but if it works it's fine.

b-studios:
This is just a workaround since we still do not translate unit literals to data types.

PR #469, opened 2024-05-24
Add Concurrent IO

b-studios:
This replaces #432 and backports its changes on top of #449.


effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -417,8 +417,9 @@ object Transformer {
   def transform(tpe: machine.Type): Type = tpe match {
     case machine.Positive()          => positiveType
     case machine.Negative()          => negativeType
-    case machine.Type.Int()          => NamedType("Int")
-    case machine.Type.Double()       => NamedType("Double")
+    case machine.Type.Int()          => IntegerType64()
+    case machine.Type.Byte()         => IntegerType8()
+    case machine.Type.Double()       => DoubleType()
```
b-studios, line 422: @phischu Is this ok, or should I revert this?

libraries/llvm/buffer.c
```diff
@@ -13,127 +14,173 @@
 // with clang-generated LLVM from this file's entire translation unit.
 
 
-// Eight bytes for the reference counter.
+// - 8 bytes for the reference counter.
 // E.g. capacity could later also be stored in the header.
 #define BUFFER_HEADER_WIDTH (8)
```
b-studios, line 19: @phischu we still need to turn buffers into valid Pos, can we do this after merging this PR?

libraries/llvm/io.c
```diff
@@ -21,4 +25,541 @@ void c_io_println_String(String text) {
...
+    // state of {
+    //   case UNRESOLVED => NULL
+    //   case RESOLVED   => Pos (the result)
+    //   case AWAITED    => Nonempty list of listeners
+    // }
+    union {
+        struct Pos pos;
+        Listeners* listeners;
+    } payload;
+} Promise;
```
b-studios, line 52: @phischu can we define all structs (Neg, Pos, etc.) like this, so that we do not have to write `struct Pos` everywhere?

libraries/llvm/io.c
```diff
@@ -21,4 +25,541 @@ void c_io_println_String(String text) {
...
+
+void erasePromise(struct Pos promise) {
```
b-studios, line 54: @phischu Should this be called `c_promise_erase`?

libraries/llvm/rts.ll
```diff
@@ -1,5 +1,9 @@
 ; Run-Time System
 
+attributes #0 = { "ccc" }
```
b-studios, line 3: @phischu I added this in experiments of trying to fix the calling convention. Either we annotate it everywhere where we call into c or drop it here. What do you think?

b-studios:
I think this is finally ready for review; @phischu there are still some things we need to address (potentially after this PR):

- consistently naming functions and types within C and LLVM (I would need your input there)
- fixing Buffer to be a Pos (I'd defer this to after merging this PR)
- adding more features to files, and also network (this can be done after merging this PR)
- deciding which of the 3 buffer implementations we now want (also after this PR)

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
phischu: No, this is fine.

PR #470, opened 2024-05-26
Report detailed missing case on non-exhaustive pattern match

### LINKED ISSUE #463: Misleading error message on a nested non-exhaustive pattern match
### Problem

The compiler reports a somewhat misleading error message which tells me that "a case for `Cons` is missing", but it _actually_ means that the `Cons(_, Cons(_, _))` case is missing:
Screenshot 2024-05-22 at 22 01 23

### Potential solution

It would be nicer to use the full scrutinee `Trace` when reporting the error to _actually_ compute which case we're missing.
Here we actually do pass it, but then we don't use it :(
https://github.com/effekt-lang/effekt/blob/2739c4791191fc18aab554d443ed3fa81d556f25/effekt/shared/src/main/scala/effekt/typer/ExhaustivityChecker.scala#L207

I went ahead and tried to debug-print the `scrutinee` in the error message which leads me to believe that it does contain the necessary information:
```
Non exhaustive pattern matching, missing case for Cons at Child(Cons,tail,Root(Var(IdRef(List(),list))))
```

I don't know how we should represent the missing case though, especially if we want to keep the other similar errors in sync.
Maybe the error messages could already try to format it in the style of the source language like:
```
Non exhaustive pattern matching, missing case for Nil()
Non exhaustive pattern matching, missing case for Cons(_, _)
Non exhaustive pattern matching, missing case for Cons(_, Cons(_, _))
```

#### Nitpick

Should we keep the underscores or should we actually provide names based on the accessors (see below)?
```
Non exhaustive pattern matching, missing case for Nil()
Non exhaustive pattern matching, missing case for Cons(head, tail)
Non exhaustive pattern matching, missing case for Cons(head, Cons(head1, tail))
```

### Details

I originally noticed this in a much larger, more complex pattern match, but I minified it for the issue itself.
Here's the minified repro:
```scala
def test(lst: List[Int]) = lst match {
  case Nil() => 42
  case Cons(x, Nil()) => 100
}
```

jiribenes:
Resolves #463 by reporting the whole missing case instead of just the sub-pattern.

### Example

```scala
type BooleyTree {
  Leaf()
  Fork(l: BooleyTree, v: Bool, r: BooleyTree)
}

def testTree(t: BooleyTree): Int = t match {
  case Leaf() => 42
  case Fork(Fork(l, true, Leaf()), _, r) => 1000
}
```

Error before:
```
Non exhaustive pattern matching, missing case for false
Non exhaustive pattern matching, missing case for Fork
Non exhaustive pattern matching, missing case for Leaf
```

Error after:
```
Non-exhaustive pattern matching, missing case Fork(Fork(_, false, _), _, _)
Non-exhaustive pattern matching, missing case Fork(Fork(_, _, Fork(_, _, _)), _, _))
Non-exhaustive pattern matching, missing case Fork(Leaf(), _, _)
```

(I like that one can now copy the `case Fork(Fork(_, false, _), _, _)` suffix to get the actual case they should copy-paste :) )

### Caveats

- It would be nice to offer a quick-fix here, similarly to how it's done in the Scala 3 compiler: https://github.com/scala/scala3/blob/e0c030ccd44089e70629b59d76962c1dfc8dbb16/compiler/src/dotty/tools/dotc/reporting/messages.scala#L871-L891.
- Related to the one above: We construct a textual representation of the missing case now, we could instead pretty-print an actual term instead!
- I made no attempt at all to handle missing guards.
- I have no idea how this works for edge-cases like existential types.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/typer/ExhaustivityChecker.scala
+++ b/effekt/shared/src/main/scala/effekt/typer/ExhaustivityChecker.scala
@@ -8,6 +8,7 @@ import effekt.source.{ Def, ExprTarget, IdTarget, MatchGuard, MatchPattern, Tree
 import effekt.source.Tree.{ Query, Visit }
 import effekt.util.messages.{ ErrorReporter, INTERNAL_ERROR }
 
+import scala.annotation.tailrec
 import scala.collection.mutable
 
 /**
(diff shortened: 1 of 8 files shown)


b-studios (approved):
Great, thanks! 

You are right in that a quick fix would be great. Same holds for case-expansion which shouldn't be too hard.

PR #475, opened 2024-05-29
LLVM first class functions

b-studios:


b-studios:
@marvinborner I mentioned that we should reenable LLVM tests. Since I needed to do this anyways (partially) in this PR, I just went through all of them.

effekt/jvm/src/test/scala/effekt/LLVMTests.scala
```diff
@@ -12,105 +12,102 @@ class LLVMTests extends EffektTests {
...
     examplesDir / "llvm",
+    examplesDir / "pos",
+  )
+
+  lazy val bugs: List[File] = List(
+    examplesDir / "pos" / "issue108.effekt", // seg faults!
+
+    // boxing
+    examplesDir / "benchmarks" / "church_exponentiation.effekt", // ... are used as type parameters but would require boxing.
+    examplesDir / "llvm" / "not-boxed.effekt", // Unboxing is performed on not-boxed int, leading to program termination
```
b-studios, line 23: @marzipankaiser could you help me looking into these bugs, which are related to polymorphism boxing not treating first-class functions correctly at the moment?

effekt/jvm/src/test/scala/effekt/LLVMTests.scala
```diff
@@ -12,105 +12,102 @@ class LLVMTests extends EffektTests {
 
   override lazy val positives: List[File] = List(
     examplesDir / "llvm",
+    examplesDir / "pos",
+  )
+
+  lazy val bugs: List[File] = List(
+    examplesDir / "pos" / "issue108.effekt", // seg faults!
```
b-studios, line 19: @phischu Could you look into this and the MALLOC one below?

effekt/jvm/src/test/scala/effekt/LLVMTests.scala
b-studios: Actually, the second one is the blocker for concurrent IO with libuv, which I minimized

b-studios:
Closing, since this is merged into #469

PR #476, opened 2024-05-29
Box/unbox boxed blocks

marzipankaiser:
When a block of type `B1 at C` is used as a block of type `B2 at C`, we (potentially) have to insert boxing code,
i.e. expand to
- unboxing the original block
- coercing this (potentially eta-expanding the call)
- boxing again, at the same capabilities

This fixes the tests mentioned in #475.
Note: Base is #475, since the llvm backend does not support first-class functions before it anyway.

cc @b-studios


b-studios:
Great, thank you for the quick fix!

PR #477, opened 2024-05-31
Mutable arrays in the LLVM backend

phischu:


libraries/llvm/forward-declare-c.ll
```diff
@@ -24,3 +24,13 @@ declare %Pos @c_buffer_show_Char(%Pos)
 declare %Pos @c_buffer_show_Double(%Pos)
 declare %Int @c_buffer_index(%Pos, i64)
 declare %Int @c_buffer_character_at(%Pos, i64)
+
```
b-studios, line 27: I don't understand why we always put everything in one file here. Doesn't it make more sense to declare them in array.effekt?

b-studios:
Really great, thanks that you implemented both arrays and refs. 

Btw, in my drafts of promises in #469 I tried to stick to a similar strategy of storing the eraser after the RC and treat Promise as Pos this way. 

One issue that I have is: do you really want to directly test the "unsafe" variants?

libraries/common/array.effekt
```diff
@@ -1,90 +1,114 @@
...
-  val arr = emptyArray(listSize);
-
-  foreachIndex(list) { (i, head) =>
-    put(arr, i, head)
-  }
-  return arr;
-}
-
-extern pure def emptyArray[T](size: Int): Array[T] =
+extern pure def unsafeNew[T](size: Int): Array[T] =
```
b-studios, line 17: To be honest, unsafeNew is not unsafe itself. Getting an element from a non-initialized array is. Right now in this PR you are testing "unsafe" implementation details, which of course is fine. However, the user-facing API should be safe. Either, we force construction to maintain the invariant that all fields are set, or we check on reading an element whether it is NULL or not.

TBH the former will almost ALWAYS force users to use `Array[Option[T]]` or a custom option type to then just initialize the array with `None` AND check on every `get`. So I'd assume the latter is much more ergonomic.

libraries/common/array.effekt
```diff
@@ -1,90 +1,114 @@
...
   js "${arr}.length"
   chez "(vector-length ${arr})"
   ml "Array.length ${arr}"
+  llvm """
+    %z = call %Int @c_array_size(%Pos %arr)
+    ret %Int %z
+  """
 
+/**
+ * Unchecked Precondition: Index in bounds
```
b-studios, line 35: Depending on what the construction asserts there is more than this here... (see above)

libraries/common/array.effekt
```diff
@@ -1,90 +1,114 @@
...
   chez "(vector-ref ${arr} ${index})"
   ml "SOME (Array.sub (${arr}, ${index}))"
+  llvm """
+    %z = call %Pos @c_array_get(%Pos %arr, %Int %index)
+    ret %Pos %z
+  """
 
 extern js """
-function put(arr, index, value) {
+function setArray(arr, index, value) {
```
b-studios, line 47: I think many helpers in JS use the naming scheme: `array$set` (since `$` is meant to not show up in user code)

libraries/llvm/array.c
```diff
@@ -0,0 +1,45 @@
+#ifndef EFFEKT_ARRAY_C
+#define EFFEKT_ARRAY_C
+
+void c_array_erase_fields(void *envPtr) {
```
b-studios, line 4: Could you add one comment above that explains in a picture / sentence how arrays are represented?

libraries/llvm/types.c
```diff
@@ -5,11 +5,19 @@
 typedef int64_t Int;
 typedef double Double;
 
+struct Header {
+  uint64_t rc;
+  void (*eraser)(void *);
+};
```
b-studios, line 11: Nice! Now does `buffer.c` also need this? (if yes, please don't fix this in this PR, since this would heavily conflict with #469 )

phischu: Yes.

libraries/common/array.effekt
```diff
@@ -1,90 +1,120 @@
...
-extern pure def copy[T](arr: Array[T]): Array[T] =
+extern global def copy[T](arr: Array[T]): Array[T] =
   js "[...${arr}]"
 
+// Derived operations:
+
+/**
+ * Creates a new Array of size `size` filled with the value `init`
+ */
+extern global def fresh[T](size: Int, init: T): Array[T] =
```
jiribenes, line 73: Note that `fill` is also already used for the same operation on lists — it might be worth it to think about synchronizing the two names.

b-studios: Hmm.. here we used `fresh` since it is the same name as for `Ref`.

b-studios:
We would like to rename:

unsafeFresh -> allocate
fresh -> array / Array  (same for ref, and bytes, and ...)

document in allocate that the default constructor should be preferred and that it is unsafe...

PR #478, opened 2024-06-04
Bugfix: Duplicate definitions in test stdlib

marzipankaiser:


jiribenes:
The test suite in `test.effekt` doesn't seem to be working in LLVM as it requires a top-level binding for `RED`.

PR #479, opened 2024-06-04
Allow new-style `on[E].result` Result reification

marzipankaiser:
We introduced a new style of handling exceptions using `on[E]`-proxy objects.
This was not introduced for `Result`, but only for `Option`.

With this PR, we now also allow code like:
```scala
    with on[MyException].result
    // ...
```


jiribenes (approved):
LGTM :shipit:

marzipankaiser:
Delaying tests to another PR, since they are also missing for the other variants (only `on[...].default` occurs in the `string` tests).

PR #482, opened 2024-06-04
Switch to LLVM's tailcc calling convention by passing arguments without env

marvinborner:
It's not clear how much this change influences the performance of the LLVM backend; we should write/merge the new benchmarking infrastructure first before merging this.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
@@ -24,7 +24,6 @@ object Transformer {
       given BC: BlockContext = BlockContext();
 
       // TODO proper initialization of runtime
-      emit(Call("environment", environmentType, malloc, List(ConstantInt(1024 * 1024))));
       emit(Call("stackPointer", stackPointerType, malloc, List(ConstantInt(256 * 1024 * 1024))));
       emit(Store(ConstantGlobal(PointerType(), "base"), LocalReference(stackPointerType, "stackPointer")));
       pushReturnAddress("topLevel", "topLevelSharer", "topLevelEraser");
@@ -74,9 +73,9 @@ object Transformer {
     statement match {
 
       case machine.Def(machine.Label(name, environment), body, rest) =>
-        defineFunction(name, List(Parameter(environmentType, "environment"), Parameter(stackPointerType, "stackPointer"))) {
+        val parameters = environment.map { case machine.Variable(name, tpe) => Parameter(transform(tpe), name) }
+        defineFunction(name, parameters :+ Parameter(stackPointerType, "stackPointer")) {
           emit(Comment("statement definition"))
-          loadEnvironment(initialEnvironmentPointer, environment)
           eraseValues(environment, freeVariables(body))
           transform(body)
         }
@@ -86,9 +85,9 @@ object Transformer {
       case machine.Jump(label) =>
         emit(Comment("statement jump"))
         shareValues(label.environment, Set())
-        storeEnvironment(initialEnvironmentPointer, label.environment)
 
-        emit(TailCall(transform(label), List(LocalReference(environmentType, "environment"), getStackPointer())))
+        val arguments = label.environment.map(transform)
+        emit(TailCall(transform(label), arguments :+ getStackPointer()))
         RetVoid()
 
       case machine.Substitute(bindings, rest) =>
@@ -150,6 +149,7 @@ object Transformer {
 
         Switch(LocalReference(IntegerType64(), tagName), defaultLabel, labels)
 
+        // this
       case machine.New(variable, clauses, rest) =>
         emit(Comment("statement new"))
         val closureEnvironment = freeVariables(clauses).toList;
@@ -177,6 +177,7 @@ object Transformer {
         eraseValues(List(variable), freeVariables(rest));
         transform(rest)
 
+        // this
       case machine.Invoke(value, tag, values) =>
         emit(Comment("statement invoke"))
         shareValues(value :: values, Set());
@@ -253,11 +254,11 @@ object Transformer {
         val frameEnvironment = freeVariables(frame).toList;
 
         val returnAddressName = freshName("returnAddress");
-        defineFunction(returnAddressName, List(Parameter(environmentType, "environment"), Parameter(stackPointerType, "stackPointer"))) {
+        val parameters = frame.parameters.map { case machine.Variable(name, tpe) => Parameter(transform(tpe), name) }
+        defineFunction(returnAddressName, parameters :+ Parameter(stackPointerType, "stackPointer")) {
           emit(Comment("statement pushFrame / return address"))
           popEnvironment(frameEnvironment);
           // eraseValues(frameEnvironment, frameEnvironment) (unnecessary)
-          loadEnvironment(initialEnvironmentPointer, frame.parameters);
           eraseValues(frame.parameters, freeVariables(frame.body))
 
           transform(frame.body);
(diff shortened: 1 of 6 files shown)


marvinborner:
I was able to (probably) fix the stack overflows after some amount of debugging. However, certain tests now take >10min or do not seem to finish at all. I'm unsure whether this can be easily fixed or is implied by our use of tailcc.

libraries/llvm/rts.ll
```diff
@@ -719,8 +709,8 @@ define void @run(%Neg %f) {
     %fp = load ptr, ptr %fpp
 
     ; call
-    %env = call %Env @malloc(i64 1048576)
-    %result = call fastcc %Pos %fp(%Obj %obj, %Env %env, %Sp %sp)
+    %buf = call %Neg @malloc(i64 1048576)
+    %result = call tailcc %Pos %fp(%Obj %obj, i64 0, %Neg %buf, %Sp %sp)
```
marvinborner, line 713: I'm very unsure about what to do here. Previously we passed uninitialized memory as second argument, what should we do now?

b-studios: 0 is exactly right!

marvinborner: Ah sorry, I actually meant the third argument (second in env). The `%buf` seems wrong (and doesn't work)

b-studios: ah, maybe don't pass anything? The environment is empty in the case and you now pass arguments individually, no?

jiribenes: This looks suspiciously relevant to #504 (there the bug is also somewhere around this suspicious `%env`)

marvinborner: According to my stack traces, the function at `%fp` (at least in some cases) looks like `@f(%Obj %obj, i64 %ev, %Neg %k, %Sp %sp)`, so we need to pass something at least. More specifically, the body of `f` would do something like
```llvm
...
%kp = getelementptr {%Neg}, %Env %env_378, i64 0, i32 0
store %Neg %k, ptr %kp
...
```

Which is then written to further down the stacktrace in some uv functions. Valgrind reports invalid writes and reads from uninitialized memory in this as well as the previous (master) version.

b-studios: Not sure about this one, but maybe. 

Regarding the issue here: for the other versions of `run` you already do the right thing and pass the correct arguments. Here just don't pass buf and you should be good.

b-studios: Sorry, race condition ;) I was still talking about the previous comments.

b-studios: How do you represent closures in the current version? Neg k suspiciously looks like a continuation it should close over

b-studios: Ah sorry, of course you need to pass `obj` here... I am on mobile and couldn't see the whole code.

marvinborner: Not passing something like `%buf` results in errors because the function at `%fp` wants a third argument (e.g. `%Neg %k`). Not sure about continuations, I don't handle them specifically. Will look more into it soon.

I found out that the current valgrind issues in master are because of use-after-frees after https://github.com/effekt-lang/effekt/blob/master/libraries/llvm/io.c#L312-L313 and not because of what I originally thought (I only test timer right now). Not directly related but I think it would be useful to include valgrind or leak/address sanitation in the CI.

b-studios: I still can't run valgrind, since I'm on M1. But I did run sanitization when developing

b-studios: Are all your changes pushed? What do I need to run to replicate?

marvinborner: Yes it should be reproducible. Try any failing test like timer.effekt

b-studios: Here is a minimal example that still segfaults:

```
import io
import io/time

def main() = {
  callback::wait(100, box { () => () })
  eventloop::start()
  eventloop::stop()
}
```

b-studios: Also `-fsanitize=address` does not report anything on this example when passing no arguments.

marvinborner: Hmm still segfaults for me after not passing the arguments. Maybe I misunderstand, what diff  fixes the segfault for you?

b-studios: Even the full `time.effekt` example works on my machine (not passing an argument in `run`) and does not report sanitization issues.

b-studios:
```diff
diff --git a/libraries/llvm/rts.ll b/libraries/llvm/rts.ll
index 5cc02792..d8542555 100644
--- a/libraries/llvm/rts.ll
+++ b/libraries/llvm/rts.ll
@@ -709,8 +709,7 @@ define void @run(%Neg %f) {
     %fp = load ptr, ptr %fpp
 
     ; call
-    %buf = call %Neg @malloc(i64 1048576)
-    %result = call tailcc %Pos %fp(%Obj %obj, i64 0, %Neg %buf, %Sp %sp)
+    %result = call tailcc %Pos %fp(%Obj %obj, i64 0, %Sp %sp)
 
     ; restore stack (TODO this shouldn't be necessary, the moment we pass stacks...; then this is a tail-call again)
     store %Sp %base, ptr @base

```

b-studios: But the sanitizer also didn't report anything on `master` for me, so maybe valgrind does more?

b-studios: More likely a strange Mac OS issue (where it is a bit more lenient with memory ;) )

b-studios: The discussion is getting pretty long, maybe we can take it to Slack.

b-studios:
To document what @phischu told me: we only have three locations where we really NEED tailcc. All of which are generated and not in hand-written LLVM code:

1. jumps
2. return
3. invoke

Since we are in CPS those need to be tail, for all others we do not care necessarily.

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -36,7 +35,8 @@ object Transformer {
       val instructions = BC.instructions; BC.instructions = null;
 
       val entryBlock = BasicBlock("entry", instructions, terminator)
-      val entryFunction = Function(VoidType(), "effektMain", List(), entryBlock :: basicBlocks)
+      // TODO strictly speaking, the entry function should use the C calling convention
+      val entryFunction = Function(Tailcc(), VoidType(), "effektMain", List(), entryBlock :: basicBlocks)
```
b-studios, line 39: I don't understand why it should use the `C` calling convention, could you explain? Isn't this a "jump" to main?

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -86,9 +86,9 @@ object Transformer {
       case machine.Jump(label) =>
         emit(Comment("statement jump"))
         shareValues(label.environment, Set())
-        storeEnvironment(initialEnvironmentPointer, label.environment)
 
-        emit(TailCall(transform(label), List(LocalReference(environmentType, "environment"), getStackPointer())))
+        val arguments = label.environment.map(transform)
+        emit(Call("", Tailcc(), VoidType(), transform(label), arguments :+ getStackPointer()))
```
b-studios, line 91: Shouldn't it be at least `"_"`? It feels wrong to have an empty string here, even though void does not return.

phischu: Done

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
phischu: For documentation: This one is literally called from C. It does a tail call to the user-defined main entry point with the tail calling convention and musttail. When trying to give it the C calling convention this musttail can not be guaranteed.

Issue #481, opened 2024-06-04
[To be discussed] Allow match guards as boolean expressions


marzipankaiser:
When coming from other languages, it is surprising that you can't, e.g. use `not(x is Cons(_,Red()))` in an `if`/`while`.
Also, splitting up longer conditionals using bindings is not possible when using match guards.
Using match guards instead of equality predicates is especially useful on llvm, since there is no `infixEq` on data types, and no generic `equals` implemented.

## Examples
```effekt
while(not(asBool(cmd is Some("quit")))) {
     cmd = readCmd()
     //...
}
```

## Proposal
Allow `x is p` in expression position and desugar it to `if(x is p) { true } else { false }`.

## Considerations and Alternatives
- It might be clearer to add some syntax for this, e.g. `asBool(x is p)` (better name required)
- Regarding binding, the behaviour is slightly unexpected when changing to booleans. This could be ameliorated by adding a check that the converted pattern does not bind anything. Note that this disallows certain patterns, though (e.g. `(x is Some(y) and valid(y)`).
- For the usecase of `while(not...)`, it would be possible to add `until`, e.g.

jiribenes:
related to this: could we somehow unify `and` with `&&`, and `or` with `||` (#805)?
(i.e. allow `val x = true and false`, `val y = false or true`)

PR #483, opened 2024-06-04
Allow matchGuards in expression position with `?(...)`

marzipankaiser:
Proof-of-concept implementation of #481 with the syntax `?(x is p)`.


b-studios:
Wouldn't it be also an alternative to support negated guards (that is not negated patterns in general, but only guards)?

marzipankaiser:
Depends on what exactly you mean by negated guards. The common usecase would be something like `x is not Some("...")`,
which I'd say is a negated pattern?
Guards which are not patterns (`is`), in my understanding, are boolean expressions and could already be negated, no?

marzipankaiser:
This has diverged far enough its original state that it is probably easier to re-implement when we want it.

PR #485, opened 2024-06-05
Use externs in `PolymorphismBoxing`

marzipankaiser:
With this PR, `PolymorphismBoxing` will search for `extern pure def unboxT` and `extern pure def boxT` to box/unbox values.
If those are not found, it will still search for a record `BoxedT`.

It also changes the search s.t. the definitions will also be found if they are in a namespace.

This allows us to, e.g., store `int`s in the tag of a `%Pos` in LLVM.

### DIFF
--- a/libraries/common/effekt.effekt
+++ b/libraries/common/effekt.effekt
@@ -604,11 +604,31 @@ def each(start: Int, end: Int) { action: (Int) => Unit / Control } = {
 
 def repeat(n: Int) { action: () => Unit / Control } = each(0, n) { n => action() }
 
+namespace internal {
+  namespace boxing {
+    // For boxing polymorphic values
+    extern type BoxedUnit
+    extern pure def boxUnit(u: Unit): Unit =
+    extern pure def unboxUnit(u: Unit): Unit =
 
-// For boxing polymorphic values
-record BoxedUnit(unboxUnit: Unit)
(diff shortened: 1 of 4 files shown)


libraries/common/effekt.effekt
```diff
@@ -604,11 +604,31 @@ def each(start: Int, end: Int) { action: (Int) => Unit / Control } = {
...
-record BoxedBool(unboxBool: Bool)
-record BoxedDouble(unboxDouble: Double)
-record BoxedString(unboxString: String)
\ No newline at end of file
+namespace internal {
+  namespace boxing {
+    // For boxing polymorphic values
+    extern type BoxedUnit
+    extern pure def boxUnit(u: Unit): Unit =
+    extern pure def unboxUnit(u: Unit): Unit =
```
b-studios, line 612: Remember that we do not need to box Unit and Boolean (and hopefully soon String) in LLVM

marzipankaiser: As for `String`: That means that we still need to box them?

b-studios: They are represented as `c_buffer` at the moment, which is not (yet) a `Pos`. We make it one though, which should be easy. Better box it here and we remove the boxing the moment this is fixed.

PR #486, opened 2024-06-05
Print types of block arguments without an Option wrapper

### LINKED ISSUE #473: Block types are pretty printed with superfluous `Option`
On current `main` (aa897e4), block types are pretty printed, but wrapped in `Option( ... )`:
Screenshot 2024-05-28 at 14 33 50
Screenshot 2024-05-28 at 14 34 27

However, it works correctly on the website (I believe the last update was a702e93):
Screenshot 2024-05-28 at 14 35 36

<details><summary>Code</summary>

```scala
def bar { prog : Int => Unit } = prog(42)
def foo { prog : => Unit } = prog()
```
</details>

jiribenes:
Resolves #473 by fixing a regression from #417 which mistakenly wrapped block types with `Option( ... )`

I think this only shows up when hovering over `bar` (see below) -- do we have a way to reliably test this?

---

Before:
Screenshot 2024-05-28 at 14 34 27

After:
Screenshot 2024-06-05 at 11 30 24


b-studios (approved):
Great, thanks!

Issue #390, opened 2024-02-06
commandLineArgs() passing doesnt work for chez backend executables


IR0NSIGHT:
```sh
# javascript backend (default)
max-priv@NOAH367-L:~/Dokumente/repos/fasteffekt$ effekt.sh -b ./chez-lift-runner.effekt && ./out/chez_lift_runner hello world i am args
Cons(hello, Cons(world, Cons(i, Cons(am, Cons(args, Nil())))))

# chez lift backend
max-priv@NOAH367-L:~/Dokumente/repos/fasteffekt$ effekt.sh -b ./chez-lift-runner.effekt --backend chez-lift && ./out/chez_lift_runner hello world i am args
Nil()

# chez-callcc
max-priv@NOAH367-L:~/Dokumente/repos/fasteffekt$ effekt.sh -b ./chez-lift-runner.effekt --backend chez-callcc && ./out/chez_lift_runner hello world i am args
Nil()

# chez-monadic
max-priv@NOAH367-L:~/Dokumente/repos/fasteffekt$ effekt.sh -b ./chez-lift-runner.effekt --backend chez-monadic && ./out/chez_lift_runner hello world i am args
Nil()

```
ran after pulling latest master:
```
commit d8a1fba426511663c97386eedc00b244744fa6e4 (HEAD -> master, origin/master, origin/HEAD)
Merge: ea9a6c64 20dd400a
Author: Jonathan Immanuel Brachthäuser <jonathan@b-studios.de>
Date:   Mon Feb 5 18:16:27 2024 +0100
```

minimal effekt implementation:
```js
import io/args

def main(): Unit = {
    println(commandLineArgs())
}
``` 
note 1:
issue can be fixed by changing the executable bash script to pass the arguments, tested only on chez-lift:
```sh
#!/bin/bash
scheme --script /home/max-priv/Dokumente/repos/fasteffekt/./out/chez_lift_runner.ss "$@"

-----------------------------------------------------------------

max-priv@NOAH367-L:~/Dokumente/repos/fasteffekt$ ./out/chez_lift_runner i am args
Cons(i, Cons(am, Cons(args, Nil())))
``` 

note: 
llvm and ml backends dont even compile because they cant typecheck println(commandLineArgs())
will make separate issues there.

jiribenes:
Current state after stdlib changes #449:
```sh
[...]

PR #487, opened 2024-06-05
Allow passing command line parameters directly

### LINKED ISSUE #300: Pass command-line arguments when running a program
# expected behaviour:
WHEN i run 
```sh
effekt.sh myFile.effekt hello world
```
THEN myFile.effekt main() gets executed
AND commandLineArgs() returns ["hello","world"]

# observed behaviour:
WHEN i run 
```sh
effekt.sh myFile.effekt hello world
```
THEN myFile.effekt main() gets executed
AND throws:
```sh
hello (Datei oder Verzeichnis nicht gefunden)
world (Datei oder Verzeichnis nicht gefunden)
```
AND commandLineargs() evaluates to Nil() 

# Notes
i could not find a way to pass cmdargs directly into an execution.
currently i have to compile to js and call that file with the cmd args using Node.js

marzipankaiser:
This changed the argument parsing to interpret arguments after `--` as arguments to the compiled program.
That is, instead of
```sh
effekt --backend ml someprogram.effekt
out/someprogram 1 2 3
```
it is now possible to directly run
```sh
effekt --backend ml someprogram.effekt -- 1 2 3
```

Resolves #300.

Note that this does not solve the issues in #390, so those are still not correct in some backends.
Also note that the current behavior will be to pass the arguments to *all* compiled files.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/EffektConfig.scala
+++ b/effekt/jvm/src/main/scala/effekt/EffektConfig.scala
@@ -5,7 +5,7 @@ import effekt.util.paths.file
 import kiama.util.REPLConfig
 import org.rogach.scallop.{ ScallopOption, fileConverter, fileListConverter, longConverter, stringConverter, stringListConverter }
 
-class EffektConfig(args: Seq[String]) extends REPLConfig(args) {
+class EffektConfig(args: Seq[String]) extends REPLConfig(args.takeWhile(_ != "--")) {
 
   // Common
   // ------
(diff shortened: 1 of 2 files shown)


effekt/jvm/src/main/scala/effekt/EffektConfig.scala
```diff
@@ -58,6 +58,8 @@ class EffektConfig(args: Seq[String]) extends REPLConfig(args) {
     group = common
   ).map(Backend.backend)
 
+  def runArgs(): Seq[String] = args.dropWhile(_ != "--").drop(1)
```
b-studios, line 61: I first thought this is gonna crash at runtime for arguments that do not contain `--`, but I wasn't aware of the contract on `drop`.

PR #488, opened 2024-06-07
Rename constructors of collections

### LINKED ISSUE #484: Rename constructors of collections
Follow-up from #477

In `array`, `ref`, `bytes`, and other collections:
1. Rename `unsafeFresh` to `allocate` and clearly document that it is unsafe
    - perhaps we should decide if we want to have `collection::unsafeFunction` or `collection::unsafe::function` in general for functions acting on the collections that are generally unsafe
2. Rename `fresh` (formerly `fill`) to `$collection_name` or `$Collection_name`, for example `array` or `Array`
    - we also discussed using `new` instead, but that's: a) ambiguous (as now we need the `array::` qualifier to determine that it's a new _array), and b) clashes with the existing `new` keyword

jiribenes:
## Motivation

Resolves #484 

## Description

In `array`, `ref`, `bytes`:
1. Rename `unsafeFresh` to `allocate` and clearly document that it is unsafe
2. Rename `fresh` (formerly `fill`) to `$collection_name` or `$Collection_name`, for example `array` or `Array`
3. Add documentation comments where needed :)

I also restructured `array` a little bit so that the "constructors" are at the top (intended for easier API discovery).

### Questions

1. Are we happy with the naming scheme of `$collection_name` for the "fill"/"fresh" constructor?
    - if so, are there any other data structures we want to support here? (what about `bytes`, for example?)
2. Should unsafe functions be in their own sub-namespace?
2. Is there anything else we want to add in this PR?

### DIFF
--- a/libraries/common/array.effekt
+++ b/libraries/common/array.effekt
@@ -4,17 +4,16 @@ import effekt
 import exception
 import list
 
-// Primitive operations:
-
 /**
  * A mutable 0-indexed fixed-sized array.
  */
 extern type Array[T]
 
 /**
- * Violated Postcondition: Fields contain values
+ * Allocates a new array of size `size`, keeping its values _undefined_.
+ * Prefer using `array` constructor instead to ensure that values are defined.
  */
-extern global def unsafeFresh[T](size: Int): Array[T] =
+extern global def allocate[T](size: Int): Array[T] =
   js "(new Array(${size}))"
   chez "(make-vector ${size})" // creates an array filled with 0s on CS
   llvm """
@@ -22,6 +21,35 @@ extern global def unsafeFresh[T](size: Int): Array[T] =
     ret %Pos %z
   """
 
+/**
+ * Creates a new Array of size `size` filled with the value `init`
+ */
+extern global def array[T](size: Int, init: T): Array[T] =
+  ml "Array.array (${size}, ${init})"
+  default {
+    val arr = allocate[T](size);
+    each(0, size) { i =>
+      unsafeSet(arr, i, init)
+    };
+    arr
+  }
+
+/**
+ * Converts a List `list` to an Array
+ */
+def fromList[T](list: List[T]): Array[T] = {
+  val listSize = list.size();
+  val arr = allocate(listSize);
+
+  foreachIndex(list) { (i, head) =>
+    arr.unsafeSet(i, head)
+  }
+  return arr;
+}
+
+/**
+ * Gets the length of the array in constant time.
+ */
 extern pure def size[T](arr: Array[T]): Int =
   js "${arr}.length"
   chez "(vector-length ${arr})"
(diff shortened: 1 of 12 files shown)


libraries/common/array.effekt
```diff
@@ -4,24 +4,52 @@ import effekt
...
+ */
+extern global def array[T](size: Int, init: T): Array[T] =
+  ml "Array.array (${size}, ${init})"
+  default {
+    val arr = allocate[T](size);
+    each(0, size) { i =>
+      unsafeSet(arr, i, init)
+    };
+    arr
+  }
```
b-studios, line 35: Is this more efficient on ml this way? Or is this even mandatory?

libraries/common/array.effekt
```diff
@@ -128,24 +157,11 @@ def build[T](size: Int) { index: Int => T }: Array[T] = {
...
-  }
-  return arr;
-}
-
 def toList[T](arr: Array[T]): List[T] = {
   var i = arr.size - 1;
   var l = Nil[T]()
   while (i >= 0) {
-    l = Cons(unsafeGet(arr, i), l)
+    l = Cons(arr.unsafeGet(i), l)
```
b-studios, line 164: If I imagine writing `unsafe::get(arr, i)` or `arr.unsafe::get(i)` I am starting to thing that `unsafeGet` is nicer.

libraries/common/io/files.effekt
```diff
@@ -95,7 +95,7 @@ def readFile(path: String): String / { AsyncIO, Exception[IOError] } = {
 
   val readSize = 1048576 // 1MB
   var size = readSize
-  var buffer = bytes::empty(size)
+  var buffer = bytes::allocate(size)
```
b-studios, line 98: Should this be `allocate` or `bytes`? There is no other constructor here.

b-studios:
Great! Thanks a lot for this PR! That cleans up a lot of things

libraries/common/array.effekt
jiribenes: Yeah, same, I originally had `arr.unsafe::get(i)`, but I didn't like the aesthetics of it.

libraries/common/array.effekt
jiribenes: I didn't add this code, I just changed the order of functions in that file, sorry. :)

libraries/common/io/files.effekt
jiribenes: It's closest to `array::allocate` in that it's unsafe.

b-studios: It's safe, no? It just initializes with 0s ;)

jiribenes: I thought about this for a bit and you're right, I've renamed it to just `bytes`.

b-studios (approved):
Great, thanks!

PR #492, opened 2024-06-07
Fix #491 by renaming leftover primitives

### LINKED ISSUE #491: Wrong primitive names in LLVM I/O
## Description

When I try to read from a file:

```scala
// repro.effekt

import io
import io/error
import io/files

def main() = eventloop(box {
  with on[IOError].panic;
  with filesystem;

  val contents = do readFile("repro.effekt")
  println(contents)
})
```

I get the following error:

```sh
bash-5.2$ effekt --backend=llvm repro.effekt
opt: ./out/repro.ll:1114:17: error: use of undefined value '@writeFile'
      call void @writeFile(i32 %fd_int, %Pos %buffer, i64 noundef %offset, ptr %onSuccess_ptr, ptr %onFailure_ptr) #0
                ^
Exception in thread "main" java.lang.RuntimeException: Nonzero exit value: 1
  [...]
```

### Investigation 🔍

Upon taking a closer look at `./out/repro.ll`, I can see that `writeFile` is truly neither defined nor declared there.
This happens even when I try to first build, then run.

If I had to guess, then the real name we're looking for is `c_file_read` in `llvm/io.c`
https://github.com/effekt-lang/effekt/blob/cd741e8a4576c62ab318efd48fc0831958ba7e8d/libraries/llvm/io.c#L476
which got renamed at the end of #469.
When investigating this, I also noticed that the LLVM definition corresponds to the C definition only with the ARM64 calling convention:
https://github.com/effekt-lang/effekt/blob/cd741e8a4576c62ab318efd48fc0831958ba7e8d/libraries/llvm/io.ll#L13

I think it should be something like `declare void @c_file_write(i32, %Pos, i64, %Neg*, %Neg*)` instead? (or `ptr, ptr` for the last two, not sure)

jiribenes:
Resolves #491 by renaming the leftover primitives:

- used `rg "call [a-z%A-Z]* @"` to try and find all of the interesting calls, then scoured through the output to see if we missed some other function besides `writeFile`
- also fixed the calling convention to correspond to how we actually use things (hopefully it works both on AMD64 & x64_86) :)
- added a regression test for the problem in question


b-studios (approved):
Thanks, the calling convention was a left over from the debugging process

PR #493, opened 2024-06-07
Fix string concatenation on the JavaScript backend

jiribenes:
## Description

I don't like the following behaviour:

```scala
> val n1 = "hello"
> val n2 = "world"
> (n1 ++ n2).length
hello5
```

Let's parenthesise the arguments of `++` instead so that this does not happen :)

> [!NOTE]
> Please read the whole thread, this is just a partial quick-fix for a much larger issue!

### Motivation

I was trying to update my Advent of Code submissions, then I noticed a really weird bug that was really hard to track down. In the end, it amounted to the following, where one version concatenated strings into a variable, the other one concatenated in place:
```patch
- $effekt.pure((0 >= tmp2630_2632 + tmp2631_3692.length));
+ $effekt.pure((0 >= n6_3822.length));
```
_Sigh._

<details>
<summary> If anybody's interested, here's a minimised version of the actual problem </summary>

```scala
import string

// NOTE: Used for debugging purposes (for sanity checking)
extern js """
function inspect$my$string(str) {
  let hexRepresentation = '';

  for (let i = 0; i < str.length; i++) {
      let codePoint = str.codePointAt(i);
      console.log(`STR Character at index ${i}: '${String.fromCodePoint(codePoint)}' (code point: ${codePoint})`);
      hexRepresentation += codePoint.toString(16).toUpperCase() + ' ';
  }

  console.log("STR Hexadecimal representation:", hexRepresentation.trim());
}
"""
extern io def inspectMyString(str: String): Unit = js "inspect$my$string(${str})"

def myToInt(str: String): Int / Exception[WrongFormat] = {
  println("Got to myToInt:")
  // Sanity check for first character:
  // BUG: Returns "Success(149)"?!
  inspect(result[Char, OutOfBounds] { str.charAt(0) })
  val zero = '0'.toInt

  def go(index: Int, acc: Int): Int = {
    result[Char, OutOfBounds] { str.charAt(index) } match {
      case Success(c) and c >= '0' and c <= '9' =>
        go(index + 1, 10 * acc + (c.toInt - zero))
      case Success(c) =>
        println("Got to Success(c) but not char:")
        inspect(c)
        inspect(str)
        inspect(index)
        inspect(acc)
        inspectMyString(str)
        inspect(on[OutOfBounds].panic { str.charAt(0) })
        wrongFormat("Not a valid number: '" ++ str ++ "' because of '" ++ c.toString ++ "' (" ++ toInt(c).show ++ ") at index " ++ index.show)
      case Error(_, _) => acc
    }
  }

  go(0, 0)
}

// NOTE: Used as a replacement for infixConcat
extern pure def concat(l: String, r: String): String = js "((${l}) + (${r}))"

def main(): Unit / {} = {
  val first: Option[String] = ["1"].headOption;
  (first, first) match {
    case (Some(x), Some(y)) =>
      println("Unpacking x, y: ")
      inspect(x); inspect(y)
      val n = concat(x, y)
      inspect(n)
      // BUG: If you replace `x ++ y` here with `n`, everything works!
      val value = on[WrongFormat].panic { myToInt(concat(x, y)) }
      println(value)
    case _ => ()
  }
}
```
</details>


libraries/common/effekt.effekt
```diff
@@ -113,7 +113,7 @@ extern io def inspect[R](value: R): Unit =
 // Strings
 // =======
 extern pure def infixConcat(s1: String, s2: String): String =
-  js "${s1} + ${s2}"
+  js "((${s1}) + (${s2}))"
```
jiribenes, line 116: We could also perhaps use `((${s1}).concat(${s2}))` instead, which might be a bit safer (but I haven't tested it thoroughly)

marvinborner:
There seem to be other missing parentheses as well. A quick look suggests for example the following bug:

```
> "abc" == "def"
false
> ("abc" == "def").not
false
> not ("abc" == "def")
false
> val huh = "abc" == "def"
> huh.not
true
```

Maybe a possible solution would be to always wrap FFI strings in parentheses automatically?

jiribenes:
Yeah, you're right, though that wouldn't prevent the bug from the title.
However, we might want to wrap _everything_ in parentheses: literals, variables, ...

libraries/common/effekt.effekt
b-studios: "Safer" in which sense?

b-studios:
This issue was only introduced when we started inlining extern defs (in js). I think @marvinborner is right and we could always parenthesize arguments. For the bug at hand, we also would need to parenthesize the whole rhs. This will be a lot of parenthesis... :)

A bit more proper, but insane solution would be to "just" parse JS...

libraries/common/effekt.effekt
jiribenes: To be honest, I had hoped that it would be more strict [wrt types] than it really is, but it's still "type-safe" in its first argument:
```js
> "abc".concat(5)
"abc5"
> (5).concat("hello")
Uncaught TypeError: 5.concat is not a function
```

PR #495, opened 2024-06-08
Recursive descent

b-studios:
A draft implementation of a recursive descent parser.

Currently missing:

- [x] not feature complete (many non-terminals are unimplemented)
- [x] positions are missing
- [ ] error reporting should be improved (original messages are in TODO comments)
- [ ] the parser for types requires quite a bit backtracking, this could be improved (also by changing the grammar itself)


effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -212,7 +227,7 @@ class RecursiveDescentParsers(positions: Positions, tokens: Seq[Token]) {
     }
 
   /*
-  <tryExpr> ::= try <stmts> <handler>+
+  <tryExpr> ::= try { <stmts> } <handler>+
```
b-studios, line 230: The braces are actually not necessary. It should be Stmt which can be braced statements

dvdvgt: Yes, you're right

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -763,13 +767,31 @@ class EffektParsers(positions: Positions) extends EffektLexers(positions) {
...
+
+  def accept[T](expected: String)(f: PartialFunction[TokenKind, T]): P[T] =
+    acceptToken[T](expected) {
+      case Token(start, end, kind) if f.isDefinedAt(kind) => f(kind)
+    }
+
+  implicit def tok2Parser(kind: lexer.TokenKind): Parser[Elem] =
+    acceptToken(kind.toString()) {
+      case t@Token(_, _, kind1) if kind == kind1 => t
+    } */
```
b-studios, line 794: @dvdvgt do we still need this?

dvdvgt: No, I'll delete it

dvdvgt:
Syntax change: Unicode literals now have the syntax `\u<HEXADECIMAL>+` instead of `\u{<HEXADECIMAL>+}`

dvdvgt:
Tests are failing, because escaping unescaped characters in strings is still missing in the various different backends.

jiribenes:
The current error is that we produce:
```ml
val NEWLINE_CHAR = #"
";
```
from:
```scala
extern ml """
val NEWLINE_CHAR = #"\n";
"""
```
which seems like a good test case :)

jiribenes:
By the way, there's a leftover debug print somewhere which outputs:
```
ModuleDecl(trailing_comment,List(),List(FunDef(IdDef(main),List(),List(),List(),None,Return(Literal((),ValueTypeApp(Unit,List()))))))
```
(can't really find where it's from, but it's in the logs: [59](https://github.com/effekt-lang/effekt/actions/runs/9610685371/job/26507669910#step:10:60))

effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
```diff
@@ -605,12 +605,12 @@ class RecursiveDescentTests extends munit.FunSuite {
...
         |""".stripMargin)
 
-    println(parseProgram(
+    parseProgram(
       """module trailing_comment
         |
         |def main() = ()
         |
-        |// foo""".stripMargin))
+        |// foo""".stripMargin)
```
b-studios, line 613: That was the debug print @jiribenes mentioned.

effekt/jvm/src/test/scala/effekt/ParserTests.scala
```diff
@@ -0,0 +1,58 @@
+package effekt
```
b-studios, line 1: @dvdvgt can we delete this file? Now that we have the recursive descent, I believe it is not necessary anymore, no?

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -353,7 +353,7 @@ class EffektParsers(positions: Positions) extends EffektLexers(positions) {
     )
 
   lazy val valDef: P[Def] =
-     `val` ~> idDef ~ (`:` ~/> valueType).? ~ (`=` ~/> stmt) ^^ ValDef.apply
+    `val` ~> idDef ~ (`:` ~/> valueType).? ~ (`=` ~/> stmt) ^^ ValDef.apply
```
b-studios, line 356: Eventually, we can delete `EffektParsers`. We need to keep `EffektLexers` around for some time, since they are also used for ANSIHighlighting (which could and *should* be changed to use the new lexer, I guess) and the core parsers.

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -0,0 +1,1364 @@
...
+        case Str(s, _) => s
+        case _ => fail("Expected string literal.")
+      }
+
+  // TODO uncomment after rebase:
+  //  def featureFlag(): FeatureFlag =
+  //    ident() match {
+  //      case "default" => FeatureFlag.Default
+  //      case flag => FeatureFlag.NamedFeatureFlag(flag)
+  //    }
```
b-studios, line 576:
```suggestion
```

This is implemented below.

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -0,0 +1,1358 @@
...
+  // Positions
+
+  inline def nonterminal[T](inline p: => T): T = {
+    val startToken = peek
+    val start = startToken.start
+    val res = p
+    val endToken = previous
+    val end = endToken.end + 1 // since positions by lexer are inclusive, but kiama is exclusive
+
+    //    val sc: Any = res
```
dvdvgt, line 1322: Can this be deleted?

b-studios: yes, it looks stupid but I wanted to keep it since we will definitely debug positions in the near future again ;)

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -130,13 +130,24 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
+  *
+  * Furthermore, we try to adhere to the following conventions:
+  * - Use `backtrack` with caution and as a last resort. Try to disambiguate the grammar by using `peek` and also try to
+  *   only use `backtrack` on "shallow" non-terminals, that is, not on a non-terminal containing an expression,
+  *   a statement or a type.
+  * - For the same reason, there is no function `manyWhile[T](p: => T): List[T]` but only 
+  *   `manyWhile[T](p: => T, predicate: => Boolean): List[T]` as this one does not use backtracking.
+  * - Use `fail` for reporting errors.
+  * - Functions consuming tokens have an empty parameter list `()`, functions that do not, have no parameter list (e.g. peek)
+  * - All non-terminals are to use the `nonterminal` function for keeping track of positions.
```
b-studios, line 149: Fantastic, thanks for documenting!

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -768,4 +770,12 @@ object Transformer {
...
   def setStackPointer(stackPointer: Operand)(using C: BlockContext) =
     C.stackPointer = stackPointer;
+
+  def escape(scalaString: String): String =
+    scalaString.foldLeft(StringBuilder()) { (acc, c) =>
+      escapeSeqs.get(c) match {
+        case Some(s) => acc ++= s
+        case None => acc += c
+      }
+    }.toString()
```
b-studios, line 780: I just noted that the implementation of escape is (with exception of the ml backend) always the same. @dvdvgt do you think it should be shared, or should we just duplicate it?

b-studios:
I am merging this PR now. We should not forget about the TODOs that are left. It would also be great to decide whether we want to share the escaping in the various backends (and maybe move it closer to the unescaping?).

b-studios:
Good job @dvdvgt !

PR #410, opened 2024-03-10
Fix multiple string issues

marvinborner:
This closes #408 and closes #415 and closes #353.
This also fixes a bug where newlines could be used in single-quoted strings, causing parsing errors in the backends.

---

I unfortunately had to do some ugly merging from fix/issue415 since its merge would've otherwise caused a merge conflict and the fix/issue415 branch was several commits ahead of fix/issue408...

PR #497, opened 2024-06-10
Fix LLVM escape sequences

marvinborner:
This is related to my [previous fixes](https://github.com/effekt-lang/effekt/pull/410/files#diff-ae47aecca5dc9ae5e4f930f8f26f09f1337346f7ee3188147c3290e0d0168777R533) for escape sequences in strings.
Apparently it's quite hard to get this 100% right since every backend handles strings differently.

Since Javascript escapes the sequences by itself and LLVM does not, I now explicitly de-escape the string using `StringContext` in a much later stage (machine transformer).

Let me know if you come up with a better solution :)

### DIFF
--- a/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
@@ -431,9 +431,10 @@ object Transformer {
       }
 
     case lifted.Literal(javastring: String, _) =>
+      val escapedJavastring = StringContext.processEscapes(javastring)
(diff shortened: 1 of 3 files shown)


jiribenes:
It looks great, but I wonder if this is the best way of doing this long-term.
Perhaps it might be better to directly:
1. unescape in the frontend (in the parser)
2. escape in the transformers for the individual backends if needed

This way, we could actually have nice, consistent string & character escapes across all backends.
However, this still seems like a good fix for now. Let's perhaps discuss this in person. :)

Issue #471, opened 2024-05-26
Inconsistent file names in output directory


marvinborner:
While working on automatic benchmarks I noticed that the names of generated files are not consistent for literate Effekt and normal Effekt files: Literate files include the full relative path while normal files don't.

Reproduction:

- `effekt.sh examples/benchmarks/nqueens.effekt` generates `out/nqueens` and `out/nqueens.js`
- `effekt.sh examples/features/adt.effekt.md` generates `out/examples_features_adt` and `out/examples_features_adt.js` (instead of expected `out/adt` and `out/adt.js`)

I'm not entirely sure which of the two naming schemes is better, but we should probably include the relative path in both cases.

dvdvgt:
Fixed with #498

PR #498, opened 2024-06-10
Use consistent naming scheme for output files

marvinborner:
This PR closes issue #471 by only using the file basename without extension for its output path.


jiribenes (approved):
LGTM, just had to double-check that `"foo".split('/').last` returns `"foo"` :)

PR #499, opened 2024-06-13
Fix generic equals in js backend

### LINKED ISSUE #429: Structural equality on JavaScript backend doesn't work
The structural equality on the JavaScript backend seems to be broken.
Although reflexivity (`x == x`) holds for simple, user-defined types, it doesn't work for `Some(x) == Some(x)`.

```scala
type MyType {
  MySingleCase()
}

record EmptyRecord()

def main() = {
  println(MySingleCase() == MySingleCase())   // ~> true
  println(EmptyRecord() == EmptyRecord())     // ~> true

  println(Some(MySingleCase()) == Some(MySingleCase())) // ~> false
  println(Some(EmptyRecord()) == Some(EmptyRecord()))   // ~> false
}
```

marzipankaiser:
Fixes #429.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/generator/js/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/js/Transformer.scala
@@ -53,7 +53,7 @@ trait Transformer {
 
     val other = freshName("other")
     def otherGet(field: JSName): js.Expr = js.Member(js.Variable(other), field)
-    def compare(field: JSName): js.Expr = js"${get(field)} !== ${otherGet(field)}"
+    def compare(field: JSName): js.Expr = js"!${$effekt.call("equals", get(field), otherGet(field))}"
     val noop    = js.Block(Nil)
     val abort   = js.Return(js"false")
     val succeed = js.Return(js"true")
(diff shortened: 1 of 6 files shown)


effekt/shared/src/main/scala/effekt/generator/js/Transformer.scala
```diff
@@ -53,7 +53,7 @@ trait Transformer {
 
     val other = freshName("other")
     def otherGet(field: JSName): js.Expr = js.Member(js.Variable(other), field)
-    def compare(field: JSName): js.Expr = js"${get(field)} !== ${otherGet(field)}"
+    def compare(field: JSName): js.Expr = js"!${$effekt.call("equals", get(field), otherGet(field))}"
```
jiribenes, line 56: This seems a bit low-level outside of the JS AST representation, but there's probably a reason for it, right?
Otherwise it would be something like `!${js.MethodCall(js"$effekt", js"equals", get(field), otherGet(field))}`?

jiribenes (approved):
LGTM, I have one small nitpick for the structure of the hot-fix, but otherwise this is mergeable and improves the current state, thanks! :shipit:

effekt/shared/src/main/scala/effekt/generator/js/Transformer.scala
jiribenes: Ah, it's Scala code! `$effekt.call` is a method in `Tree.scala`.
Nevermind, retracting.

PR #501, opened 2024-06-17
Change standard library and handle corner case

phischu:


effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -224,6 +224,18 @@ object Transformer {
           Invoke(Variable(transform(id), transform(tpe)), opTag, values)
         }
 
+      case lifted.App(lifted.Member(lifted.Unbox(lifted.ValueVar(id, lifted.ValueType.Boxed(tpe))), op, annotatedTpe), targs, args) =>
```
b-studios, line 227: Is there a more principled way of doing this? I understand (somewhat), that we need to special case `App(Member(...))`, but shouldn't we ANF unbox before this translation -- or during this translation? Similar to how we unbox values?

examples/ml/raytracer.check
```diff
@@ -1038,38 +1038,38 @@
...
-86, 199, 134
+86, 198, 133
 81, 187, 126
-76, 176, 118
+76, 175, 118
 71, 163, 110
-66, 151, 102
-60, 138, 93
-54, 124, 83
+65, 150, 101
```
b-studios, line 1062: Nice, all images got darker ;)

libraries/common/effekt.effekt
```diff
@@ -401,22 +415,22 @@ extern pure def _pi(): Double =
...
-  js "${n}"
-
 extern pure def toInt(d: Double): Int =
-  js "Math.round(${d})"
-  chez "(round ${d})"
-  ml "Real.round ${d}"
+  js "Math.trunc(${d})"
+  chez "(flonum->fixnum ${d})"
+  ml "Real.trunc ${d}"
+  llvm "%z = fptosi double %d to %Int ret %Int %z"
```
b-studios, line 422: Could you add a test for this that passes all backends, to be sure?

libraries/common/effekt.effekt
```diff
@@ -557,6 +583,32 @@ def infixAnd { first: => Bool } { second: => Bool }: Bool =
   if (first()) second() else false
 
 
+// Bitwise operations
+// ===============
+extern pure def bitwiseShl(x: Int, y: Int): Int =
```
b-studios, line 588: I would have added `infixShiftLeft` and << to the parser, but that's fine for now. 

Trivia, did you know that apparently it was called `shl` in Pascal (written infix)?

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
phischu: Yes, eventually we will have to

libraries/common/effekt.effekt
phischu: We will need more bitwise operations and I don't know a good design yet.

b-studios: Which ones are you thinking of?

phischu: `bitwiseAnd`, `bitwiseOr`, `bitwiseNot`, `bitSet`, `bitGet`, sign extended shift right, zero extended shift right, literals, count leading zeroes, rank, select, ...

b-studios: We could at least add:

- `<<` shift left
- `>>` shift right
- `|` or
- `&` and
- `^` xor

and potentially
- `>>>` unsigned shift right (that's what JS does)

b-studios: We do not allow prefix operators at the moment so bitwise negation would need to be a function.

b-studios (approved):

PR #510, opened 2024-06-25
Add debug and valgrind to LLVM tests and config

marvinborner:
This makes use of the `--debug` flag and adds a new flag `--valgrind`:

- `--debug`: Use GCC's sanitization and other debugging flags (LLVM only for now)
- `--valgrind`: Run the final executable using valgrind and disable optimizations during compilation. This is enabled by default for LLVM tests.

This will make a lot of LLVM tests fail (intended).


b-studios:
Sadly this means I cannot run tests anymore :(

PR #511, opened 2024-06-25
Emit ES modules from JavaScript backend

### LINKED ISSUE #405: Emit CommonJS in JS backends

jiribenes:
Resolves #405

Disclaimer: most of this PR was written by Claude 3.5 Sonnet

TODO:
- [ ] test separate compilation
- [ ] test Effekt web
- [ ] test ES modules in the browser
- [ ] run CI
- [ ] tidy up

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Runner.scala
+++ b/effekt/jvm/src/main/scala/effekt/Runner.scala
@@ -122,7 +122,7 @@ trait Runner[Executable] {
 object JSRunner extends Runner[String] {
   import scala.sys.process.Process
 
-  val extension = "js"
+  val extension = "mjs"
 
   def standardLibraryPath(root: File): File = root / "libraries" / "common"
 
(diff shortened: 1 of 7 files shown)


jiribenes:
On Windows, the hack to load a file dynamically without ES modules doesn't work, because there it's a `.mjs` file, not a file-ending-less file.
```
file:///D:/a/effekt/effekt/out/tests/effekt.javascripttests/sideeffects__main.mjs:2
const { pathToFileURL } = require('url');
                          ^
ReferenceError: require is not defined
```
Instead, we might want to:
1. investigate whether we really need the `pathToFileURL` import at all
2. write the import directly nicely in the `.mjs` file:
```scala
    val jsScript = s"""
      |import { main } from '${mjsFilePath}';
      |main();
    """.stripMargin
```

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -158,27 +158,32 @@ object JSNodeRunner extends Runner[String] {
...
+    val mjsFilePath = (out / path).canonicalPath.escape
+    
     // create "executable" using shebang besides the .js file
-    val jsScript = s"require('./${jsFileName}').main()"
+    val mjsFileName = path.unixPath.split("/").last
+    
+    // NOTE: This is a hack since this file cannot use ES imports & exports
+    //       because it doesn't have the .mjs ending. Sigh.
+    //       Also, we add the 'file://' prefix to satisfy Windows.
+    val jsScript = s"import('file://${mjsFileName}').then(({main}) => { main(); })"
```
jiribenes, line 174: I'll need to revisit this: the path should be relative (see #566).

jiribenes: Let's just try using `file:filename` instead 🤔
```suggestion
    // NOTE: This is a hack since this file cannot use ES imports & exports
    //       because it doesn't have the .mjs ending. Sigh.
    //       Also, we add the 'file:' prefix to satisfy Windows.
    val jsScript = s"import('file:${mjsFileName}').then(({main}) => { main(); })"
```

jiribenes: Uhoh.
> (node:8164) UnhandledPromiseRejectionWarning: TypeError [ERR_INVALID_FILE_URL_PATH]: File URL path must be absolute

I'll need to think about this, worstcase we need to dynamically get the current file in JS...

jiribenes:
When we do this, we'll need to change the emitted `out/package.json` added by https://github.com/effekt-lang/effekt/pull/989

jiribenes:
This PR has by now bit-rotted way too much to be usable. Let's close it (the source is still visible, perhaps it can still help somebody in #405)

PR #518, opened 2024-07-09
Additions to the string stdlib

marzipankaiser:
- Int.toChar
- String.toInt(base: Int)

### DIFF
--- a/libraries/common/string.effekt
+++ b/libraries/common/string.effekt
@@ -250,6 +250,13 @@ extern pure def toInt(ch: Char): Int =
   ml "${ch}"
   llvm "ret %Int ${ch}"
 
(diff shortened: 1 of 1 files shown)


libraries/common/string.effekt
```diff
@@ -139,6 +139,46 @@ def toInt(str: String): Int / Exception[WrongFormat] = {
   }
 }
 
+// TODO optimize (right now this will be horribly slow (compared to the native JS version)
+def toInt(str: String, base: Int): Int / Exception[WrongFormat] = {
+
+  if( base > 36 || base < 1 ) {
```
b-studios, line 145:
```
if (base > 36 || base < 1) {
```

PR #509, opened 2024-06-25
Improve readability of generated LLVM code

marvinborner:
This refactoring fixes several small annoyances that occured while trying to debug/understand the generated LLVM code:

- previously we switched between different naming schemes, so I chose camelCase for the names (in accordance to other code in the project). This is up for discussion.
- confusingly shortened names are expanded fully, e.g. `retadrp`->`returnAddressPointer`, `fpp`->`functionPointerPointer`. I realize I may have overdone this, so I'd happily revert some of it after discussion.
- arbitrary fresh names are named according to its general purpose, e.g. `k_123`->`returnAddress_123` or `x_4234`->`intLiteral_4234`
- comments indicate the source in relation to the machine/llvm structs

I believe these changes can massively speed up the process of trying to understand the LLVM code generation (e.g. for new students).

Some ugly fresh names still remain, especially ones that are overloaded from previous phases (e.g. `%l65_3658p_89 = ...`). Fixing this would require additional refactoring across phases.

PR #520, opened 2024-07-17
Even more descriptive LLVM codegen

### LINKED ISSUE #513: More descriptive comments when generating LLVM from Machine
Follow-up from #509

After #509, there are really cool comments like:
```llvm
        ; statement substitution
        ; statement literalEvidence
        %evidenceZero_7509 = add i64 0, 0
```
in the LLVM output.

It would be really cool if instead of `; statement substitution`, it would pretty-print the actual term from Machine like, for example, `; substitution[%x_305 !-> @v_102_30p]` using the LLVM names.

Also, we could consider generating these descriptive comments only when the `--debug` flag is on to prevent the compiler from being slower then necessary.

marvinborner:
This is a follow-up from #509 and therefore closes #513.

I believe descriptive code generation to be the one of the most important aspects in making LLVM debugging simpler and more approachable to new contributors/students. This PR extends upon my previous improvements:

- refactor previously missed abbreviations
- name temporary and single-letter variables according to their rough context in source/core
- add context to comments (e.g. for substitution as suggested by @jiribenes)
- add newlines before comments (as suggested by @b-studios)
- only generate comments in `--debug` mode (as suggested by @jiribenes)

I think some of the "doubly freshened" variables could be removed since previous stages already give some guarantees about their distinctness. I didn't touch that though as this seems rather risky.

Comparing the snippet from the previous PR from `triples.effekt` shows some of the improvements:

<table>
<tr>
<th>Previously</th>
<th>Now</th>
</tr>
<tr>
<td>

```llvm
define fastcc void @returnAddress_27(%Environment %environment, %StackPointer %stackPointer) {
        
    entry:
        ; statement pushFrame / return address
        %stackPointer_28 = getelementptr {%Stack}, %StackPointer %stackPointer, i64 -1
        %resume12_7430p_29 = getelementptr {%Stack}, %StackPointer %stackPointer_28, i64 0, i32 0
        %resume12_7430 = load %Stack, ptr %resume12_7430p_29
        %run_7511p_30 = getelementptr {%Pos}, %Environment %environment, i64 0, i32 0
        %run_7511 = load %Pos, ptr %run_7511p_30
        ; statement substitution
        ; statement literalEvidence
        %evidenceZero_7509 = add i64 0, 0
        ; statement construct
        %tmp_31 = insertvalue %Pos zeroinitializer, i64 0, 0
        %booleanLiteral_7510 = insertvalue %Pos %tmp_31, %Object null, 1
        ; statement pushStack
        %stack_32 = call %Stack @uniqueStack(%Stack %resume12_7430)
        %stackPointer_33 = call %StackPointer @pushStack(%Stack %stack_32, %StackPointer %stackPointer_28)
        ; statement return
        %booleanLiteral_7510p_34 = getelementptr {%Pos}, %Environment %environment, i64 0, i32 0
        store %Pos %booleanLiteral_7510, ptr %booleanLiteral_7510p_34
        %stackPointer_36 = getelementptr %FrameHeader, %StackPointer %stackPointer_33, i64 -1
        %returnAddressPointer_37 = getelementptr %FrameHeader, %StackPointer %stackPointer_36, i64 0, i32 0
        %returnAddress_35 = load %ReturnAddress, ptr %returnAddressPointer_37
        tail call fastcc void %returnAddress_35(%Environment %environment, %StackPointer %stackPointer_36)
        ret void
}

...

define fastcc void @effektMain() {
        
    entry:
        ; program
        %environment = call %Environment @malloc(i64 1048576)
        %stackPointer = call %StackPointer @malloc(i64 268435456)
        store %StackPointer %stackPointer, ptr @base
        %returnAddressPointer_1 = getelementptr %FrameHeader, %StackPointer %stackPointer, i64 0, i32 0
        %sharerPointer_2 = getelementptr %FrameHeader, %StackPointer %stackPointer, i64 0, i32 1
        %eraserPointer_3 = getelementptr %FrameHeader, %StackPointer %stackPointer, i64 0, i32 2
        store %ReturnAddress @topLevel, ptr %returnAddressPointer_1
        store %Sharer @topLevelSharer, ptr %sharerPointer_2
        store %Eraser @topLevelEraser, ptr %eraserPointer_3
        %stackPointer_4 = getelementptr %FrameHeader, %StackPointer %stackPointer, i64 1
        ; statement jump
        tail call fastcc void @main_7256(%Environment %environment, %StackPointer %stackPointer_4)
        ret void
}
```
</td>
<td>

```llvm
define fastcc void @returnAddress_27(%Environment %environment, %StackPointer %stackPointer) {
        
    entry:
        
        ; statement pushFrame / return address, 1 free variables
        %stackPointer_28 = getelementptr {%Stack}, %StackPointer %stackPointer, i64 -1
        %resume12_3393_pointer_29 = getelementptr {%Stack}, %StackPointer %stackPointer_28, i64 0, i32 0
        %resume12_3393 = load %Stack, ptr %resume12_3393_pointer_29
        %run_3475_pointer_30 = getelementptr {%Pos}, %Environment %environment, i64 0, i32 0
        %run_3475 = load %Pos, ptr %run_3475_pointer_30
        
        ; statement substitution
        
        ; substitution [_13_3390 !-> run_3475]
        
        ; statement literalEvidence evidenceZero_3473, n=0
        %evidenceZero_3473 = add i64 0, 0
        
        ; statement construct booleanLiteral_3474, tag 0, 0 values
        %booleanLiteral_3474_temporary_31 = insertvalue %Pos zeroinitializer, i64 0, 0
        %booleanLiteral_3474 = insertvalue %Pos %booleanLiteral_3474_temporary_31, %Object null, 1
        
        ; statement pushStack resume12_3393
        %stack_32 = call %Stack @uniqueStack(%Stack %resume12_3393)
        %stackPointer_33 = call %StackPointer @pushStack(%Stack %stack_32, %StackPointer %stackPointer_28)
        
        ; statement return, 1 values
        %booleanLiteral_3474_pointer_34 = getelementptr {%Pos}, %Environment %environment, i64 0, i32 0
        store %Pos %booleanLiteral_3474, ptr %booleanLiteral_3474_pointer_34
        %stackPointer_36 = getelementptr %FrameHeader, %StackPointer %stackPointer_33, i64 -1
        %returnAddressPointer_37 = getelementptr %FrameHeader, %StackPointer %stackPointer_36, i64 0, i32 0
        %returnAddress_35 = load %ReturnAddress, ptr %returnAddressPointer_37
        tail call fastcc void %returnAddress_35(%Environment %environment, %StackPointer %stackPointer_36)
        ret void
}

...

define fastcc void @effektMain() {
        
    entry:
        
        ; program, 10 declarations
        %environment = call %Environment @malloc(i64 1048576)
        %stackPointer = call %StackPointer @malloc(i64 268435456)
        store %StackPointer %stackPointer, ptr @base
        %returnAddressPointer_1 = getelementptr %FrameHeader, %StackPointer %stackPointer, i64 0, i32 0
        %sharerPointer_2 = getelementptr %FrameHeader, %StackPointer %stackPointer, i64 0, i32 1
        %eraserPointer_3 = getelementptr %FrameHeader, %StackPointer %stackPointer, i64 0, i32 2
        store %ReturnAddress @topLevel, ptr %returnAddressPointer_1
        store %Sharer @topLevelSharer, ptr %sharerPointer_2
        store %Eraser @topLevelEraser, ptr %eraserPointer_3
        %stackPointer_4 = getelementptr %FrameHeader, %StackPointer %stackPointer, i64 1
        
        ; statement jump main_2244
        tail call fastcc void @main_2244(%Environment %environment, %StackPointer %stackPointer_4)
        ret void
}
```

</td>
</tr>
</table>

Let me know if you have further ideas to make the generated code more readable.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/DirectStyleMutableState.scala
+++ b/effekt/shared/src/main/scala/effekt/core/DirectStyleMutableState.scala
@@ -36,12 +36,12 @@ object DirectStyleMutableState extends Phase[CoreTransformed, CoreTransformed] {
       case Stmt.Val(y, ytpe, Put(x, tpe, v), s) => Let(y, ytpe, Put(x, tpe, rewrite(v)), rewrite(s))
 
       case Get(x, tpe) =>
-        val id = Id("tmp")
+        val id = Id("temporary_get")
         val binding = Get(x, tpe)
         Let(id, binding.tpe, binding, Stmt.Return(Pure.ValueVar(id, tpe.result)))
 
       case Put(x, tpe, v) =>
-        val id = Id("tmp")
+        val id = Id("temporary_put")
         val binding = Put(x, tpe, v)
         Let(id, binding.tpe, binding, Stmt.Return(Pure.ValueVar(id, Type.TUnit)))
     }
--- a/effekt/shared/src/main/scala/effekt/core/LambdaLifting.scala
+++ b/effekt/shared/src/main/scala/effekt/core/LambdaLifting.scala
@@ -84,8 +84,8 @@ class LambdaLifting(m: core.ModuleDecl)(using Context) extends core.Tree.Rewrite
     // e.g. f : (Int) => Unit @ {io,exc}   ===>   { (n) => f(n, exc) }
     //   the type of f after transformation is `(Int, Exc) => Unit @ {io}`
     case f @ core.BlockVar(id, core.BlockType.Function(tps, cps, vps, bps, res), capt) if needsCallsiteAdaptation(id) =>
-      val vparams: List[core.ValueParam] = vps map { tpe => core.ValueParam(Id("x"), tpe) }
-      val bparams: List[core.BlockParam] = (cps zip bps) map { case (capt, tpe) => core.BlockParam(Id("f"), tpe, Set(capt)) }
+      val vparams: List[core.ValueParam] = vps map { tpe => core.ValueParam(Id("valueParam"), tpe) }
+      val bparams: List[core.BlockParam] = (cps zip bps) map { case (capt, tpe) => core.BlockParam(Id("blockParam"), tpe, Set(capt)) }
 
       val targs = tps map { tpe => core.ValueType.Var(tpe) }
       val vargs = vparams.map { p => core.ValueVar(p.id, p.tpe) } ++ infos(id).valueArgs
--- a/effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala
+++ b/effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala
@@ -198,7 +198,7 @@ object PolymorphismBoxing extends Phase[CoreTransformed, CoreTransformed] {
       if (coerce.isIdentity) {
         List(Definition.Let(id, transform(tpe), transform(binding)))
       } else {
-        val orig = TmpValue()
+        val orig = TmpValue("letDefinition")
         val origTpe = binding.tpe
         List(
           Definition.Let(orig, origTpe, transform(binding)),
@@ -248,7 +248,7 @@ object PolymorphismBoxing extends Phase[CoreTransformed, CoreTransformed] {
       if (coerce.isIdentity) {
         Stmt.Val(id, transform(tpe), transform(binding), transform(body))
       } else {
-        val orig = TmpValue()
+        val orig = TmpValue("valStatement")
         Stmt.Val(orig, binding.tpe, transform(binding),
           Let(id, transform(binding.tpe), coerce(Pure.ValueVar(orig, binding.tpe)),
             transform(body)))
@@ -508,9 +508,9 @@ object PolymorphismBoxing extends Phase[CoreTransformed, CoreTransformed] {
         override def to = totpe
 
         override def apply(block: B): B = {
-          val vparams: List[Param.ValueParam] = vcoercers.map { c => Param.ValueParam(TmpValue(), transform(c.from)) }
-          val bparams: List[Param.BlockParam] = bcoercers.map { c => val id = TmpBlock(); Param.BlockParam(id, transform(c.from), Set(id)) }
-          val result = TmpValue()
+          val vparams: List[Param.ValueParam] = vcoercers.map { c => Param.ValueParam(TmpValue("valueParam"), transform(c.from)) }
+          val bparams: List[Param.BlockParam] = bcoercers.map { c => val id = TmpBlock("blockParam"); Param.BlockParam(id, transform(c.from), Set(id)) }
+          val result = TmpValue("applicationResult")
           val inner = TmpBlock()
           val vargs = (vcoercers zip vparams).map { case (c, p) => c(Pure.ValueVar(p.id, p.tpe)) }
           val bargs = (bcoercers zip bparams).map { case (c, p) => c(Block.BlockVar(p.id, p.tpe, Set.empty)) }
@@ -525,7 +525,7 @@ object PolymorphismBoxing extends Phase[CoreTransformed, CoreTransformed] {
         }
 
         override def callDirect(block: B, vargs: List[Pure], bargs: List[Block])(using PContext): Expr = {
-          val result = TmpValue()
+          val result = TmpValue("callDirectResult")
           Run(Let(result, rcoercer.from, DirectApp(block, targs map transformArg,
             (vcoercers zip vargs).map {case (c,v) => c(v)},
             (bcoercers zip bargs).map {case (c,b) => c(b)}),
@@ -533,7 +533,7 @@ object PolymorphismBoxing extends Phase[CoreTransformed, CoreTransformed] {
         }
 
         override def call(block: B, vargs: List[Pure], bargs: List[Block])(using PContext): Stmt = {
-          val result = TmpValue()
+          val result = TmpValue("callResult")
           Stmt.Val(result, rcoercer.from, Stmt.App(block, targs map transformArg,
             (vcoercers zip vargs).map { case (c, v) => c(v) },
             (bcoercers zip bargs).map { case (c, b) => c(b) }),
--- a/effekt/shared/src/main/scala/effekt/core/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/core/Transformer.scala
@@ -297,7 +297,7 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
           // if this block argument expects to be called using PureApp or DirectApp, make sure it is
           // by wrapping it in a BlockLit
           val targs = tparams.map(core.ValueType.Var.apply)
-          val vparams: List[Param.ValueParam] = vparamtps.map { t => Param.ValueParam(TmpValue(), transform(t))}
+          val vparams: List[Param.ValueParam] = vparamtps.map { t => Param.ValueParam(TmpValue("valueParam"), transform(t))}
           val vargs = vparams.map { case Param.ValueParam(id, tpe) => Pure.ValueVar(id, tpe) }
 
           // [[ f ]] = { (x) => f(x) }
@@ -321,11 +321,11 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
           // [[ f ]] = { (x){g} => let r = f(x){g}; return r }
           def etaExpandDirect(f: ExternFunction): BlockLit = {
             assert(effects.isEmpty)
-            val bparams: List[Param.BlockParam] = bparamtps.map { t => val id = TmpBlock(); Param.BlockParam(id, transform(t), Set(id)) }
+            val bparams: List[Param.BlockParam] = bparamtps.map { t => val id = TmpBlock("etaParam"); Param.BlockParam(id, transform(t), Set(id)) }
             val bargs = bparams.map {
               case Param.BlockParam(id, tpe, capt) => Block.BlockVar(id, tpe, capt)
             }
-            val result = TmpValue()
+            val result = TmpValue("etaBinding")
             val resultBinding = DirectApp(BlockVar(f), targs, vargs, bargs)
             BlockLit(tparams, bparams.map(_.id), vparams, bparams,
               core.Let(result, resultBinding.tpe, resultBinding,
@@ -411,7 +411,7 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
     //   def loop$13() = if ([[cond]]) { [[ body ]]; loop$13() } else { return () }
     //   loop$13()
     case source.While(guards, body, default) =>
-      val loopName = TmpBlock()
+      val loopName = TmpBlock("whileLoop")
       val loopType = core.BlockType.Function(Nil, Nil, Nil, Nil, core.Type.TUnit)
 
       // TODO double check: probably we are forgetting the capture of the guards!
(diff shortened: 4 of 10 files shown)


b-studios:
Fantastic, thanks!

effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala
```diff
@@ -97,9 +97,11 @@ ${indentedLines(instructions.map(show).mkString("\n"))}
     case ExtractValue(result, aggregate, index) =>
       s"${localName(result)} = extractvalue ${show(aggregate)}, $index"
 
-    case Comment(msg) =>
+    case Comment(msg) if C.config.debug() =>
       val sanitized = msg.map((c: Char) => if (' ' <= c && c != '\\' && c <= '~') c else '?').mkString
-      s"; $sanitized"
+      s"\n; $sanitized"
```
jiribenes, line 102: It seems like `C.config.debug()` isn't defined, the CI reports:

```
[error] -- [E008] Not Found Error: /home/runner/work/effekt/effekt/effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala:100:34 
[error] 100 |    case Comment(msg) if C.config.debug() =>
[error]     |                         ^^^^^^^^^^^^^^
[error]     |                    value debug is not a member of effekt.EffektConfig
```

jiribenes (approved):
Looks good to me: although it's very verbose, it seems like it might be quite helpful for debugging :)
I don't use the LLVM backend too much though, so it might be good to ask somebody who does for a review.

effekt/shared/src/main/scala/effekt/core/DirectStyleMutableState.scala
```diff
@@ -36,12 +36,12 @@ object DirectStyleMutableState extends Phase[CoreTransformed, CoreTransformed] {
       case Stmt.Val(y, ytpe, Put(x, tpe, v), s) => Let(y, ytpe, Put(x, tpe, rewrite(v)), rewrite(s))
 
       case Get(x, tpe) =>
-        val id = Id("tmp")
+        val id = Id("temporary_get")
```
b-studios, line 39: It think`tmp_get` is fine here.

effekt/shared/src/main/scala/effekt/core/LambdaLifting.scala
```diff
@@ -84,8 +84,8 @@ class LambdaLifting(m: core.ModuleDecl)(using Context) extends core.Tree.Rewrite
     // e.g. f : (Int) => Unit @ {io,exc}   ===>   { (n) => f(n, exc) }
     //   the type of f after transformation is `(Int, Exc) => Unit @ {io}`
     case f @ core.BlockVar(id, core.BlockType.Function(tps, cps, vps, bps, res), capt) if needsCallsiteAdaptation(id) =>
-      val vparams: List[core.ValueParam] = vps map { tpe => core.ValueParam(Id("x"), tpe) }
-      val bparams: List[core.BlockParam] = (cps zip bps) map { case (capt, tpe) => core.BlockParam(Id("f"), tpe, Set(capt)) }
+      val vparams: List[core.ValueParam] = vps map { tpe => core.ValueParam(Id("valueParam"), tpe) }
```
b-studios, line 87: Honestly, I rather have `x` and `f`.

b-studios:
I agree that the names are overly verbose. The problem is: those will also show up in all IRs inbetween! Often I rather prefer `x` for values and `f` for blocks when no better names are available since these are the metavariables in our papers.

effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -611,7 +611,7 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
 
     // create joinpoint
     val params = patterns.flatMap { case (sc, p) => boundInPattern(p) } ++ guards.flatMap(boundInGuard)
-    val joinpoint = Context.bind(TmpBlock(), BlockLit(Nil, Nil, params, Nil, body))
+    val joinpoint = Context.bind(TmpBlock("joinpoint"), BlockLit(Nil, Nil, params, Nil, body))
```
b-studios, line 614: Typically, I would use `k` for joinpoints. But we can also use `j` if `k` is used too often.

effekt/shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -212,8 +212,8 @@ case class CallTarget(symbols: List[Set[BlockSymbol]]) extends BlockSymbol { val
  * Introduced by Transformer
  */
 case class Wildcard() extends ValueSymbol { val name = Name.local("_") }
-case class TmpValue() extends ValueSymbol { val name = Name.local("tmp" + Symbol.fresh.next()) }
-case class TmpBlock() extends BlockSymbol { val name = Name.local("tmp" + Symbol.fresh.next()) }
+case class TmpValue(hint: String = "temporary") extends ValueSymbol { val name = Name.local(hint + "_" + Symbol.fresh.next()) }
+case class TmpBlock(hint: String = "temporary") extends BlockSymbol { val name = Name.local(hint + "_" + Symbol.fresh.next()) }
```
b-studios, line 216: Please revert to 
```suggestion
case class TmpBlock(hint: String = "tmp") extends BlockSymbol { val name = Name.local(hint + "_" + Symbol.fresh.next()) }
```

effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -809,7 +809,7 @@ trait TransformerOps extends ContextOps { Context: Context =>
   private[core] def bind(s: Stmt): ValueVar = {
 
     // create a fresh symbol and assign the type
-    val x = TmpValue()
+    val x = TmpValue("statementBind")
```
b-studios, line 812: Let's just call this `r` for `result` :) same for `expressionBind`. Maybe we want to prefix ALL synthesized names with something special to mark them as such. Previously `tmp` told me: "Hey this is synthesized".

jiribenes:
Reading the review, it seems that there are two different use-cases:
1. human-friendly "enough" names (which are then seen everywhere): that's a place where a `x` and `f` make sense
2. provenance information for variables in core: knowing _why_ (for what purpose) we generated the variable

Can we separate these two use-cases somehow? Maybe keep the `hint` field and rename it to `provenance`, using it only in some (?) cases (perhaps rendering it as a comment in the textual IR?)

effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala
```diff
@@ -198,7 +198,7 @@ object PolymorphismBoxing extends Phase[CoreTransformed, CoreTransformed] {
       if (coerce.isIdentity) {
         List(Definition.Let(id, transform(tpe), transform(binding)))
       } else {
-        val orig = TmpValue()
+        val orig = TmpValue("letDefinition")
```
marzipankaiser, line 201: This is the name of a binding to be boxed/unboxed. So a long name would be `toBeCoerced`. Alternatively, also a short name like `x` would work. - it should be very clear from the (immediately) following call to the coercer that this is what's happening.

effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala
```diff
@@ -248,7 +248,7 @@ object PolymorphismBoxing extends Phase[CoreTransformed, CoreTransformed] {
       if (coerce.isIdentity) {
         Stmt.Val(id, transform(tpe), transform(binding), transform(body))
       } else {
-        val orig = TmpValue()
+        val orig = TmpValue("valStatement")
```
marzipankaiser, line 251: Same as above. - All the ones in boxing should probably be called the same, as they are all generated for the value before coercion.

effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala
```diff
@@ -525,15 +525,15 @@ object PolymorphismBoxing extends Phase[CoreTransformed, CoreTransformed] {
...
+          val result = TmpValue("callDirectResult")
           Run(Let(result, rcoercer.from, DirectApp(block, targs map transformArg,
             (vcoercers zip vargs).map {case (c,v) => c(v)},
             (bcoercers zip bargs).map {case (c,b) => c(b)}),
             Return(rcoercer(Pure.ValueVar(result, rcoercer.from)))))
         }
 
         override def call(block: B, vargs: List[Pure], bargs: List[Block])(using PContext): Stmt = {
-          val result = TmpValue()
+          val result = TmpValue("callResult")
```
marzipankaiser, line 536: Same as above. - All the ones in boxing should probably be called the same, as they are all generated for the value before coercion.

marzipankaiser:
> Reading the review, it seems that there are two different use cases:
> 
> 1. human-friendly "enough" names (which are then seen everywhere): that's a place where a `x` and `f` make sense
> 2. provenance information for variables in core: knowing _why_ (for what purpose) we generated the variable
> 
> Can we separate these two somehow? Maybe keep the `hint` field and rename it to `provenance`, using it only in some (very debug-ey?) cases (or perhaps by rendering it as a comment in the textual IR?)

For the second kind, those could also be output into a separate file if there are enough of those infos to be too much noise in the IR, and they are useful.

b-studios:
Another piece of information is the information that these names are generated, at all. TmpValue etc could be printed in llvm with a leading special character to signal that they are not part of the original program.

jiribenes:
For reference, here's the regex used to lex identifiers in LLVM: 
```
[%@][-a-zA-Z$._][-a-zA-Z$._0-9]*
```

marvinborner:
Thanks for the extensive review. I've added all suggestions. I think this is a good compromise between readability, conciseness and formal notations.

Still unsure about `toBeCoerced`, but `x` would also be confusing I think.

effekt/shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -212,8 +212,8 @@ case class CallTarget(symbols: List[Set[BlockSymbol]]) extends BlockSymbol { val
  * Introduced by Transformer
  */
 case class Wildcard() extends ValueSymbol { val name = Name.local("_") }
-case class TmpValue() extends ValueSymbol { val name = Name.local("tmp" + Symbol.fresh.next()) }
-case class TmpBlock() extends BlockSymbol { val name = Name.local("tmp" + Symbol.fresh.next()) }
+case class TmpValue(hint: String = "tmp") extends ValueSymbol { val name = Name.local("v_" + hint + "_" + Symbol.fresh.next()) }
+case class TmpBlock(hint: String = "tmp") extends BlockSymbol { val name = Name.local("b_" + hint + "_" + Symbol.fresh.next()) }
```
jiribenes, line 216: Observation: we could also use `v.tmp.457` (so `Name.local(s"v.${hint}.${Symbol.fresh.next()}")`) as the name if we're fine with identifiers including dots in Core and follow-up phases. :)

effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala
```diff
@@ -198,7 +198,7 @@ object PolymorphismBoxing extends Phase[CoreTransformed, CoreTransformed] {
       if (coerce.isIdentity) {
         List(Definition.Let(id, transform(tpe), transform(binding)))
       } else {
-        val orig = TmpValue()
+        val orig = TmpValue("toBeCoerced")
```
jiribenes, line 201: I think it's also fine to have just something like `coe` there, especially since the result looks like `v_coe_1234` (as opposed to `v_toBeCoerced_1234`, but again, I'm personally fine with both.

jiribenes (commented):
I'll refrain from approving this again since I don't really use this code that much anyways 🙃, but I have some tiny notes.

b-studios (approved):

PR #523, opened 2024-07-23
Experiment: a more resilient lexer

### LINKED ISSUE #522: Improve lexer resilience
It would be really lovely to have a slightly more resilient _lexer_.

There's two different kinds of problems that the lexer can encounter:
1. "completely recoverable": instead of crashing when encountering an invalid escape, just continue lexing the given string and only store the error somewhere to report it later (we don't really care about an invalid escape even in Typer!)
2. "partially recoverable": just create a proper error token and continue lexing; the tricky part here is finding the delimiter until which the error token should be (where we can recover)

jiribenes:
Resolves #522

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/LexerTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/LexerTests.scala
@@ -11,23 +11,40 @@ import munit.Location
 class LexerTests extends munit.FunSuite {
 
   def assertTokensEq(prog: String, expected: TokenKind*)(using Location): Unit = {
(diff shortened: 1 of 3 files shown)


jiribenes:
I think this approach isn't great: separating out soft and hard errors doesn't make things easier.
Instead, I think we should:
1) always advance
2) produce error tokens recognised by the backend
3) leave the actual string parsing, escaping etc. to the Parser (=> moving to a `Token = { offset: Int, length: Int }` representation or something similar)

Issue #252, opened 2023-04-03
Handling without control


phischu:
It would be great if objects could be passed implicitly without introducing an effect handler.

```
interface State {
  def get(): Int
  def set(x: Int): Unit
}

def myComputation(): Int / State = do get()

def main() = {
  var s = 0;
  with new State {
    def get() = s
    def set(x: Int) = s = x
  };
  myComputation()
}
```

The typing rule for `with` should be like

```
G |- b : F    G |- s : 𝜏 / ε
-----------------
G |- with b; s : 𝜏 / (ε - F)
```

b-studios:
Yes, this is on our map for quite some time. I have thought a bit about it: most of it should be fairly straight forward. There are a few things that need to be sorted out:

- what syntactic category (in source) is `with`? Can be a `Stmt`, or a `Def`, or a `Term` -- all would make sense. Maybe `Def` is the most appropriate one.
- much more important: how does this interact with capture tracking (System C). I would assume the outer context captures whatever b and s capture, but the inferred capability would probably need to be transparent. This is a problem, since in Typer associates effect types with tracked block parameters: https://github.com/effekt-lang/effekt/blob/13af16fc1215814905cba9d0839af862d8bd00a1/effekt/shared/src/main/scala/effekt/Typer.scala#L1289

kyay10:
Since `with` exists already, all you'd really need is some way to turn an explicit capability into an implicit one, like #406 suggests (and also #67 block type polymorphism to make it simpler)

If we want usage like:
```effekt
{
  with given{myEffect}
  do someMyEffectOp()
}
```

Currently you can do:
```effekt
def given[R]{cap: MyEffect} {block: => R / {MyEffect}} = try block() with MyEffect {
  // delegate all methods tail-resumptively
}
```
you'd have to write it for every type though.
With #406, you can do:
```effekt
def given[R]{cap: MyEffect} {block: => R / {MyEffect}} = try block() with cap // magic syntax that doesn't exist yet
```
but you'd still have to write it for each type
and with #67 you can finally do:
```effekt
def given[E, R]{cap: E} {block: => R / {E}} = try block() with cap
```
which can just go in the prelude for everyone to use!

phischu:
Syntax-wise I believe we can use `with new MyEffect { ... }; ...`.

Last time we checked there was some difficulty because the captures of implicit things must always be opaque but here they should be transparent.

PR #527, opened 2024-07-26
Basic streaming

phischu:
This one would greatly benefit from #252.


libraries/common/stream.effekt
```diff
@@ -0,0 +1,234 @@
...
+}
+
+def for[A, R] { stream: () => R / Emit[A] } { action: A => Unit }: R =
+  try {
+    stream()
+  } with Emit[A] {
+    def emit(value) = resume(action(value))
+  }
+
+// TODO what if the number is negative? undefined behavior?
```
marzipankaiser, line 16: I'd want the program to either panic or act as if it is 0. (This is not something where I'd expect forever, and this is a place where I could see myself accidentally passing a negative number).

libraries/common/stream.effekt
```diff
@@ -0,0 +1,234 @@
...
+        resume(do emit(v))
+      } else {
+        do done()
+      }
+    }
+  }
+}
+
+
+def addition[R] { stream : () => R / Emit[Int] }: (R, Int) = {
```
marzipankaiser, line 135: I usually call this function either `sum` or `summing` (to distinguish it from `sum(List[Int]): Int` etc).

libraries/common/stream.effekt
```diff
@@ -0,0 +1,234 @@
...
+def gatherList[A, R] { stream: () => R / Emit[A] }: (R, List[A]) =
+  try {
+    (stream(), Nil())
+  } with Emit[A] {
+    def emit(v) = {
+      val (r, vs) = resume(());
+      (r, Cons(v, vs))
+    }
+  }
+
```
marzipankaiser, line 156: I would have expected this to do the standard "build a reversed list then reverse" trick for me, so it does not build up stack in the handler.
That is, if this is actually helpful in our current implementation.

libraries/common/stream.effekt
b-studios: Yes, please check and panic

libraries/common/stream.effekt
```diff
@@ -0,0 +1,234 @@
...
+interface Done {
+  def done(): Nothing
+}
+
+def noneOnDone[R] { program: () => R / Done }: Option[R] =
+  try {
+    Some(program())
+  } with Done {
+    def done() = None()
+  }
```
b-studios, line 74: Would be nice to see a few examples to come up with a less punny name.

libraries/common/stream.effekt
```diff
@@ -0,0 +1,234 @@
...
+
+
+record Indexed[A](index: Int, value: A)
```
b-studios, line 77: I like indexed somehow (even though it introduces boxing).

Anyway, either we use it everywhere (including list, array, etc.) or nowhere. But then indexed should be moved to `effekt.effekt`.

libraries/common/stream.effekt
```diff
@@ -0,0 +1,234 @@
...
+
+// In Effekt lower bounds are inclusive and upper bounds are exclusive
```
b-studios, line 79: Same for this, this should be in `effekt.effekt` somewhere in the toplevel documentation.

jiribenes:
It would be really nice to have a couple of examples showcasing the various features so that we can get an idea of how to use it :)

libraries/common/stream.effekt
```diff
@@ -0,0 +1,234 @@
...
+
+def range(lower: Int, upper: Int): Unit / Emit[Int] =
+  if (lower < upper) {
+    do emit(lower)
+    range(lower + 1, upper)
+  }
+
+def rangeTo(upper: Int): Unit / Emit[Int] =
+  range(0, upper)
```
b-studios, line 88: While not call this `until`? Then the usage would look like:

```
for { until(10) } { ... }
```

b-studios: hm... `rangeTo` does make sense now I think more about it.

We could even use it as

```
for { rangeTo(1); rangeTo(2); rangeTo(3) } { ... }
```
which streams

1, 1, 2, 1, 2, 3

And reads better than `until`.

libraries/common/stream.effekt
```diff
@@ -0,0 +1,234 @@
...
+    }
+  }
+}
+
+def ever[A] { action: () => A }: Nothing / Emit[A] = {
+  def go(): Nothing = {
+    do emit(action())
+    go()
+  }
+  go()
```
b-studios, line 114: Common....this is waaaay to punny. Like `ret.urn`...

jiribenes:
By the way, here's Unison's [`Stream` effect](https://share.unison-lang.org/@unison/base/code/main/latest/types/@rfi1v9429f9qluv533l2iba77aadttilrpmnhljfapfnfa6sru2nr8ibpqvib9nc4s4nb9s1as45upsfqfqe6ivqi2p82b2vd866it8)  and [`Each` effect](https://share.unison-lang.org/@unison/base/code/releases/3.13.0/latest/types/abilities/Each) (lazy nondeterminism from a stream)

```hs
Each.toList do
  use Nat <
  a = Each.range 0 5
  b = each [1, 2, 3]
  guard (a < b)
  (a, b)
```

libraries/common/stream.effekt
b-studios: Maybe `always`?

libraries/common/stream.effekt
jiribenes: I've posted the related work in Unison below :)
https://github.com/effekt-lang/effekt/pull/527#issuecomment-2253022331

libraries/common/stream.effekt
```diff
@@ -0,0 +1,234 @@
...
+      val c = i;
+      if (c < number) {
+        i = c + 1;
+        resume(do emit(v))
+      } else {
+        do done()
+      }
+    }
+  }
+}
```
b-studios, line 132: This is the same as:

```suggestion
def limit[A, R](number: Int) { stream: () => R / Emit[A] }: R / { Emit[A], Done } = {
  var i = 0;
  for { stream() } {
    if (c < number) { i = c + 1; do emit(v) }
    else { do done() }
  }
}
```

isn't it?

b-studios: Also `limit` seems like it would limit the emitted values, not their number. Why not the standard `take`?

b-studios:
Can we maybe translate some Unison examples to Effekt to compare? That would be nice. @phischu could you please look into the Unison implementation as well?

libraries/common/stream.effekt
b-studios: Also: I really, really, really, would prefer only returning a list and not a pair.

libraries/common/stream.effekt
```diff
@@ -0,0 +1,234 @@
...
+    def emit(v) = {
+      val c = i;
+      if (c < number) {
+        a.unsafeSet(i, v)
+        i = c + 1
+        resume(())
+      }
+    }
+  }
+}
```
b-studios, line 188: This is buggy and needs to be commented out to try the library at all.

The error I get is:

```
Expected Tuple2[R, Array[A]] but got Unit.
```

libraries/common/stream.effekt
```diff
@@ -0,0 +1,234 @@
...
+        return v
+      }
+    }
+  }
+  try {
+    reader()
+  } with Read[A] {
+    def read() = resume(next())
+  }
+}
```
b-studios, line 234: This also doesn't work at all and results in a lot of type errors

b-studios:
If you comment out the broken functions you can actually write an example which might look like this:

```
  val (_, l) = gatherList[Int, Unit] {
    for[Int, Unit] { range(0, 5) } { i =>
      for[Int, Unit] { range(1, 3) } { j =>
        println(i + j)
      }
    }
  }

  println(l)
  ```
  
  Please note the explicit type arguments which are necessary because I need to say WHICH stream I want to handle.

File: libraries/common/stream.effekt (hunk 1 above)
Comment (phischu) [reply libraries/common/stream.effekt:16]: If we panic, we have to call it `unsafeRepeat`. I'm doing nothing for now. I wish we had `Nat`.

File: libraries/common/stream.effekt (hunk 3 above)
Comment (phischu) [reply libraries/common/stream.effekt:156]: > That is, if this is actually helpful in our current implementation.
Is this about performance? We can benchmark later.

File: libraries/common/stream.effekt (hunk 5 above)
Comment (phischu) [reply libraries/common/stream.effekt:77]: Let's leave this open until we have more experience

**Comment by phischu** [comment]
> Please note the explicit type arguments which are necessary because I need to say WHICH stream I want to handle.

How about

```
on[Int].for { ...
```

Or we try to fix this at the language level.

**Comment by phischu** [comment]
> (lazy nondeterminism from a stream)

This is planned in Phase 3

**Comment by phischu** [comment]
> could you please look into the Unison implementation as well?

I am looking at the [`streaming`](https://hackage.haskell.org/package/streaming) library, which is great.

**Comment by b-studios** [comment]
on the language level probably means fixing type inference which is planned but will take a while

File: libraries/common/bytearray.effekt [hunk 12]
```diff
@@ -70,13 +70,13 @@ def foreachIndex(arr: ByteArray){ action: (Int, Byte) {Control} => Unit }: Unit
...
 
-extern pure def fromUTF8(str: String): ByteArray =
+extern pure def fromString(str: String): ByteArray =
   js "(new TextEncoder().encode(${str}))"
   llvm """
     ret %Pos ${str}
   """
 
-extern pure def toUTF8(arr: ByteArray): String =
+extern pure def intoString(arr: ByteArray): String =
```
b-studios, line 79: `into` sounds like the string would be mutable.

examples/benchmarks/input_output/word_count_ascii.effekt
```diff
@@ -0,0 +1,69 @@
...
+def isNewline(b: Byte) = b.toInt == 10 // \n
+
+def countWords(): Output / read[Byte] = {
+
+  var chars = 0
+  var words = 0
+  var lines = 0
+  var wasSpace = true
+
+  exhaustively[Byte] { do read() } { c =>
```
b-studios, line 25: Honestly, I would annotate `read` to disambiguate more locally
```suggestion
  exhaustively { do read[Byte]() } { c =>
```

examples/benchmarks/input_output/word_count_ascii.effekt
```diff
@@ -0,0 +1,69 @@
...
+
+def run(n: Int) = {
+  with on[IOError].panic;
+
+  val filename = "/tmp/word_count_ascii.txt"
+
+  val _ = {
+    with writeFile(filename)
+    with repeat(n)
+    for[Int] { range(32, 127) } { c =>
```
b-studios, line 52:
```suggestion
    for { range(32, 127) } { c =>
```

not necessary anymore.

examples/benchmarks/input_output/word_count_ascii.effekt
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+
+  Output(chars, words, lines)
+}
+
+def run(n: Int) = {
+  with on[IOError].panic;
+
+  val filename = "/tmp/word_count_ascii.txt"
```
b-studios, line 47: Does this work on windows machines?

examples/benchmarks/input_output/word_count_ascii.effekt
```diff
@@ -0,0 +1,69 @@
...
+
+  val output = {
+    with readFile(filename)
+    countWords()
+  }
+
+  return output.chars + output.words + output.lines
+}
+
+def main() = benchmark(5){run}
```
b-studios, line 68: Fantastic case study, thanks!

examples/benchmarks/input_output/word_count_utf8.effekt
```diff
@@ -0,0 +1,71 @@
...
+import stream
+
+record Output(chars: Int, words: Int, lines: Int)
+
+def formatWith(output: Output, filename: String): String =
+  output.lines.show ++ " " ++ output.words.show ++ " " ++ output.chars.show ++ " " ++ filename
+
+def isSpace(c: Char) =
+  c == ' ' || c.toInt == 9 || c.toInt == 10 || c.toInt == 11 || c.toInt == 12 || c.toInt == 13
+  // \n \r \t \v \f
```
b-studios, line 14: See above, we use it a second time here, maybe move?

examples/benchmarks/input_output/word_count_utf8.effekt
```diff
@@ -0,0 +1,71 @@
...
+def isNewline(c: Char) = c.toInt == 10 // \n
+
+def countWords(): Output / read[Char] = {
+
+  var chars = 0
+  var words = 0
+  var lines = 0
+  var wasSpace = true
+
+  exhaustively[Char] { do read() } { c =>
```
b-studios, line 25: Same as above.

examples/benchmarks/input_output/word_count_utf8.effekt
```diff
@@ -0,0 +1,71 @@
...
+def run(n: Int) = {
+  with on[IOError].panic;
+
+  val filename = "/tmp/word_count_utf8.txt"
+
+  val _ = {
+    with writeFile(filename)
+    with encodeUTF8
+    with repeat(n)
+    for[Int] { range(128512, 128522) } { c =>
```
b-studios, line 53:
```suggestion
    for { range(128512, 128522) } { c =>
```

examples/benchmarks/input_output/word_count_utf8.effekt
```diff
@@ -0,0 +1,71 @@
...
+  val output = {
+    with readFile(filename)
+    with decodeUTF8
+    countWords()
+  }
+
+  return output.chars + output.words + output.lines
+}
+
+def main() = benchmark(5){run}
```
b-studios, line 70: Again, great example.

libraries/common/array.effekt
```diff
@@ -126,6 +103,44 @@ def build[T](size: Int) { index: Int => T }: Array[T] = {
...
+/**
+ * Creates a copy of `arr`
+ */
+def copy[T](array: Array[T]): Array[T] =
+  array.resize(array.size)
+
+/**
+ * Copies `length`-many elements from `from` to `to`
+ * starting at `start` (in `from`) and `offset` (in `to`)
+ */
```
b-studios, line 128: Minor: we should be starting to use `///`

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+
+import array
+import bytearray
+import io/filesystem
+import io/error
+
+
+effect emit[A](value: A): Unit
+
+def for[A, R] { stream: () => R / emit[A] } { action: A => Unit }: R =
```
b-studios, line 11: Beautiful!

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+  try {
+    stream()
+  } with emit[A] { value =>
+    resume(action(value))
+  }
+
+def for[A] { stream: () => Unit / emit[A] } { action: A => Unit }: Unit =
+  for[A, Unit]{stream}{action}
+
+def continually[A] { action: () => A }: Nothing / emit[A] = {
```
b-studios, line 21: `continually` is a strange word to remember and type.

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+    }
+  }
+  go(0)
+}
+
+// not option
+// not dequeue
+// not queue
+// not result
+// not seq
```
b-studios, line 64: What does this mean? You don't want to support it, or you don't yet?

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+
+
+effect stop(): Nothing
```
b-studios, line 67: We should have some common best practice on where to define effects. In the middle of files, all at the top?

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+
+
+def boundary[R] { program: () => R / stop }: Option[R] =
+  try {
+    Some(program())
+  } with stop {
+    def stop() = None()
+  }
+
+def exhaustively[A] { program: () => A / stop } { action: A => Unit }: Unit =
```
b-studios, line 77: Again, is there a better word than `exhaustively`?

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+def rangeTo(upper: Int): Unit / emit[Int] =
+  range(0, upper)
+
+def rangeFrom(lower: Int): Unit / emit[Int] = {
+  do emit(lower)
+  rangeFrom(lower + 1)
+}
+
+
+def index[A, R] { stream: () => R / emit[A] }: R / emit[Indexed[A]] = {
```
b-studios, line 107: Maybe `indexed`?

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+    i = c + 1;
+    resume(do emit(Indexed(c, v)))
+  }
+}
+
+
+def take[A, R](number: Int) { stream: () => R / emit[A] }: R / { emit[A], stop } = {
+  var i = 0;
+  try {
+    stream()
```
b-studios, line 122: If number is 0, do we even need to run `stream`?

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+      i = c + 1;
+      resume(do emit(v))
+    } else {
+      do stop()
+    }
+  }
+}
+
+
+// If `number` is zero or negative it does nothing (like `stream::take`)
```
b-studios, line 135: Not consistent with the implementation above. Also please use `///`

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+  } with emit[Int] { v =>
+    s = s + v;
+    resume(())
+  }
+}
+
+def sum { stream : () => Unit / emit[Int] }: Int =
+  sum[Unit]{stream}.second
+
+def collectList[A, R] { stream: () => R / emit[A] }: (R, List[A]) =
```
b-studios, line 156: You know I am not a big fan of returning two things here. Same below.

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+      i = i + 1
+      resume(())
+    }
+  }
+}
+
+def collectArray[A] { stream: () => Unit / emit[A] }: Array[A] =
+  collectArray[A, Unit]{stream}.second
+
+def collectBytes[R] { stream: () => R / emit[Byte] }: (R, ByteArray) = {
```
b-studios, line 187: Should probably be called `collectByteArray`.

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+  var i = 0
+  var a = bytearray::allocate(1)
+  try {
+    (stream(), a.resize(i))
+  } with emit[Byte] {
+    def emit(v) = {
+      if (i >= a.size) { a = a.resize(2 * a.size) }
```
b-studios, line 194: Is this now the best resizing that we agreed on?

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+      resume(())
+    }
+  }
+}
+
+def collectBytes { stream: () => Unit / emit[Byte] }: ByteArray =
+  collectBytes[Unit]{stream}.second
+
+
+effect read[A](): A / stop
```
b-studios, line 206: Again, as above: do we want to sprinkle effect definitions all over files?

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+    } with emit[A] {
+      def emit(v) = {
+        next = box { () => resume(()) }
+        Some(v)
+      }
+    }
+  }
+  try {
+    reader()
+  } with read[A] { () =>
```
b-studios, line 270:
```suggestion
  } with read[A] {
```

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+  }
+}
+
+
+def writeFile[R](path: String) { stream: () => R / emit[Byte] }: R / Exception[IOError] = {
+
+  val file = open(path, WriteOnly());
+  with on[IOError].finalize { close(file) }
+
+  val chunkSize = 1048576 // 1MB
```
b-studios, line 283: It would be cool to have a separate standard library that helps with file sizes and other units, like
```
1.MB
100.ms
etc.
```

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+      push(i + r, n - r)
+    }
+  }
+
+  try {
+    val r = stream()
+    push(0, offset)
+    return r
+  } with emit[Byte] {
+    def emit(byte) = {
```
b-studios, line 299: Why do you use the long form here?

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+        do stop()
+      case n =>
+        offset = 0
+        length = n
+    }
+  }
+
+  try {
+    reader()
+  } with read[Byte] { () =>
```
b-studios, line 330:
```suggestion
  } with read[Byte] {
```

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+    def emit(byte) = {
+      if (offset >= buffer.size) { push(0, buffer.size) }
+      buffer.unsafeSet(offset, byte)
+      offset = offset + 1
+      resume(())
+    }
+  }
+}
+
+def readFile[R](path: String) { reader: () => R / read[Byte] }: R / Exception[IOError] = {
```
b-studios, line 308: That's really cool!

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+    val part3 = bitwiseAnd(do read().toInt, 63)
+    bitwiseOr(bitwiseOr(part1, part2), part3).toChar
+  } else {
+    val part1 = bitwiseShl(bitwiseAnd(b, 7), 18)
+    val part2 = bitwiseShl(bitwiseAnd(do read().toInt, 63), 12)
+    val part3 = bitwiseShl(bitwiseAnd(do read().toInt, 63), 6)
+    val part4 = bitwiseAnd(do read().toInt, 63)
+    bitwiseOr(bitwiseOr(bitwiseOr(part1, part2), part3), part4).toChar
+  }
+}
```
b-studios, line 360: Why do you mix `do read().toInt` and the decoding? Shouldn't the decoding go into strings or char or whatever?

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
...
+    do emit(bitwiseOr(224, bitwiseShr(code, 12)).toByte)
+    do emit(bitwiseOr(128, bitwiseAnd(bitwiseShr(code, 6), 63)).toByte)
+    do emit(bitwiseOr(128, bitwiseAnd(code, 63)).toByte)
+  } else {
+    do emit(bitwiseOr(240, bitwiseShr(code, 18)).toByte)
+    do emit(bitwiseOr(128, bitwiseAnd(bitwiseShr(code, 12), 63)).toByte)
+    do emit(bitwiseOr(128, bitwiseAnd(bitwiseShr(code, 6), 63)).toByte)
+    do emit(bitwiseOr(128, bitwiseAnd(code, 63)).toByte)
+  }
+}
```
b-studios, line 379: Same as above.

libraries/common/stream.effekt
omentic: I agree... if I understand the design of having an overt generic `R` and supporting streams that step to it (that then, you get streams stepping to `Unit` for-free) then I think it necessitates duplicating all these methods here that want to *do* something with the overt `R`. Since when `stream` is `() => Unit / emit[A]` (probably the most common use case?) you're always getting back this empty tuple `((), [...])` instead of just `[...]`

I don't really see the use case for having an overt generic `R` to begin with, to be honest. A Stream seems like something that should always step to the unit type. But I also haven't really worked with explicit Stream APIs all that much.

It might be worth considering having an alias `Stream` to `() => Unit` and a `Stream[R]` to `() => R`? And implementing separate functions for `Stream`, which mostly are direct wrappers around the corresponding functions for `Stream[R]`, but discard the empty tuple in these cases?

libraries/common/stream.effekt
```diff
@@ -0,0 +1,417 @@
+module stream
+
+import array
+import bytearray
+import io/filesystem
+import io/error
+
+
+effect emit[A](value: A): Unit
```
omentic, line 9: I would personally prefer this to be Uppercase for consistency with the other effects I've come across...

omentic (commented):
This is really cool! I gave some surface-level feedback. I implemented a kinda similar API recently and will go through and give deeper feedback when less busy (if this hasn't been merged by then :smile:)

libraries/common/stream.effekt
b-studios: Please explain what this function does

effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
```diff
@@ -21,13 +21,13 @@ abstract class ChezSchemeTests extends EffektTests {
...
-    // bidirectional handlers
-    examplesDir / "pos" / "maps.effekt",
-
     // bidirectional effects are not yet supported in our Chez backend
+    examplesDir / "pos" / "maps.effekt",
     examplesDir / "pos" / "bidirectional",
     examplesDir / "pos" / "object",
     examplesDir / "pos" / "type_omission_op.effekt",
+    examplesDir / "benchmarks" / "input_output" / "word_count_ascii.effekt",
+    examplesDir / "benchmarks" / "input_output" / "word_count_utf8.effekt",
```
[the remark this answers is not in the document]
phischu: Perhaps, but right now extern definitions are missing.

effekt/jvm/src/test/scala/effekt/LLVMTests.scala
```diff
@@ -84,6 +84,8 @@ class LLVMTests extends EffektTests {
     examplesDir / "pos" / "bidirectional",
     examplesDir / "pos" / "type_omission_op.effekt",
     examplesDir / "pos" / "object"/ "higher_order_compose_op.effekt",
+    examplesDir / "benchmarks" / "input_output" / "word_count_ascii.effekt",
+    examplesDir / "benchmarks" / "input_output" / "word_count_utf8.effekt",
```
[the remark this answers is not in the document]
phischu: There is a separate issue that should be resolved separately.

libraries/common/stream.effekt
b-studios: We talked about this already... it can't be do to the character wise streaming `do read()`

effekt/jvm/src/test/scala/effekt/LLVMTests.scala
```diff
@@ -29,6 +29,10 @@ class LLVMTests extends EffektTests {
     examplesDir / "llvm" / "polymorphism_map.effekt",
     examplesDir / "pos" / "type_parameters_blocks.effekt",
 
+    // Type inference during polymorphism boxing
+    examplesDir / "benchmarks" / "input_output" / "word_count_ascii.effekt",
+    examplesDir / "benchmarks" / "input_output" / "word_count_utf8.effekt",
```
b-studios, line 34: Is there an issue already? It would be horrible if this can't be run on LLVM only due to some bug.

phischu: No, but I am too lazy to minimize these. We can always fix this later.

b-studios: Here is the function type that is failing when instantiated:

```
Function(List(),List(),List(),List(Interface(stop,List())),Data(Byte,List()))
```
which is the inferred type of

```
{ (){g7113} =>
  shift(p7114) { (){k7115} =>
    def resume3708(){h7116} = resume(k7115) { h7116(g7113) }

    resume3708({ (){stop4511} =>
      val v_r_49857117: Int398 = !i3707;
      if (infixLt175(v_r_49857117, size2454(bytes3621))) {
        val v_r_49867118: Int398 = !i3707;
        let v_r_49877119 = unsafeGet2457(bytes3621, v_r_49867118)

        v_r_49877119
      } else {
        stop4511.stop()
      }
    })
  }
}
```

b-studios: This is the function which is causing the problem:

```
def feed[T](array: Array[T]) { reader: () => Unit / read[T] } = {
  var i = 0
  try {
    reader()
  } with read[T] {
    resume {
      if (i < array.size) {
        array.unsafeGet(i)
      } else {
        do stop()
      }
    }
  }
}
```

b-studios: Fixed by f30d83a. I just forgot to bind `cps`...

b-studios:
@phischu I took the liberty to add some doc comments.

b-studios:
Interesting, the tests pass on my machine but apparently they now leak memory...

b-studios:
To avoid having to rebase over and over again, I am in favor of just merging this and revising later. It is in a state that is already useful and we can continue improving it.

@phischu I'll give you the approve and if you deem it ready, just merge.

b-studios (approved):

jiribenes:
Before merging, please rename the PR (just so that the autogenerated Changelog for the next release is a little more understandable 😇)

Issue #404, opened 2024-03-04
Change semantics of `return` to be an early return


b-studios:
Right now `return` is a keyword but only used to embed expressions into statements. This is fine for a core representation like System C, but not appropriate for source.

That is, right now
```
def foo() = {
  val x = if (true) return 42 else 0;
  x + 1
}
```
currently returns `43` and not `42` (which might be confusing to people).

b-studios:
yes, main point is that it is very confusing the way it is now. For an IR `return` is fine, for the source language, the way it currently is, is not.

Currently `Control` as in the stdlib is not good enough for boundary/break, since it does not take arguments. I guess we should change this, then.

jiribenes:
Agreed, parametrising `Control` sounds good, the biggest problem I can foresee is the following:

What should be the type of `foreach`?
```scala
// currently
def foreach[A](l: List[A]) { f: A => Unit / Control } : Unit

// Option 1: we can only use `do break(())`, i.e. `Control[Unit]`
// ... but what if I _want_ to return some value
def foreach[A](l: List[A]) { f: A => Unit / Control[Unit] } : Unit

// Option 2: we align the type of `break` with the return type
// ... but what if I do _not_ want to return a value? Perhaps this needs to return `Option[R]`?
// ... also we might have errors/warnings because we're ignoring a returned value?
def foreach[A, R](l: List[A]) { f: A => Unit / Control[R] } : R
// which is strictly more general than Option 1 and reduces to it if we break with `()`
```
Are there any other possibilities / designs?

b-studios:
It needs to be `Control[Unit]` since in the case of not returning early we would otherwise need to make up an `R`.

jiribenes:
I had a quick thought: If we use `Control[R]` as `boundary/break`, what should the semantics of `continue` be in the `boundary` handler?

b-studios:
There are only two options, right? Either it aligns with break, or it restarts :)

b-studios:
I just tried adding an argument to `break`, but it is really annoying in the cases, where `Control[Unit]` and we would like to call `label.break()`; here we need to write `label.break(())`, then.

omentic:
Hi @b-studios & co - I'm coming back to Effekt after a little while and am hitting conceptual issues with `return`, now that handlers are overt functions. How might one return early out of the top-level function, instead of just the handler? Is the intended way to set overt blocks and mutate variables and break out?

This is my example code, which doesn't compile due to a type mismatch:
[...]

PR #528, opened 2024-07-26
Make control an explicit capability to overload the traversal

b-studios:


jiribenes:
If you get the time, could you also add a basic `boundary` handler for `Control` as per the discussion in https://github.com/effekt-lang/effekt/issues/404#issuecomment-2194555200?
(so that we can express an early exit from a function in examples without the need for a `return`)

jiribenes (approved):
Looks good to me, let's merge it!

PR #530, opened 2024-07-29
LLVM: Pass stack instead of stack pointer

marvinborner:
This isn't doing anything yet, most things are incomplete/wrong.

Edit:

All tests are passing now locally. We may have introduced new leaks though, so the Valgrind tests will likely fail.

### DIFF
--- a/libraries/llvm/rts.ll
+++ b/libraries/llvm/rts.ll
@@ -96,7 +96,6 @@
 ; Foreign imports
 
 declare ptr @malloc(i64)
-declare ptr @calloc(i64, i64)
 declare void @free(ptr)
 declare ptr @realloc(ptr, i64)
 declare void @memcpy(ptr, ptr, i64)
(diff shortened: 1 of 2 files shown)


libraries/llvm/rts.ll
```diff
@@ -291,7 +292,7 @@ define %StackPointer @stackDeallocate(%Stack %stack, i64 %n) {
 ; Meta-stack management
 
 define %Memory @newMemory() {
-    %stackPointer = call %StackPointer @malloc(i64 268435456)
+    %stackPointer = call %StackPointer @calloc(i64 268435456, i64 1)
```
marvinborner, line 295: We should really grow the stack dynamically, otherwise initializing with 0 will decrease performance a lot. I added it here since Valgrind warned about uninitialized memory.

phischu (approved):

PR #531, opened 2024-07-30
JS: Parenthesize arguments of raw expresions

### LINKED ISSUE #505: Overeager JS inliner forgets to parenthesize
Continuing from https://github.com/effekt-lang/effekt/pull/493, we haven't solved the general issue: the JS inliner is great at inlining, but it might inline `(s1 + s2).length` into `s1 + s2.length` without the parentheses.

Here's an example from @marvinborner from #493 which showcases similar buggy behaviour:
```scala
> "abc" == "def"
false
> ("abc" == "def").not
false
> not ("abc" == "def")
false
> val huh = "abc" == "def"
> huh.not
true
```

marzipankaiser:
Fixes #505 by parenthesizing the arguments of `RawExpr`s unless they are already atomic expressions.

Also tries to add few parentheses into the generated code.


marzipankaiser:
Future work:
- [ ] Double-check whether we need to parenthesize objects or literals in certain places.
- [ ] Delete douuble-parens in stdlib (hotfixes from earlier).

marzipankaiser:
> Future work:
> 
> * [ ]  Double-check whether we need to parenthesize objects or literals in certain places.

Update: You were right btw, that the receiver in a method call has to be parenthesized even if it is an object or number literal (but it seems to work for arrays).

PR #532, opened 2024-07-30
Do not crash the compiler in server mode

b-studios:


b-studios:
I forgot what the state of this was, but I think it would be very important to merge.

jiribenes:
I'm fine with merging this as-is, it's definitely an improvement over the current state.

PR #535, opened 2024-07-31
Cleanup for #531

marzipankaiser:
This fixes the things noted in #531 (mostly a draft PR so I don't forget):

- [x] Double-check whether we need to parenthesize objects or literals in certain places.
    - Yes, we have to:
        - [x] objects
        - [ ] int and double literals
- [x] Delete double-parens in stdlib (hotfixes from earlier).


b-studios:
Hey @marzipankaiser what do you want to do with this PR?

jiribenes:
Bump @marzipankaiser

marzipankaiser:
`RawLiteral` can stay unparenthesized, this is documented at the definition site and it's only used for strings and undefined.

Issue #422, opened 2024-03-22
Allow braces in a pattern match


jiribenes:
When an arm of an pattern match gets too long, I'd like to wrap it in _braces_:
```scala
def main(opt: Option[Int]) = {
  opt match {
    case None() => println("Nothing")
    case Some(x) => {
      println("We have a number: ${x.show}")
    }
  }
}
```

Unfortunately, the parser doesn't like that very much:
<img width="418" alt="Screenshot 2024-03-22 at 11 26 18" src="https://github.com/effekt-lang/effekt/assets/11269173/1b9c7ef3-0e4f-4ca5-b4c1-bac4e4df94eb">

(As far as I know, this is one of the few places where we only allow delimiting by `case`)

jiribenes:
You can find the relevant parts of the parser here:
https://github.com/effekt-lang/effekt/blob/68d40378fea59cb66074de3eb61df7048578a4e1/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala#L722-L724

I think this is easy enough to qualify as a good first issue.

PR #536, opened 2024-08-05
Effekt by example

dvdvgt:
This PR aims to add literate Effekt files showcasing the various interesting aspects of Effekt. The main target audience are researchers (including, most notably, compiler engineers). I mostly tried to limit explanations to purely syntactical aspects. Considering our target audience, the semantics should mostly be straightforward to understand.

## Table of contents

Here are the planned and partially already drafted topics:

- [X] Hello World (println, inspect)
- [X] Primitive Values
- [X] Strings and characters
- [X] Records
- [X] ADTs & pattern matching
- [X] Mutable variables
- [X] Loops
- [X] Functions
  - local function definitions
  - known jumps
- [X] Effects & handlers(*)
- [ ] Effect collection?
  - Non-deterministic
  - ...
- [X] Objects
- [x] Blocks & Computations(*)
- [x] Boxing(*)
- [x] Namespaces
- [ ] Modules
- [x] File! IO
  - show word count example, load and read file
- [x] Regions(*)
- [x] FFI

Topics marked with a `*` are considered difficult and yet the most crucial topics of this collection. 

## TODO

- [x] Perhaps include more citations

## Changelog

- Added `foldLeft` and `foldRight` function in the standard library for lists
- Fixed a bug causing the lexer to crash when encountering unmatched curly braces

## Request for feedback

Please do leave feedback. It is surprisingly difficult to gauge what to explain and mention and what to leave out.

While writing these examples, I may extend the standard library with new functions. It would be nice to keep an eye on those and discuss their inclusion.

**Disclaimer**: I am aware that there may be grammatically errors and typos. The formatting (e.g. limiting line width to ~120 characters) partly is subpar as well. Since much is still in flux, I have not bothered yet to go through the prose in more detail. When the rest is finalised, I will take care of these issues.

## Future Work

Upon closer inspection, you may notice that the fenced code blocks not only have the types like `effekt:repl` but also

- `effekt:hide`: This content is not visible on the website
- `effekt:clear`: The following content is _not_ compiled together with previous.
- `effekt:def`: This is the default when there is no type after three `\`` and contains definitions.

These are preprocessing directives mainly targeting the website ~and still need to be implemented~.


examples/features/records.effekt.md
```diff
@@ -0,0 +1,32 @@
...
+
+```
+val t = ("a", 42)
+```
+
+For destructing a tuple, you may use pattern matching:
+
+```effekt:repl
+t match { case (first, second) => first }
+```
```
jiribenes, line 13: Nitpicks:
- should we also show the type of a tuple? (perhaps something like "t has type `(String, Int)" as a comment?
- you can also use the _other_ deconstruction syntax:
```
val (first, second) = t
```

examples/features/records.effekt.md
```diff
@@ -0,0 +1,32 @@
...
+
+Before creating a record, you first need to declare it:
+
+```
+record Vec2d(x: Int, y: Int)
+val vec = Vec2d(1, 2)
+```
+
+A record can either again be destructed by employing pattern matching
```
jiribenes, line 22:
```suggestion
A record can either again be destructed by employing pattern matching:
```

examples/features/adts.effekt.md
```diff
@@ -0,0 +1,52 @@
...
+```
+def containsFile(fs: FileSystem, name: String): Bool =
+  fs match {
+    case File(name1, _) and name1 == name => true
+    case Directory(_, children) => 
+      children
+        .map { child => containsFile(child, name) }
+        .foldLeft(false) { (lhs, rhs) => lhs || rhs } 
+  } else { false }
+```
```
jiribenes, line 38: Nitpick: This is not a bad example per se, but I dislike the `.map( <recurse> ).foldLeft(false) { _ || _ }` as a pattern here -- the actual operation here is `.filter( <recurse> ).nonEmpty` or even just `.any {  <recurse> }` from the standard library ;)

examples/features/adts.effekt.md
```diff
@@ -0,0 +1,52 @@
...
+
+A pattern match with `match` consists of one or more clauses delimited by `case`. There is an exhaustivity check such 
+that you need to handle all cases of the scrutinee. Pattern guards, starting with `and`, can be used to refine matches
+with further conditions that evaluate to a `Bool`. Optionally, like with `if` expressions, a `match` expression can be 
+followed by an `else` branch.
```
jiribenes, line 43: Question: Is it worth it to also showcase the other way of a default match (`case _ => ...`) here, or is it truly obvious that it exists?

examples/features/adts.effekt.md
```diff
@@ -0,0 +1,52 @@
...
+
+```effekt:repl
+containsFile(Directory("/", file), "README.md")
+```
+Furthermore, you can use partial matches in positions where a `Bool` is expected, like here:
+
+```effekt:repl
+if (file is File(_, _)) println("it's a file!")
+```
```
jiribenes, line 52: Here I'd really like to see an example that actually _uses_ the newly bound value somehow -- perhaps by printing its name or the content length or something?

examples/features/loops.effekt.md
```diff
@@ -0,0 +1,46 @@
...
+}
+record Job()
+```
+
+```
+def worker() = loop {
+  val packet = do receive[Job]()
+  packet match {
+    case Some(p) => println("processing request")
+    case None => do break()
```
jiribenes, line 41: Note that from #528 onward, you'll need to bind the label:
```scala
each(0, 100) { (i) {label} =>
     total = total + i;
     if (total > 15) label.break()
     println(i)
}
```

examples/features/loops.effekt.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+```repl
+fib(10)
+```
+
+A `while` loop consists a condition on which to continue the loop and an optional `else` branch that is run when the
+`while` loop ended.
+
+Furthermore, there is a `loop` function defined in the standard library, you can use for infinite, breakable loops.
```
jiribenes, line 27: I think it's also worth it to show `each` / `repeat` here :)
I was also thinking about `.foreach` on lists, as that's a looping operation I use _all the time_, but I'm not sure it fits into this chapter.

examples/features/loops.effekt.md
```diff
@@ -0,0 +1,46 @@
+# Loops
+
+In Effekt, there is only a `while` loop expression built-in (that is completely standard).
```
jiribenes, line 3: Nitpick: the formulation of this sentence is pretty weird and it took me a while to understand it.
```suggestion
Effekt includes a (completely standard) built-in `while` loop expression as its sole native looping construct.
```

examples/features/mutable-variables.effekt.md
```diff
@@ -0,0 +1,19 @@
...
+  }
+  sum
+}
+```
+
+It is not permitted to have unintialized variables, that is, there are no mere declarations of variables. Therefore, the following is not allowed:
+
+```effekt:repl
+var x: Int
+```
```
jiribenes, line 19: Here it might be nice to have a forward reference to regions in the future

jiribenes (commented):
This looks like a great start! I read through some of these and wrote down a few notes in the review (whilst trying to keep your disclaimer in mind :))

libraries/common/list.effekt
```diff
@@ -207,6 +207,24 @@ def any[A](list: List[A]) { predicate: A => Bool }: Bool = {
...
 }
 
+// Fold a list using `f`, starting from the left given a starting value.
+//
+// O(N)
+def foldLeft[A, B](l: List[A], init: B) { f: (B, A) => B }: B = {
+  var acc = init;
+  l.foreach { x => acc = f(acc, x) };
+  acc
+}
```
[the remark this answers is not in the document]
dvdvgt: This how other functions like `sum` are also defined, so I just assumed this works

dvdvgt:
I just noticed that all tests in `examples/stdlib` either have a module path that is not consistent with the folder structure or none at all. Does this matter?

examples/features/functions.effekt.md
```diff
@@ -0,0 +1,51 @@
+# Functions
+
+For defining a function, you may use similar syntax for know from other languages such as Scala:
```
omentic, line 3: Nitpick: "...similar syntax for know from other languages..." doesn't make sense. Maybe remove "for know"?

omentic: (also thanks for this, currently learning some Effekt and this is useful)

examples/features/effect-handlers.effekt.md
```diff
@@ -0,0 +1,87 @@
+# Effects and handlers
+
+For getting started with effects and handlers, we define the popular exception effect, consisting just of one effect operation `throw`:
+
+```
+interface Exception {
+  def throw(msg: String): Nothing
```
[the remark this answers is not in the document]
dvdvgt: We have a very flat subtyping lattice in Effekt. `Nothing` is bottom in this lattice and thus is a subtype of everything. However, there is no inhabitant of this type. Yet, it is still useful. It is commonly used as the return type for functions that do not return. This is exactly what happens in other languages with exceptions. `throw` will never return anything.

examples/features/adts.effekt.md
```diff
@@ -0,0 +1,52 @@
...
+file.name // fails
+```
+
+Instead, pattern matching has to be used.
+
+```
+def containsFile(fs: FileSystem, name: String): Bool =
+  fs match {
+    case File(name1, _) and name1 == name => true
+    case Directory(_, children) =>
```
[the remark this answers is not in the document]
dvdvgt: I am not exactly sure what you mean. The body of each arm may be a single expression or a statement delimited by curly braces. So what you are describing should work unless I misunderstood your remark. Feel free to elaborate in this case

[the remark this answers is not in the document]
dvdvgt: That is, you also could have written `{ true }` as the first match arm's body.

examples/features/effect-handlers.effekt.md
```diff
@@ -0,0 +1,87 @@
...
+The return type of non-recursive functions can be left out and is inferred. In this it is `Double / { Exception }`, that is, the return type `Double` followed by a slash and the effects that need to be handled by the calling context.
+
+Whereas operations introduce effects, handlers offer a way of discharging them:
+
+```
+def unsafeDiv(a: Double, b: Double): Double / {} =
+  try {
+    div(a, b)
+  } with Exception {
+    def throw(msg) = {
```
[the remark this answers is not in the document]
dvdvgt: Good question!

If we had defined the `Exception` effect as follows

```
effect Exception(msg: String): Nothing
```

we could have used the syntax you described. However, this is IMO in a bit of a weird place right now because of the different  keyword (`effect` vs `interface`). I am therefore reluctant to mention it. I'll give it some thought.

examples/features/loops.effekt.md
```diff
@@ -0,0 +1,46 @@
...
+def worker() = loop {
+  val packet = do receive[Job]()
+  packet match {
+    case Some(p) => println("processing request")
+    case None => do break()
+  }
+}
+```
+
+The operations `break` and `continue` have their usual semantics, except that they are handled by `loop` as operations of the algebraic effect `Control`.
```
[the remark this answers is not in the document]
dvdvgt: Thanks for taking your time to ask these insightful questions! It is very valuable for us to get feedback for newcomers to algebraic effects.

- `while` is a natively built-in constructor and does not use the `Control` effect by default. However, no one is stopping you from wrapping a while loop using this effect in a handler and defining your own behaviour :)
- See previous answer. Interestingly, `loop` is a handler. I invite you to look at the (relatively concise) implementation to perhaps reconcile it with your mental model:
https://github.com/effekt-lang/effekt/blob/master/libraries%2Fcommon%2Feffekt.effekt#L651-L677
- Yes, indeed. You can do that. Note, however, that the semantics of `Control` depend on the lexically nearest handler. If there is one "deeper nested" than loop, that one will handle `Control` instead. This is true for all effects, of course.

examples/features/adts.effekt.md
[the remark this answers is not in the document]
jiribenes: see #422, I don't think the new parser resolved this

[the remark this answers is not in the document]
dvdvgt: Ha, thanks. I would have expected this to work. This should be easy to support if we want to.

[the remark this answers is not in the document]
dvdvgt: Ha, thanks. I would have expected this to work. This should be easy to support if we want to.

examples/features/loops.effekt.md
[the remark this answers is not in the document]
omentic: Cool! Is there any reason you all decided to separate out `while` into a built-in primitive in the first place?

examples/features/effect-handlers.effekt.md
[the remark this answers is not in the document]
omentic: Oh interesting, okay! That clears up a lot of confusion I had. I thought `effect Exception(msg: String): Nothing` was a direct alias to `interface Exception { def Exception(msg: String): Nothing }` and couldn't figure out why it wasn't working...

examples/features/effect-handlers.effekt.md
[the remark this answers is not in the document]
jiribenes: Thanks for the question! We could also explain that somewhere ("it's the type for unreachable things") -- it might even be worthy to say that it's Rust's & TypeScript's `never`, C++'s `noreturn`.

examples/features/functions.effekt.md
```diff
@@ -0,0 +1,51 @@
...
+```effekt:repl
+fib(5)
+```
+
+Perhaps unusual, you can also call `fib` using Effekt's implementation of [Uniform Function Call Syntax
+](https://en.wikipedia.org/wiki/Uniform_Function_Call_Syntax#cite_note-4):
+
+```effekt:repl
+5.fib
+```
```
jiribenes, line 25: I feel like this is subtly combining two different things:
1. UFCS
2. the ability to omit parens for functions without parameters 

Could we split it into two? Show `5.fib()` first and then show the parenthesis-omitting part some time later?

examples/features/blocks.effekt.md
```diff
@@ -0,0 +1,28 @@
+# Blocks
+
+```effekt:hidden
+interface Yield[A] {
+  def yield(x: A): Unit
+}
+```
+
+As we have established in a previous section, functions in Effekt are second-class. While you cannot pass them as first-class values to other functions as argument, you can pass them as second-class blocks.
```
omentic, line 9: I might suggest moving this whole section into the `functions.effekt.md` section... it's not entirely clear at the moment how UFCS and blocks interact (i.e. I was surprised that you can't do `{ 5.div(0) }.default(0)`)

examples/features/adts.effekt.md
dvdvgt: Yes, good idea! I didn't even know the fields are bound here :)

jiribenes: It's very much like Rust's `while let` and `if let` :)

examples/features/loops.effekt.md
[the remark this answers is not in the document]
dvdvgt: I am not aware of any particular reason. I would guess that `while` loops map nicely to some of the backends and can leverage the respective backend's optimisations.

examples/features/blocks.effekt.md
dvdvgt: Good point, I will add some explanation for blocks and UFCS

dvdvgt:
Thanks to @jiribenes for workshopping the captures and regions sections with me!

examples/features/adts.effekt.md
```diff
@@ -0,0 +1,51 @@
...
+type FileSystem {
+  File(name: String, content: String)
+  Directory(name: String, children: List[FileSystem])
+}
+```
+
+Each definition starts with the `type` keyword and is followed by the name of the datatype. Enclosed by curly braces
+and separated by newlines, variants consisting of a name and optional fields can be defined.
+
+Creating instances of this datatype functions just as you might expect:
```
b-studios, line 15:
```suggestion
A datatype definition starts with the `type` keyword and is followed by the name of the datatype. 
Variants are enclosed by curly braces and separated by newlines (or semicolons) and consist of a name and a (non-optional) list of fields.

Constructing values of this datatype works as you might expect:
```

examples/features/adts.effekt.md
```diff
@@ -0,0 +1,51 @@
...
+    case Directory(_, children) => 
+      children.any { c => containsFile(c, name) }
+  } else { false }
+```
+
+A pattern match with `match` consists of one or more clauses delimited by `case`. There is an exhaustivity check such 
+that you need to handle all cases of the scrutinee. Pattern guards, starting with `and`, can be used to refine matches
+with further conditions that evaluate to a `Bool`. Optionally, like with `if` expressions, a `match` expression can be 
+followed by an `else` branch. Instead of an `else` branch in a `match` expression, you may also use a wildcard pattern
+(`case _ => ...`).
```
b-studios, line 42:
```suggestion
A pattern match (initiated with the keyword `match`) consists of one or more clauses delimited by `case`. 
There is an exhaustivity check helping you to handle all cases of the scrutinee. 
Pattern guards, starting with `and`, can be used to refine matches with further conditions that evaluate to a `Bool`. 
Optionally, like with `if` expressions, a `match` expression can be followed by an `else` branch to provide a default case. 
Instead of an `else` branch in a `match` expression, you may also use a wildcard pattern
(`case _ => ...`).
```

examples/features/blocks.effekt.md
```diff
@@ -0,0 +1,34 @@
+# Blocks
+
+```effekt:hidden
+interface Yield[A] {
+  def yield(x: A): Unit
+}
+```
+
```
b-studios, line 8:
```suggestion
## Values vs. Computation
Following Paul Levy's [Call-By-Push-Value](https://www.cs.bham.ac.uk/~pbl/papers/thesisqmwphd.pdf) Effekt distinguishes between **values** (such as `42`, `true`, or instances of datatypes) and **computation**.

Examples of "computation" in Effekt are:
- blocks (this is what we call functions),
- instances of interfaces (objects), and
- regions

Functions (and all other computation) are _second-class_ in Effekt. To make this difference explicit, we pass values in parenthesis (e.g. `f(42)`) and computation in braces (e.g. `f { x => println(x) }`).
```

examples/features/blocks.effekt.md
```diff
@@ -0,0 +1,34 @@
...
+Functions are second-class in Effekt. While you cannot pass them as first-class values to other functions as argument, you can pass them as second-class blocks.
```
b-studios, line 9:
```suggestion
```

b-studios (commented):
Reviewed adts and blocks:

examples/features/ffi.effekt.md
```diff
@@ -0,0 +1,119 @@
...
+
+Sometimes, you want to include a whole JavaScript files containing your favourite JavaScript library
+so that you can write bindings for it later.
+You can do that in Effekt with the following code:
+
+```effekt
+extern include js "./mycoollibrary.js"
+```
+
+> NOTE: Do not run this code on the website ^
```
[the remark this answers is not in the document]
jiribenes: Yeah, that was an internal note for myself. (Like all of the `> NOTE`s; this was meant to be as a rough draft)

examples/features/captures.effekt.md
```diff
@@ -0,0 +1,135 @@
...
+  - on term level unbox, box
+  - `at` at type level
+  - capability, resource, region, lifetime, coeffect
+
+## Captures
+
+Functions are second-class in Effekt, but for a good reason. Functions closing over capabilities must not be returned, otherwise it can not be guaranteed that these capabilities are still in scope.
+Thus, we explicitly have to keep track of these captures and ensure they are in scope upon envoking.
+
+Consider the following example where `divide` is being passed a explicit capability:
```
b-studios, line 63:
```suggestion
Functions (and other computation like objects) are second-class in Effekt, but for a good reason. Functions _capturing_ (that is "closing over") capabilities must not be returned, otherwise it can not be guaranteed that these capabilities are still in scope.
Thus, we explicitly have to keep track of these captured capabilities and ensure they are in scope, when calling the function.

Consider the following example where `divide` is being passed as an explicit capability:
```

examples/features/ffi.effekt.md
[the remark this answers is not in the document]
jiribenes: To explain, I thought there might be some "show this Effekt code but don't actually run it" mode I could apply to the code block, so the note was there to remind me to ask :)

examples/features/captures.effekt.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+```effekt:repl
+try { 
+  val res = divide(10, 0) {exc}
+  println(res)
+} with exc: Exception {  
+  def throw(msg) = println("ERROR: " ++ show(msg))
+}
+```
```
b-studios, line 78: Here the effect handler for `Exception` introduces a capability `exc` that we can use in the scope delimited by `try { ... }`.

examples/features/captures.effekt.md
```diff
@@ -0,0 +1,135 @@
...
+} with exc: Exception { 
+  def throw(msg) = panic("ohno")
+}
+```
+
+By directly returning `exc` from the `try` expression, we break the lexical effect handling reasoning since `exc` is only defined within the `try` body.
+For the same reason, returning a closure is not an option either.
+
+For preventing capabilities escaping their scope, Effekt uses captures. In case of the previous example, the return type of `example` is `Exception at {exc}`. This is called
+a boxed computation. A boxed computation may only be used after unboxing it, which is only permitted when all its captures are in scope.
```
b-studios, line 94:
```suggestion
To prevent capabilities escaping their scope, Effekt tracks their capture. 
In case of the previous example, the return type of `example` would be `Exception at {exc}`. 
That is, it expresses that the capability of type `Exception` can only be used when `exc` is in scope. 
The Effekt type checker prevents this example from type checking since `exc` is not even a valid name outside of its binding effect handler.
```

jiribenes:
As a quick comment, do we explain the following syntax somewhere? (sorry for the subpar example)
```
val numbersMaybe = [Some(100), Some(42), Some(0), None(), Some(42), Some(1), None()]
numbersMaybe.map {
  case Some(42) => None()
  case Some(x) => Some(x * 2)
  case None() => Some(42)
}
```
Does it fit into some of the existing pages or would that be a new page?
I don't necessarily want to increase the scope, I just haven't seen this covered.

examples/features/captures.effekt.md
```diff
@@ -0,0 +1,135 @@
...
+- or being passed as arguments where first-class values are expected.
+
+```effekt:repl
+parallel(
+  box { () => println("Hello, ") },
+  box { () => println("world!") }
+)
+```
+
+Here both arguments have the capture set `{io}` while the expected capture set is expected to be empty `{}`. Hence, this call does not type-check.
```
b-studios, line 127:
```suggestion
## Boxing
Only values (which are "first-class") can be returned from computation, be passed as arguments to a datatype constructor, or stored in a mutable reference.

Does this mean that computation (which is "second-class") like functions, capabilities, or objects cannot be returned or stored?

As illustrated above, allowing this without any restriction would lead to unsafety and potential runtime errors. 
Effekt thus takes the following approach: 

> Computation can be packaged up into inert values by **boxing**.
> Computation can be extracted (and reactivated) from boxed values by **unboxing**.

In our example, this means that we can turn the capability `exc` into a value by using the term-level syntax `box c`.

```
def example() = try {
  val cap = box exc;
  println("done")
} with exc: Exception { 
  def throw(msg) = panic("ohno")
}
```

Hovering over cap shows the inferred type is `Exception at {exc}`. While `Exception` is an interface and lives in the universe of computation types, `Exception at {exc}` is a "boxed" type and lives in the universe of value types. 
To prevent capabilities escaping their scope, Effekt tracks their capture in the type of boxed computations. 
In case of our example, the type of `cap` expresses that the capability of type `Exception` can only be used when `exc` is in scope.

TODO continue...

While this may seem like a restriction, this also has upsides. For example, using boxing, you can have first-class functions. You can even ensure that these first-class functions
only use a certain set of effects.

Consider a simple scheduler that runs two functions in parallel:

```
def parallelWrong { f: () => Unit } { g: () => Unit }: Unit = <>
```

Since `parallelWrong` expects two arbitrary blocks as arguments, these blocks may capture arbitrary capabilities.
Running both in parallel can have non-deterministic side-effects or introduce data races.
Thus, we need to enforce that both `f` and `g` are pure, that is, do not capture any capabilities.

```
def parallel (f: () => Unit at {}, g: () => Unit at {}): Unit = <>
```

Converting second-class computations to first-class values can be done either explicitly by `box`ing them or implicitly when
- they occur in return position,
- on the right-hand side of assignments,
- or being passed as arguments where first-class values are expected.

```effekt:repl
parallel(
  box { () => println("Hello, ") },
  box { () => println("world!") }
)
```

Here both arguments have the capture set `{io}` while the expected capture set is expected to be empty `{}`. Hence, this call does not type-check.
```

**Review by b-studios — COMMENTED** [review]
Revise boxes a bit, but still some work to do

File: examples/features/captures.effekt.md (hunk 26 above)
Comment (jiribenes) [reply examples/features/captures.effekt.md:127]: I've fixed the formatting of the suggestion, it should be something like this if you want nested code blocks:
`````
````suggestion
words here
```effekt
code here
```
more words here
````
`````

File: examples/features/regions.effekt.md [hunk 27]
```diff
@@ -0,0 +1,141 @@
...
+---
+
+# Regions
+
+## State and continuations
+
+Consider the following handle for non-determinism:
+
+```
+interface Amb { def flip(): Boolean }
```
jiribenes, line 14:
```suggestion
interface Amb { def flip(): Bool }
```
:)

examples/features/regions.effekt.md
```diff
@@ -0,0 +1,141 @@
...
+layout: docs
+title: Regions
+permalink: docs/tutorials/regions
+---
+
+# Regions
+
+## State and continuations
+
+Consider the following handle for non-determinism:
```
marzipankaiser, line 11:
```suggestion
Consider the following handler for non-determinism:
```
?

examples/features/regions.effekt.md
```diff
@@ -0,0 +1,141 @@
...
+  var x in r = 1
+  if (do flip()) { x = 2 }
+  else { () }
+  println(x)
+}
+
+def example1Region() =
+  region r {
+    allChoices {
+      exampleProgram {r}
```
marzipankaiser, line 106: I can't find a definition of `exampleProgram` in this file - did you test this? (maybe I just didn't see it)

examples/features/regions.effekt.md
```diff
@@ -0,0 +1,141 @@
...
+  }
+```
+
+By doing so, the type of `closure` becomes `() => Int at {r}` and thereby signifying that `closure` closes over something allocated in the region `r`.
+`closure` may only ever be unboxed where region `r` is still in scope. Otherwise, accessing `closure` outside of `r` would exceed the lifetime given to `x` by allocating it in `r`.
+
+Additionally, regions are second-class like blocks and objects and thus can be passed to functions. We use this for rewriting the opening example from the previous section:
+
+```
+def prog {r: Region} = {
```
marzipankaiser, line 96: Or should this be
```suggestion
def exampleProgram {r: Region} = {
```

dvdvgt:
I noticed that I failed to mention the `with handler` syntax.

dvdvgt:
The `control` built-in should also be mentioned.

examples/features/regions.effekt.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+Multiple resumptions and mutable state need to be handled with care such that the state is appropiately restored upon each resumption.
+
+```
+def example1() = {
+  var x = 1
+  allChoices {
+    if (do flip()) { x = 2 }
+    else { () }
```
jiribenes, line 33:
```suggestion
```
no `else` should be needed here (similarly for the other examples) :)

examples/features/effect-handlers.effekt.md
```diff
@@ -93,3 +97,65 @@ def genFibs(limit: Int): List[Int] / {} = {
...
+  }
+```
+
+```effekt:repl
+tellTime {
+  val start = do tell()
+  val end = do tell()
+  println("took " ++ show(end - start))
+}
+```
```
jiribenes, line 130: Nitpick: this example feels very artificial, we're not even measuring anything here.
We have `genFibs` in the same file, so we can use it here, no?
``````suggestion
```effekt:repl
tellTime {
  val start = do tell()
  val fibs = genFibs(8)
  val end = do tell()
  println("took " ++ show(end - start))
}
```
``````

File: examples/features/effect-handlers.effekt.md [hunk 33]
```diff
@@ -93,3 +97,65 @@ def genFibs(limit: Int): List[Int] / {} = {
...
+
+The handler for this interface uses a slightly different more concise syntax like it is used in the literature:
+
+```
+def tellTime[A] { prog: () => A / tell }: A =
+  try { prog() }
+  with tell { () =>
+    resume(time())
+  }
+```
```
jiribenes, line 122: This is slightly confusing as the `time` is invisible. Could we rename the extern def to something more primitive-looking? Something like
``````suggestion
```effekt:hide
extern def primitiveTimeNow(): Int =
  js "Date.now()"
```

```
effect tell(): Int
```

The handler for this interface uses a slightly different more concise syntax like it is used in the literature:

```
def tellTime[A] { prog: () => A / tell }: A =
  try { prog() }
  with tell { () =>
    resume(primitiveTimeNow())
  }
```
``````
(The name could probably still be improved...)

Also, this nitpick could be invalidated by us using the `time()` somewhere else before...

File: examples/features/effect-handlers.effekt.md [hunk 34]
```diff
@@ -93,3 +97,65 @@ def genFibs(limit: Int): List[Int] / {} = {
...
+```effekt:repl
+tellTime {
+  val start = do tell()
+  val end = do tell()
+  println("took " ++ show(end - start))
+}
+```
+
+## `with` handler
+
```
jiribenes, line 133: Missing text?
``````suggestion
## `with` handler

If you just want to use one handler for the rest of the given scope, and/or you want to reduce nesting, use `with` syntax where:
```effekt:sketch
with limited[Int](1);
...
```
is the same as
```effekt:sketch
limited[Int](1) {
  ...
}
```
``````
(The text could probably be better, this was the first quick thing I wrote down).

Comment (jiribenes) [reply examples/features/effect-handlers.effekt.md:133]: Maybe we could even reuse the `tellTime` example above and just have the exact same code, but now with `with`:
```
with tellTime;
val start = do tell()
val end = do tell()
println("took " ++ show(end - start))
```

That would probably make it easier and we wouldn't need to introduce a new -- albeit nice! -- handler.

**Review by jiribenes — COMMENTED** [review]
A few nitpicks :)

File: examples/features/effect-handlers.effekt.md (hunk 33 above)
Comment (jiribenes) [reply examples/features/effect-handlers.effekt.md:122]: As a different nitpick: it could be interesting to show why the `() =>` is needed (let the effect operation take an argument), but then that destroys this example :'(

File: examples/features/objects.effekt.md [hunk 35]
```diff
@@ -6,48 +6,80 @@ permalink: docs/tutorial/objects
...
-interface Exception {
-  def throw(msg: String): Nothing
+interface Counter {
+  def increment(): Unit
+  def get(): Int
 }
 ```
 
-besides implementing a handler for it, there is also another way of defining the semantics of an effect:
+Interfaces are [_computation types_](./computation.md) and we can write a program that expects a counter to work on:
```
jiribenes, line 18: Will such a link work on the website? 🤔

jiribenes: In the parser case study, we link to the lexer case study like this:
```
# Parsing
In this case study, we show how to implement a parser, using the lexer from the
[Lexer case study](lexer).
```

So here, it should probably be:
```suggestion
Interfaces are [_computation types_](computation) and we can write a program that expects a counter to work on:
```
?

jiribenes: Ditto for the links in captures 👀

dvdvgt:
Todo: rename `feature` folder to `tour` before merge

dvdvgt:
I locally deployed the website with compiler built from the latest master commit. All seems to be well and all examples included here are runnable.

Only the IO section needs to be revised due to #553.

b-studios (approved):
Great work!

PR #538, opened 2024-08-09
Move to a weekly release scheme

### LINKED ISSUE #533: Move to a rolling release scheme
### Motivation

We are way too slow to release new versions, but at the same time it's unnecessarily difficult to install Effekt for students and other users. Let's resolve that by moving to a weekly ("nightly") release.

### Details

On GitHub CI, set up a cronjob for every Monday at 3am to make a new release.

Version format: `M.YY.WW`, so for example `0.24.10` for the 10th week of 2024 where the major version is `0`.
A related commonly used format is called [Calendar Versioning](https://calver.org).

#### Possible problems

What if we want  to release a new (sub)version quickly?

Regarding prior art, Minecraft uses `YYwWWN` (where `N` is some unique identifier) for their weekly snapshots, so for example `24w10a`.
We could similarly do `0.24.10.0` or `0.24.10.a` or something like that. :)

jiribenes:
Resolves #533

**UNTESTED, DO NOT MERGE**

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,13 +1,37 @@
-# This is a copy-and-past version of ci.yml but additionally creating a release
+# This is a copy-and-paste version of ci.yml but additionally creating a release
 name: Release Artifacts
 
 on:
+  schedule:
+    - cron: '0 3 * * 1'  # Every Monday at 03:00 UTC
   push:
     tags:
       - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+
 jobs:
+  calculate-version:
+    name: Calculate Version
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.set-version.outputs.VERSION }}
+    steps:
+
+      - name: Calculate new version
+        id: set-version
+        run: |
+          if [[ "${{ github.ref }}" == refs/tags/* ]]; then
+            echo "VERSION=${GITHUB_REF#refs/tags/v}" >> $GITHUB_OUTPUT
+          else
+            MAJOR=0
+            YEAR=$(date +%y)
+            WEEK=$(date +%W)
+            WEEK_PADDED=$(printf "%02d" $WEEK)
+            echo "VERSION=${MAJOR}.${YEAR}.${WEEK_PADDED}" >> $GITHUB_OUTPUT
+          fi
+
   build-jar:
     name: Build Effekt compiler and run tests
+    needs: calculate-version
     runs-on: ubuntu-latest
     steps:
 
@@ -39,6 +63,10 @@ jobs:
     - name: Install Chez Scheme
       run: sudo apt-get install chezscheme
 
+    - name: Update version in build.sbt
+      run: |
+        sed -i 's/lazy val effektVersion = ".*"/lazy val effektVersion = "${{ needs.calculate-version.outputs.version }}"/' build.sbt
+
     - name: Run tests and assemble jar file
       run: sbt clean deploy
 
@@ -61,7 +89,7 @@ jobs:
   release:
     name: Create Release
     runs-on: ubuntu-latest
-    needs: [build-jar]
+    needs: [calculate-version, build-jar]
     steps:
       - name: Checkout code
         uses: actions/checkout@master
@@ -84,11 +112,11 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          release_name: Prerelease ${{ github.ref }}
-          tag_name: ${{ github.ref }}
-          body: Automatic release for ${{ github.ref }}
+          release_name: Release ${{ needs.calculate-version.outputs.version }}
+          tag_name: v${{ needs.calculate-version.outputs.version }}
+          body: Release ${{ needs.calculate-version.outputs.version }}
           draft: false
-          prerelease: true
+          prerelease: false
 
       - name: Upload jar file
         id: upload_jar
@@ -109,5 +137,5 @@ jobs:
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
           asset_path: ./distribution/effekt.tgz
-          asset_name: effekt.tgz
+          asset_name: effekt-${{ needs.calculate-version.outputs.version }}.tgz
           asset_content_type: application/gzip


.github/workflows/deploy.yml
```diff
@@ -1,13 +1,37 @@
...
+
+      - name: Calculate new version
+        id: set-version
+        run: |
+          if [[ "${{ github.ref }}" == refs/tags/* ]]; then
+            echo "VERSION=${GITHUB_REF#refs/tags/v}" >> $GITHUB_OUTPUT
+          else
+            MAJOR=0
+            YEAR=$(date +%y)
+            WEEK=$(date +%W)
```
jiribenes, line 27: How do we want to determine the week number?
These are the possible options:
```
  %U   week number of year, with Sunday as first day of week (00..53)
  %V   ISO week number, with Monday as first day of week (01..53)
  %W   week number of year, with Monday as first day of week (00..53)
```

.github/workflows/deploy.yml
```diff
@@ -1,13 +1,37 @@
...
+            WEEK_PADDED=$(printf "%02d" $WEEK)
+            echo "VERSION=${MAJOR}.${YEAR}.${WEEK_PADDED}" >> $GITHUB_OUTPUT
```
jiribenes, line 29: Do we always want to release as `0.YY.WW.0` so that we can make subsequent releases?

.github/workflows/deploy.yml
```diff
@@ -84,11 +112,11 @@ jobs:
...
         with:
-          release_name: Prerelease ${{ github.ref }}
-          tag_name: ${{ github.ref }}
-          body: Automatic release for ${{ github.ref }}
+          release_name: Release ${{ needs.calculate-version.outputs.version }}
+          tag_name: v${{ needs.calculate-version.outputs.version }}
+          body: Release ${{ needs.calculate-version.outputs.version }}
           draft: false
-          prerelease: true
+          prerelease: false
```
jiribenes, line 119: Do we want to have automatic releases as "true" releases instead of pre-releases?

b-studios:
I don't understand the issue with the tag: isn't it possible to automatically create and push a tag, which then triggers the release process?

jiribenes:
> I don't understand the issue with the tag:

It might be nicer to decompose it into two operations in order to share the actual "make a release" CI code.

> isn't it possible to automatically create and push a tag, which then triggers the release process?

I'm not sure that actions made by Github Actions can _automatically_ trigger other actions, I'd need to verify that.

jiribenes:
https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs
Here are more details on actions triggering other actions automatically:
> When you use the repository's [default] `GITHUB_TOKEN` to perform tasks, events triggered by the `GITHUB_TOKEN` will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's `GITHUB_TOKEN`, a new workflow will not run even when the repository contains a workflow configured to run when push events occur.

There are a few ways to go around this: we could for example add a personal token to be used in CI instead.

b-studios:
Since this auto-releases to npm (which is great!), can we also change the README to tell people to install effekt via the npm registry?

jiribenes:
> Since this auto-releases to npm (which is great!), can we also change the README to tell people to install effekt via the npm registry?

It's severely untested now (the CI never ran!), but if everything goes well, then we can change the README in a few weeks (once we're confident in the release process). The installer in the VSCode plugin https://github.com/effekt-lang/effekt-vscode/pull/20 is also relying on `npm` as it's much easier to get the list of versions, check if a version is up to date, etc.

By the way, we'll need to add the NPM secrets to the repository.

.github/workflows/autorelease.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: |
+          sed -i 's/lazy val effektVersion = ".*"/lazy val effektVersion = "${{ steps.set-version.outputs.VERSION }}"/' build.sbt
+
+      - name: Commit and push changes
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add build.sbt
+          git commit -m "Update version to ${{ steps.set-version.outputs.VERSION }}"
+          git push
```
jiribenes, line 46: Is there anything else we need to do to change the version?
I think we also might want to call `sbt updateVersions` and then commit the `package.json` and `pom.xml` files?
https://github.com/effekt-lang/effekt/blob/8905a16cfa48ba9631d0bdcde9dcf92315cf7bfc/build.sbt#L159-L162

I mean, technically, we probably don't: the `sbt clean deploy` in `deploy.yml` will take care of it, but it's perhaps better to keep the version numbers on `master` up-to-date.

.github/workflows/deploy.yml
```diff
@@ -111,3 +115,29 @@ jobs:
...
+
+      - name: Publish to NPM as effekt-lang
+        run: npm publish effekt.tgz --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Publish to NPM as @effekt-lang/effekt
+        run: npm publish effekt.tgz --access public --scope @effekt-lang
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
jiribenes, line 143: I haven't tested this; I never even published an NPM package myself, so this is just what I read in the docs.
I'm quite unsure whether this will work, but I'm fine with tweaking this later. :)

If someone with access is bored, could you try it with `0.2.2`: it's released, but _not_ on `npm` yet, which makes it a perfect candidate.

Also, we need to add the NPM token for `effekt-lang` and `@effekt-lang/effekt` to the repository secrets.

jiribenes:
I'm testing this CI in https://github.com/jiribenes/effekt-ci-test (except for the NPM part for now).

jiribenes:
✨ _what fun_ ✨ 
the CI managed to stumble into the "text file busy" problem _twice_ in a single run!
https://github.com/effekt-lang/effekt/actions/runs/10634552922/job/29482682022

b-studios:
A controversial question: does the CI that deploys also need to run the tests again?

jiribenes:
> does the CI that deploys also need to run (all) the tests again?

Au contraire, I think it's the only CI that actually really \*does\* have to pass all the tests.

I understand what you mean, _technically_ not as long as we maintain the invariant that everything on the main branch went through a successful PR. But we have been known to merge PRs without CI finishing or commit directly, both of which resulted in failing test on the main branch.

I'd keep it so that we watch for flaky tests and similar "oops, failing `master`" escapades, especially since it runs at 3AM and we don't really care how long it takes as long as the new version is ready when we wake up :)

jiribenes:
Two problems remain:
1. I haven't tested the `npm` deployment
2. `sbt test` sometimes fails... what do we do then? by that time, we've already created a git tag with a new version, we just never release it

I think the correct way to proceed is:
1. Just merge it, if there's an error during `npm` deployment, we deploy by hand and then fix the CI.
2. Either disable the tests and agree not to push to `master`, or maybe doing the tests in CI that bumps a version? But what do we do then if it fails? (then we don't have a release at all)

bin/effekt.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-java -jar "$(npm root -g)/effekt/bin/effekt" $@
+java -jar "$(npm root -g)/@effekt-lang/effekt/bin/effekt" $@
```
jiribenes, line 2: https://docs.npmjs.com/cli/v10/using-npm/scope#installing-scoped-packages
> scoped modules will be installed in `node_modules/@myorg/packagename`

b-studios:
I think moving the tests to the ci that creates the version makes sense. Can we automatically rerun the tests a second time? If they still fail it would be good to be notified and abort the release.

build.sbt
```diff
@@ -160,6 +161,27 @@ lazy val effekt: CrossProject = crossProject(JSPlatform, JVMPlatform).in(file("e
...
+      // Update the build.sbt file
+      val buildSbt = (ThisBuild / baseDirectory).value / "build.sbt"
+      IO.writeLines(buildSbt, IO.readLines(buildSbt).map {
+         case line if line.startsWith("lazy val effektVersion =") =>
+          s"""lazy val effektVersion = "$newVersion""""
+        case other => other
+      })
+
+      println(newVersion)
+    },
```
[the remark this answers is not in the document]
b-studios: It might be a bit more robust if we store (only) the version in a separate file in project/version.scala or so.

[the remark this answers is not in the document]
b-studios: This way you wouldn't need to iterate through all lines but could check whether the contents have the exact correct shape and fail otherwise.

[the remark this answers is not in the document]
jiribenes: https://github.com/jiribenes/effekt-ci-test/commit/8d39efb8924a6243b8be47fa94eeba5243f58b9c
Here's the commit on the test repo, let's see how/if it works :)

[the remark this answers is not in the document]
jiribenes: Took a lot of time to dial this in, `sbt` is a ✨ _special_ ✨  one 🙃
(The file really must be called `EffektVersion`, `version` or `Version` won't do.)

jiribenes:
I've moved the tests to `autorelease.yml` (= before the tag is created).
I'll investigate if we can rerun the tests on failure somehow tomorrow.

.github/workflows/autorelease.yml
```diff
@@ -41,8 +42,13 @@ jobs:
...
 
-    - name: Run tests
-      run: sbt clean test
+    - name: Run tests with retry
+      uses: nick-fields/retry@v3
+      with:
+        timeout_minutes: 120 # NOTE: This needs _some_ value. As of writing this, 2 hours should be future-proof. :)
+        max_attempts: 3
+        retry_on: error
+        run: sbt clean test
```
jiribenes, line 51: If we don't want to depend on an external action, we could also add something like the following:
```yaml
    - name: Run tests (Attempt 1)
      id: test_attempt_1
      continue-on-error: true
      run: sbt clean test

    - name: Run tests (Attempt 2)
      id: test_attempt_2
      if: steps.test_attempt_1.outcome == 'failure'
      continue-on-error: true
      run: sbt clean test

    - name: Run tests (Attempt 3)
      id: test_attempt_3
      if: steps.test_attempt_2.outcome == 'failure'
      run: sbt clean test
```

jiribenes:
This PR should now be completely ready for review / discussion. I've modified the PR description to reflect the current state and what the autorelease & deploy CI actually do.

PR #541, opened 2024-08-16
Add 'process' module with 'exit' function

jiribenes:
Added a quick way to exit the running process by calling `exit(exitCode: Int)`.
I don't know what should the module name here should be:
- `process`
- `io/process`
- `sys`
- `os`
- `exit`

Should the capture be just `io`, or should there also be some other `exit` resource so that we know nothing weird happens inside of promises?

### DIFF
--- a/libraries/common/process.effekt
+++ b/libraries/common/process.effekt
@@ -7,4 +7,4 @@ extern io def exit(errorCode: Int): Nothing =
     ret %Pos zeroinitializer
   """
   chez "(exit ${errorCode})"
-  ml "(POSIX.Process.exit (Word8.fromInt ${errorCode}))"
+  ml "(Posix.Process.exit (Word8.fromInt ${errorCode}))"
--- a/libraries/common/process.effekt
+++ b/libraries/common/process.effekt
@@ -7,3 +7,4 @@ extern io def exit(errorCode: Int): Nothing =
     ret %Pos zeroinitializer
   """
   chez "(exit ${errorCode})"
+  ml "(POSIX.Process.exit (Word8.fromInt ${errorCode}))"
--- a/libraries/common/process.effekt
+++ b/libraries/common/process.effekt
@@ -1,0 +1,9 @@
+module process
+
+extern io def exit(errorCode: Int): Nothing =
+  js   "(function() { process.exit(${errorCode}) })()"
+  llvm """
+    call void @exit(i64 %errorCode)
+    ret %Pos zeroinitializer
+  """
+  chez "(exit ${errorCode})"
--- a/libraries/common/test.effekt
+++ b/libraries/common/test.effekt
@@ -93,6 +93,6 @@ def suite(name: String) { body: => Unit / Test }: Bool = {
 /// Use as `def main() = mainSuite("...") { ... }`
 def mainSuite(name: String) { body: => Unit / Test }: Unit = {
   val result = suite(name) { body() }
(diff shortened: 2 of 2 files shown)


libraries/common/test.effekt
```diff
@@ -87,4 +88,11 @@ def suite(name: String) { body: => Unit / Test }: Bool = {
     println(ANSI::GREEN ++ "All tests passed (" ++ passed.show ++ " passed)" ++ ANSI::RESET)
     true
   }
+}
+
+/// Use as `def main() = mainSuite("...") { ... }`
+def mainSuite(name: String) { body: => Unit / Test }: Unit = {
```
jiribenes, line 94: I have no idea what the name of this function should be, but the goal is to use it in automated tests / in standalone test binaries, for example like:
```scala
def main() = mainSuite("My Test Suite") {
  test("Addition works") {
    assertEqual(2 + 2, 4)
  }
}
```

jiribenes: Also, should this return `Unit` or `Nothing`?

libraries/common/process.effekt
```diff
@@ -0,0 +1,9 @@
+module process
+
+extern io def exit(errorCode: Int): Nothing =
+  js   "(function() { process.exit(${errorCode}) })()"
+  llvm """
+    call void @exit(i64 %errorCode)
+    ret %Pos zeroinitializer
+  """
+  chez "(exit ${errorCode})"
```
jiribenes, line 9: There's no MLton version, because I haven't figured out what to do there:
- I found that [`OS.Process.exit OS.Process.failure/success` could be used](https://stackoverflow.com/a/47849043), but it fails on my machine:
```
$ effekt --backend=ml libraries/common/process.effekt
Warning: <bogus> ~1.~1.
  Type of variable was not inferred and could not be generalized: it.
    type: ???
    in: val it = main_2612 (fn a => a)
```
- ... and I can't see a way to return a custom exit code at all.

jiribenes: How do we define `exit` for `jsWeb`? #454

b-studios: Regarding ML and exit, in #551 we try to enforce that main returns `unit`. So the type of `it` should be Unit (not sure how to annotate this here, though).

libraries/common/test.effekt
b-studios: I would go with Unit considering that main currently needs to return Unit

b-studios: Another thought related to the design of `suite` and tests: would it make sense to first create a reified representation of the test results that can be queried to then render / print it / translate to exit code?

jiribenes: Sure, the only problem here is that in the ideal case, what I call `mainSuite` could also be able to install CLI args in order to filter tests / run only a single one.

But yes, the `suite` itself could just return reified results, then we'd have something else that would translate them.

b-studios: The output could / should even be "streaming" using effects :)

jiribenes: I'm fine with doing that, but I would prefer to do that part in #542 instead of here.

libraries/common/process.effekt
jiribenes: Funnily enough, the warning is really just a warning & now I have a version which (on a POSIX system) allows returning any POSIX return code. :)

b-studios (approved):

Issue #496, opened 2024-06-10
Top-level constants in the LLVM backend


jiribenes:
## Motivation

The LLVM backend currently does not support top-level let bindings of any kind.
As a first step, it would be nice to at least support top-level bindings of literals that don't use any capabilities (what I'll call "top-level constants" here), like for example:
```scala
val myString = "Hello, world!"
// ^^^
// [error] Toplevel def and let bindings not yet supported: Let(myString,Literal(Hello, world!,Data(String,List())))

def main() = {
  println(myString)
}
```

### Existing use in the standard library

We even use top-level constants in the standard library in the `string` module:
https://github.com/effekt-lang/effekt/blob/acbec872ae6190bd9662d1eaccc1b00342109851/libraries/common/string.effekt#L206-L227

which are then used by the `test` module:
https://github.com/effekt-lang/effekt/blob/acbec872ae6190bd9662d1eaccc1b00342109851/libraries/common/test.effekt#L79

Unfortunately since the LLVM backend doesn't support these top-level constants, one may use neither ANSI escape codes nor, by transitivity, the `test` module. :(

## Implementation details

Based on my Clang-based testing, I think it should be enough to do the following to declare a top-level `Pos` constant:
```llvm
@topLevelPos = constant %struct.Pos { i64 42, ptr null }, align 8
```

The biggest problem might be that we'd need to add these top-level constants to the Machine representation as that's where the current error is being thrown from:
https://github.com/effekt-lang/effekt/blob/acbec872ae6190bd9662d1eaccc1b00342109851/effekt/shared/src/main/scala/effekt/machine/Transformer.scala#L39-L47
(One could start by pattern-matching on something like `(rest, lifted.Definition.Let(constantName, lit@Literal(_, _)))`...)

phischu:
Yes, we should do it. The hardest part is finding the difference between top-level variables and local variables. Consider the following program:

```
val myString1 = "Hello, world 1! "

def main() = {
  val myString2 = "Hello, world 2!"
  println(myString1)
  println(myString2)
}
```

The occurence of `myString1` and `myString2` look the same to the compiler, but operationally they are vastly different. My strategy would therefore be to transform the program into the following:

```
def myString1() = "Hello, world 1! "

def main() = {
  val myString2 = "Hello, world 2!"
  println(myString1())
  println(myString2)
}
```

[...]

PR #542, opened 2024-08-16
Prettify the stdlib testing framework

jiribenes:
### Motivation

I'm playing around a lot with the stdlib testing framework and it would be nice for it to be ever-so-slightly prettier. :)
I was very inspired by [Bun](https://bun.sh/guides/test/run-tests) when making this.

### Features

> [!NOTE]
> This PR might still need some refactoring before it gets merged. Nitpicks are welcome.
> See [Caveats](#Caveats) below.

- `string` now has many more ANSI escapes for text formatting (bold, italic, etc.)
- `test` now has:
    - a tiny formatting interface using which we can turn all highlighting off (useful for CI / older terminals)
    - a few combinators for printing pretty text in bold, italic, etc.
    - a formatter for time durations in ms
- related: #541 enables a wrapper which makes everything much easier -- there we could make another one which does not use any formatting [= this is **blocked** on #541 which should get merged first]
- tests for `test` are now in `stdlib/test` instead of `pos` (which also means they don't run on MLton, I think)

#### Before:
Screenshot 2024-08-16 at 16 28 10

#### After:
Screenshot 2024-08-16 at 16 28 55

### Caveats

- I don't quite like the ad-hoc formatting library: where could we move this?
    - Also, it _is_ very ad-hoc: `RESET` deletes all formatting, which means that something like `red("Hello " ++ bold("world") ++ "!")` will actually result in `<RED>Hello <BOLD>world<RESET>!<RESET>`, so the exclamation mark is surprisingly no longer red.
- ~~Can/should we move the Duration-related things into `bench`?~~ _EDIT: I've moved most of them except for `diff`_
- Are we okay with `✓` and `✕` instead of `+` and `-`? Can we trust old terminals/CI with these?
- Should we handle `Formatter` straight-away or just expose it to the user and let them handle it?
- ~~The timing needs to be opt-in only (= not in the CI version, sadly, since we're comparing text!)~~ _EDIT: Solved._

### TODO

- [ ] from #541 by @b-studios: Another thought related to the design of suite and tests: would it make sense to first create a reified representation of the test results that can be queried to then render / print it / translate to exit code?


libraries/common/string.effekt
```diff
@@ -245,25 +245,38 @@ def printing[T] { prog: => T / Stream }: T = <>
...
+  val RESET = escape("0")
+
+  val BOLD = escape("1")
+  val FAINT = escape("2")
+  val ITALIC = escape("3")
+  val UNDERLINE = escape("4")
+  val BLINK = escape("5")
+  val REVERSE = escape("7")
+  val CROSSOUT = escape("9")
+  val OVERLINE = escape("53")
```
jiribenes, line 279: On one hand, I'd like to leave this split up, on the other, I'm a little worried that this doesn't get inlined.
The LLVM backend cannot even compile simple top-level constants (see https://github.com/effekt-lang/effekt/issues/496), much less this :(

jiribenes:
This seems to have uncovered two different problems:
1. WebTests don't fail when they should
2. We're not escaping correctly on the MLton backend
```
+ effekt.WebTests.Can read files from the stdlib 8ms  
/home/runner/work/effekt/effekt/out/tests/effekt.javascripttests/lib_test.js:854
              const v_r_5457733111_546489 = Math.round(performance.now() * 1000000);
                                                       ^
ReferenceError: performance is not defined
    at /home/runner/work/effekt/effekt/out/tests/effekt.javascripttests/lib_test.js:854:56
    at /home/runner/work/effekt/effekt/out/tests/effekt.javascripttests/lib_test.js:856:15
    at Object.apply (/home/runner/work/effekt/effekt/out/tests/effekt.javascripttests/lib_test.js:515:19)
    at trampoline (/home/runner/work/effekt/effekt/out/tests/effekt.javascripttests/lib_test.js:435:19)
    at Object.run (/home/runner/work/effekt/effekt/out/tests/effekt.javascripttests/lib_test.js:542:18)
    at Object.main (/home/runner/work/effekt/effekt/out/tests/effekt.javascripttests/lib_test.js:1095:29)
    at Object.<anonymous> (/home/runner/work/effekt/effekt/out/tests/effekt.javascripttests/lib_test:2:89)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
+ effekt.WebTests.Evaluate simple expressions in REPL 14384ms  
42
+ effekt.WebTests.Evaluate expressions that uses a builtin which relies on an extern literal 14812ms  [object Object]
+ effekt.WebTests.Evaluate expressions using stdlib in REPL 11153ms  
+ effekt.WebTests.Show core on website 10711ms  
+ effekt.WebTests.Load file with multiline extern strings 12618ms  
Error: /home/runner/work/effekt/effekt/out/tests/effekt.mltests/lib_test.sml 391.69-391.76.
  String constant with character too large for type: #"\u2715".
    type: string
Error: /home/runner/work/effekt/effekt/out/tests/effekt.mltests/lib_test.sml 561.67-561.74.
  String constant with character too large for type: #"\u2713".
    type: string
    ```

**Comment by jiribenes** [comment]
There's a test failure that I don't quite understand: 🤔 
```
==> X effekt.JavaScriptTests.examples/pos/lib_test.effekt (js)  0.641s munit.ComparisonFailException: /home/runner/work/effekt/effekt/effekt/jvm/src/test/scala/effekt/EffektTests.scala:102
[...]
diff assertion failed
=> Obtained
    """|Arithmetic
Error:        |[error] examples/neg/def_instead_of_val.effekt:3:13: Process exited with non-zero exit code 1.
       |def three = addOne(2)
       |            ^^^^^^^^^
       |""".stripMargin
 Arithmetic
-[error] examples/neg/def_instead_of_val.effekt:3:13: Process exited with non-zero exit code 1.
-def three = addOne(2)
-            ^^^^^^^^^
+✓ addition
+✕ faulty addition
+  2 + 2 should be 4!
+✓ multiplication
+
+ 2 pass
+ 1 fail
+ 3 tests total
```

File: libraries/common/string.effekt (hunk 1 above)
Comment (b-studios) [reply libraries/common/string.effekt:279]: I believe currently this will not be constant-folded. I think we should support top-level constants in LLVM for this to work; WDYT @phischu?

Comment (jiribenes) [reply libraries/common/string.effekt:279]: I'm fine with inlining the calls to `escape` and `++` here manually, but I'd still like #496. Otherwise this code is still unusable in LLVM.

**Comment by jiribenes** [comment]
Rebased on `master` after #541 got merged.

File: libraries/common/test.effekt [hunk 2]
```diff
@@ -29,70 +64,131 @@ def assert(obtained: Char, expected: Char, msg: String): Unit / Assertion =
...
+/// Recommended for standalone test files ran by CI.
+///
+/// Exits after running all tests:
+/// - if all tests succeed, exits the program with success (exit code 0)
+/// - otherwise exits the program with failure (exit code 1)
+def mainSuite(name: String) { body: => Unit / { Test, Formatted } }: Unit = {
+  val result = suite(name, true) { body }
   val exitCode = if (result) 0 else 1
   exit(exitCode)
 }
```
jiribenes, line 194: Unfortunately, this cool exiting runner is pretty useless if one of your files has multiple test suites.
In general, having multiple test suites is this testing framework's weak spot.

We could mandate `suite`s and use them just for compartmentalisation/grouping, making a new `runTests` handler, but what if I really just want to write three tests and be done with it all?

jiribenes:
Not sure if we want something like the ability to set tests as ignored/skipped when developing. If we ever do, here's a quick patch:

<details>
  <summary><i>Patch for setting a test as skipped</i></summary>

```diff
diff --git i/libraries/common/test.effekt w/libraries/common/test.effekt
index c30f305c..28644bea 100644
--- i/libraries/common/test.effekt
+++ w/libraries/common/test.effekt
@@ -94,6 +94,7 @@ def assertEqual[A](obtained: A, expected: A) { equals: (A, A) => Bool } { show:
 interface Test {
   def success(name: String, duration: Int): Unit
   def failure(name: String, msg: String, duration: Int): Unit
+  def skipped(name: String): Unit
 }
 
 /// Runs the `body` as a test under the given `name`
@@ -114,6 +115,10 @@ def test(name: String) { body: => Unit / Assertion } = {
   }
 }
 
+/// Does *not* run the `body` as a test under the given `name`: 
+/// Useful for ignoring a test.
+def skip(name: String) { body: => Unit / Assertion } = do skipped(name)
+
 /// Run a test suite with a given `name`.
 /// - If `printTimes` is `true` (or missing), prints out time in milliseconds.
 /// - Formats automatically using ANSI escapes.
@@ -164,14 +169,26 @@ def suite(name: String, printTimes: Bool) { body: => Unit / { Test, Formatted }
         println("  " ++ msg.red)
         resume(())
       }
+
+      // 2c) Handle a skipped test
+      def skipped(name) = {
+        skipped = skipped + 1
+        println(("»" ++ " " ++ name).dim)
+        // no need to resume
+      }
     }
   }
 
+  val total = passed + failed + skipped
+
   // 3) Format the test results
   println("")
   println(" " ++ (passed.show ++ " pass").dimWhenZeroElse(passed) { green })
+  if (skipped > 0) { 
+    println(" " ++ (skipped.show ++ " skip").dim)
+  }
   println(" " ++ (failed.show ++ " fail").dimWhenZeroElse(failed) { red })
-  println(" " ++ (passed + failed).show ++ " tests total" ++ totalDuration.ms)
+  println(" " ++ total.show ++ " tests total" ++ totalDuration.ms)
 
   // 4) Return true if all tests succeeded, otherwise false
   return failed == 0
```

</details>

---

I find this interesting, since ideally, we would have some basic CLI args support for `mainSuite`, allowing it to run only certain tests (this way, we could have simple IDE support even to click & run all / click & run a single test).

b-studios:
This is so nice, can't we just merge it?

effekt/jvm/src/test/scala/effekt/StdlibTests.scala
```diff
@@ -76,5 +79,8 @@ class StdlibLLVMTests extends StdlibTests {
     examplesDir / "stdlib" / "list" / "build.effekt",
     examplesDir / "stdlib" / "string" / "strings.effekt",
     examplesDir / "stdlib" / "string" / "unicode.effekt",
+
+    // missing support for top-level constants for ANSI escapes
+    examplesDir / "stdlib" / "test",
```
jiribenes, line 84: Since top-level definitions #496 are fixed now, this should Just :tm: go through :)

jiribenes:
```suggestion
```
(after rebuilding)

effekt/jvm/src/test/scala/effekt/StdlibTests.scala
```diff
@@ -45,9 +45,12 @@ class StdlibMLTests extends StdlibTests {
     // unicode is not supported in the ML backend
     examplesDir / "stdlib" / "string" / "unicode.effekt",
 
+    // missing support for multibyte character escape in a string
+    examplesDir / "stdlib" / "test",
+
```
jiribenes, line 50: Waiting for #544.

libraries/common/bench.effekt
```diff
@@ -27,6 +27,11 @@ extern llvm """
   declare i32 @clock_gettime(i32, ptr)
 """
 
+// This will not be needed from NodeJS 16 and newer.
+extern jsNode """
+  const { performance } = require('perf_hooks');
+"""
```
jiribenes, line 33: Do we want to keep this import now that we're on NodeJS 16 and newer?
```suggestion
```

libraries/common/test.effekt
```diff
@@ -1,5 +1,40 @@
...
+    try { prog() } with Formatted {
+      def supportsEscape(escape: String) = resume(false)
+    }
+
+  def tryEmit(escape: String): String / Formatted =
+    if (do supportsEscape(escape)) escape else ""
+
+  def colored(text: String, colorEscape: String): String / Formatted =
+    tryEmit(colorEscape) ++ text ++ tryEmit(ANSI::RESET)
+}
```
jiribenes, line 27: I don't know whether this should stay here or not: should it go into a different module?

jiribenes:
I don't understand the errors here, it works fine on my machine.
It's definitely something to do with the new round, ceil, floor functions added to LLVM, but those are just intrinsics (https://releases.llvm.org/12.0.0/docs/LangRef.html#llvm-round-intrinsic), though they need to be declared.
I thought it might be a version mismatch since I'm on LLVM 18, but these intrinsics have existed since LLVM 12. Strange.

If I don't figure this out soon, I'll cherry-pick these last two commits out and make a separate PR to discuss this.

jiribenes:
Rebased on the newly merged #606.

b-studios (approved):
:shipit:

PR #545, opened 2024-08-17
Fix off-by-one error in array::copy

### LINKED ISSUE #508: `array::copy(from, start, to, offset, length)` copies one element more
According to the documentation:
```effekt
/**
 * Copies `length`-many elements from `from` to `to`
 * starting at `start` (in `from`) and `offset` (in `to`)
 */
def copy[T](from: Array[T], start: Int, to: Array[T], offset: Int, length: Int): Unit / Exception[OutOfBounds]
```

However, actually:
```effekt
import array

def main() = {
  with on[OutOfBounds].panic

  val a = array(10, 0)
  val b = array(10, 1)

  copy(a, 0, b, 0, 5)
  println(b)
}
```
consistently outputs:
```effekt
Array(0, 0, 0, 0, 0, 0, 1, 1, 1, 1)
```
(i.e., with 6 times 0 instead of 5) in all backends I tried.

Probably, here:
https://github.com/effekt-lang/effekt/blob/a4d5f589f1a59a8091f03410f75df735ae77cecc/libraries/common/array.effekt#L119
it should be `>` instead of `>=`.

rzikm:
Fixes #508.


jiribenes (approved):
Thank you for the contribution, everything checks out! :)

PR #551, opened 2024-08-21
Require main to always return unit

b-studios:
In the LLVM backend, we experienced problems with boxing of the result of main. Here, we require that `main` always returns `Unit` and check this.

Related to #541


b-studios:
This breaks `effekt.WebTests` since they rely on the (only) exported function main to return a result that can be inspected. 

We could potentially use a mutable reference in JS to sidechannel the result and compare it this way for testing purposes.

b-studios:
There are some more tests that fail now. However, the most interesting one is the `stdlib/test/test` which uses the unit testing library that returns a boolean.

jiribenes:
> However, the most interesting [failing test] is the `stdlib/test/test` which uses the unit testing library that returns a boolean.

I saw this coming, #541 introduces `def mainSuite(name: String) { body: => Unit / Test }: Unit` which uses `exit` explicitly instead of returning a boolean which is supposed to be used exactly like `def main() = mainSuite("my tests") { <tests> }` :)

phischu (approved):

Issue #548, opened 2024-08-20
Capability escapes from higher-order function


b-studios:
@dvdvgt identified the following program

```
interface Time {
  def now(): Int
}

def withTime[A] { prog: { Time } => A }: A =
  try {
    prog {timeCap}
  } with timeCap: Time {
    def now() = resume(1)
  }

def main() = {
  def b { t: Time }: Time at {t} = box t;  // this is fine
  val cap = withTime {b} // ERROR escapes
  def t2: Time = unbox cap
  t2.now()
}
```

which type-checks but crashes at runtime. It appears, we forgot the wellformedness check on higher-order functions.

b-studios:
We need to check that all inferred type arguments (and potentially return types) are well-formed.
In this case both are not, for instance:

```
withTime[Time at {t}] {b}
//       ^^^^^^^^^^^
//      t not in scope
```

b-studios:
Same wellformedness bug with higher rank types:

```
def hof[A] { prog: [B](B) => A }: A = prog[String]("hello")

def main() = {
  val res = hof { [C](x: C) => x };
  println(res.genericShow)
}
```

dvdvgt:
https://github.com/effekt-lang/effekt/blob/68d40378fea59cb66074de3eb61df7048578a4e1/effekt/shared/src/main/scala/effekt/typer/Wellformedness.scala#L12

needs to be changed to `class WFContext(var capsInScope: List[symbols.BlockParam])` and then updated at every binding instance while type checking. We still need to investigate whether we need to check the well-formedness in every "type judgement".

PR #552, opened 2024-08-22
Implement a proper wellformedness check

b-studios:
This is an attempt at fixing #548

Create a proper context tracking which bindings (types and captures) are in scope and then run the wellformedness check on inferred types.

TODO:

- [x] make sure that we catch all binding occurrences
- [x] make sure that we check all inferred types (type arguments, return types of functions and matches, ...)
- [x] treat existentials properly


b-studios:
Still todo:

- [ ] improve implementation and deduplicate
- [ ] improve error messages (previously they have been specialized to what kind of capture escaped "local variable ... escapes" instead of "captures {x} escapes ..."

Especially the worse error messages are frustrating. It would be great to look at the diff of this PR and see how we can get this kind of (or better) messages back.

PR #553, opened 2024-08-23
Direct IO which is asynchronous

phischu:
TODO
- [x] Fix in JS backend `io/console.effekt`
- [ ] Implement in LLVM backend `io/console.effekt`
- [x] Fix in JS backend `io/network.effekt`
- [ ] Implement in LLVM backend `io/network.effekt`
- [x] Rename `io/files.effekt` to `io/filesystem.effekt`
- [ ] Rename `run_Pos` to `runPositive` etc.
- [x] Check and handle error result in call to `uv_fs_open` etc.
- [x] Find and fix memory leaks


libraries/common/io/files.effekt
```diff
@@ -264,6 +264,43 @@ namespace callback {
...
+    call i32 @uv_run(ptr %loop, i32 0)
+    call void @uv_loop_close(ptr %loop)
+    ret %Pos zeroinitializer
+  """
+
+  extern control def openFile(path: String, mode: String): Int =
+  llvm """
+    call void @c_open_file(%Pos %path, %Pos %mode, %Stack %stack)
+    ret void
+  """
```
b-studios, line 282: This starts to look really nice! Good job!

libraries/common/io/files.effekt
```diff
@@ -89,17 +89,17 @@ def filesystem[R] { program: => R / Files }: R / { AsyncIO, Exception[IOError] }
 /**
  * Reads a file at given path as utf8 encoded string.
  */
-def readFile(path: String): String / { AsyncIO, Exception[IOError] } = {
-  val fd = openFile(path, ReadOnly());
-  with on[IOError].finalize { closeFile(fd) }
+def readFile(path: String): String / Exception[IOError] = {
+  val fd = direct::openFile(path, modeName(ReadOnly()));
+  with on[IOError].finalize { direct::closeFile(fd); () }
```
b-studios, line 94: It's nice that we can make this work in LLVM, but in JS this sadly cannot work. There, you cannot make a synchronize file operation asynchronous.

b-studios: Ok, I retract. Of course we can, but this will poison the coeffecrs with `control` (which is maybe fine)

libraries/llvm/io.c
```diff
@@ -293,18 +254,11 @@ void c_resume_unit_timer(uv_timer_t* handle) {
...
 void c_timer_start(Int millis, Stack stack) {
 
-  // Get the default loop
-  uv_loop_t* loop = uv_default_loop();
-
-  // Allocate memory for the timer handle
-  uv_timer_t* timer = (uv_timer_t*)malloc(sizeof(uv_timer_t));
-
-  // // Initialize the timer handle
-  uv_timer_init(loop, timer);
```
b-studios: These comments might seem trivial, but I wouldn't just delete all of them.

libraries/common/io/error.effekt
```diff
@@ -280,7 +280,7 @@ def name(error: IOError): String = error match {
  *
  * See documentation of type IOError for further information.
  */
-def fromCode(code: Int): IOError = code match {
+def fromNumber(number: Int): IOError = number match {
```
b-studios, line 283: why number? this is literally called "error codes"

libraries/js/io.js
```diff
@@ -1,93 +1,94 @@
...
-        EAI_SOCKTYPE: 213,
-        ECHARSET: 215,
-        ENONET: 216,
-        UNKNOWN: 217,
-        EOF: 218,
-        EUNATCH: 219,
-        ESHUTDOWN: 220
+function errorNumber(errno) {
+    const errnoMap = {
+        [os.constants.errno.EPERM]: 1,
```
b-studios, line 12: computed keys will prevent us from using the closure compiler

effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala
```diff
@@ -18,10 +18,9 @@ define ${show(callingConvention)} ${show(returnType)} ${globalName(name)}(${comm
     ${indentedLines(basicBlocks.map(show).mkString)}
 }
 """
-    case VerbatimFunction(returnType, name, parameters, body) =>
-      // TODO what about calling convention?
+    case VerbatimFunction(callingConvention, returnType, name, parameters, body) =>
       s"""
-define ${show(returnType)} ${globalName(name)}(${commaSeparated(parameters.map(show))}) {
+define ${show(callingConvention)} ${show(returnType)} ${globalName(name)}(${commaSeparated(parameters.map(show))}) {
```
b-studios, line 23: This will go away when rebasing, right?

effekt/shared/src/main/scala/effekt/lifted/LiftInference.scala
```diff
@@ -179,6 +176,13 @@ object LiftInference extends Phase[CoreTransformed, CoreLifted] {
 
     case core.Region(_) => ErrorReporter.panic("Should not happen. Regions always take block literals as body.")
 
+    case core.App(core.BlockVar(b: symbols.ExternFunction, tpe, _), targs, vargs, bargs) =>
+      if (bargs.nonEmpty) {
+        ErrorReporter.panic("Extern control definitions must not take block arguments")
+      } else {
+        App(BlockVar(b, transform(tpe)), targs.map(transform), vargs map transform)
+      }
```
b-studios, line 184: Typically we just check preconditions and then proceed on the normal path
```suggestion
    case core.App(core.BlockVar(b: symbols.ExternFunction, tpe, _), targs, vargs, bargs) =>
      if (bargs.nonEmpty) ErrorReporter.panic("Extern control definitions must not take block arguments")
      App(BlockVar(b, transform(tpe)), targs.map(transform), vargs map transform)
```

effekt/shared/src/main/scala/effekt/lifted/Tree.scala
```diff
@@ -53,7 +53,7 @@ enum Extern {
   // WARNING: builtins do not take evidence. If they are passed as function argument, they need to be eta-expanded.
   //   (however, if they _would_ take evidence, we could model mutable state with this)
   // TODO revisit
-  case Def(id: Id, tparams: List[Id], params: List[Param], ret: ValueType, bodies: ExternBody)
+  case Def(id: Id, tparams: List[Id], params: List[Param], ret: ValueType, control: Boolean, body: ExternBody)
```
b-studios, line 56: This will not be necessary when rebasing, since core has annotated captures already.

examples/stdlib/io/files/async_file_io.effekt
```diff
@@ -3,17 +3,17 @@ import io/error
 import io/files
 
 def program(path1: String, path2: String) =  {
-  val f1 = do promise(box {
+  val f1 = promise(box {
     with on[IOError].panic;
     readFile(path1)
   })
```
b-studios, line 9: 🥇 

TBH I really dislike `do promise`

examples/stdlib/io/promise.effekt
```diff
@@ -1,12 +1,13 @@
...
-  promise::resolve(p, 42)
+  spawn(box { println("Resolved 1: " ++ promise::await(p)) });
+  spawn(box { println("Resolved 2: " ++ promise::await(p)) });
+  spawn(box { println("Resolved 3: " ++ promise::await(p)) });
 
-}
\ No newline at end of file
+  promise::resolve(p, "42")
+
+})
```
b-studios, line 13: In the Effekt meeting, we said that we want to always automatically insert libuv::start if `main` includes capture `control`.

libraries/common/io.effekt
```diff
@@ -6,49 +6,80 @@ import queue
...
+
+def spawn(task: Task[Unit]): Unit / Eventloop =
+  // TODO installing a handler here is unnecessary
+  direct::spawn(box{ try task() with Eventloop {} })
+
+def yield(): Unit / Eventloop =
+  direct::yield()
+
+def abort(): Nothing / Eventloop =
+  direct::abort()
```
b-studios, line 29: Why the indirection and not move those out of `direct`?

libraries/common/io.effekt
```diff
@@ -6,49 +6,80 @@ import queue
...
+// Event Loop
+// ----------
+
+interface Eventloop {}
+
+type Task[T] = () => T / Eventloop at {io, control, global}
+
+def eventloop(program: Task[Unit]): Unit = direct::eventloop(box {
+  try program() with Eventloop {}
+})
```
b-studios, line 19: As discussed in the meeting, let's drop this effect and automatically insert `eventloop` when `main` has capture `control`.

libraries/common/io.effekt
```diff
@@ -6,49 +6,80 @@ import queue
...
+  // = llvm "{tag: 0, obj: Promise*}"
 
-interface Promises {
-  def promise[T](p: Task[T]): Promise[T]
-  def await[T](p: Promise[T]): T
+def promise[T](task: Task[T]): Promise[T] / Eventloop = {
+  val p = promise::make[T]();
+  spawn(box { promise::resolve(p, task()) });
+  return p
 }
```
b-studios, line 78: I think it is fine, but please be aware that with all of these changes the scheduling and semantics of promises is baked into the language and cannot be changed by users.

libraries/common/io.effekt
```diff
@@ -6,49 +6,80 @@ import queue
...
 
+def await[T](promise: Promise[T]): T =
+  promise::await(promise)
+
```
b-studios, line 82: Why single out `await` here? Users _could_ in principle also call `resolve`, right? Why hide one behind a namespace? This is not your decision, I know. I am just questioning the design.

libraries/common/io.effekt
```diff
@@ -63,96 +94,23 @@ namespace promise {
...
-      def promise[T](task: Task[T]) = {
-        val p = promise::make[T]();
-        handle(box {
-          val t = task()
-          promise::resolve(p, t)
-        })
-        eventloop::schedule( box { resume(p) })
-      }
-      def await[T](p: Promise[T]) = promise::await(p, box { t => resume(t) })
-    }
```
b-studios, line 120: One point of this exercise was to see how well we can write an eventloop within Effekt. I agree that for IO it could be built in. We just need to make sure (and test) that the semantics aligns between LLVM, JS, and what we intend.

libraries/common/io/files.effekt
```diff
@@ -55,51 +29,27 @@ type Mode {
...
-
 /**
  * Reads a file at given path as utf8 encoded string.
  */
-def readFile(path: String): String / { AsyncIO, Exception[IOError] } = {
-  val fd = openFile(path, ReadOnly());
-  with on[IOError].finalize { closeFile(fd) }
+def readFile(path: String): String / Exception[IOError] = {
+  val fd = direct::open(path, ReadOnly());
+  with on[IOError].finalize { direct::close(fd); () }
```
b-studios, line 44: Note, that if you don't hide it behind the namespace `direct`, you can call it `fd.close()`.

libraries/common/io/files.effekt
```diff
@@ -124,9 +74,9 @@ def readFile(path: String): String / { AsyncIO, Exception[IOError] } = {
 /**
  * Writes the (utf8 encoded) string `contents` into the specified file.
  */
-def writeFile(path: String, contents: String): Unit / { AsyncIO, Exception[IOError] } = {
-  val fd = openFile(path, Write());
-  with on[IOError].finalize { closeFile(fd) }
+def writeFile(path: String, contents: String): Unit / Exception[IOError] = {
+  val fd = direct::open(path, WriteOnly());
```
b-studios, line 78: Is this really intended to be `direct::open`? It looks like `direct::open` doesn't check results while  `open` does.

libraries/common/io/files.effekt
```diff
@@ -149,120 +99,87 @@ def writeFile(path: String, contents: String): Unit / { AsyncIO, Exception[IOErr
...
-      call void @c_file_write(i32 %fd_int, %Pos %buffer, i64 noundef %offset, ptr %onSuccess_ptr, ptr %onFailure_ptr) #0
-      ret %Pos zeroinitializer
-    """
-}
+  extern control def close(fd: Int): Int =
+  js "$effekt.callcc(callback => close(${fd}, callback))"
+  llvm """
+    call void @c_fs_close(%Int %fd, %Stack %stack)
+    ret void
+  """
```
b-studios, line 172: The indentation seems of here and on `checkResult`.

libraries/js/io.js
```diff
@@ -1,137 +1,155 @@
...
+  case 10: // ReadWriteSync()
+    return 'rs+';
+  case 11: // WriteExclusive()
+    return 'wx';
+  case 12: // ReadWriteExclusive()
+    return 'wx+';
+  default:
+    // Invalid tag value
+    return null;
+  }
```
b-studios, line 125: Why can't we write this in Effekt?

libraries/llvm/io.c
```diff
@@ -27,159 +27,128 @@ void c_io_println_String(String text) {
...
-            break;
-        }
-    }
-    // do we need to erase promise now? Is it shared before?
-    erasePositive(promise);
+    // Convert the Effekt String to a 0-terminated string
+    char* path_str = c_buffer_as_null_terminated_string(path);
+    erasePositive((struct Pos) path);
+
+    // Convert the Effekt String representing the opening mode to libuv flags
```
b-studios, line 105:
```suggestion
    // Convert the Effekt Mode-Datatype representing the opening mode to libuv flags
```

libraries/llvm/io.c
```diff
@@ -27,159 +27,128 @@ void c_io_println_String(String text) {
...
+
+    uv_fs_t* request = malloc(sizeof(uv_fs_t));
+    request->data = stack;
+
+    uv_fs_open(uv_default_loop(), request, path_str, flags, 0666, c_resume_int_fs);
+    // TODO report result (UV_EINVAL)
+
+    // We must free the string, since libuv copies it into the request
+    // TODO double check!
+    free(path_str);
```
b-studios, line 117: This should be fine, I double checked that multiple times

libraries/llvm/io.c
```diff
@@ -27,159 +27,128 @@ void c_io_println_String(String text) {
...
+
+    // Convert the Effekt String representing the opening mode to libuv flags
+    int flags = modeFlags(mode);
+    erasePositive((struct Pos) mode);
+
+    uv_fs_t* request = malloc(sizeof(uv_fs_t));
+    request->data = stack;
+
+    uv_fs_open(uv_default_loop(), request, path_str, flags, 0666, c_resume_int_fs);
+    // TODO report result (UV_EINVAL)
```
b-studios, line 113: Do we need to free stuff when this errors? This probably means the continuation is never called and also needs to be ref-count-decremented.

libraries/llvm/io.c
```diff
@@ -277,289 +246,155 @@ int uv_error_to_errno(int uv_err) {
...
+				Listeners* current = tail;
+				while (current != NULL) {
+					Stack head = current->head;
+					Listeners* tail = current->tail;
+					free(current);
+					eraseStack(head);
+					current = tail;
+				};
+			};
+            break;
```
b-studios, line 310: What is up with the indentation here?

libraries/llvm/io.c
```diff
@@ -277,289 +246,155 @@ int uv_error_to_errno(int uv_err) {
...
-    uv_fs_read(loop, req, fd, &buf, 1, offset, c_file_read_handler);
+void c_promise_resume_listeners(Listeners* listeners, struct Pos value) {
+	if (listeners != NULL) {
+		Stack head = listeners->head;
+		Listeners* tail = listeners->tail;
+		free(listeners);
+		c_promise_resume_listeners(tail, value);
+		sharePositive(value);
+		resume_Pos(head, value);
+	}
```
b-studios, line 325: Since when do we use tabs?

b-studios: Please check to not use tabs anywhere, this shows up in the whole file.

b-studios (commented):
In general I like the direction of this PR, a lot.

Could you please:

- rebase it (this will get rid of the additional `control` flag on externs in lifted, you can delete that again, since we don't plan on supporting the other backends any longer)
- automatically insert a call to `eventloop` if `main` contains `control`?
- check indentation (tabs are used in C files quite a lot)

In general, we need to decide which functions should be hidden behind namespaces like `direct` and `promises` and which don't.

Can you showcase somewhere how to recover the callback-based version, based on the direct-style one? I think a test or something to document it would be great.

b-studios:
There seems to be a valgrind error on the `promise.effekt` example. @phischu could you investigate this? Or maybe @marvinborner ?

b-studios:
@dvdvgt the latest commit renames `control` to `async`. We should also reflect this in external documentation.

libraries/js/io.js
phischu: This is a problem, I don't have a solution...

libraries/common/io.effekt
phischu: `resolve` should be called exactly once, while `await` can be called arbitrarily.

b-studios: Do we want to enforce this by hiding it behind a namespace?

libraries/common/io/files.effekt
phischu: That's a mistake, fixed.

libraries/llvm/io.c
phischu: Sorry!

libraries/llvm/io.c
```diff
@@ -109,11 +109,15 @@ void c_fs_open(struct Pos path, struct Pos mode, Stack stack) {
...
-    // TODO report result (UV_EINVAL)
+    int result = uv_fs_open(uv_default_loop(), request, path_str, flags, 0666, c_resume_int_fs);
 
-    // We must free the string, since libuv copies it into the request
-    // TODO double check!
+	if (result < 0) {
+        uv_fs_req_cleanup(request);
+        free(request);
+        resume_Int(stack, result);
+	}
```
b-studios, line 118: The indentation seems to be off here

libraries/llvm/io.c
```diff
@@ -127,8 +131,13 @@ void c_fs_read(Int fd, struct Pos buffer, Int offset, Stack stack) {
...
 
-    uv_fs_read(uv_default_loop(), request, fd, &buf, 1, offset, c_resume_int_fs);
-    // TODO report result (UV_EINVAL)
+    int result = uv_fs_read(uv_default_loop(), request, fd, &buf, 1, offset, c_resume_int_fs);
+
+	if (result < 0) {
+        uv_fs_req_cleanup(request);
+        free(request);
+        resume_Int(stack, result);
+    }
```
b-studios:
```suggestion
    if (result < 0) {
        uv_fs_req_cleanup(request);
        free(request);
        resume_Int(stack, result);
    }
```

libraries/llvm/io.c
```diff
@@ -139,17 +148,27 @@ void c_fs_write(Int fd, struct Pos buffer, Int offset, Stack stack) {
...
 
-    uv_fs_close(uv_default_loop(), request, fd, c_resume_int_fs);
-    // TODO report result (UV_EINVAL)
+    int result = uv_fs_close(uv_default_loop(), request, fd, c_resume_int_fs);
+
+	if (result < 0) {
+        uv_fs_req_cleanup(request);
+        free(request);
+        resume_Int(stack, result);
+    }
```
b-studios, line 171:
```suggestion
    if (result < 0) {
        uv_fs_req_cleanup(request);
        free(request);
        resume_Int(stack, result);
    }
```

libraries/llvm/io.c
```diff
@@ -142,12 +144,14 @@ void c_fs_read(Int fd, struct Pos buffer, Int offset, Stack stack) {
 
 void c_fs_write(Int fd, struct Pos buffer, Int offset, Stack stack) {
 
+    char* bytes = c_buffer_bytes(buffer); // libuv expects signed integers
+    uv_buf_t buf = uv_buf_init(bytes, c_buffer_length(buffer));
+    // erasePositive(buffer);
+    // TODO we should erase the buffer but abort if this was the last reference
```
b-studios, line 150: You mean: decrement the RC but don't free?

b-studios: Postponed solution:

a)
- allocate a user-defined struct `Stack x Buffer`
- erase in the continuation / callback

b) 
- erase here, but make sure that it is USED on the Effekt side.

b-studios:
Please delete / fix all occurrences of `AsyncIO`, then this is ready to merge.

b-studios:
Thanks, this is a huge step forward!

PR #555, opened 2024-08-24
Fix: Lexer crashes on unmatched `}`

dvdvgt:
Currently, the compiler crashes if the lexer encounters the program consisting of
- just a single `}`
- an unmatched `}`, that is without a matching `{` or `${`
- and probably more complex cases

This is due to these unsafe lines here 
https://github.com/effekt-lang/effekt/blob/d96a17b70f99099d9a77e7059ccde99be6511421/effekt/shared/src/main/scala/effekt/Lexer.scala#L296-L298
I forgot to fix when writing lexer.


jiribenes (approved):
LGTM, I tried this out locally: although it (naturally) doesn't help the early exiting parser, this should at least prevent a bunch of early exits in the lexer phase 👍

PR #556, opened 2024-08-26
Add `--version` CLI flag

### LINKED ISSUE #540: Add `--version` CLI flag
In tooling (especially in https://github.com/effekt-lang/effekt-vscode/pull/20), it would be nice to know the version of the current Effekt executable so that we can offer a newer version.

We already surface the version number via the REPL, but it would be _much_ easier if tooling could rely on an easily parseable format.

### Specification

When the user specifies the `--version` flag, the program prints the `effekt.util.Version.effektVersion` and exits successfully. Note that the successful exit is very important, this allows us to automate a bunch of things.

I'm not really sure if it's better to implement this in `Kiama` for `REPLConfig` directly or here for `EffektConfig`.

### Alternatives

We could also just assume every user installed Effekt via `npm` and check the version there, but that feels extremely finicky to me -- it would also break my own workflow, as my Effekt is _never_ installed via `npm` :)

### Related work 

For example in `npm` and `node`:
```
$ npm --version
10.5.0

$ node --version
v20.12.2
```

jiribenes:
Resolves #540

Scallop already supports printing out a version just by calling `version(<your version identifier>)` (see docs [here](https://github.com/scallop/scallop/wiki/Help-information-printing#version-summary)), let's just use it :)

### Usage
```
$ effekt --version
Effekt 0.99.99+nightly
```
(I changed my version in `build.sbt` in order to make this very visible)

### Help page

The version also gets printed out on `--help`:
```
$ effekt --help
Effekt 0.99.99+nightly
 Common Options
  -c, --compile                  Compile the Effekt program to the backend
                                 specific representation
      --no-compile
  -b, --build                    Compile the Effekt program and build a backend
                                 specific executable
      --nobuild
  -o, --out  <arg>               Path to write generated files to (defaults to
                                 ./out)
      --includes  <arg>...       Path to consider for includes (can be set
                                 multiple times)
  -l, --lib  <arg>               Path to the standard library to be used
      --backend  <arg>           The backend that should be used Choices: js,
                                 js-web, chez-callcc, chez-monadic, chez-lift,
                                 llvm, ml
```

Note that we could add a larger "Usage: effekt ..."-style banner, see [here](https://github.com/scallop/scallop/wiki/Help-information-printing#help-page).

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/EffektConfig.scala
+++ b/effekt/jvm/src/main/scala/effekt/EffektConfig.scala
@@ -6,8 +6,8 @@ import kiama.util.REPLConfig
 import org.rogach.scallop.{ ScallopOption, fileConverter, fileListConverter, longConverter, stringConverter, stringListConverter }
 
 class EffektConfig(args: Seq[String]) extends REPLConfig(args.takeWhile(_ != "--")) {
-  // Print version on `--version`
-  version(effekt.util.Version.effektVersion)
+  // Print version information on `--version` and `--help`
+  version(s"Effekt ${effekt.util.Version.effektVersion}")
 
   // Common
   // ------
--- a/effekt/jvm/src/main/scala/effekt/EffektConfig.scala
+++ b/effekt/jvm/src/main/scala/effekt/EffektConfig.scala
@@ -6,6 +6,8 @@ import kiama.util.REPLConfig
 import org.rogach.scallop.{ ScallopOption, fileConverter, fileListConverter, longConverter, stringConverter, stringListConverter }
 
 class EffektConfig(args: Seq[String]) extends REPLConfig(args.takeWhile(_ != "--")) {
+  // Print version on `--version`
+  version(effekt.util.Version.effektVersion)
 
   // Common
   // ------


effekt/jvm/src/main/scala/effekt/EffektConfig.scala
```diff
@@ -6,6 +6,8 @@ import kiama.util.REPLConfig
 import org.rogach.scallop.{ ScallopOption, fileConverter, fileListConverter, longConverter, stringConverter, stringListConverter }
 
 class EffektConfig(args: Seq[String]) extends REPLConfig(args.takeWhile(_ != "--")) {
+  // Print version on `--version`
+  version(effekt.util.Version.effektVersion)
```
b-studios, line 10: If we prefix it with `"Effekt "`, then it wouldn't look so ugly in the help dialog and remains somewhat easy to parse.

b-studios:
Ah you already discuss the prefixing in the PR description...

b-studios:
I agree it is simpler to process, but you probably need to parse the version anyways, no? If not, then lexicographic ordering would also work for a stable prefix, still.

I am just saying: I now how to strip a prefix in javascript, but don't know whether it is possible to configure Scallop the right way without forking it.

jiribenes:
It complicates the tooling [IMO a bit unnecessarily], but I really need this PR to go through to unblock the other ones, so I'll concede the point and just do `Effekt 0.1.2.3` instead.

Now the question is: are we happy with this format? It will be really difficult to change it afterwards, so please, if there's anybody who wants the output to look differently, now's your chance, please speak up!

b-studios:
I don't have a strong preference so choose what you prefer and I'll handle the rest :)

jiribenes:
I went ahead and customised the tooling so that it can handle a prefix, now I just want some confidence vote on the specific prefix to be used so that I don't have to support multiple different ones in the future.

Issue #428, opened 2024-03-22
Investigate CI setup caching


jiribenes:
It takes a long time (~2 minutes) for our developer environment to load in CI.
Let's investigate different ways of speeding this up:
1. [GitHub Action caching](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)
2. Custom Docker/Nix image

EDIT: Please see the comments in https://github.com/effekt-lang/effekt/pull/560 if you're interested in tackling this.

jiribenes:
I think we're caching enough as of #560 to close this.

PR #560, opened 2024-08-29
Cache sbt files in CI, update versions of CI actions

jiribenes:
Investigating #428 since I noticed that `setup-java` allows simple `sbt` caching when playing with https://github.com/effekt-lang/effekt/pull/538 :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,9 +20,11 @@ jobs:
         submodules: 'true'
 
     - name: Set up JDK 11
-      uses: actions/setup-java@v1
+      uses: actions/setup-java@v3
       with:
         java-version: 11
+        distribution: 'zulu'
+        cache: 'sbt'
 
     - name: Set up NodeJS
       uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -45,9 +47,11 @@ jobs:
         submodules: 'true'
 
     - name: Set up JDK 11
-      uses: actions/setup-java@v1
+      uses: actions/setup-java@v3
       with:
         java-version: 11
+        distribution: 'zulu'
+        cache: 'sbt'
```
jiribenes, line 54: The newer `setup-java` action requires an explicit JDK distribution. I set it to `zulu` since that's what it used implicitly, but I'm open to using a different distribution. (I honestly have no idea which distribution is the best here.)

Here's the list of supported distributions https://github.com/actions/setup-java?tab=readme-ov-file#supported-distributions

jiribenes:
It's very difficult to measure if this helps at all, I'm guesstimating based on only 5 runs.
On a cache hit, this PR makes the CI go from around ~11m40-12m55 to ~11m20-11m50. On a cache miss, it takes **much** longer (15m00), but we shouldn't have a cache miss too often (our sbt dependencies, Java/Scala versions are stable)

Caching MLton makes very little sense (it takes only 1s already), similarly for `apt` dependencies (takes about 10s; what we could do there is to put them into a single command...)

jiribenes:
I _think_ it's worth merging this, even if the speed-up isn't too huge, but I'll leave #428 open as yet unresolved.

b-studios (approved):
Great, thanks!

PR #563, opened 2024-08-30
Move from lift to prompts

b-studios:
This supersedes #456

### DIFF
--- a/effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
@@ -302,9 +302,9 @@ object Transformer {
         emit(callLabel(returnAddress, values.map(transform)))
         RetVoid()
 
-      case machine.NewStack(variable, frame, rest) =>
+      case machine.NewStack(variable, prompt, frame, rest) =>
         emit(Comment(s"newStack ${variable.name}"))
-        emit(Call(variable.name, Ccc(), transform(variable.tpe), newStack, List()));
+        emit(Call(variable.name, Ccc(), transform(variable.tpe), newStack, List(transform(prompt))));
 
         val frameEnvironment = freeVariables(frame).toList;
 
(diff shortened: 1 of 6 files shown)


b-studios:
Current status: the only big thing missing is mutable state. @serkm maybe this branch is now a good starting point for you. How would you implement regions now on top of this? We certainly need to change the representation in machine.

b-studios:
Looks like there is still some memory leak... Otherwise this PR is ready to be discussed.

b-studios:
This PR is now ready for review.

effekt/jvm/src/test/scala/effekt/StdlibTests.scala
```diff
@@ -68,6 +68,7 @@ class StdlibLLVMTests extends StdlibTests {
     examplesDir / "stdlib" / "list" / "zip.effekt",
     examplesDir / "stdlib" / "list" / "deleteat.effekt",
     examplesDir / "stdlib" / "list" / "join.effekt",
+    examplesDir / "stdlib" / "list" / "modifyat.effekt",
```
b-studios, line 71: I am  little concerned by this one..

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -315,7 +334,7 @@ object LLVMRunner extends Runner[String] {
     val executableFile = basePath
     var gccArgs = Seq(gcc, gccMainFile, "-o", executableFile, objPath) ++ libuvArgs
 
-    if (C.config.debug()) gccArgs ++= Seq("-fsanitize=address,leak,undefined", "-fstack-protector-all", "-Og", "-g")
+    if (C.config.debug()) gccArgs ++= Seq("-fsanitize=address,undefined", "-fstack-protector-all", "-Og", "-g")
```
b-studios, line 337: `leak` is not supported on macos (at least on my machine)

phischu (approved):

PR #566, opened 2024-09-02
Use a relative path for loading the JS main

marzipankaiser:
Currently, the generated script for loading the compiled JavaScript uses an absolute path.
This prevents moving the generated files.

This makes it so a relative path is used instead.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Runner.scala
+++ b/effekt/jvm/src/main/scala/effekt/Runner.scala
@@ -147,8 +147,9 @@ object JSNodeRunner extends Runner[String] {
   def build(path: String)(using C: Context): String =
     val out = C.config.outputPath().getAbsolutePath
     val jsFilePath = (out / path).canonicalPath.escape
+    val jsFileName = jsFilePath.split("/").last
     // create "executable" using shebang besides the .js file
-    val jsScript = s"require('${jsFilePath}').main()"
+    val jsScript = s"require('./${jsFileName}').main()"
     os match {
       case OS.POSIX =>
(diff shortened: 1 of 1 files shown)


jiribenes:
Motivation: If you compile some Effekt file like `main.effekt` on the JS backend using `effekt --build --backend js main.effekt`, the resulting file `out/main` is just a thin wrapper containing the following:
```js
#!/usr/bin/env node
require('/a/long/path/to/./out/main.js').main()
```

Unfortunately, this prevents any attempts at packaging the resulting application in a reasonable way since the path is absolute.
Instead, it should perhaps be a relative path like `require('./main.js').main()` and we then trust the packager to bundle the runner script together with the JS file.

jiribenes:
Currently errors out on Windows with:
```
Error: Cannot find module './D:\a\effekt\effekt\out\tests\effekt.javascripttests\sideeffects.js'
Require stack:
- D:\a\effekt\effekt\out\tests\effekt.javascripttests\sideeffects__main.js
  [...]
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    'D:\\a\\effekt\\effekt\\out\\tests\\effekt.javascripttests\\sideeffects__main.js'
  ]
}
```

jiribenes (approved):
LGTM :shipit:

jiribenes:
We also might want to check what happens on the Chez backends wrt this issue 👀

jiribenes:
https://github.com/effekt-lang/effekt/blob/8c1c05780108664095cf0b2147edd67e121d9e49/effekt/jvm/src/main/scala/effekt/Runner.scala#L217-L226

PR #567, opened 2024-09-02
Move benchmark configs from effekt-plots

marvinborner:
This is related to https://github.com/effekt-lang/effekt-plots/pull/4 and should be merged in parallel.


jiribenes (approved):
I can attest that the files `benchmark_config_js` and `benchmark_config_llvm` have been moved from https://github.com/effekt-lang/effekt-plots/pull/4 here.

PR #568, opened 2024-09-03
Fix buggy LLVM tests

marvinborner:
This is an ongoing effort to fix some of the broken LLVM tests.

Already fixed:

- stack smashing in most files (solved by transitioning from CCC to Tailcc explicitly)
- segfaults caused by machine typing bug in records
- also return unit in forgotten excluded tests (see #551)

For better diagnosing and a more reliable RTS overall, I also enabled and fixed GCC warnings.

---

Also, I believe there's a bug in polymorphism boxing:

```scala
interface Eff[A] {
    def eff(a: A): Unit
}

def foo { f: [A](A) => A / Eff[A] }: Unit =
    try {
        f(42)
        ()
    } with Eff[Int] {
        def eff(a) = println(a)
    }

def main() = {
    foo { [T](x: T) =>
        do eff(x)
        x
    }
}
```

This generates a function invocation

``` llvm
musttail call tailcc void %functionPointer_62(%Object %closure_60, i64 %evidenceZero_3523, %Pos %pureApp_3524, %Stack %stack_14)
```

where `functionPointer_62` is actually a reference to

``` llvm
define tailcc void @Eff$capability4_3495_clause_15(%Object %closure, i64 %ev3506_3506, i64 %a7_3500, %Stack %stack) {
```

(note the third argument boxed int `%Pos` vs `i64`)

Maybe @marzipankaiser can help?


effekt/shared/src/main/scala/effekt/generator/llvm/Tree.scala
```diff
@@ -17,6 +17,7 @@ export Definition.*
 enum CallingConvention {
   case Ccc()
   case Tailcc()
+  case TailccPerhaps()
```
marvinborner, line 20: Does anyone have a better idea for calling convention transition from CCC to Tailcc? Here, `TailccPerhaps` is unenforced (not `musttail`) Tailcc, such that it can be used in the initial jump to `main` from the CCC entry.

b-studios: Do I understand correctly that this is ONLY used for the main entry point?

marvinborner: Yes, exactly. I don't have ideas for other solutions though.

marzipankaiser:
After PolymorphismBoxing, the above is:
```
interface Eff2384[A2383] {
  eff2390: (A2383) => Unit374
}

def foo2387(){f2386} =
  try { (){Eff$capability3800} => 
    val v_coe_34833801: BoxedInt283 = f2386[BoxedInt283](boxInt285(42), Eff$capability3800);
    unboxInt287(v_coe_34833801);
    ()
  } with Eff2384[Int378] {
    def eff2390(a2391: Int378){resume2392} = 
      println6(a2391)
  };
def main2388() =
  foo2387({ [T2393](x2394: T2393){Eff$capability3802} => 
    Eff$capability3802.eff(x2394);
    x2394
  })
```
Where the parameter `a` is incorrectly annotated / not transformed.
This would need to be fixed somewhere here, I think: https://github.com/effekt-lang/effekt/blob/458fc9cb0cfbeed87fd98273cc5e23414ad5f0c7/effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala#L226-L240
by making sure that `blockType` is the actual (polymorphic) type of the operation in the interface (which we'd probably need to look up at the definition via a DeclarationContext / the PContext)

effekt/shared/src/main/scala/effekt/generator/llvm/Tree.scala
b-studios: One could also refactor this to 

```
case Tailcc(musttail: Boolean)
```

which is isomorphic, but documents the semantics a bit better than `TailccPerhaps`.

marvinborner:
Remaining bugs:

```
// leaks cons
def main() = [1,2,3].map { a => println(a) }; ()

// array is never freed
def main() = allocate(5000); ()
```

libraries/llvm/array.c
```diff
@@ -13,7 +13,7 @@ void c_array_erase_fields(void *envPtr) {
   uint64_t *sizePtr = envPtr;
   struct Pos *dataPtr = envPtr + sizeof(uint64_t);
   uint64_t size = *sizePtr;
-  for(int i = 0; i < size; i++) {
+  for(uint64_t i = 0; i < size; i++) {
```
b-studios, line 16:
```suggestion
  for (uint64_t i = 0; i < size; i++) {
```

marvinborner:
I think we can merge this for now and work on the remaining issues later.

b-studios:
I guess a few splices are missing, hence the errors (the LLVM backend now also properly splices parameters).

marvinborner:
Only one, I just chose the wrong version during merging.

PR #570, opened 2024-09-03
Also refer to main files and chez using a relative path

marzipankaiser:
This continues the work from #566 to also work for scheme and also use relative paths in the generated bat/sh scripts.
For this to work, in some cases we want to be relative to the directory that contains the script (as opposed to cwd), which we compute dynamically.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Runner.scala
+++ b/effekt/jvm/src/main/scala/effekt/Runner.scala
@@ -42,15 +42,44 @@ trait Runner[Executable] {
   /**
    * Creates a OS-specific script file that will execute the command when executed,
    * forwarding command line arguments.
+   * `$SCRIPT_DIR` refers to the directory the script is in.
+   * 
    * @return the actual name of the generated script (might be `!= name`)
    */
   def createScript(name: String, command: String*): String = os match {
     case OS.POSIX =>
-      IO.createFile(name, s"#!/bin/sh\n${command.mkString(" ")} \"$$@\"", true)
+      val computeScriptDir =
+        """# Determine the directory of the script
+          |if [ -n "$BASH_SOURCE" ]; then
+          |    # Bash
+          |    SCRIPT_DIR=$(dirname "$BASH_SOURCE")
+          |elif [ -n "$ZSH_VERSION" ]; then
+          |    # Zsh
+          |    SCRIPT_DIR=$(dirname "$0")
+          |elif [ -n "$KSH_VERSION" ]; then
+          |    # Ksh
+          |    SCRIPT_DIR=$(dirname "${.sh.file}")
+          |else
+          |    # POSIX shells (sh, dash, etc.)
+          |    SCRIPT_DIR=$(dirname "$0")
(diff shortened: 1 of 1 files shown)


effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -42,15 +42,44 @@ trait Runner[Executable] {
...
+          |    # Zsh
+          |    SCRIPT_DIR=$(dirname "$0")
+          |elif [ -n "$KSH_VERSION" ]; then
+          |    # Ksh
+          |    SCRIPT_DIR=$(dirname "${.sh.file}")
+          |else
+          |    # POSIX shells (sh, dash, etc.)
+          |    SCRIPT_DIR=$(dirname "$0")
+          |fi
+          |""".stripMargin
```
jiribenes, line 66: Do we need this? Can't we just always use `SCRIPT_DIR=$(dirname "$0")`?
(I _really_ don't know enough cross-platform `sh` to be honest)

marzipankaiser: (me neither, so I just kept it 😅)
Probably we don't and you're right 🤔

jiribenes, line 66: We could just try removing it and running the tests -- I don't think we really care about `ksh` support :)
```suggestion
      val computeScriptDir =
        """# Determine the directory of the script
          |SCRIPT_DIR=$(dirname "$0")
          |""".stripMargin
```

jiribenes: A potential downside might be symlinks: `$BASH_SOURCE` should supposedly have the "true" source as opposed to `$0`

marzipankaiser: I first thought "do we care?" - but we might if we want to package an Effekt-application with nix, since it might just symlink the one file (and arguably should only symlink that) 🤔

marzipankaiser: But then, `BASH_SOURCE` is not portable and in this case we might want to use e.g. `realpath`

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -42,15 +42,32 @@ trait Runner[Executable] {
...
+   * `$SCRIPT_DIR` refers to the directory the script is in.
+   * 
    * @return the actual name of the generated script (might be `!= name`)
    */
   def createScript(name: String, command: String*): String = os match {
     case OS.POSIX =>
-      IO.createFile(name, s"#!/bin/sh\n${command.mkString(" ")} \"$$@\"", true)
+      val computeScriptDir =
+        """# Determine the directory of the script
+          |SCRIPT_DIR=$(dirname "$0")
```
marzipankaiser, line 53:
```suggestion
          |SCRIPT_DIR=$(dirname "$(realpath "$0")")
```

PR #575, opened 2024-09-04
Fix PolymorphismBoxing for more complex uses of interfaces and effect handlers

marzipankaiser:
This fixes the program mentioned by @marvinborner in #568.

In order to do this, this also fixes #574 (thats just 257c07a, if we'd want to cherry-pick)

cc @phischu

PR #581, opened 2024-09-10
Fix type arguments when PolymorphismBoxing parametrized handlers

marzipankaiser:
Fixup of #575.


marzipankaiser:
cc @marvinborner - this fixes what you noticed here: https://github.com/effekt-lang/effekt/pull/575#issuecomment-2339203149

PR #584, opened 2024-09-10
Local Variables

serkm:
Allocate local variables directly on the stack, with a special stack frame

### DIFF
--- a/libraries/llvm/rts.ll
+++ b/libraries/llvm/rts.ll
@@ -308,6 +308,57 @@ define ptr @getPointer(%Reference %reference, i64 %index, %Stack %stack) {
     ret ptr %pointer
 }
 
+define %Stack @getStack(%Stack %stack, %Prompt %prompt) {
+    %promptPointer = getelementptr %StackValue, %Stack %stack, i64 0, i32 3
+    %thisPrompt = load %Prompt, ptr %promptPointer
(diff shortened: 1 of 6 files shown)


libraries/llvm/rts.ll
```diff
@@ -308,6 +308,57 @@ define ptr @getPointer(%Reference %reference, i64 %index, %Stack %stack) {
...
+    %base = load %Base, ptr %basePointer
+    %varPointer = getelementptr i8, %Base %base, i64 %offset
+    ret ptr %varPointer
+}
+
+define %Reference @newReference(%Stack %stack) alwaysinline {
+    %stackPointerPointer = getelementptr %StackValue, %Stack %stack, i64 0, i32 1, i32 0
+    %basePointer = getelementptr %StackValue, %Stack %stack, i64 0, i32 1, i32 1
+
+    %stackPointer = load %StackPointer, ptr %stackPointerPointer
```
b-studios, line 344: Nitpick, we now try to use the convention `Foo_pointer`. So this will be `stackPointer_pointer`

b-studios:
Great job! Do you have results on running benchmarks on this?

phischu:
Could you rebase onto current master? I can not reproduce the bad benchmarks, we should discuss offline.

b-studios (approved):
LGTM :)

PR #585, opened 2024-09-10
Mutable dynamic arrays, heaps, and simple union find data structures

marzipankaiser:
For writing examples, I wanted to have some standard mutable data structures, currently:
- dynamic/resizable arrays
- min-heap (=> priority queue)
- union-find on integers

They all still need more documentation, testing and maybe some additional functions (I only implemented what I needed).

### DIFF
--- a/examples/stdlib/resizable_array/resizable_array.effekt
+++ b/examples/stdlib/resizable_array/resizable_array.effekt
@@ -1,0 +1,33 @@
+import resizable_array
+import test
+
+def main() = {
+  suite("ResizableArrayTests", false) {
+    test("usage as stack") {
+      with on[OutOfBounds].default { assertTrue(false, "out of bounds") }
+      val a = resizableArray()
+      a.add(1)
+      a.add(1)
+      a.add(2)
+      a.add(3)
+      a.add(13)
+      a.add(21)
+      a.add(34)
+      a.add(55)
+      assert(a.popRight(), 55)
+      assert(a.popRight(), 34)
+      assert(a.popRight(), 21)
+      assert(a.popRight(), 13)
+      a.add(5)
+      a.add(8)
+      assert(a.popRight(), 8)
+      assert(a.popRight(), 5)
+      assert(a.popRight(), 3)
+      assert(a.popRight(), 2)
+      assert(a.popRight(), 1)
+      assert(a.popRight(), 1)
+    }
+    // TODO
+  };
+  ()
+}
--- a/examples/stdlib/resizable_array/resizable_array.effekt
+++ b/examples/stdlib/resizable_array/resizable_array.effekt
@@ -27,7 +27,6 @@ def main() = {
       assert(a.popRight(), 1)
       assert(a.popRight(), 1)
     }
-    // TODO
   };
   ()
 }
--- a/libraries/common/heap.effekt
+++ b/libraries/common/heap.effekt
@@ -1,15 +1,15 @@
 module heap
 import resizable_array
 
-// Resizable 2-ary min-heap, backed by a resizable array
-// `cmp` defines the ordering of elements
+/// Resizable 2-ary min-heap, backed by a resizable array
+/// `cmp` defines the ordering of elements
 record Heap[T](rawContents: ResizableArray[T], cmp: (T, T) => Ordering at {})
 
-// Make a new Heap with the given comparison operation
+/// Make a new Heap with the given comparison operation
 def heap[T](cmp: (T,T) => Ordering at {}) = 
   Heap[T](resizableArray(), cmp)
 
-// Make a new Heap with the given comparison operation and initial capacity
+/// Make a new Heap with the given comparison operation and initial capacity
 def heap[T](cmp: (T,T) => Ordering at {}, capacity: Int) = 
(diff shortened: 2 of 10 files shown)


libraries/common/union_find.effekt
```diff
@@ -0,0 +1,133 @@
...
+// Make it so a and b are in the same set
+//
+// O(inverse-ackermann(n)) amortized
+def union(u: UnionFind, a: Int, b: Int): Int / Exception[MissingValue] = {
+  val x = u.find(a)
+  val y = u.find(b)
+  u.unionRoots(x, y)
+}
+
+def main() = {
```
jiribenes, line 116: See above, `examples` namespace?

jiribenes (commented):
Looks good so far!
I had some time during the meeting, so I quickly wrote some observations. :)

libraries/common/resizable_array.effekt
```diff
@@ -0,0 +1,266 @@
+module resizable_array
+
+import ref
+import array
+import test
+
+record ResizableArray[T](size_: Ref[Int], rawContentPtr: Ref[Array[T]])
```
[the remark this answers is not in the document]
marzipankaiser: Yeah, me neither - that would probably work 🤔

b-studios:
What happened to this PR, it looks really cool! :)

marzipankaiser:
Rebased to check if the Valgrind-issues happen to be fixed by now.

examples/stdlib/resizable_array/resizable_array.effekt
```diff
@@ -0,0 +1,33 @@
...
+      a.add(5)
+      a.add(8)
+      assert(a.popRight(), 8)
+      assert(a.popRight(), 5)
+      assert(a.popRight(), 3)
+      assert(a.popRight(), 2)
+      assert(a.popRight(), 1)
+      assert(a.popRight(), 1)
+    }
+    // TODO
```
marzipankaiser, line 30:
```suggestion
```

I won't write more tests atm unless someone asks me to, so I'll remove that TODO.

libraries/common/heap.effekt
```diff
@@ -0,0 +1,103 @@
...
+///
+/// O(1)
+def findMin[T](heap: Heap[T]): T / Exception[OutOfBounds] = {
+  heap.rawContents.get(0)
+}
+
+/// find and remove the minimal element in this heap
+///
+/// O(log n)
+def deleteMin[T](heap: Heap[T]): T / Exception[OutOfBounds] = {
```
b-studios, line 91: Bike shed, in the comment you write "remove" but the function is called `deleteMin`. Some languages also call this operation `pop` (and `insert` then `push`):

https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html

libraries/common/resizable_array.effekt
```diff
@@ -0,0 +1,234 @@
...
+    val newRaw = array::allocate(capacity)
+    oldRaw.copy(0, newRaw, 0, arr.size)
+    arr.rawContentPtr.set(newRaw)
+  }
+}
+
+/// Change the dynamic to have exactly the given capacity
+///
+/// O(n)
+def setCapacity[T](arr: ResizableArray[T], capacity: Int): Unit / Exception[OutOfBounds] = {
```
b-studios, line 108: What about `resize`? If the type is called resizable array?

b-studios (approved):

libraries/common/resizable_array.effekt
marzipankaiser: I feel like `resize` might be interpreted as setting the size of the array (and filling with `0`s, e.g.), whereas this just resizes the backing array, not the data structure (it's almost a no-op when looking at it as an abstract data structure).

marzipankaiser: We could implement `resize` (as in "make length k and fill"), too, of course.

b-studios: I am also fine with keeping as is. Maybe the reasoning can be explained a bit further in the comment to help users?

libraries/common/heap.effekt
marzipankaiser: I chose the names based on what https://en.wikipedia.org/wiki/Binary_heap calls the operations, but we can rename them - I don't have a strong opinion here.

libraries/common/resizable_array.effekt
```diff
@@ -0,0 +1,234 @@
...
+  with on[OutOfBounds].panic
+  val oldRaw = arr.rawContentPtr.get
+  if (oldRaw.size != capacity) { 
+    val newRaw = array::allocate(capacity)
+    oldRaw.copy(0, newRaw, 0, arr.size)
+    arr.rawContentPtr.set(newRaw)
+  }
+}
+
+/// Change the dynamic to have exactly the given capacity
```
marzipankaiser, line 105:
```suggestion
/// Change the resizable array to have exactly the given capacity.
/// This only changes the size of the backing array, not the `size`.
```
Maybe this helps?

jiribenes:
🪤 It could be fun to take the unification benchmark from `other/unify` and try it with the union find from this PR.

PR #587, opened 2024-09-11
Use splices in LLVM templates

b-studios:
Note that we only splice the names of variables (e.g., `%x17`) and not their type (e.g., `%Pos %x17`). This is mostly due to instructions like

```
 %z = icmp sge %Int %x, %y
 ```
 
 where generating
 
 ```
  %z = icmp sge %Int %x, %Int %y
  ```
  would be wrong.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
@@ -46,15 +46,12 @@ object Transformer {
     case core.Extern.Def(name, tps, cparams, vparams, bparams, ret, capture, body) =>
       if (bparams.nonEmpty) then ErrorReporter.abort("Foreign functions currently cannot take block arguments.")
 
-      val transformedParams = vparams.map {
-        //  we do not use transform(ValueParam) here because we use the original name (e.g. %x and not %x1234)
-        case core.ValueParam(id, tpe) => Variable(id.name.name, transform(tpe))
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by b-studios** [comment]
@phischu I am aware that this will result in merge conflicts with #553, but at some point we need to bite the bullet.

File: effekt/shared/src/main/scala/effekt/machine/Transformer.scala [hunk 1]
```diff
@@ -46,15 +46,12 @@ object Transformer {
...
-        case core.ValueParam(id, tpe) => Variable(id.name.name, transform(tpe))
-      }
-      noteDefinition(name, transformedParams, Nil) // TODO maybe use vparams.map(transform) here
+      val transformedParams = vparams.map(transform)
+      noteDefinition(name, transformedParams, Nil)
       val tBody = body match {
         case core.ExternBody.StringExternBody(ff, Template(strings, args)) =>
           ExternBody.StringExternBody(ff, Template(strings, args map {
-            case core.ValueVar(id, tpe) => Variable(id.name.name, transform(tpe))
+            case x @ core.ValueVar(id, tpe) => Variable(transform(id), transform(tpe))
```
phischu, line 54:
```suggestion
            case core.ValueVar(id, tpe) => Variable(transform(id), transform(tpe))
```

phischu (approved):

PR #589, opened 2024-09-11
JS: Strip trailing/leading whitespace in externs (+ warn)

### LINKED ISSUE #569: Multiline control externs broken by JS semicolon insertion
When the user defines a JS extern with control effects like so:
```effekt
extern control def foo(): Int = 
  js """
     $effekt.pure(2)
  """
```
Then the call fails at runtime with an error like:
```
    foo_2839().then((v_r_2852_3977) =>
              ^

TypeError: Cannot read properties of undefined (reading 'then')
    at Object.head (/Users/gaisseml/dev/effekt/experiments/out/js_multiline_externs.js:828:15)
    at apply (/Users/gaisseml/dev/effekt/experiments/out/js_multiline_externs.js:459:25)
    at Object.apply (/Users/gaisseml/dev/effekt/experiments/out/js_multiline_externs.js:549:34)
    at trampoline (/Users/gaisseml/dev/effekt/experiments/out/js_multiline_externs.js:435:19)
    at Object.run (/Users/gaisseml/dev/effekt/experiments/out/js_multiline_externs.js:542:18)
    at Object.main (/Users/gaisseml/dev/effekt/experiments/out/js_multiline_externs.js:836:27)
    at Object.<anonymous> (/Users/gaisseml/dev/effekt/experiments/out/js_multiline_externs:2:79)
    at Module._compile (node:internal/modules/cjs/loader:1480:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1564:10)
    at Module.load (node:internal/modules/cjs/loader:1287:32)

Node.js v22.1.0
```

## Expected behaviour
This should return 2.

## Why does this happen
The above definition gets translated to the following JS:
```js
function foo_2839() {
  return 
       $effekt.pure(2)
    ;
}
```
Semicolon insertion will insert a `;` after the return, so this is equivalent to something like:
```js
function foo_2839() {
  return undefined; // early return
  $effekt.pure(2); // dead code
  ;
}
```

## Full example code
```effekt
extern control def foo(): Int = 
  js """
     $effekt.pure(2)
  """

extern control def bar(): Int = 
  js """$effekt.pure(1)"""


def main() = {
  println(bar())
  println(foo())
}
```

marzipankaiser:
Fixes #569.

This basically applies this suggestion: https://github.com/effekt-lang/effekt/issues/569#issuecomment-2326363108
but generates a warning instead of an error.

We do not seem to use this anywhere.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/generator/js/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/js/Transformer.scala
@@ -134,6 +134,17 @@ trait Transformer {
       }
     }.toString()
 
+  // Externs
+  // -------
+  def normalizeExternStrings(l: List[String])(using Context): List[String] = {
+    val start = l.head
+    val end = l.last
+    val mid = l.tail.init
+    if (start.matches("""^\s.*""") || end.matches(""".*\s$""")) {
+      Context.warning("Extern string in js has trailing / leading whitespace. This will be removed.")
+    }
+    start.stripLeading() +: mid :+ end.stripTrailing()
+  }
 
   // Separate Compilation (Website)
   // ------------------------------
--- a/effekt/shared/src/main/scala/effekt/generator/js/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/js/Transformer.scala
@@ -136,14 +136,21 @@ trait Transformer {
 
   // Externs
   // -------
-  def normalizeExternStrings(l: List[String])(using Context): List[String] = {
-    val start = l.head
-    val end = l.last
-    val mid = l.tail.init
-    if (start.matches("""^\s.*""") || end.matches(""".*\s$""")) {
-      Context.warning("Extern string in js has trailing / leading whitespace. This will be removed.")
-    }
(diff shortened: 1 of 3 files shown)


effekt/shared/src/main/scala/effekt/generator/js/Transformer.scala
```diff
@@ -134,6 +134,17 @@ trait Transformer {
...
     }.toString()
 
+  // Externs
+  // -------
+  def normalizeExternStrings(l: List[String])(using Context): List[String] = {
+    val start = l.head
+    val end = l.last
+    val mid = l.tail.init
+    if (start.matches("""^\s.*""") || end.matches(""".*\s$""")) {
+      Context.warning("Extern string in js has trailing / leading whitespace. This will be removed.")
```
b-studios, line 144: I think this warning can be very annoying, no?

marzipankaiser: It could be - we're not using it anywhere yet, so...
I can also remove it.

PR #592, opened 2024-09-16
Enhance Regex Support in Standard Library for JS Backend

leonfuss:
This pull request significantly improves regex support in Effekt's standard library for the JavaScript backend. It introduces capture groups, group indices, and regex flags, addressing current limitations and providing a more powerful regex interface.

**Key Enhancements:**
1. Support for capture groups and their indices
2. Implementation of regex flags
3. Stateful regex object for sequential matching

**Current Limitations:**
Effekt's regex support is currently limited, lacking support for capture groups and advanced regex features. This implementation bridges this gap for JavaScript environments.

**New Features:**
- Capture Groups: Ability to use and access capture groups in regex patterns
- Group Indices: Retrieve the starting index of each capture group
- Regex Flags: Support for standard regex flags (e.g., 'i' for case-insensitive matching)
- Stateful Matching: Global flag ('g') set by default for sequential matching

**Backwards compatibility**
This change is not backwards compatible.

I'm opening this as a draft as I'm not an experienced Effekt developer. 
So feedback on the API design and usability from experienced Effekt developers is highly appreciated.

### DIFF
--- a/libraries/common/option.effekt
+++ b/libraries/common/option.effekt
@@ -49,6 +49,19 @@ def option[A, E](proxy: on[E]) { p: => A / Exception[E] }: Option[A] =
   try { Some(p()) }
   with Exception[E] { def raise(_, _) = None() }
 
+// zip two options with a transform to one. Only (Some[A] Some[B]) will result
+// in Some[C]
+def zip[A, B, C](a: Option[A], b: Option[B]) {transform: (A,B) => C}: Option[C] = {
+    a match {
+        case None() => None()
+        case Some(x) =>
+            b match {
+                case None() => None()
+                case Some(y) => Some(transform(x,y))
+            }
+    }
+}
+
 // Interop with FFI
 // ---------------
 // These functions should only be used by libraries that interface with FFI
--- a/libraries/js/text/regex.effekt
+++ b/libraries/js/text/regex.effekt
@@ -1,33 +1,186 @@
 module text/regex
 
 import string
+import option
+import list
+
 
 extern type Regex
 
-record Match(matched: String, index: Int)
+// matches and groups
+record Range(start: Int, end: Int)
+record Group(content: String, index: Option[Range])
+record Match(content: String, start: Int, end: Int,  groups: List[Group])
 
-extern pure def regex(str: String): Regex =
-  js "new RegExp(${str})"
+/**
+ * Constructs a Regex object from a pattern string.
+ *
+ * This function creates a Regex object with the global flag set by default.
+ * The resulting Regex object is stateful, meaning its internal state
+ * changes after each match when used with methods like `exec`.
+ *
+ * @param pattern The regular expression pattern as a string.
+ * @return A Regex object that can be used with the `exec` function.
+ *
+ * Note: To extract substring indices for capturing groups, the 'd' flag
+ * must be explicitly set using the overloaded `regex(pattern: String, flags: String)` function.
+ * Without the 'd' flag, indices of capture groups will always be `None`.
+ *
+ * Example usage:
+ * ```effekt
+ * val simpleRegex = regex("\\w+")
+ * val math = exec(simpleRegex, "Hello World")
+ * ```
+ *
+ * For more control over flags, use the overloaded version:
+ * ```effekt
+ * val regexWithIndices = regex("(\\w+)", "gd")
+ * ```
+ */
+def regex(pattern: String): Regex = regex(pattern, "g")
 
-def exec(reg: Regex, str: String): Option[Match] =
-   undefinedToOption(reg.unsafeExec(str)) match {
-    case None() => None()
-    case Some(v) => Some(Match(v.matched, v.index))
-  }
 
-// internal representation { matched: String, index: Int }
+/**
+ * Constructs a Regex object from a pattern string and optional flags.
+ *
+ * @param pattern The regular expression pattern as a string.
+ * @param flags A string containing any combination of the following flags:
+ *   - `g`: Global search (find all matches rather than stopping after the first match)
+ *   - `i`: Case-insensitive search
+ *   - `m`: Multi-line search (^ and $ match start/end of each line)
+ *   - `d`: Generate indices for substring matches
+ *
+ * @return A Regex object that can be used with the `exec` function.
+ *
+ * Example usage:
+ * ```effekt
+ * val simpleRegex = regex("\\w+")
+ * val complexRegex = regex("(\\w+)\\s(\\d+)", "gi")
+ * ```
+ *
+ * Note: The `d` flag is required for extracting substring indices of groups.
+ * If not present, the indices of capture groups will always be `None`.
+ */
+def regex(pattern: String, flags: String): Regex = {
+    extRegex(pattern, flags)
+}
+
+extern pure def extRegex(str: String): Regex = js "new RegExp(${str})"
+extern pure def extRegex(str: String, flags: String): Regex = js "new RegExp(${str}, ${flags})"
+
+/**
+ * Executes the regex against the given string.
+ *
+ * @param reg The Regex object to execute.
+ * @param str The string to match against.
+ * @return Option[Match] A Some(Match) if a match is found, None otherwise.
+ *
+ * The returned Match object contains:
+ * - content: The full matched string.
+ * - start: The starting index of the match.
+ * - end: The ending index of the match.
+ * - groups: A list of Group objects for each capturing group.
+ *
+ * Each Group object contains:
+ * - content: The matched substring for this group.
+ * - index: An Option[Range] with start and end indices if the 'd' flag is set, None otherwise.
+ *
+ * Note: This function is stateful when the global flag 'g' is set on the Regex object.
+ * Subsequent calls with the same Regex and string will return the next match in the string.
+ *
+ * Example usage:
+ * ```effekt
+ * val rx = regex("(\\w+)\\s(\\d+)", "g")
+ * val text = "Hello 123 World 456"
+ * while (true) {
+ *   exec(rx, text) match = {
+ *      Some(m) =>
+ *          println("content: " ++ m.content ++ "range: " ++ m.start.show() ++ ", " ++ m.end.show()))
+ *      None() => break
+ * }
+ * ```
+ */
+def exec(reg: Regex, str: String): Option[Match] = {
+    def toGroup(g: GroupMatch): Group = Group(
+        content(g),
+        zip(undefinedToOption(start(g)), undefinedToOption(end(g))) { (s, e) => Range(s, e)}
+    )
+
+    undefinedToOption(reg.unsafeExec(str)) match {
+        case None() => None()
+        case Some(v) =>
+        val fullMatch = fullMatch(v)
+        Some(Match(
+            fullMatch.content,
+            fullMatch.start,
+            fullMatch.end,
+            v.subGroups.toList().map{toGroup}
+        ))
+    }
+}
+
+// internal representation
+extern type GroupMatch
+extern pure def content(g: GroupMatch): String = js "${g}.content"
+extern pure def start(g: GroupMatch): Int = js "${g}.start"
+extern pure def end(g: GroupMatch): Int = js "${g}.end"
+
 extern type RegexMatch
-extern pure def matched(r: RegexMatch): String = js "${r}.matched"
-extern pure def index(r: RegexMatch): Int = js "${r}.index"
+extern pure def fullMatch(r: RegexMatch): GroupMatch = js "${r}.fullMatch"
+extern pure def subGroups(r: RegexMatch): Array[GroupMatch] = js "${r}.groups"
 
 extern js """
 function regex$exec(reg, str) {
-  var res = reg.exec(str);
-  if (res === null) { return undefined }
-  else { return { matched: res[0], index: res.index } }
+  var match = reg.exec(str);
+  if (match === null)  return undefined
+
+  var fullMatch = { content: match[0], start: match.index, end: reg.lastIndex };
+  var groups = [];
+  for (var i = 1; i < match.length; i++) {
+    if (match[i] !== undefined) {
+        if (match.indices !== undefined && match.indices[i] !== undefined) {
+            groups.push({
+                content: match[i],
+                start: match.indices[i][0],
+                end: match.indices[i][1]
+            });
+        } else {
+            groups.push({
+                content: match[i],
+            });
+        }
+    }else {
+        // Push an empty string if an intermediate group is undefined
+        groups.push({content: ""});}
+  }
+
+  return { fullMatch: fullMatch, groups: groups };
 }
+
 """
 
 // internals
 extern io def unsafeExec(reg: Regex, str: String): RegexMatch =
   js "regex$exec(${reg}, ${str})"
+
+
+// Debug show
+def debug(m: Match): String = {
+    val start = m.start.show()
+    val end = m.end.show()
+    val groups = m.groups.map{it => it.debug()}.join("\n")
+
+    "matched: " ++ m.content ++
+    "\nfrom: (" ++ start ++ ", " ++ end ++ ")" ++
+    "\nGroups: \n" ++ groups
+}
+
+def debug(g: Group): String = {
+    val range = g.index.map{ it => it.debug() }.getOrElse{"None"}
+
+    g.content ++  "\nfrom: " ++ range
+}
+
+def debug(r: Range): String = {
+    "(" ++ r.start.show() ++ ", " ++ r.end.show() ++ ")"
+}


libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,186 @@
...
+        case None() => None()
+        case Some(v) =>
+        val fullMatch = fullMatch(v)
+        Some(Match(
+            fullMatch.content,
+            fullMatch.start,
+            fullMatch.end,
+            v.subGroups.toList().map{toGroup}
+        ))
+    }
```
jiribenes, line 119: Nitpick: since you were asking about style, I'd personally ANF the scrutinee out and then use `.map` on the result. I also like UFCS perhaps a bit too much. (suggestion untested)
```suggestion
    val regexMatch: Option[RegexMatch] = reg.unsafeExec(str).undefinedToOption
    regexMatch.map { v =>
      val groupMatch = v.fullMatch
      Match(
        groupMatch.content,
        groupMatch.start,
        groupMatch.end,
        v.subGroups.toList.map{toGroup}
      )
    }
```

libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,186 @@
...
+
+def debug(g: Group): String = {
+    val range = g.index.map{ it => it.debug() }.getOrElse{"None"}
+
+    g.content ++  "\nfrom: " ++ range
+}
+
+def debug(r: Range): String = {
+    "(" ++ r.start.show() ++ ", " ++ r.end.show() ++ ")"
+}
```
jiribenes, line 186: I think you could even use `show` here instead of debug :)
We currently use it for a more debug-ey output anyways.

By the way, you should be able to do `["matched: " ++ m.content, "from: (" ++ start ...].join("\n")` or something similar via [`string::join`](https://github.com/effekt-lang/effekt/blob/1a1d34199e2edce3a4e7b5ea70209feb5d618cf7/libraries/common/string.effekt#L97-L106)

libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,186 @@
...
 
 extern js """
 function regex$exec(reg, str) {
-  var res = reg.exec(str);
-  if (res === null) { return undefined }
-  else { return { matched: res[0], index: res.index } }
+  var match = reg.exec(str);
+  if (match === null)  return undefined
+
+  var fullMatch = { content: match[0], start: match.index, end: reg.lastIndex };
```
jiribenes, line 137: Could these be `const`?

jiribenes (commented):
Thanks for the contribution! :)
I can't review it fully, but I have a few nitpicks I pointed out in the review: most of them are just stylistic.

What I would really like to see before merging are a few relevant tests in [`effekt/examples/stdlib/`](https://github.com/effekt-lang/effekt/tree/master/examples/stdlib), feel free to make a new folder.
Our tests are always regression tests, so there's an `.effekt` file and a corresponding `.check` file. You can either just print out the results (see [here](https://github.com/effekt-lang/effekt/blob/master/examples/stdlib/list/all.effekt)) or use the test suite written in Effekt (see [here](https://github.com/effekt-lang/effekt/blob/master/examples/stdlib/test/test.effekt)).

jiribenes:
The tests are failing since [`examples/casestudies/lexer.effekt.md`](https://github.com/effekt-lang/effekt/blob/master/examples/casestudies/lexer.effekt.md) uses `text/regex`.
Here are some more modules importing `text/regex`: https://github.com/search?q=repo%3Aeffekt-lang%2Feffekt+%22text%2Fregex%22+path%3Aexamples%2F**%2F*.effekt&type=code

leonfuss:
> The tests are failing since [`examples/casestudies/lexer.effekt.md`](https://github.com/effekt-lang/effekt/blob/master/examples/casestudies/lexer.effekt.md) uses `text/regex`. Here are some more modules importing `text/regex`: https://github.com/search?q=repo%3Aeffekt-lang%2Feffekt+%22text%2Fregex%22+path%3Aexamples%2F**%2F*.effekt&type=code

I refactored the lexer example to account for the changes. The other occurrences shouldn't be affected since I'm only changing the JS backend? Please correct me if I oversaw anything.

leonfuss:
Thank you for your thorough review. I've implemented your suggestions, making the code more idiomatic. The addition of typed flags should notably enhance usability. I plan to add the test cases tomorrow.

libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,221 @@
...
+import list
+
 
 extern type Regex
 
-record Match(matched: String, index: Int)
+// matches and groups
+record Range(start: Int, end: Int)
+record Group(content: String, index: Option[Range])
+record Match(content: String, start: Int, end: Int,  groups: List[Group])
```
jiribenes, line 13: 👀 Renaming the first field of `Match` to `content` is a change that ripples through the other backends.
For example in the Chez backends, the `Match` type is unchanged, which triggers the test failures you're seeing:
https://github.com/effekt-lang/effekt/blob/cd18ad896b9f1c856a9d6002fd0a5733e5cb9a15/libraries/chez/common/text/regex.effekt#L9

(and the same API is probably [?] assumed for the ML backend:)
https://github.com/effekt-lang/effekt/blob/cd18ad896b9f1c856a9d6002fd0a5733e5cb9a15/libraries/ml/text/regex.effekt#L36

WDYT is the best step forward? I can see two straightforward ways here:
- rename the `matched` field in all backends to `content`
- rename the `content` field(s) here to `matched` to align with the other backends

(I'm fine with both of these solutions, I think)

b-studios: _Side note: Let me note again that ultimately we will drop Chez and ML._

leonfuss: I'll go with the second option :) Matched is still expressive enough provide a good user experience

jiribenes (dismissed):
I went through the PR again and LGTM :shipit:
Thank you for the contribution :)

jiribenes:
I'll take a look. This error usually means that UFCS (`foo.bar(baz)` standing for `bar(foo, baz)`) failed.

examples/stdlib/string/regex.effekt
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+def main() = {
+  val src = "ab cd efg hij k l m n"
+
+  // stateless regex
+  val rgx1 = regex("\\w+ \\w+")
+  rgx1.exec(src).printMatched()
+  rgx1.exec(src).m2.printMatched()
```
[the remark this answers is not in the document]
jiribenes: Darn, you were quicker :)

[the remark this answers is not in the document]
jiribenes: anyways:
```suggestion
  rgx1.exec(src).printMatched()
```

leonfuss:
Is there a way to run certain tests only for the JS-Backend? I would like to add test for the new features, but it seems that the chez-backend is not happy with that. At least I have no trouble running the test succesful on my machine using only the JS-Backend.

jiribenes:
You can add the file into `ignored` here for the backends it should not run on together with a comment why it's disabled: https://github.com/effekt-lang/effekt/blob/master/effekt/jvm/src/test/scala/effekt/StdlibTests.scala

So for example for Chez-$something backends:
```scala
abstract class StdlibChezTests extends StdlibTests {
  override def ignored: List[File] = List(
    // Not implemented yet
    examplesDir / "stdlib" / "bytes",
    examplesDir / "stdlib" / "io",

    // Not implemented: advanced regex features
    examplesDir / "stdlib" / "string" / "regex.effekt",
  )
}
```

jiribenes:
> It looks like this test even fails on JS:
> 
> https://github.com/effekt-lang/effekt/actions/runs/10919082132/job/30306075187?pr=592#step:11:402

It seems to work for me locally so it might be a NodeJS version issue?
Did the API for regexes change there?

jiribenes:
Don't know how to resolve this if we want to keep the functionality. We could update NodeJS: that might be a wise move anyway since even the extended support for NodeJS 12.x ended 2 years and 4 months ago (30 Apr 2022). Current oldest somewhat supported LTS version is NodeJS 18.x (where security support ends in 7 months).

leonfuss:
> EDIT 2: I _think_ that NodeJS 12.x does not support the `d` (`GenerateIndices()`) flag since it was only introduced in ES9 and implemented in NodeJS 16.x according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/hasIndices#browser_compatibility).

I just tested it locally and can confirm that. I will try to add a condition that the `d` flag will only works if the node version is   >16.x and change the test to only run if node is >16.x

Do you see any problems with that?

libraries/js/text/regex.effekt
```diff
@@ -184,8 +193,20 @@ namespace internal {
...
 }
 
+def regexIndicesSupported?(): Bool = internal::nodeMajorVersion() >= 16
+
+extern js """
+  function process$version() {
+    const v = process.version
+    const parts = v.split('.')
+    return parseInt(parts[0].replace('v', ''))
+  }
```
jiribenes, line 207: As a tiny nitpick: according to StackOverflow, `process.versions.node` should have the version without `v` :)
https://stackoverflow.com/questions/6656324/check-for-current-node-version

b-studios:
We can just require a newer node version, that shouldn't be a problem.

b-studios:
Thanks for the PR! I think it is great to have improved regex support in the language. 

I have one major concern, which is the compatibility with the LLVM backend. In general, I do not care so much anymore about the Chez and ML backends. However, we should aim for feature completeness of the LLVM backend. 

Adding JS-only features now will make this more and more difficult in the future.

@phischu what is your opinion when viewing this from the LLVM perspective?

b-studios:
Which of the extensions that are proposed in this PR are compatible with the regular expressions in POSIX? 

(https://www.man7.org/linux/man-pages/man3/regex.3.html)

jiribenes:
`master` is updated to NodeJS 16.x, you can just revert the last commit and rebase :)

b-studios:
To clarify: I am not proposing that you should implement the LLVM part. I just want to know what the intersection between posix regex and your proposed features is.

libraries/common/option.effekt
```diff
@@ -49,6 +49,15 @@ def option[A, E](proxy: on[E]) { p: => A / Exception[E] }: Option[A] =
   try { Some(p()) }
   with Exception[E] { def raise(_, _) = None() }
 
+/// Applies a function `f` to two `Option` values:
+/// returning `None` if either of the arguments is `None`,
+/// otherwise returning `Some` of the result.
```
b-studios, line 54:
```suggestion
/**
 * Applies a function `f` to two `Option` values:
 * returning `None` if either of the arguments is `None`,
 * otherwise returning `Some` of the result.
 **/
```

libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,242 @@
...
+import list
+
 
 extern type Regex
 
-record Match(matched: String, index: Int)
+// matches and groups
+record Range(start: Int, end: Int)
+record Group(matched: String, index: Option[Range])
+record Match(matched: String, start: Int, end: Int,  groups: List[Group])
```
b-studios, line 13:
```suggestion
record Match(matched: String, start: Int, end: Int, groups: List[Group])
```

libraries/common/option.effekt
jiribenes: This one is my fault since it's from my `suggestion`: if we're going to phase out block comments, it might be better to go full "`///` means doc-comment". But yes, it's technically not consistent with the rest of the style.
I'd have also expected some nitpicking: is this `zip`, `map2`, `zipWith`...?

libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,242 @@
...
+import option
+import list
+
 
 extern type Regex
 
-record Match(matched: String, index: Int)
+// matches and groups
+record Range(start: Int, end: Int)
+record Group(matched: String, index: Option[Range])
```
b-studios, line 12: Can we just always generate indices and turn this into `index: Range`?

libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,242 @@
...
+          matched: match[i],
+          start: match.indices[i][0],
+          end: match.indices[i][1]
+        });
+      } else {
+        groups.push({
+          matched: match[i],
+        });
+      }
+    }else {
```
b-studios, line 181:
```suggestion
    } else {
```

libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,242 @@
...
+  return { fullMatch: fullMatch, groups: groups };
+}
+
+"""
+
+// internal calls
+namespace internal {
+  extern io def unsafeExec(reg: Regex, str: String): RegexMatch = js "regex$exec(${reg}, ${str})"
+
+  extern pure def unsafeRegex(str: String, flags: String): Regex = js "new RegExp(${str}, ${flags})"
```
b-studios, line 195: This needs to be `io`, because otherwise the same regex will potentially be inlined over and over again and not have the "stateful" behavior.

libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,242 @@
...
+ *
+ * @param reg The Regex object to execute.
+ * @param str The string to match against.
+ * @return Option[Match] A Some(Match) if a match is found, None otherwise.
+ *
+ * The returned Match object contains:
+ * - matched: The full matched string.
+ * - start: The starting index of the match.
+ * - end: The ending index of the match if `Global` flag is set, otherwise 0
+ * - groups: A list of Group objects for each capturing group.
```
b-studios, line 112: Maybe move this documentation to `Match` and refer to it here?

Also, why is end `0` if global isn' set?

libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,242 @@
...
+  [
+    "matched: " ++ m.matched,
+    "from: (" ++ start ++ ", " ++ end ++ ")",
+    "Groups:",
+    groups
+  ].join("\n")
+}
+
+def show(g: Group): String = {
+  val range = g.index.show{it => show(it)}
```
b-studios, line 226:
```suggestion
  val range = g.index.show{ it => show(it) }
```

libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,242 @@
...
+ *
+ * Example usage:
+ * ```effekt
+ * val rx = regex("(\\w+)\\s(\\d+)", "g")
+ * val text = "Hello 123 World 456"
+ * while (true) {
+ *  exec(rx, text) match = {
+ *    Some(m) =>
+ *      println("matched: " ++ m.matched ++ "range: " ++ m.start.show() ++ ", " ++ m.end.show()))
+ *    None() => break
```
b-studios, line 129: This code probably doesn't work, right? Could you add this to the tests (verbatim) to make sure it is correct?

libraries/js/text/regex.effekt
```diff
@@ -1,33 +1,242 @@
...
+ *
+ * Note: GenerateIndices is required for capture group indexing.
+ * Without it, capture group indices will be None.
+ */
+type RegexFlags {
+  Global();
+  IgnoreCase();
+  Multiline();
+  GenerateIndices();
+}
```
b-studios, line 38: Out of curiosity: 

1. is it possible to implement global matching in terms of local one (starting from a given index) _within_ Effekt?
2. is it possible to implement multiline matching _within_ Effekt by preprocessing the regular expression string? 
3. Can we always require `GenerateIndices` and drop it here?

leonfuss:
> Which of the extensions that are proposed in this PR are compatible with the regular expressions in POSIX?
> 
> (https://www.man7.org/linux/man-pages/man3/regex.3.html)

The POSIX standard for Regex does only include basic support for capture groups. Only the extraction of the content is possible here, but generating indices for those is not. Stateful matching is definitely not known in the POSIX world. 

Flags are a bit more difficult in compatibility. Usually flags are passed as additional command line arguments which makes it dependent on the used program if they are compatible or not. 

Does that answer the question?

leonfuss:
> u can just revert the last commit and rebase

Thank you :)

libraries/js/text/regex.effekt
leonfuss: That's definitely possible and should decrease complexity

libraries/js/text/regex.effekt
leonfuss: 1. Is certainly possible and would possible allow for shared logic across different backend. 
2. multiline should also be possible to implement within Effekt even though I'm not sure how that would impact performance. In my understanding most implementations use simd acceleration for the line ending search or embed a newline stopping criteria directly in the regex engine. 
3. Yes we can.

leonfuss:
I have implemented the following changes as suggested previously:

1. The Global Flag is now handled directly. The `exec` function returns a Regex object that captures the string in the `capture` field. This allows subsequent calls on the returned object to iterate (lazily) through all matches. A shorthand for `map` on the Regex Object is provided, which maps directly onto the captured `Match` for an ergonomic API. This also eliminates the need for a global or `io` modifier to prevent inlining. Both attributes previously prevented the use of regex queries in areas where these effects are not allowed.
2. The examples have been added to the tests and verified to work as expected.
3. The `GenerateIndices` flag has been removed as an option and is now enabled by default.

Please feel free to share your thoughts on the current design.

leonfuss:
Currently, the tests fail for both LLVM and the Chez backend because they lack the flags interface. To proceed, I will disable the regex tests on those backends. As long as the regex engine of a backend supports flags, implementing those should be straightforward, as the only stateful Flag (`Global`) is now directly expressed in Effekt.

b-studios:
I'll unassign myself, but let's keep this PR open for now.

PR #594, opened 2024-09-17
Experiment: try optimising core/Deadcode and core/Inline

jiribenes:
Using `Vector`s instead of `List`s in `Inline` and `Deadcode` and a few more mutable structures in `Deadcode`.
Not 100% certain correctness was preserved, PTAL.

I'd love if somebody good at benchmarking could just run these on the casestudies and see if it helps. In my testing, I think I got some ~12% speedup, but it would be nice if somebody could verify that independently.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/Deadcode.scala
+++ b/effekt/shared/src/main/scala/effekt/core/Deadcode.scala
@@ -5,6 +5,9 @@ import effekt.core.Block.BlockLit
 import effekt.core.Pure.ValueVar
 import effekt.core.normal.*
 
+import scala.annotation.tailrec
+import scala.collection.mutable
+
 class Deadcode(entrypoints: Set[Id], definitions: Map[Id, Definition]) extends core.Tree.Rewrite {
 
   val reachable = Reachable(entrypoints, definitions)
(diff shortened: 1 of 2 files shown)


effekt/shared/src/main/scala/effekt/core/Deadcode.scala
```diff
@@ -41,69 +44,60 @@ object Deadcode {
...
-    }
+  private def process(d: Definition): Unit = d match {
+    case Definition.Def(id, block) =>
+      within(id) { process(block) }
+    case Definition.Let(id, _, binding) =>
+      process(binding)
+  }
+
+  @tailrec
+  private def process(b: Block): Unit = b match {
```
jiribenes, line 89: This `private` is actually needed for `@tailrec`, the other ones are there for symmetry.
We could also mark it `final` instead.

jiribenes:
👀 

After: 
- inliner on ANF: 85.86
- inliner on build system: 2.71
- inliner on prettyprinter: 103.01
- inliner on stdlib/list/modifyat: 34.97

Before:
- inliner on ANF: 92.11
- inliner on build system: 2.18
- inliner on prettyprinter: 77.13
- inliner on stdlib/list/modifyat: 33.39

PR #597, opened 2024-09-23
Privatize inlined functions

serkm:
This avoids code generation for inlined functions and reduces program size slightly.


b-studios:
LGTM, note that @phischu planned to mark most operations as non-inline anyway and instead annotate them as allocation operations.

PR #598, opened 2024-09-23
Desugar toplevel constants to functions

b-studios:
This PR addresses #496 by 

- desugaring toplevel vals to functions
- desugaring references to toplevel vals to function calls

The translation is performed as part of `core.Transformer` and thus core by construction NEVER has toplevel values on the toplevel. There is only a strange case when translating toplevel object definitions, like:

```
def foo = makeFoo(42)
```
which could end up with
```
let tmp = makeFoo(42)
def foo = unbox tmp
```

Honestly, I do not know how to support this case...

Another thing I do not like is that this minor change now required new syntax in source and a new symbol. Specifically, while for `val foo = 42` on the toplevel `foo` is a value symbol, after translation it is a block symbol. I cannot foresee what this silent transmogrification will imply...


jiribenes:
It would be especially nice to have a few unit-tests here since it's such a treacherous territory. 👀

b-studios:
We already had one simple test, that I expanded on a bit now. What doesn't work is `embed.effekt`, since there is a conceptual conflict between the desugaring of `val` and `def` on the toplevel.

effekt/shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -197,6 +197,7 @@ enum Binder extends TermSymbol {
   case RegBinder(name: Name, tpe: Option[ValueType], region: BlockSymbol, decl: RegDef) extends Binder, RefBinder
   case VarBinder(name: Name, tpe: Option[ValueType], decl: VarDef) extends Binder, RefBinder, TrackedParam
   case DefBinder(name: Name, tpe: Option[BlockType], decl: DefDef) extends Binder, BlockSymbol
+  case ConstBinder(name: Name, tpe: Option[ValueType], decl: ConstDef) extends Binder, ValueSymbol
```
phischu, line 200: Would it work to introduce them as a `BlockSymbol` (morally nullary functions)? This confuses people because they are introduced with the keyword `val`.

b-studios:
Overall, I am not very happy with this PR. I would like to see an alternative, where we do this transformation in the translation from core to machine.

b-studios:
Superseded by #602

PR #602, opened 2024-09-24
Alternative: implement transformation in machine-transformer

b-studios:
This is an alternative to #598 

It still needs to be cleaned up (accesses to `global` should be through proper functions).


jiribenes:
Both conceptually and wrt the amount of code, I like this more than #598.

jiribenes (approved):
LGTM, explained on the Effekt meeting

PR #606, opened 2024-09-25
Implement round, floor, ceil on the LLVM backend

jiribenes:
Cherry-picked from #542.

~~I don't really know why the tests fail on CI: the intrinsics are there from at least LLVM 12, and it does work locally...~~
The CI ~~fails~~ _failed_ with:
```
/usr/bin/ld: ./out/tests/effekt.llvmtests/doubles.o: in function `round_1113':
doubles.ll:(.text+0xff2): undefined reference to `round'
/usr/bin/ld: ./out/tests/effekt.llvmtests/doubles.o: in function `floor_1118':
doubles.ll:(.text+0x1002): undefined reference to `floor'
/usr/bin/ld: ./out/tests/effekt.llvmtests/doubles.o: in function `ceil_1120':
doubles.ll:(.text+0x1012): undefined reference to `ceil'
collect2: error: ld returned 1 exit status
```

_See below for the resolution._


jiribenes:
Oops, I thought we were linking `-lm` already!

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -321,6 +321,9 @@ object LLVMRunner extends Runner[String] {
     val llPath  = basePath + ".ll"
     val optPath = basePath + ".opt.ll"
     val objPath = basePath + ".o"
+    val linkedLibraries = Seq(
+      "-lm", // Math library
+    ) ++ libuvArgs
```
jiribenes, line 326: cc @phischu, do you foresee any problems with linking in `-lm` in the LLVM backend?
This is only relevant when using `gcc`, it works fine without it on `clang`.

phischu: Not an actual problem, just that this is the beginning of a never ending stream of C libraries we are linking against.

phischu (approved):
There could be subtle differences between the backends, but I don't care.

PR #610, opened 2024-09-25
Experiment: run tests forked and in parallel

jiribenes:
Let's see if the CI is faster, I guess?


jiribenes:
This doesn't seem to work, perhaps we can revisit this later.

Issue #611, opened 2024-09-25
Avoid extern includes


phischu:
Right now, we are using `extern include` in the standard library for IO functionality. We would like to inline the files `libraries/llvm/io.ll` and also `libraries/js/io.js` into the position where they are included. We could actually be more fine-grained and inline the parts of them closer to where they are used.

Moreover, we have to be more fine-grained with regard to what runs on node only versus in the web.

PR #612, opened 2024-09-25
Inline includes and mark them as jsNode where necessary

b-studios:
This ought to address #611


libraries/common/io/console.effekt
```diff
@@ -24,26 +24,26 @@ def console[R] { program: () => R / Console }: R = {
 
 namespace js {
 
-  extern js """
+  extern jsNode """
     const readline = require('node:readline');
   """
```
b-studios, line 29: Specific to node js and doesn't work in the webbrowser, for now simply disable

libraries/common/io/error.effekt
```diff
@@ -366,8 +366,106 @@ def fromNumber(number: Int): IOError = number match {
 
 namespace internal {
 
+  extern llvm """
+    declare %Int @c_error_number(%Int)
+  """
+
+  extern jsNode """
+    const os = require('os');
```
b-studios, line 374: Same, 'os' is only available in node

libraries/common/io/error.effekt
```diff
@@ -366,8 +366,106 @@ def fromNumber(number: Int): IOError = number match {
...
+        [os.constants.errno.EUNATCH]: 219,
+        [os.constants.errno.ESHUTDOWN]: 220
+      };
+      return errnoMap[-errno] || -1; // Default to -1 for unknown error names
+    }
+  """
+
   extern pure def errorNumber(errno: Int): Int =
-    js "errorNumber(${errno})"
+    jsNode "errorNumber(${errno})"
```
b-studios, line 468: Using `os`

libraries/common/io/filesystem.effekt
```diff
@@ -139,32 +139,109 @@ def close(fd: File): Unit / Exception[IOError] = {
...
+  extern llvm """
+    declare void @c_fs_open(%Pos, %Pos, %Stack)
+    declare void @c_fs_read(%Int, %Pos, %Int, %Stack)
+    declare void @c_fs_write(%Int, %Pos, %Int, %Stack)
+    declare void @c_fs_close(%Int, %Stack)
   """
 
   extern async def open(path: String, mode: Mode): Int =
-    js "$effekt.callcc(callback => open(${path}, ${mode}, callback))"
+    jsNode "$effekt.callcc(callback => open(${path}, ${mode}, callback))"
```
b-studios, line 223: Using `fs` which is only available in node, but not web

libraries/common/io/network.effekt
```diff
@@ -4,7 +4,7 @@ import bytes
 import io
 
 namespace js {
-  extern js """
+  extern jsNode """
     const net = require('node:net');
```
b-studios, line 8: `node:net` is only available on node

libraries/common/io/error.effekt
jiribenes: We could also use this:
```suggestion
    const os = require('node:os');
```
for consistency.

libraries/common/io.effekt
```diff
@@ -3,15 +3,15 @@ module io
...
-
-
 // Event Loop
 // ----------
 
 type Task[T] = () => T at {io, async, global}
 
+extern llvm """
+  declare void @c_timer_start(%Int, %Stack)
+"""
```
b-studios, line 13: This wouldn't be duplicated, if instead we define a function
`c_spawn(%Stack)` that just calls c_timer_start with `0`.

b-studios:
@phischu I tried to establish the principle that each file only includes its own forward declarations. To this end, and to deduplicate the forward declaration of `c_timer_start`, I introduced `c_yield` which is just `c_timer_start(0, _)`. 

However, while this is a tiny function that just forwards and seems useless, semantically I think it makes sense to distinguish the two.

b-studios:
We can merge this PR in my opinion, the moment CI passes.

PR #616, opened 2024-09-26
Remove lifted and backends that use it

b-studios:
In order to move faster on other parts of the language I am deleting lifted and all backends that use it.

This is a conscious decision with the drawbacks that:

- we cannot demonstrate the compilation pipeline presented in our OOPSLA 2023 paper nicely
- MLton was our fastest backend (by far)
- it will make it difficult to benchmark against the deleted backends
- we cannot show lifted core in the editor and on the website anymore

Especially due to the last point we also need to update the website and vscode plugin.


jiribenes:
I think it would be worth it to write some page on the website for researchers that:
- relates the papers to the actual language (changes since paper XYZ?)
- tells you how to benchmark things long gone (like "in order to benchmark the MLton backend from the paper XYZ, use version A.B.C of Effekt available from \<link\>")

Also, here's a link for showing lifted core on the website on the Internet Archive: http://web.archive.org/web/20240823062750/https://effekt-lang.org/docs/implementation/lift-inference
(that could be on the page outlined above)

examples/stdlib/string/unicode.effekt
```diff
@@ -46,7 +46,7 @@ def main() = {
       charOrOutOfBounds { "f😅o😅o".charAt(expectedLength - 1) }
       println("f😅o😅o".length == expectedLength)
 
-    // ML does not support unicode
+    // Unknown platform probably doesn't support unicode
```
b-studios, line 49: :)

PR #617, opened 2024-09-26
Fix cast warning in LLVM io.c

marvinborner:
Fixes a cast warning in clang. The warning only appeared with `-Wall -Wextra` on GCC. Maybe we should consider using `-Wall -Wextra -Werror` on CI?

### DIFF
--- a/libraries/llvm/io.c
+++ b/libraries/llvm/io.c
@@ -125,7 +125,7 @@ void c_fs_open(struct Pos path, struct Pos mode, Stack stack) {
 
 void c_fs_read(Int fd, struct Pos buffer, Int offset, Stack stack) {
 
-    char* bytes = c_buffer_bytes(buffer); // libuv expects signed integers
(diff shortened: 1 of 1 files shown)


jiribenes:
If I understand C correctly than there are potentially platform-dependent issues here: our buffers are (for some reason) `unsigned char*`, but native platform-dependent string type in C is `char*` (where, of course, `char` isn't the same thing as `signed char`, the signedness is iirc the platform-dependent thing)
This is a great fix for the warning, but not necessarily for the underlying issue. WDYT?

marvinborner:
I'm not entirely sure. I don't really understand why libuv uses (not unsigned) chars anyway and to be honest I didn't even know that chars aren't necessarily signed by default. I don't see any other solution though.

jiribenes:
A solution would be to store platform-dependent strings as mere char* in the Buffer (or uchar8_t or whatever is the name nowadays and then convert)

PR #619, opened 2024-09-30
Js cps backend

b-studios:
Still TODO:

- [x] bidirectional effects
- [x] IO (but needs to be cleaned up, still)
- [x] separate compilation for the web (we do not compile separately anymore)
- [x] cleanup code of `cps.Transformer` and `js.TransformerCps`
- [x] implement show-core and show-target for LSP again (JS backend only)

After this PR:
- [ ] improved introduction of thunks
- [ ] fix lambda lifting (and enable it again?)

Here are some preliminary performance results compared to current master

image

Despite too many thunks, this already looks good.


b-studios:
There is a sibling PR that needs to be merged after this one has been rebased:  https://github.com/effekt-lang/effekt-website/pull/76

b-studios:
This PR is ready for review.

Please be aware that the calling convention changed:

- of toplevel functions (now requires `RUN_TOPLEVEL`)
- of async functions (now requires explicit aborting, since we use exceptions to terminate the trampoline)

Also:
- this PR deprecates separate compilation to favor simplicity of the compiler pipeline. This means, we need to change how things are evaluated on the website (see sibling https://github.com/effekt-lang/effekt-website/pull/76)

The async API is still up for discussion; but we can also change that later again. If you have ideas, let me know.

libraries/js/effekt_runtime.js
```diff
@@ -1,385 +1,208 @@
...
+ * - another value (like `undefined`): the Effekt runtime will terminate.
+ */
+$effekt.capture = CAPTURE
+
+/**
+ * Used to call Effekt function arguments in the JS FFI, like in `io::spawn`.
+ *
+ * Requires an active Effekt runtime (trampoline).
+ */
+$effekt.run = RUN
```
b-studios, line 208: I changed the API once more. It now trades-off bare performance (there might be a few more additional runtime checks now) for easy of use.

There now is only _one_ continuation capturing primitive for `async` FFI functions (see docs in `effekt_runtime.js`).

libraries/js/effekt_runtime.js
```diff
@@ -1,385 +1,218 @@
...
+
+/**
+ * Used to call Effekt function arguments in the JS FFI, like in `network::listen`.
+ *
+ * This function should be used when _no_ Effekt runtime is available. For instance,
+ * in callbacks passed to the NodeJS eventloop.
+ *
+ * If a runtime is available, use `$effekt.run`, instead.
+ */
+$effekt.runToplevel = RUN_TOPLEVEL
```
b-studios, line 218: Sadly, for examples like `network` to work, we need a second way to run Effekt functions, where we install the "Effekt-runtime" (which simply is a trampoline).

jiribenes (approved):
Looks alright, tests are passing :shipit:

Issue #521, opened 2024-07-20
Escapes are broken


jiribenes:
Currently, unicode escapes themselves are valid characters which is somewhat confusing, I'd expect the following escape to have to be written as `'\u0a00'` instead of `\u0a00`:
![image](https://github.com/user-attachments/assets/aa602717-04d4-4ddb-a217-02ef36a1b44d)

I'd propose only allowing unicode escapes inside of string or char literals.

Here's the relevant code:
https://github.com/effekt-lang/effekt/blob/d98362eb405e21a26edf0a35dbb8f15c2ec77c07/effekt/shared/src/main/scala/effekt/Lexer.scala#L584
https://github.com/effekt-lang/effekt/blob/d98362eb405e21a26edf0a35dbb8f15c2ec77c07/effekt/shared/src/main/scala/effekt/Lexer.scala#L448-L469

---

Related to this: we allow wildly nonstandard escapes in string literals ("anything goes"):
https://github.com/effekt-lang/effekt/blob/d98362eb405e21a26edf0a35dbb8f15c2ec77c07/effekt/shared/src/main/scala/effekt/Lexer.scala#L408-L414

This requires some design, but I'd prefer to agree on a common format of an escape, for example Zig has:
[`\u{NNNNNN}` as "hexadecimal Unicode scalar value UTF-8 encoded (1 or more digits)"](https://ziglang.org/documentation/master/#Escape-Sequences)
I like it because it allows multiple digits (subsuming the need for separate `\xNN` and `\uNNNN` and `\uNNNNNN` and whatever else) and it is easy to parse (`<escape> ::= '\u{' <hex>+ '}'`)

I think it's also fine to invest a little bit into lexing escapes properly by ourselves :)

jiribenes:
Related: 
```scala
effekt> '\n' == '\t'
true
```

cc @marzipankaiser

Let's just please fix the lexer.

jiribenes:
Another issue: `val c = '\'` could have two possible meanings:
1. a character that is not completely parsed yet that contains an escaped ' inside
2. a character that completely parsed and contains just \ inside

Currently, we choose 2, but it’s inconsistent with what I presumed when making the highlighting (1 — what Scala does), and with common expectations of programmers & LLMs.
We agreed on Slack to ban this as an unclosed character and require writing `val backslash = '\\'` for a backslash and `val apostrophe = '\''` for an apostrophe.

marzipankaiser:
[...]

PR #622, opened 2024-10-02
Improve handling of (unicode) escapes

### LINKED ISSUE #596: Inconsistency between JS and Scheme backends regarding Unicode (double-)escapes
Hello dear team, 

I seem to have encountered an inconsistency between the behaviour of programs compiled for the JS backend and for the Chez Scheme backend in the context of doubly escaped Unicode codepoints in strings.

The following line, compiled for the Node.js backend and executed in a Windows console:

`println("\\u000A is the Unicode representation of \\n")`

prints, as expected:

> \u000A is the Unicode representation of \n

However, when compiled for the Chez Scheme backend, it prints:

> \012 is the Unicode representation of \n

The generated Scheme code is `(println_1 "\\012 is the Unicode representation of \\n")`.

Apparently, the Scheme version partly honours the double escape (insofar as it preserves it, and doesn't print an actual newline), but nevertheless converts the hexadecimal Unicode value into its octal representation. 
Triple-quoting doesn't prevent that either.

Unless I'm missing or misunderstanding something, this currently makes it necessary to use a workaround like 

`println("\\" ++ "u000A")`

to print 

> \u000A

literally on the Scheme backend.

(Tested on Effekt from 2024-09-13, as of commit 19089d3)

Kind regards,
Michael

marzipankaiser:
This gets us half-way there for escapes in Strings:
- Allow escapes in character literals, i.e. `'\n'` is a newline character
    - Note: This disallows `'\'` for \, but requires `'\\'`
    - Fixes `'\n' == '\t'`  (cmp. #521)
- Check that character literals consist of only one codepoint
- Report wrong escape sequences in parser, but generate the error in lexer - this only happens for SOME of those
    - We should generate error tokens everywhere instead, different issue
- Escape ourselves in Lexer and build up a string
- Add additional escapes: `\$` (because of unquotes), `\u{...}`
- In chez, escape all characters in one function, use chez's unicode handling functions for characters beyond 3 octals
   - Fixes #596

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/LexerTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/LexerTests.scala
@@ -97,6 +97,7 @@ class LexerTests extends munit.FunSuite {
     assertTokensEq("\" \\t \"", Str(" \t ", false), EOF)
     assertFailure("\"\\k\"")
     assertTokensEq("\"\\u001b\"", Str("\u001b", false), EOF)
+    assertTokensEq("\"\\u{001b}\"", Str("\u001b", false), EOF)
   }
 
   test("characters") {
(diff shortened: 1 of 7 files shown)


jiribenes:
See the (failed) experiment in https://github.com/effekt-lang/effekt/pull/523 by the way :)

jiribenes (approved):
LGTM, this is a good start for improving the situation regarding escapes.
Does it fix some things in #521 (like `'\n' == '\r'`) and/or #596? If so, could you add it as a test somewhere?

PR #623, opened 2024-10-02
Remove "user-defined effects" terminology 

### LINKED ISSUE #400: Avoid using "user-defined effects" terminology
Currently, we use "user defined effects" in an error message when signalling that a required effect isn't handled in `main`.

https://github.com/effekt-lang/effekt/blob/6f8973ae77e4962b67b3cb626142fab7430a6cd8/effekt/shared/src/main/scala/effekt/context/ModuleDB.scala#L86

Perhaps we should say simply "effects" instead :)

Note that it's also necessary to change the following neg-test:
https://github.com/effekt-lang/effekt/blob/6f8973ae77e4962b67b3cb626142fab7430a6cd8/examples/neg/unhandledmain.check#L1

oberth-effect:
Fixes #400.


jiribenes (approved):
LGTM, thank you for your contribution! :shipit:

PR #624, opened 2024-10-02
Add  support for documentation comments

BinderDavid:


jiribenes:
@b-studios, I took a look at how to continue, and the biggest problem I could see is that when we'll need _two_ different annotations.
One for `Tree -> DocComment` in parser, and one for `Symbol -> DocComment` which gets populated in Namer, and then used in the LSP (on hover -> resolve to symbol -> retrieve the doc comment from symbol).

Is there a better way? We could also store the DocComment on the AST node of a declaration, but I remember you were somewhat opposed to this. Or should we somehow keep only the `Tree -> DocComment` annotation, and then when hovering, do a "resolve to symbol -> resolve to tree -> retrieve the doc comment from tree"?

WDYT?

b-studios:
Will we only have doc comments on things that are symbols? If yes, then we can always go from tree to symbol and only store `Symbol -> DocComment`, no?

b-studios:
Maybe easiest to store it in the tree, but I have to see how this looks and whether this is too annoying when doing source to source rewritings.

jiribenes:
Incorporated as part of #930

PR #625, opened 2024-10-02
Recover from malformed operation implementation

### LINKED ISSUE #369: Unhelpful error message on missing braces in an interface implementation
When trying to write an instance of an interface:
```scala
interface Eff {
  def op(): Int
}

def e: Eff = new Eff {
  def op() = println("42"); 42
}
```

shows an error: "Operation Eff is not part of interface Eff."

Here's a screenshot:
Screenshot 2024-01-23 at 17 19 27

One way to work around this is to add braces into the method body:
```scala
interface Eff {
  def op(): Int
}

def e: Eff = new Eff {
  def op() = { println("42"); 42 }
}
```

jiribenes:
Resolves #369 

Before:
image

After:
image


effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -711,6 +724,24 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
     nonterminal:
       (`def` ~> idRef()) ~ paramsOpt() ~ maybeReturnAnnotation() ~ (`=` ~> stmt()) match {
         case id ~ (tps, vps, bps) ~ ret ~ body =>
+         if (isSemi) {
+           semi()
+
+           val startPosition = position
+
+           // TODO(jiribenes): would be nice to abstract this and have a `consumeUntil` function
```
jiribenes, line 732:
```suggestion
```

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -29,7 +32,17 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
       //val after = System.currentTimeMillis()
       //println(s"${input.source.name}: ${after - before}ms")
 
-      res
+      // Report soft fails
+      softFails.foreach {
+        case SoftFail(msg, from, to) =>
+          val source = input.source
+          val fromPos = source.offsetToPosition(tokens(from).start)
+          val toPos = source.offsetToPosition(tokens(to).end) // NOTE(jiribenes): I think this is correct, we don't need the +1, right?
```
jiribenes, line 40:
```suggestion
          val toPos = source.offsetToPosition(tokens(to).end)
```

jiribenes (commented):
The TODOs are okay 👀

dvdvgt (approved):
LGTM for this special error

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -711,6 +724,23 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
+           val startPosition = position
+
+           if (!peek(`}`) && !peek(`def`) && !peek(EOF)) {
+              // consume until the next `def` or `}` or EOF
+              while (!peek(`}`) && !peek(`def`) && !peek(EOF)) {
+                next()
+              }
+
+              val endPosition = position
+              val msg = "Unexpected tokens after operation definition. Expected either a new operation definition or the end of the implementation."
```
jiribenes, line 739: Ideally, we would have some hint with a quick fix like "did you mean to do `def op() = { ... }`?", but we don't have the infrastructure for this.

PR #626, opened 2024-10-02
Allow braces in match clauses

### LINKED ISSUE #422: Allow braces in a pattern match

dvdvgt:
Resolves #422


jiribenes (approved):
LGTM, thanks! :shipit:

PR #628, opened 2024-10-02
Migrate doc comments to use '///'

jiribenes:
In preparation for @BinderDavid's #624, I went through the stdlib and tried to change doc comments to use `///` :)

### DIFF
--- a/libraries/common/seq.effekt
+++ b/libraries/common/seq.effekt
@@ -1,41 +1,32 @@
 module seq
+/// This file implements the [[Seq]] type as a general purpose functional sequence.
+///
+/// Implemented as a 2-3 finger tree, it supports
+/// - prepend and append in amortized O(1),
+/// - concat(m, n) in O(log(min(m, n)))
+/// - first and last in amortized O(1)
+/// - size in amortized O(1)
+///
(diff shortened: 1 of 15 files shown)


libraries/common/seq.effekt
```diff
@@ -1,41 +1,32 @@
...
+/// This file implements the [[Seq]] type as a general purpose functional sequence.
+///
+/// Implemented as a 2-3 finger tree, it supports
+/// - prepend and append in amortized O(1),
+/// - concat(m, n) in O(log(min(m, n)))
+/// - first and last in amortized O(1)
+/// - size in amortized O(1)
+///
+/// More information on finger trees:
+///   https://www.staff.city.ac.uk/~ross/papers/FingerTree.pdf
```
jiribenes, line 11: @BinderDavid have you thought of supporting a module-level doc comment?
Do you think it should be under the `module ...` definition or above it? 🤔

jiribenes: ACK, makes sense, thanks

jiribenes:
I'm pretty sure nobody wants to review this, so I'm just merging.

PR #631, opened 2024-10-02
Stdlib: Add simple libraries: Scanner, Json

marzipankaiser:
This adds:
- Some (minimal) features in the effekt stdlib and the subset of #527 that it depends on
- A simple Scanner library (lexing string literals and numbers for now)
- A library to parse and print (minified, not pretty) json


libraries/common/json.effekt
```diff
@@ -0,0 +1,342 @@
...
+  def field[R](k: String){ contents: => R / JsonBuilder }: R
+}
+
+interface Emit[E] {
+  def emit(e: E): Unit
+}
+
+def escape(s: String): String = "\"" ++ s ++ "\"" // TODO
+
+// Make explicit JsonBuilders implicit
```
b-studios, line 24: We now have great doc comments

marzipankaiser:
This seems to run into a bug in overload resolution.
https://github.com/effekt-lang/effekt/actions/runs/11150574191/job/30992018269#step:10:329 says that there are *no overloads which typecheck*, but annotating a type in 2a11c350 fixes this for this test: https://github.com/effekt-lang/effekt/actions/runs/11150733604/job/30992562652?pr=631#step:10:487
but resurfaces it for another one.

This is strange. (But it might be me and a late evening... But documenting this so I remember when I look into this in a few weeks.)

jiribenes:
FYI, I cherrypicked `char` out in #664 to speed it up a little (so that other code can use it)

libraries/common/scanner.effekt
```diff
@@ -0,0 +1,251 @@
...
+        pre * pow(10.0, e)
+      } else { pre }
+    } with stop {
+      pre
+    }
+  }
+  def fraction(pre: Int) = {
+    try {
+      if (do peek[Char]() is '.') {
+        do skip[Char]()
```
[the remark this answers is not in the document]
jiribenes:
```suggestion
      if (readIf('.') is Some(_)) {
```
?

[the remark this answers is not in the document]
marzipankaiser: @jiribenes there is
```effekt
def readIf(e: Char): Bool / Scan[Char] =
   readIf { c => c == e}.isDefined
```
 https://github.com/effekt-lang/effekt/pull/631/files#diff-99fd06f89975ea055bc74f9b7fae3dcaa56ca69e7544717aeadfd749c81d9115R87-R88

libraries/common/scanner.effekt
```diff
@@ -34,65 +34,87 @@ def scanner[A, R] { scanner: () => R / Scan[A] }: R / read[A] = {
...
+  }
+
+def successfull[R] { program: () => R / fail }: Option[R] =
+  try {
+    Some(program())
+  } with fail {
     None()
   }
 
+def successfull { program: () => Any / fail }: Bool =
```
b-studios, line 71:
```suggestion
def successful { program: () => Any / fail }: Bool =
```

libraries/common/char.effekt
```diff
@@ -15,32 +15,30 @@ def isWhitespace(c: Char): Bool = c match {
...
-def isDigit(c: Char): Bool = { c >= '0' && c <= '9' }
-
-/// Checks if the given character is an ASCII digit in the given base
-/// Use `digitValue(c: Char, base: Int)` to get the numeric value out.
-def isDigit(c: Char, base: Int): Bool = {
-  with on[WrongFormat].default { false }
-  val _ = digitValue(c, base)
-  true
-}
+def isDigit(c: Char): Bool = digitValue(c).isDefined
```
jiribenes, line 37: The tests are still using the `isDigit(char: Char, base: Int)` version. 👀
I think it's kind of a shame that we're getting rid of the versions with an explicit base for seemingly no reason, but I'll be able to live with it...

jiribenes:
Potential issue: I have trouble reading a format like:
```
1 2
3 4
5 6<EOF>
```
(so no newline at the end)
with the following code, as the `readDecimal` always just fails:
```scala
def intoStream(input: String): Unit / emit[(Int, Int)] = {
  with boundary
  with feed(input)
  with scanner[Char, Unit]

  while (true) {
    with on[WrongFormat].default { do stop() }
    println("iter!")

    val n = readDecimal()
    println("saw n: " ++ show(n))
    skipWhitespace()
    println("saw whitespace")
    val m = readDecimal()
    println("saw m: " ++ show(m))
    try readIf('\n') with fail { do stop() }
    println("saw newline")
    do emit((n, m))
  }
}
```

What is the best way to do such a thing with this library, @phischu?
Also, how do I say something like "expect a newline here, but if there's no newline, then that's definitely the end of the stream"? Does the scanner even make sense?

libraries/common/scanner.effekt
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+/// Read a positive decimal number.
+def readDecimal(): Int / { Scan[Char], stop, Exception[WrongFormat] } = {
+  var result = expect("digit") { readDigit() }
+  for[Int] { successfully { readDigit() } } { digit =>
+    result = result * 10 + digit
+  }
+  result
+}
```
jiribenes, line 164: This ^ doesn't really work, the `successfully` eats up everything it sees, so `readDecimal` always returns only the first digit of the parsed number.
Instead, I'd think you want something more like:
```suggestion
/// Read a positive decimal number.
def readDecimal(): Int / { Scan[Char], stop, Exception[WrongFormat] } = {
  var result = expect("digit") { readDigit() }
  successfully { 
    result = result * 10 + readDigit()
  }
  result
}
```

libraries/common/scanner.effekt
```diff
@@ -0,0 +1,288 @@
...
+  if (predicate(t)) {
+    do skip[A]()
+    return t
+  } else {
+    do fail()
+  }
+}
+
+/// Check that the next token satisfies the predicate and skip it when it does.
+def skipIf[A] { predicate: A => Bool }: Unit / { Scan[A], fail, stop } = {
```
b-studios, line 99: Is it really necessary that `skipIf` uses `fail`? It shouldn't fail if the next one does not match, right?

effekt/jvm/src/test/scala/effekt/LLVMTests.scala
```diff
@@ -39,6 +39,8 @@ class LLVMTests extends EffektTests {
     examplesDir / "benchmarks" / "are_we_fast_yet" / "queens.effekt",
     examplesDir / "benchmarks" / "input_output" / "word_count_ascii.effekt",
     examplesDir / "benchmarks" / "input_output" / "word_count_utf8.effekt",
+    examplesDir / "benchmarks" / "input_output" / "dyck_one.effekt",
+    examplesDir / "benchmarks" / "input_output" / "number_matrix.effekt",
```
jiribenes, line 43: This one should really **not** be ignored straight away!
Valgrind complains about something freed too early when reading a file: cc @marvinborner 
```
-[error] Process exited with non-zero exit code 139.
-[error] Valgrind log:
-==11975== Invalid read of size 8
-==11975==    at 0x10FE9B: returnAddress_1207 (in /home/runner/work/effekt/effekt/out/tests/effekt.llvmtests/number_matrix)
-==11975==    by 0x10B598: resume_Int (in /home/runner/work/effekt/effekt/out/tests/effekt.llvmtests/number_matrix)
-==11975==    by 0x109EB5: c_resume_int_fs (io.c:62)
```

libraries/common/stream.effekt
```diff
@@ -179,6 +179,19 @@ def sum[R] { stream : () => R / emit[Int] }: (R, Int) = {
 def sum { stream : () => Any / emit[Int] }: Int =
   sum[Any]{stream}.second
 
+def count[R] { stream : () => R / emit[Bool] }: (R, Int) = {
```
jiribenes, line 182: 🚲🏠 
better name?
```suggestion
def countTrue[R] { stream : () => R / emit[Bool] }: (R, Int) = {
```

or alternatively different semantics:
```suggestion
def count[A, R] { stream : () => R / emit[A] }: (R, Int) = {
// just count how many times it's there
```
?

jiribenes:
Here's a quick post about some functions that I found pretty useful / functions I'd want. I'm not at all suggesting to add these, it's just for inspiration. It's also very possible that some of them are very "anti-pattern-ey" -- in such a case, I'd love to know what I should use instead. :)

### Helpful combinators I've been using:
- short-circuiting `any` and `all`:
```scala
def all { stream: => Unit / emit[Bool] }: Bool = 
  try { stream(); true } with emit[Bool] {
    case true => resume(())
    case false => false
  }

def any { stream: => Unit / emit[Bool] }: Bool = 
  try { stream(); false } with emit[Bool] {
    case true => true
    case false => resume(())
  }
```

- hot-swappable combinators for feeding a reader from a string / from a file, specifically adding the `feedFile` here to mirror `stream`'s `feed` (called `feedString` here)
```scala
def feedFile[R](filename: String) { reader: () => R / read[Char] }: R = {
  with on[IOError].panic
  with readFile(filename)
  with decodeUTF8

  reader()
}

// This already exists as "feed" with the same type and implementation:
def feedString[R](input: String) { reader: () => R / read[Char] }: R = {
  with feed(input.fromString)
  with decodeUTF8

  reader()
}
```

- `each` and `eachSorted` for reader _functions_:
```scala
def eachSorted { reader: () => Int / stop }: Unit / emit[Int] = {
  var cache = []
  exhaustively {
    val a = reader()
    cache = Cons(a, cache)
  }
  cache.sort.each // TODO: should probably be more generic in its type and take a comparator used in `sortBy`!
}

def each[A] { reader: () => A / stop }: Unit / emit[A] = {
  var cache = []
  exhaustively {
    val a = reader()
    cache = Cons(a, cache)
  }
  cache.reverse.each
}
```


---

### Helpful combinators I want:

There are a few more helpful combinators that I think I'd like to have to do with separating and unseparating strings. All of these are currently expressible, but not in a reasonably streaming way. I think these would go a long way to make at least partial streaming somewhat easier for noobs (like me):

1. a variant of `collectString` that also takes a separator to join with, example use is Haskell's `unlines` (`unlines ["hello, "world"] == "hello\nworld"`)
    - law: `collectList[String] { ... }.join("\n") === collectStringsWith("\n") { ... }` (but RHS is hopefully streaming)
2. the dual (Haskell's lines) as a variant of `each` for Strings: something that doesn't give me each _character_, but each _line_
    - ideally with a custom separator, then it's much easier to do something like a CSV parser (`with val field = for[String] { line.eachSplit(",") }; ...`)
    - law is something like: `for[String] { input.eachSplit("\n") } { ... } === for[String] { input.split("\n").each } { ... }` (but LHS is streaming)
    - in the best case, it would be usable with things like the scanner -- preprocess by giving the scanner a whole buffered line?

jiribenes:
I also started using Haskell-style `takeWhile` and `iterate` a lot, they are super helpful for AoC!
```scala
def iterate[A](init: A) { step: A => A }: Unit / emit[A] = {
  var current = init
  while (true) {
    do emit(current)
    current = step(current)
  }
}

def takeWhile[A] { cond: A => Bool } { stream: => Unit / emit[A] }: Unit / emit[A] = {
  try stream() with emit[A] {
    case value and cond(value) => do emit(value); resume(())
    case _                     => ()
  }
}
```

libraries/common/scanner.effekt
```diff
@@ -0,0 +1,197 @@
...
+  returning::scanner[A, Unit]{scanner}
+
+
+def many[A] { action: () => A / stop }: Unit / emit[A] =
+  exhaustively { do emit(action()) }
+
+def some[A] { action: () => A / stop }: Unit / { emit[A], stop } = {
+  action()
+  many { action() }
+}
```
jiribenes, line 32: Should these be in `scanner` or rather in `stream`?

b-studios (approved):

PR #634, opened 2024-10-02
Add repl to scheduler casestudy

dvdvgt:
I forgot to add it in #627 :(


jiribenes (dismissed):
LGTM, let's run the case study ;) :shipit:

examples/casestudies/scheduler.effekt.md
```diff
@@ -103,3 +103,7 @@ def main() = {
 ```
 
 Try guessing first what you expect the output to be and then run it:
+
+```effekt:repl
+main()
+```
```
jiribenes, line 109: What do the other studies do? Something like
> ``````
> Please call `main`:
> ```effekt:repl
>```
> ``````
?

examples/casestudies/scheduler.effekt.md
```diff
@@ -103,3 +103,6 @@ def main() = {
 ```
 
 Try guessing first what you expect the output to be and then run it:
```
jiribenes, line 105:
```suggestion
Try guessing first what you expect the output to be and then run it with `main()`:
```
👀

jiribenes (changes requested):

jiribenes (approved):

PR #637, opened 2024-10-14
Fix SBT installation

### LINKED ISSUE #639: CI: Setup `sbt`
As reported by @marvinborner, `ubuntu-latest` no longer contains `sbt`.
We should fix that by also using the `setup-sbt` action like in `effekt-plots`:
https://github.com/effekt-lang/effekt-plots/blob/b3a19c8a077c61f7d8f3c226cde16e02314bdf89/.github/workflows/sync.yml#L41-L42

Bonus: think about collapsing all of the different setups into one single reusable action which is then called later.
This would make the whole setup less brittle.

marvinborner:
Due to updates in the `ubuntu-latest` image, `sbt` is [no longer included](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) in the default installation. Now we also use the [`sbt/setup-sbt`](https://github.com/sbt/setup-sbt) action in all workflows. Closes #639.

CI fails because it still runs with the old config, I believe.

### DIFF
--- a/.github/workflows/autorelease.yml
+++ b/.github/workflows/autorelease.yml
@@ -27,6 +27,9 @@ jobs:
         distribution: 'zulu'
         cache: 'sbt'
 
+    - name: Setup SBT
+      uses: sbt/setup-sbt@v1
+
     - name: Set up NodeJS ${{ env.NODE_VERSION }}
       uses: actions/setup-node@v4
       with:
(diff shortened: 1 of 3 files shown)


jiribenes:
Let's merge it and see!

PR #642, opened 2024-10-14
Allocate region variables as "local variables" on previous stack

serkm:
@phischu I hate how this turned out.

When allocating into a region, we cannot know the answer type of that region. This makes it impossible to push a regular return address.

As a workaround, variables are allocated with a special return function, that removes the region from the stack. When underflowing a region, we first call this special function, and then return regularly.

Also, this breaks globals, as the first stack has no previous stack.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala
@@ -58,8 +58,10 @@ ${indentedLines(instructions.map(show).mkString("\n"))}
       s"call ccc void ${globalName(name)}(${commaSeparated(arguments.map(show))})"
     case Call(result, Ccc(), tpe, ConstantGlobal(name), arguments) =>
       s"${localName(result)} = call ccc ${show(tpe)} ${globalName(name)}(${commaSeparated(arguments.map(show))})"
+    case Call(_, Ccc(), VoidType(), LocalReference(_, name), arguments) =>
+      s"call ccc void ${localName(name)}(${commaSeparated(arguments.map(show))})"
     case Call(_, Ccc(), _, nonglobal, _) =>
-      C.abort(s"cannot call non-global operand: $nonglobal")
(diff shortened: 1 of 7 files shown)


effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala
```diff
@@ -58,8 +58,10 @@ ${indentedLines(instructions.map(show).mkString("\n"))}
       s"call ccc void ${globalName(name)}(${commaSeparated(arguments.map(show))})"
     case Call(result, Ccc(), tpe, ConstantGlobal(_, name), arguments) =>
       s"${localName(result)} = call ccc ${show(tpe)} ${globalName(name)}(${commaSeparated(arguments.map(show))})"
+    case Call(_, Ccc(), VoidType(), LocalReference(_, name), arguments) =>
+      s"call ccc void ${localName(name)}(${commaSeparated(arguments.map(show))})"
     case Call(_, Ccc(), _, nonglobal, _) =>
-      C.abort(s"cannot call non-global operand: $nonglobal")
+      C.abort(s"cannot call non-global operand: $nonglobal") // why not?
```
b-studios, line 64: @phischu maybe you could discuss this in the meeting briefly?

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -385,7 +349,7 @@ object Transformer {
         emit(callLabel(returnAddress, values.map(transform)))
         RetVoid()
 
-      case machine.NewStack(variable, prompt, frame, rest) =>
+      case machine.NewStack(variable, prompt, frame, isRegion, rest) =>
```
b-studios, line 352: will this flag go away when we allocate regions directly on the stack?

serkm: No, we need to know whether we are returning from a region.

serkm:
The last commit depends on #651

phischu (approved):
Nice, so much code was deleted.

PR #643, opened 2024-10-14
Add missing sbt setups

marvinborner:
The previous (merged) PR #637 missed a few 🫣

@jiribenes suggestion would really help in such occasions.

> think about collapsing all of the different setups into one single reusable action which is then called later. This would make the whole setup less brittle.


jiribenes:
I'll merge this for now, but I really think the refactoring would help here...

PR #644, opened 2024-10-15
Make LLVM's stack growable

marvinborner:
This enables automatic growing of LLVM's stack once the end is reached (or memory larger than the remaining space is allocated).

We can now have much smaller stack sizes than before. For now, I've set the previous 256M stacks to 1KB (which, after initial tests, seems fine).

Due to some code repetition I wanted to merge the growing logic of regions with this. I've reverted my attempts since #642 removes the duplicated code anyway.


libraries/llvm/rts.ll
```diff
@@ -403,7 +436,7 @@ define private %Stack @newStack(%Prompt %prompt) {
     ; TODO find actual size of stack
     %stack = call ptr @malloc(i64 120)
 
-    ; TODO initialize to zero and grow later
+    ; TODO initialize to zero?
```
marvinborner, line 439: Do we actually want to initialize the size to zero?

b-studios: Please no! Ideally, try a few sizes and perform benchmark.

marvinborner:
Figure_1

The benchmarks use `config_llvm.txt` similar to `effekt-plots`. Current master is in the left bar.

libraries/llvm/rts.ll
```diff
@@ -366,30 +366,63 @@ define private %Reference @newReference(%Stack %stack) alwaysinline {
...
+
+realloc:
+    %base = extractvalue %Memory %memory, 1
+
+    %intStackPointer = ptrtoint %StackPointer %current to i64
+    %intBase = ptrtoint %Base %base to i64
+
+    %size = sub i64 %intStackPointer, %intBase
+    %double = mul i64 %size, 2
+    %newSize = add i64 %double, %n ; TODO: should we be smarter here?
```
jiribenes, line 393: As someone who doesn't understand the LLVM runtime, why do we need to add the `%n` here?

marvinborner: It's not strictly related to the runtime, but rather the growing logic. Since the bytes to be allocated (`n`) could be larger than the double of the current stack size, I first double the size and then add `n` to it. We could (should?) of course do something smarter here.

jiribenes: Intuitively, I'd expect something like "if we know we have to resize, then `new_size <- nextStrictlyBiggestPowerOfTwo(current_size, n)" where the mysterious function is https://llvm.org/doxygen/namespacellvm.html#afb65eef479f0473d0fe1666b80155237 or `clz` (get the highest bit, choose a number one bigger in binary), but I have to think about it.

marvinborner: Yes, that also sounds good. My thought behind the strategy was that, intuitively, large stack allocations are quite rare. So if I allocate 1GB on a 10MB stack, I'd prefer the next stack size to be 20MB + 1GB rather than 2GB.

b-studios: Might be worth looking at 

https://github.com/golang/go/blob/24cb743d1faa6c8f612faa3c17ac9de5cc385832/src/runtime/stack.go#L336

and it's callsites.

b-studios:
Thanks for the benchmarks, do I read it correctly that `triples` and `tree_explore` basically disappear, because they are much faster?

b-studios:
Most continuation heavy benchmarks seem to be much faster, except parsing dollars. That's strange

libraries/llvm/rts.ll
marvinborner: Very interesting, although I agree that maybe you're overthinking this a bit :octocat:

Regarding the profile of the allocations: We (currently?) only use `@stackAllocate` to make space for pushes of pos/neg, builtin types and other pointers, never for arbitrarily large data allocations. (Maybe @phischu can confirm this?)

As far as I understand, large (de-)allocations only happen when some function has a lot of arguments that need to be stored on the stack (~8-16 byte each). In `nqueens`, for example, this leads to stack (de-)allocations of 108 bytes.

To make this more clear: Allocation sizes/amount in all generated files in `effekt.llvmtests`:

|bytes allocated|how often|
|-----|-----|
|8|318|
|16|305|
|24|1856|
|32|101|
|40|57|
|48|22|
|56|24|
|64|4|
|72|1|
|80|1|

So I don't think we really need to be creative with the growing logic, since `%n` will almost never be larger than the doubled current stack size, especially if we settle on an initial stack size like 1024. Minimal solutions like adding the current size (as I did), or by using `NextPowerOf2` are probably more than enough.

marvinborner: @phischu asked about the stack allocation profile at runtime. Here's the data:

For the entire test suite (`test`):

| bytes allocated | how often |
|-----------------|-----------|
| 8               | 2918774   |
| 16              | 815964    |
| 24              | 5386483   |
| 32              | 841972    |
| 40              | 169619    |
| 48              | 6444      |
| 56              | 79        |
| 64              | 2085      |
| 72              | 126       |
| 80              | 707       |
| 96              | 8         |
| 112             | 1         |

Only the LLVM tests (`testOnly effekt.LLVMTest`)

| bytes allocated | how often |
|-----------------|-----------|
| 8               | 2918721   |
| 16              | 815664    |
| 24              | 5385418   |
| 32              | 841852    |
| 40              | 169604    |
| 48              | 6435      |
| 56              | 79        |
| 64              | 2080      |
| 72              | 126       |
| 80              | 702       |

libraries/llvm/rts.ll
```diff
@@ -367,39 +367,34 @@ define private %Reference @newReference(%Stack %stack) alwaysinline {
 
 define private %StackPointer @stackAllocate(%Stack %stack, i64 %n) {
     %stackPointer_pointer = getelementptr %StackValue, %Stack %stack, i64 0, i32 1, i32 0
+    %base_pointer = getelementptr %StackValue, %Stack %stack, i64 0, i32 1, i32 1
+    %limit_pointer = getelementptr %StackValue, %Stack %stack, i64 0, i32 1, i32 2
 
-    %stackMemory = getelementptr %StackValue, %Stack %stack, i64 0, i32 1
-    %memory = load %Memory, ptr %stackMemory
+    %currentStackPointer = load %StackPointer, ptr %stackPointer_pointer
+    %base = load %Base, ptr %base_pointer
```
marvinborner, line 374: Does this impact performance? I assume LLVM optimizes this anyway, but I've deliberately left `base_pointer`, `limit_pointer`, `base`, etc. in the `realloc` branch since they're not needed in the more likely `continue` branch

phischu: You are right, please move `base_pointer` and `base` to the `realloc` branch.

b-studios:
Thanks @marvinborner !

PR #653, opened 2024-10-25
Require default for finite domains instead of giving up

### LINKED ISSUE #652: Overzealous exhaustiveness checker for booleans
The following (IMO valid) program
```scala
def foo(opt: Option[Bool]) = opt match {
  case Some(false) => Some(false)
  case opt => Some(true)
}
```
returns an error `Non-exhaustive pattern matching, missing case Some(true)` which is clearly nonsense here.

Using the `else` for a `match` also doesn't work
```scala
def foo(opt: Option[Bool]) = opt match {
  case Some(false) => Some(false)
} else Some(true)
```

---

Moreover, even the following program (and its `else` variant) fails with the `Non-exhaustive pattern matching, missing case true` error:
```scala
def foo(b: Bool) = b match {
  case false => false
  case b => b
}
```

b-studios:
Should fix #652

### DIFF
--- a/examples/pos/issue652.effek
+++ b/examples/pos/issue652.effek
@@ -1,0 +1,4 @@
+def foo(opt: Option[Bool]) = opt match {
+  case Some(false) => Some(false)
+  case opt => Some(true)
+}
--- a/effekt/shared/src/main/scala/effekt/typer/ExhaustivityChecker.scala
+++ b/effekt/shared/src/main/scala/effekt/typer/ExhaustivityChecker.scala
@@ -253,19 +253,16 @@ object ExhaustivityChecker {
         }
       }
 
-      def requireDefault(): Unit =
-        defaults.toList match {
-          case Nil => E.missingDefault(tpe, scrutinee)
-          case head :: tail => matchClauses(head, tail)
-        }
-
       finite match {
         // we have a finite domain
         case Some(cases) =>
           cases.foreach { v =>
             matches.get(v) match {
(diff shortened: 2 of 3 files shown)


examples/pos/issue652.effek
jiribenes, line 1: oops 😇

jiribenes:
It seems that it might be breaking my nice custom error messages in `neg/match_nested_literal.effekt` 🥲 

```
==> X effekt.JavaScriptTests.examples/ (js)  0.04s munit.ComparisonFailException: 
-[error] examples/neg/match_nested_literal.effekt:9:21: Non-exhaustive pattern matching, scrutinees of type Bool require a default case Or(_, And(Lit(_), _))
+[error] examples/neg/match_nested_literal.effekt:9:21: Non-exhaustive pattern matching, missing case Or(_, And(Lit(true), _))
 def test(b: BExp) = b match {
                     ^
-[error] examples/neg/match_nested_literal.effekt:9:21: Non-exhaustive pattern matching, scrutinees of type Bool require a default case Lit(_)
+[error] examples/neg/match_nested_literal.effekt:9:21: Non-exhaustive pattern matching, missing case Lit(false)
```

examples/pos/issue652.effek
jiribenes: @b-studios 👀 You still left this file here

b-studios: haha, oh that's true

PR #654, opened 2024-10-26
Fix bytearray construct

phischu:


b-studios (approved):
Great thanks for fixing this! I tried it on my machine and it resolved the issue.

b-studios:
We should really add something to CI to detect this automatically.

PR #657, opened 2024-10-29
More sharing of erasers and sharers

marvinborner:
Currently we generate a lot of erasers and sharers repeatedly even if they erase or share the same objects. This PR reduces the redundancy, although it has a smaller effect than (I) expected.

CLOC of `effekt.llvmtests` before:

```
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
LLVM IR                        188          98411          15714         225451
```

CLOC of `effekt.llvmtests` after (~20kLOC reduction):

```
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
LLVM IR                        188          86310          15714         207307
-------------------------------------------------------------------------------
```

This is a first step towards making erasing and sharing in the LLVM backend more elegant. There are many more (more complex) improvements that I/we would like to make, but I suggest we do this in a separate PR.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
@@ -714,31 +714,21 @@ object Transformer {
     }
 
   def shareValue(value: machine.Variable)(using FunctionContext, BlockContext): Unit = {
-    value.tpe match {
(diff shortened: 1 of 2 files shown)


phischu (approved):
Beautiful! Much better!

PR #662, opened 2024-11-03
Fix #661 by improving type inference

### LINKED ISSUE #661: TypeError: $effekt.handleMonadic is not a function
The following code crashes Effekt:

```scala
interface Iter {
  def yield(value: Int): Unit
}

def iter(self: List[Int]): Unit / {Iter} =
  self match {
    case Cons(head, tail) => do yield(head); iter(tail)
    case Nil() => ()
  }

def my_map {action: () => Unit / {Iter}} {f: Int => Int}: Unit / {Iter} =
  try action()
  with Iter {
    def yield(value) = do yield(f(value))
  }

def main() =
  try my_map { iter([1, 2, 3]) } { x => x+1 }
  with Iter {
    def yield(x) = println(x)
  }
```

```javascript
[apropos@arch effekt]$ java -jar bin/effekt effects/crash.effekt 
/home/apropos/Projects/contrib/effekt/out/crash.js:661
  return $effekt.handleMonadic((p4101) => {
                 ^

TypeError: $effekt.handleMonadic is not a function
    at main_2823 (/home/apropos/Projects/contrib/effekt/out/crash.js:661:18)
    at Object.main (/home/apropos/Projects/contrib/effekt/out/crash.js:685:15)
    at Object.<anonymous> (/home/apropos/Projects/contrib/effekt/out/crash:2:63)
    at Module._compile (node:internal/modules/cjs/loader:1469:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1548:10)
    at Module.load (node:internal/modules/cjs/loader:1288:32)
    at Module._load (node:internal/modules/cjs/loader:1104:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:174:12)
    at node:internal/main/run_main_module:28:49

Node.js v20.18.0
```

b-studios:
We now order block arguments to first type check those that do not subtract effects. This hopefully gives enough info about type arguments to uniquely determine effects of other block arguments.


b-studios:
Hopefully, this also helps @phischu in #527

b-studios:
Even though this is a simple heuristic that could be discussed and improved in the future, I am merging this now so it appears in the next release on Monday.

PR #664, opened 2024-11-04
Add 'char' library by cherrypicking from #631

### LINKED ISSUE #632: Use proper character literals in the wordcount example
In the wordcount test/example, use real character literals `'\n'` and `'\t'` after #622 gets merged:
https://github.com/effekt-lang/effekt/blob/7ce8ab19fbe5d38e4cefc1698a1132d96329cd76/examples/stdlib/io/filesystem/wordcount.effekt#L12-L16

jiribenes:
Quick backport (forward-port?) of the 'char' library from #631 by @marzipankaiser plus:
- doc comments,
- using the library in the `wordcount` example, (resolves #632)
- and _very_ basic tests.

One big limitation is that it's ASCII only which we should remedy in the future...


libraries/common/char.effekt
```diff
@@ -0,0 +1,58 @@
+/// Warning: This library currently only works with ASCII characters, **not** unicode!
```
b-studios, line 1: This sounds dangerous, indeed.

b-studios (approved):
Looks good, but maybe @phischu can also look at it again?

jiribenes:
@phischu has a beautiful function in #527 that takes a UTF-8 char and produces bytes and vice versa. Where should something like that go? Into `char` or into `stream`?

phischu:
> @phischu has a beautiful function in #527 that takes a UTF-8 char and produces bytes and vice versa. Where should something like that go? Into `char` or into `stream`?

There are 2 possible ways to structure the standard library.

1. Right now, `stream` uses lots of other modules and defines streaming for e.g. container types and strings.
2. Alternatively, we could make the `stream` module expose only the `emit` and `read` effects and not much more and define the streaming functionality in each of the standard library modules like `list` and `string`.

I propose we stick with (1) for now and see how it pans out.

b-studios:
@jiribenes is there anything you would like to change or should I merge this one? What about the unicode warning?

jiribenes:
I think we should merge this.

Unicode support is hard to do properly (and needs updating pretty often). Essentially, there are machine readable tables which we could use to generate a mapping for character classes, that in turn get used for things such as "is this character lowercase" and "how do I transform this lowercase character into an uppercase one?" (like every other language does).

PR #666, opened 2024-11-05
Improve benchmark arguments

marvinborner:
Lately much of our benchmark data has become less useful since the measurements got really small. Our JS and LLVM backends also have fewer errors like stack overflows, allowing us to use larger arguments in general.

We should still consider when and how often we want to normalize the arguments, since it makes historical data less comparable.

I've also started writing a benchmark config for reference arguments, such that we can compare the data with the benchmarks from other languages (this is not finished right now, since it's hard to find good reference values).


examples/benchmarks/config_default.txt
```diff
@@ -0,0 +1,26 @@
...
+are_we_fast_yet/sieve.effekt 5000
+are_we_fast_yet/storage.effekt 1
+are_we_fast_yet/towers.effekt 13
+duality_of_compilation/erase_unused.effekt 5
+duality_of_compilation/factorial_accumulator.effekt 5
+duality_of_compilation/fibonacci_recursive.effekt 5
+duality_of_compilation/iterate_increment.effekt 5
+duality_of_compilation/lookup_tree.effekt 5
+duality_of_compilation/match_options.effekt 5
+duality_of_compilation/sum_range.effekt 5
```
phischu, line 16:
```suggestion
duality_of_compilation/erase_unused.effekt 10000
duality_of_compilation/factorial_accumulator.effekt 10000000
duality_of_compilation/fibonacci_recursive.effekt 40
duality_of_compilation/iterate_increment.effekt 100000000
duality_of_compilation/lookup_tree.effekt 10000000
duality_of_compilation/match_options.effekt 10000000
duality_of_compilation/sum_range.effekt 10000000
```

phischu (approved):

phischu:
Can I just merge this or not?

marvinborner:
Yes, I think we can just merge.

Issue #670, opened 2024-11-06
Unambiguous field access leads to a confusing 'Unbox requires a boxed type' error


jiribenes:
In the following program (already minified), there's an error "Unbox requires a boxed type, but got Person." over the `person.age` call:

```scala
interface Foo {
  def foo(): Unit
  def age(s: String): Int
}

record Person(name: String, age: Int)

def foo(person: Person) { prog: () => Unit / Foo }: Unit = {
  try { prog() } with Foo {
    def foo() = {
      val age: Int = person.age
      //             ~~~~~~
      // Unbox requires a boxed type, but got Person.

      if (age < 100) {
        println(person.name)
      }
      resume(())
    }
    def age(s) = resume(42)
  }
}
```

The error goes away if you remove the `age` operation.
If you instead make the `age` operation into a separate function returning 42, then the overload becomes ambiguous with a nice error message:

```
Ambiguous overload.
There are multiple overloads, which all would type check:
- module0::age: Person => Int
- module0::age: Person => Int
```

---

In the original, the compiler is really saying that it cannot distinguish between the `Foo::age` operation and the `Person::age` field access.
Not only is the error message a bit confusing (what is it unboxing?), but the type-based name resolution here should be unambiguous, I think, since:
1. we've annotated the type `Int`, so the result of `person.age` must be an `Int`, which is only true for the `Person::age(person)` interpretation
2. I don't think it could've been the `age` _operation_ here anyways as we'd need to use the `do` keyword, something like `do age(person)`

Perhaps I'm misunderstanding something here, but for now, I'll use both the `bug` and the `errormessage` labels.

EDIT: Maybe this has something to do with a generated getter? (some `age: Person => Int` function generated by the compiler?)

jiribenes:
@marzipankaiser noted that it could be because of some {person: Foo} capability -> person.age

PR #671, opened 2024-11-06
Try to improve resolution by adding a bit context

b-studios:
This is an attempt to improve on #670. However, there are some trade-offs as can be seen in the modified test.

Another thing is that the error messages are really bad now if something cannot be found.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Namer.scala
+++ b/effekt/shared/src/main/scala/effekt/Namer.scala
@@ -491,25 +491,23 @@ object Namer extends Phase[Parsed, NameResolved] {
     case source.MethodCall(receiver, target, targs, vargs, bargs) =>
       resolveGeneric(receiver)
 
-      Context.resolveMethodCalltarget(target)
-
       // We are a bit context sensitive in resolving the method
       Context.focusing(target) { _ =>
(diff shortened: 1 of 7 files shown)


b-studios:
Currently, there is an issue where in `bytearray` we have

```
def resize(source: ByteArray, size: Int): ByteArray = {
  val target = allocate(size)
  val n = min(source.size, target.size)
  def go(i: Int): ByteArray =
    if (i < n) {
      target.unsafeSet(i, source.unsafeGet(i))
      go(i + 1)
    } else {
      target
    }
  go(0)
}
```

There the parameter `size` (a value) shadows the selector `.size`.

jiribenes:
Would specialising on arity help there? If there's something like `foo.bar`, then perhaps it's more likely to be field access than a method/UFCS'd function?

b-studios:
In 4982b9c, I was able to improve the precision a bit further and revert some changes to tests. 

The strategy is now to first try the respective function / operation lookup and if this fails **completely** then also try the other... you never know ;)

b-studios:
Maybe the special casing of error messages is useless, but I'll merge this so that we can have this in the next release.

Issue #660, opened 2024-11-02
Go to definition for imported modules


jiribenes:
It would be nice to be able to jump to a definition of an imported module:
<img width="286" alt="Screenshot 2024-11-02 at 14 54 33" src="https://github.com/user-attachments/assets/840c4ce1-3e7b-4926-b844-49dafb845baa">

Having something like this would ease the exploration of primarily stdlib modules, esp. before we have a proper documentation generator. :)

Current workaround is using one of the functions provided by the module and jumping to its definition, thus jumping to the module the function was defined in.

b-studios:
Turns out this might need to be special-cased since the current "jump-to-definition" searches for an Id and its symbol. We do have a symbol (e.g. `Module`) for the import, but currently do not attach it to the import itself.

PR #676, opened 2024-11-07
Add jump-to-definition for imports

b-studios:
This addresses #660


b-studios:
Would be great if somebody could try this out, before merging.

jiribenes (approved):
I've tested this, it works perfectly, even for submodules like `string/tty` and still acts correctly on namespaced identifiers, both when it's `module::ident` and `namespace::ident`. Thanks!

PR #677, opened 2024-11-07
Note parameters

serkm:
I'm not sure why this was still TODO, but this fixes the bidirectional tests.


b-studios:
Haha, great fix :) Thanks!

PR #679, opened 2024-11-07
Direct Style Shift

b-studios:
This is WIP PR where we translate shift, reset, and resume to direct style in JS...

Overall, the performance stays the same; only `tree explore` looks faster. Strangely, `resume non-tail` get's MUCH slower (factor 2-3). Maybe the higher-order functions weren't that bad before for inlining by the JIT.

image

### DIFF
--- a/effekt/shared/src/main/scala/effekt/cps/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/cps/Transformer.scala
@@ -147,8 +147,9 @@ object Transformer {
     case core.Stmt.Resume(cont, body) =>
       val ks2 = Id("ks")
       val k2 = Id("k")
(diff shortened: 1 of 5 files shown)


b-studios:
Conclusion: we save some JS stack, but it might not be worth it. (cc @phischu)

b-studios:
We could revisit this, once tuples landed in JS: https://tc39.es/proposal-record-tuple/tutorial/

Maybe this will be more obvious to the JITter

b-studios:
Also, regardless of the result, we should cherry-pick resumptions as statement in CPS.

b-studios:
Resume has been cherry picked in #708

b-studios:
Tuples have been withdrawn: https://github.com/tc39/proposal-record-tuple/issues/394

PR #656, opened 2024-10-29
Add more GCC flags to CI builds

marvinborner:
Within the last few weeks we've repeatedly had compilation warnings in the LLVM backend compiled with the master branch. The warnings could have been caught earlier by running Effekt with the `--debug` flag.

To enforce this, the CI now runs with `--debug` by default (in addition to `--valgrind`) and `--debug` adds the additional `-Werror` flag -- thus failing the CI if the compilation produces warnings.

PR #685, opened 2024-11-12
Replace legacy sprintf with snprintf

marvinborner:
Although not officially deprecated, the usage of `sprintf` throws warnings *on MacOS* (see #656). They are not reproducible on GCC/Clang, but we may want to fix them anyway since switching to `snprintf` has no disadvantages.

### DIFF
--- a/libraries/llvm/bytearray.c
+++ b/libraries/llvm/bytearray.c
@@ -81,7 +81,7 @@ char* c_bytearray_into_nullterminated_string(const struct Pos arr) {
 // TODO do this in Effekt
 struct Pos c_bytearray_show_Int(const Int n) {
     char str[24];
-    sprintf(str, "%" PRId64, n);
+    snprintf(str, sizeof(str), "%" PRId64, n);
     return c_bytearray_from_nullterminated_string(str);
 }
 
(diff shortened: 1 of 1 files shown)


libraries/llvm/bytearray.c
```diff
@@ -112,14 +112,14 @@ struct Pos c_bytearray_show_Char(const uint64_t n) {
...
 }
 
 // TODO do this in Effekt
 struct Pos c_bytearray_show_Double(const Double x) {
-    char str[64]; // TODO is this large enough? Possibly use snprintf first
-    sprintf(str, "%g", x);
+    char str[64]; // TODO is this large enough?
+    snprintf(str, sizeof(str), "%g", x);
     return c_bytearray_from_nullterminated_string(str);
 }
```
jiribenes, line 124: Would something like:
```
    // Get required size
    int size = snprintf(NULL, 0, "%g", x);

    // Allocate exact buffer size (+1 for null terminator)
    char buf[size + 1];
    
    // Do the actual conversion
    snprintf(str, size + 1, "%g", x);
```
make sense? How many characters can a Double even be? 🤔

PR #691, opened 2024-11-13
Also desugar with-do

b-studios:


effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -221,6 +221,8 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
      case Var(id) =>
        val tgt = IdTarget(id)
        Return(Call(tgt, Nil, Nil, (BlockLiteral(Nil, vparams, bparams, body)) :: Nil))
+     case Do(effect, id, targs, vargs, bargs) =>
+      Return(Do(effect, id, targs, vargs, (BlockLiteral(Nil, vparams, bparams, body)) :: Nil))
```
phischu, line 225: You are dropping `bargs`, why not (like in the other cases `bargs :+ (BlockLiteral(Nil, vparams, bparams, body))`?

b-studios: Good point!

b-studios, line 225:
```suggestion
      Return(Do(effect, id, targs, vargs, bargs :+ (BlockLiteral(Nil, vparams, bparams, body)) :: Nil))
```

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -221,6 +221,8 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
      case Var(id) =>
        val tgt = IdTarget(id)
        Return(Call(tgt, Nil, Nil, (BlockLiteral(Nil, vparams, bparams, body)) :: Nil))
+     case Do(effect, id, targs, vargs, bargs) =>
+      Return(Do(effect, id, targs, vargs, bargs :+ (BlockLiteral(Nil, vparams, bparams, body)) :: Nil))
```
b-studios, line 225:
```suggestion
      Return(Do(effect, id, targs, vargs, bargs :+ BlockLiteral(Nil, vparams, bparams, body)))
```

phischu (approved):

PR #693, opened 2024-11-13
Add a regression check for #692

jiribenes:
Adds a quick regression test for a handler in tail position that uses a mutable variable (minimised reproduction of the underlying issue).

The test passes on current `master` after #692.
Crashes the compiler on `v0.9.0`, but works on `v0.8.0`.


examples/pos/mutabletailrec.effekt
```diff
@@ -0,0 +1,19 @@
+module examples/pos/mutabletailrec
+
+effect tagging(tag: String): Unit
+
+def handle { prog: () => Unit / tagging }: Unit = {
+  try {
+    prog()
+  } with remove { tag =>
```
jiribenes, line 8:
```suggestion
  } with tagging { tag =>
```

jiribenes:
More tests are good, this PR has been up for a month now, merging.

> _All the planning charts and demolition orders have been on display at your local planning department in Alpha Centauri for 50 of your Earth years._

PR #694, opened 2024-11-14
Refactor core for optimizer

b-studios:


b-studios:
Thanks @phischu and @marzipankaiser for your help!

PR #695, opened 2024-11-15
Only inline into active positions

b-studios:


b-studios:
This PR is now ready for review, some minor cleanup changes might be pending.

It would be great if we could test it with all the problematic examples that caused the previous inliner to fail, before merging.

effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
```diff
@@ -1,22 +1,25 @@
...
         case d: Definition.Def if reachable.isDefinedAt(d.id) => rewrite(d)
         // we only keep non-pure OR reachable let bindings
         case d: Definition.Let if d.capt.nonEmpty || reachable.isDefinedAt(d.id) => rewrite(d)
       }, rewrite(stmt))
+
+    case Reset(body) =>
+      rewrite(body) match {
+        case BlockLit(tparams, cparams, vparams, List(prompt), body) if !reachable.isDefinedAt(prompt.id) => body
+        case b => Stmt.Reset(b)
+      }
```
b-studios, line 22: Not sure, but we could also move this to a separate phase. 

in favor:
+ deadcode would really only do deadcode elimination

against:
- removing an unused reset is a bit like removing a dead binding, so it does fit
- a separate phase would potentially cost us some time and conceptual overhead.

effekt/shared/src/main/scala/effekt/core/optimizer/Optimizer.scala
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    def normalize(m: ModuleDecl) = {
+      val anfed = BindSubexpressions.transform(m)
+      val normalized = Normalizer.normalize(Set(mainSymbol), anfed, Context.config.maxInlineSize().toInt)
+      Deadcode.remove(mainSymbol, normalized)
+    }
+
+    // (3) normalize once and remove beta redexes
+    tree = Context.timed("normalize-1", source.name) { normalize(tree) }
```
b-studios, line 45: @marvinborner will these new timings cause problems with plots?

effekt/shared/src/main/scala/effekt/core/optimizer/StaticArguments.scala
```diff
@@ -120,22 +123,22 @@ object StaticArguments {
...
         // if arguments are static && recursive call: call worker with reduced arguments
         case BlockVar(id, annotatedTpe, annotatedCapt) if hasStatics(id) && within(id) =>
           val IsStatic(staticT, staticV, staticB) = C.statics(id)
-          app(C.workers(id),
+          Stmt.App(C.workers(id),
             dropStatic(staticT, targs),
             dropStatic(staticV, vargs).map(rewrite),
             dropStatic(staticB, bargs).map(rewrite))
-        case _ => app(rewrite(b), targs, vargs.map(rewrite), bargs.map(rewrite))
+        case _ => Stmt.App(rewrite(b), targs, vargs.map(rewrite), bargs.map(rewrite))
```
b-studios, line 137: Here we do not perform beta reduction as this will be done later by `Normalizer`.

effekt/shared/src/main/scala/effekt/core/optimizer/StaticArguments.scala
```diff
@@ -120,22 +123,22 @@ object StaticArguments {
 
   def rewrite(s: Stmt)(using C: StaticArgumentsContext): Stmt = s match {
     case Stmt.Scope(definitions, body) =>
-      scope(rewrite(definitions), rewrite(body))
+      normal.Scope(rewrite(definitions), rewrite(body))
```
b-studios, line 126: This regularly is the only smart constructor we use since it is stupid to accidentally create a scope with no bindings.

effekt/shared/src/main/scala/effekt/generator/chez/Transformer.scala
```diff
@@ -117,9 +117,11 @@ trait Transformer {
     // currently bidirectional handlers are not supported
     case Resume(k, Return(expr)) => chez.Call(toChez(k), List(toChez(expr)))
 
+    case Resume(k, other) => sys error s"Not supported yet: ${util.show(stmt)}"
+
     case Region(body) => chez.Builtin("with-region", toChez(body))
 
-    case other => chez.Let(Nil, toChez(other))
+    case s: Scope => chez.Let(Nil, toChez(s))
```
b-studios, line 124: Annotated since otherwise we might be missing cases (which happened) and then loop during code gen.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -168,7 +170,7 @@ object Transformer {
         transform(vargs, bargs).run { (values, blocks) =>
           callee match {
             case Block.BlockVar(id, annotatedTpe, annotatedCapt) =>
-              BPC.info.getOrElse(id, sys.error(s"Cannot find block info for ${id}.\n${BPC.info}")) match {
+              BPC.info.getOrElse(id, sys.error(pp"In ${stmt}. Cannot find block info for ${id}: ${annotatedTpe}.\n${BPC.info}")) match {
```
b-studios, line 173: This shouldn't happen, but it does (sadly quite frequently), so I dump a bit of additional info here.

b-studios:
Before the last (two) commit(s), `product_early` was much better on LLVM. This might be worth investigating.

b-studios:
Turns out in d44618f I removed some of the `normal`ization functions from `core.Tree`. In turn, some empty continuation pushes of the form"

```
val x = STMT
x
```

were not removed anymore.

Instead of adding this back, [84f1db6](https://github.com/effekt-lang/effekt/pull/695/commits/84f1db64f501534ca1ba944e2beaee2ea313d08b) now only removes empty continuations.

The reasoning is that some the previous "normalization" also performed commuting conversions of `val x = { val y = stmt1 }; stmt2`. However, this extends the lifetime and is not "conservative".

phischu:
> The reasoning is that some the previous "normalization" also performed commuting conversions of `val x = { val y = stmt1 }; stmt2`. However, this extends the lifetime and is not "conservative".

If it is not too much work, please do these commuting conversions too. This might be a problem for other people but not for us, because we are using garbage-free reference counting.

effekt/shared/src/main/scala/effekt/core/optimizer/StaticArguments.scala
```diff
@@ -82,11 +85,17 @@ object StaticArguments {
...
+    val freshBparams: List[BlockParam] = (staticB zip blockLit.bparams zip freshCparams).map {
+      case ((true, param), capt) => param
+      case ((false, BlockParam(id, tpe, _)), capt) => BlockParam(Id(id), tpe, Set(capt))
     }
 
+    // the worker now closes over the static block arguments (`c` in the example above):
+    val newCapture = blockLit.capt ++ selectStatic(staticB, freshCparams).toSet
+
+    val workerVar: Block.BlockVar = BlockVar(Id(id.name.name + "_worker"), workerType, newCapture)
+    ctx.workers(id) = workerVar
```
b-studios, line 97: This was actually a crucial fix.

b-studios:
The PR is ready to merge. Here are the latest results (after merging #736, which made a few things slower):

image

There is no need to worry about `iterate_increment`, which is in the order of 10ms.

b-studios:
Ok, doesn't look like it is mergable due to new valgrind issues...

b-studios:
I ignore the valgrind issues and segfaults that resurfaced after merging master into this PR.

Issue #758 tracks these and can be resolved independently.

PR #704, opened 2024-11-22
Use noalias on StackValues

serkm:


libraries/llvm/rts.ll
```diff
@@ -776,3 +777,14 @@ define void @run_Pos(%Neg %f, %Pos %argument) {
     tail call tailcc %Pos %functionPointer(%Object %object, %Evidence 0, %Pos %argument, %Stack %stack)
     ret void
 }
+
+
+; Scope domains
+!0 = !{!"types"}
+!1 = !{!1}
```
phischu, line 784: This is unnecessary, and then the `!3` could become `!2` and `!2` could become `!1`.

phischu (commented):
Very good job!

phischu (approved):

PR #705, opened 2024-11-22
Refactor 'stream', add a few examples

jiribenes:
Update: 'stream' got slightly refactored (using `Any` everywhere now), fixed a few bugs, added a few examples of push streams, pull streams, their infinite variants, `limit`, `zip`, ...

---

I added a few examples for the new stream stdlib into `examples/stdlib`.
I'd also be happy to add some _Nice_ ®️ example into the `stream.effekt` file itself (either into the doc comments or as a `namespace examples { ... }` thing)

~~I couldn't figure out how to use `take` properly in a sane (nice idiomatic) way, since it requires me to handle `stop` and I couldn't find any nicely looking way to do that. Therefore all of these examples are just simple, finite push streams.~~ _EDIT: solved_

I also tried adding `zip` in order to iterate through multiple streams in step, but I'm not sure how idiomatic it is (see comments below).

@phischu, please let me know if I'm using the `stream` library as you envisioned it.
If not, feel free to go ahead and change the examples so that they are closer to what you'd want. :)

### TODO (in an ideal world)
- [x] at least one simple example with pull streams
- [x] one example using both pull streams and push streams
- [x] some example with infinite streams
- [x] make `zip` nicer to use?


examples/stdlib/stream/neighbours.effekt
```diff
@@ -0,0 +1,38 @@
...
+/// Gets the neighbours of a given position.
+/// with radius=1, those are immediate neighbours (including the diagonal)
+/// with radius=2, these are neighbours&their neighbours
+/// ...
+def neighboursOf(pos: Pos, radius: Int) = {
+  with val dx = for[Int] { range(neg(radius), radius + 1) }
+  with val dy = for[Int] { range(neg(radius), radius + 1) }
+  val newPosition = Pos(pos.x + dx, pos.y + dy)
+  do emit(newPosition)
+}
```
jiribenes, line 20: Would be really cool if it were first the neighbours of radius 1, then the neighbours of radius 2, ...

examples/stdlib/stream/sum_of_squares.effekt
```diff
@@ -0,0 +1,12 @@
...
+def squares(max: Int): Unit / emit[Int] = {
+  with val n = for[Int] { range(1, max) }
+  do emit(n * n)
+}
+
+def main() = {
+  val max = 10
+  println("The sum of squares from 1 to " ++ show(max) ++ " is:")
+  println(sum { squares(max + 1) })
+}
```
jiribenes, line 12: This would be especially nice as an infinite stream, I think.
I tried `rangeFrom` and `take`, but it didn't seem very ergonomic to me :/

libraries/common/stream.effekt
```diff
@@ -282,6 +293,18 @@ def source[A, R] { stream: () => Unit / emit[A] } { reader: () => R / read[A] }:
...
+def zip[A, B] { stream1: () => Unit / emit[A] } { stream2: () => Unit / emit[B] }: Unit / emit[(A, B)] = {
+  with source[A, Unit] { stream1 }
+  with source[B, Unit] { stream2 }
+
+  exhaustively {
+    val a = do read[A]()
+    val b = do read[B]()
+    do emit((a, b))
+  }
+}
```
jiribenes, line 307: What's the idiomatic way to write `zip` and use it?

1. There's _something_ like `exhaustively` being used, but without the `action` parameter (added temporarily as a separate overload). What would be the idiomatic way to do this kind of thing?
2. I also don't really like that I have to specify the type params both in `for` and in `zip` when I use it. Would it instead make sense to make a `for` wrapper taking two streams? My intent is to mostly use it for iteration anyways...
3. Should we return a `These` instead? (`Left a | Right b | Both (a, b)`)

libraries/common/stream.effekt
```diff
@@ -102,9 +102,20 @@ def exhaustively[A] { program: () => A / stop } { action: A => Unit }: Unit =
     }
     go()
   } with stop {
-    def stop() = ()
+    ()
   }
```
jiribenes, line 106: This confused me like three different times while reading the code -- I don't think we should necessarily rely on the fact that this more explicit `def stop() = ...` syntax is allowed when using a singleton `effect`...

examples/stdlib/stream/zip.effekt
```diff
@@ -0,0 +1,12 @@
+import stream
+
+def main() = {
+  def stream1() = [1, 2, 3].each
+  def stream2() = array::fromList(["a", "b", "c", "d"]).each
+
+  with val tup = for[(Int, String)] {
+    zip[Int, String] { stream1 } { stream2 }
+  }
+  val (a, b) = tup
```
jiribenes, line 10: Perhaps `zip` should really exist as a multi-parameter `for`?
Then the code would be 
```suggestion
  with val (a, b) = for[Int, String] { stream1 } { stream2 }
```
... but then `for` has _a lot_ of different weird overloads (we already have a `for[A] {...} {...}` and `for[A, R] {...} {...}`, with this, we'd have also at least `for[A, B] {...} {...} {...}` and `for[A, B, R] {...} {...} {...}`)

---

Why not just `with val (a, b) = for ...`? See #707 to find out!

b-studios: I think it would be interesting to revisit @bracevac's work on event correlation for great motivating examples and inspiration: 

https://bracevac.org/assets/pdf/icfp18.pdf

jiribenes:
Also, I just tried using a "for-each" on characters, but it doesn't act the way I'd expect:
```scala
import stream

def main() = {
  for[Char] { each("Hello") } { c =>
    println(show(c) ++ " (" ++ show(c.toInt) ++ ")")
  }
}
```

as it prints an infinite stream of:
```
H (72)
H (72)
H (72)
H (72)
H (72)
H (72)
H (72)
H (72)
H (72)
H (72)
...
```

examples/stdlib/stream/zip.effekt
```diff
@@ -0,0 +1,12 @@
+import stream
```
jiribenes, line 1:
```suggestion
import stream
import array
```

examples/stdlib/stream/zip.effekt
```diff
@@ -0,0 +1,14 @@
...
+import array
+
+def main() = {
+  def stream1() = rangeFrom(1)
+  // Ideally this would be an array to demonstrate the capabilities. 
+  def stream2() = ['H', 'e', 'l', 'l', 'o'].each
```
jiribenes, line 7: I'd love this to be an array like:
```suggestion
  // Ideally this would be an array to demonstrate the capabilities. 
  def stream2() = array::fromList(['H', 'e', 'l', 'l', 'o']).each
```
but it runs into the classic Valgrind issue we discussed at a meeting.

jiribenes: Or a string like:
```suggestion
  // Ideally this would be an array to demonstrate the capabilities. 
  def stream2() = "Hello".each
```
but it doesn't seem to work the way I would expect, then the zipping returns:
```
1, H
2, H
3, H
4, H
...
```
(an infinite stream of `($number, 'H')`

libraries/common/stream.effekt
phischu: This is a left-over from an earlier version, sorry.

examples/stdlib/stream/sum_of_squares.effekt
phischu: Example is now
```
def squares(): Unit / emit[Int] =
  for[Int] { rangeFrom(1) } { n =>
    do emit(n * n)
  }


def main() = {
  val max = 10
  println("The sum of squares from 1 to " ++ show(max) ++ " is:")
  println(sum {
    with limit[Int](max + 1)
    squares()
  })
}
```
going through an infinite stream. I will rename `limit` back to `take`.

libraries/common/stream.effekt
```diff
@@ -91,23 +85,27 @@ def boundary[R] { program: () => R / stop }: Option[R] =
...
+    None()
   }
 
-def exhaustively[A] { program: () => A / stop } { action: A => Unit }: Unit =
+def boundary { program: () => Any / stop }: Unit = {
+  boundary[Any]{program}
+  ()
+}
+
+def exhaustively { program: () => Any / stop }: Unit =
```
jiribenes, line 96: This comment got lost in history:
```suggestion
/// Run `program` forever until `stop` is thrown.
def exhaustively { program: () => Any / stop }: Unit =
```

phischu (approved):

effekt/jvm/src/test/scala/effekt/StdlibTests.scala
```diff
@@ -23,7 +23,9 @@ abstract class StdlibChezTests extends StdlibTests {
   override def ignored: List[File] = List(
     // Not implemented yet
     examplesDir / "stdlib" / "bytearray",
-    examplesDir / "stdlib" / "io"
+    examplesDir / "stdlib" / "io",
+    examplesDir / "stdlib" / "stream" / "characters.effekt",
+    examplesDir / "stdlib" / "stream" / "fuse_newlines.effekt"
```
jiribenes, line 28: We should make a follow up issue for this (it's just Valgrind problems with strings...)

examples/benchmarks/input_output/word_count_ascii.effekt
```diff
@@ -28,7 +28,9 @@ def countWords(): Output / read[Byte] = {
   var lines = 0
   var wasSpace = true
 
-  exhaustively { do read[Byte]() } { c =>
+  exhaustively {
+
+    val c = do read[Byte]()
```
jiribenes, line 33:
```suggestion
  exhaustively {
    val c = do read[Byte]()
```

examples/benchmarks/input_output/word_count_utf8.effekt
```diff
@@ -17,7 +17,9 @@ def countWords(): Output / read[Char] = {
   var lines = 0
   var wasSpace = true
 
-  exhaustively[Char] { do read() } { c =>
+  exhaustively {
+
+    val c = do read[Char]()
```
jiribenes, line 22:
```suggestion
  exhaustively {
    val c = do read[Char]()
```

jiribenes (commented):
LGTM (modulo small whitespace changes)

effekt/jvm/src/test/scala/effekt/StdlibTests.scala
jiribenes: => #715.

examples/stdlib/stream/zip.effekt
jiribenes: => #716.

PR #706, opened 2024-11-22
Fix/static arguments

b-studios:


effekt/shared/src/main/scala/effekt/core/StaticArguments.scala
```diff
@@ -96,7 +97,7 @@ object StaticArguments {
         dropStatic(staticB, blockLit.cparams),
         dropStatic(staticV, blockLit.vparams),
         dropStatic(staticB, blockLit.bparams),
-        rewrite(blockLit.body)
+        rewrite(blockLit.body)(using enterFunction(id))
```
b-studios, line 100: The recursive calls were never rewritten and thus all programs had the form:

```
def wrapper() = {
  def worker() = {
    ... wrapper() ...
  }
  worker()
}
```
which resulted in a useless `worker`.

effekt/shared/src/main/scala/effekt/core/StaticArguments.scala
```diff
@@ -203,20 +204,20 @@ object StaticArguments {
     val statics: Map[Id, IsStatic] = recursiveFunctions.map {
       case (id, RecursiveFunction(BlockLit(tparams, cparams, vparams, bparams, body), targs, vargs, bargs)) =>
         val isTypeStatic = tparams.zipWithIndex.collect {
-          case (param, index) => targs.map(args => args(index)).forall {
+          case (param, index) => targs.nonEmpty && targs.map(args => args(index)).forall {
```
b-studios, line 207: We transformed also functions that do not have any recursive calls, so I now explicitly check whether there are arguments at all.

effekt/shared/src/main/scala/effekt/core/StaticArguments.scala
```diff
@@ -30,7 +30,8 @@ object StaticArguments {
 
   def hasStatics(id: Id)(using ctx: StaticArgumentsContext): Boolean = ctx.statics.get(id).exists {
     // types ALL need to be static
-    case IsStatic(types, values, blocks) => types.forall(x => x) && (values.exists(x => x) || blocks.exists(x => x))
+    case s  @ IsStatic(types, values, blocks) =>
+      types.forall(x => x) && (values.exists(x => x) || blocks.exists(x => x))
```
b-studios, line 34: This is just a whitespace change.

PR #713, opened 2024-11-25
Add first draft of instrumented interpreter

b-studios:
In this PR I develop a tracing interpreter for Core. This will allow us to:

- check whether optimizations actually optimize things
- measure different aspects of the evaluation as part of CI and make sure that changes do not invalidate optimizations
- implement a debugger / visualizer in the long run

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/core/InterpreterTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/core/InterpreterTests.scala
@@ -1,0 +1,304 @@
+package effekt
+package core
+
+import effekt.core.Interpreter.{ InterpreterError, State }
+import effekt.source.FeatureFlag
+import effekt.symbols.QualifiedName
+import effekt.symbols.given
+import kiama.util.FileSource
+
+class InterpreterTests extends munit.FunSuite {
+
+  import effekt.context.{ Context, IOModuleDB }
+  import effekt.util.AnsiColoredMessaging
+  import kiama.util.{ Positions, StringSource }
+
+  //  object driver extends effekt.Driver
+  //
+  //  def run(content: String): String =
+  //    var options = Seq(
+  //      "--Koutput", "string",
+  //      "--backend", "js",
+  //    )
+  //    val configs = driver.createConfig(options)
+  //    configs.verify()
+  //
+  //    val compiler = new TestFrontend
+  //    compiler.compile(StringSource(content, "input.effekt"))(using context).map {
+  //      case (_, decl) => decl
+  //    }
+  //    configs.stringEmitter.result()
+  val positions = new Positions
+  object ansiMessaging extends AnsiColoredMessaging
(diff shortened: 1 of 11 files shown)


effekt/jvm/src/test/scala/effekt/core/InterpreterTests.scala
```diff
@@ -0,0 +1,304 @@
...
+      |  }
+      |
+      |def main() = benchmark(10){run}
+      |
+      |
+      |""".stripMargin
+
+  // doesn't work: product_early (since it SO due to run run run)
+
+  val Some(main, mod, decl) = runFile("examples/benchmarks/effect_handlers_bench/triples.effekt"): @unchecked
```
b-studios, line 237: You can already experiment with it by changing the filename here and then running `testOnly effekt.core.InterpreterTests`

b-studios: For example, here is the result of running `triples`:

```
Static dispatches: 232
Dynamic dispatches: 231
Branches: 701
Pattern matches: 5
Frames (pushed: 405, popped: 530)
Allocations: 5
Closures: 112
Field lookups: 0
Variable reads: 0
Variable writes: 0
Installed delimiters: 1
Captured continuations: 347
Resumed continuations: 350
```

effekt/jvm/src/test/scala/effekt/core/InterpreterTests.scala
```diff
@@ -0,0 +1,304 @@
...
+
+  try {
+    object data extends Counting {
+      override def step(state: Interpreter.State) = state match {
+        case State.Done(result) => ???
+        case State.Step(stmt, env, stack) =>
+          //println(Interpreter.show(stack))
+      }
+    }
+    Interpreter(data).run(main, inlined)
```
b-studios, line 251: By changing this to 

```suggestion
    Interpreter(data).run(main, gced)
```

you can compare the numbers to those without inlining.

effekt/shared/src/main/scala/effekt/core/Inline.scala
```diff
@@ -200,7 +200,7 @@ object Inline {
   }
 
   def rewrite(e: Expr)(using InlineContext): Expr = e match {
-    case DirectApp(b, targs, vargs, bargs) => directApp(rewrite(b), targs, vargs.map(rewrite), bargs.map(rewrite))
+    case DirectApp(b, targs, vargs, bargs) => directApp(b, targs, vargs.map(rewrite), bargs.map(rewrite))
```
b-studios, line 204: TBH I forgot why I removed the `rewrite` here...

b-studios:
TODO we could also run polymorphism boxing in the tests in order to count how often things are boxed...

b-studios:
This is very much not finished. However, I am running out of time and this is useful already as is, so I am gonna merge this soon...

PR #718, opened 2024-11-27
Fix default arguments of benchmarks

marvinborner:
While working on https://github.com/effekt-lang/effekt-plots/pull/10, I've realized that the default arguments for the benchmarks are not working as expected (mostly because of formatting issues that we somehow missed).

Also, we need to consider a different argument for `effect_handlers_bench/resume_nontail` because the JS backend can't handle `10000` (for now I've set it to 1000)


b-studios:
Right now I think it would be great to have the same numbers for JS and LLVM if possible. This way we could compare much better. If the numbers are the "standard" number (like from effect-handlers-bench), even better.

marvinborner:
In the attached plots PR we now also support the are_we_fast_yet JS reference benchmarks. I think we should only merge this if we settled on default arguments here -- notably the ones that overflow in JS.

Otherwise we could also remove all the lines that are not relevant for are_we_fast_yet. This would allow us to compare this data and would not cause overflows since its default args are quite small.

PR #720, opened 2024-11-30
HTTP client in jsNode and jsWeb

jiribenes:
This is _very_ preliminary, extremely buggy, but at least `GET https://effekt-lang.org` seems to work on Node.js...

This PR exists only just to be adopted and fixed by someone who actually needs/wants to use it for something. :)

TODO:
- [ ] jsWeb
- [ ] make the Effekt API nicer to use
- [x] add support for custom headers
- [ ] reduce the native API area
- [ ] solve the TODOs in code
- [ ] LLVM? (via `libuv`)

### DIFF
--- a/libraries/common/io/client.effekt
+++ b/libraries/common/io/client.effekt
@@ -1,0 +1,161 @@
+/// HTTP Clients: send a request via HTTP, get a response.
+module io/client
+
+import bytearray
+import io
+
+// TODO: mutable (native) map of headers instead?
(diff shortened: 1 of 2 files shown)


jiribenes:
Once again, it would be very nice to support HTTP requests: it would unlock many potential fun use-cases :)

jiribenes:
Stale, closing (would probably need to rewritten anyways)

PR #561, opened 2024-08-30
Add immutable ordered finite 'map' and 'set' to the stdlib

jiribenes:
I've been sitting on this code for almost ~~9~~ _13_ months now (started on Xmas 2023), it's about time we finalise it.
This PR adds immutable ordered maps and sets based on weight balanced trees into the standard library.

The `map` and `set` now work on all main backends: JS, LLVM, Chez.

Progress ~~is **blocked**~~ was blocked on:
1. ~~very annoying inliner bug(s?) [#733]~~ _resolved_ 🥳 
2. ~~missing comparisons on strings on LLVM [#748]~~ _resolved_ 🎉

PR #725, opened 2024-12-03
Uncomment rewrite

marvinborner:
As discussed in https://github.com/effekt-lang/effekt/pull/561#issuecomment-2513199799


jiribenes (approved):
:shipit:

PR #728, opened 2024-12-07
Add new escapes to 'string/tty', rename to 'tty', deprecate `ANSI` namespace

jiribenes:
I've been sitting on these changes for a good while now, so let's upstream 'em.

Among others adds:
- ANSI256 colours  Screenshot 2024-12-07 at 18 37 24
- all sorts of combinators for manipulating the terminal: change title, move cursor, clear line, clear screen, hyperlink, ...

Be warned, this might break some users who rely on escapes directly.

Bikeshedding around names/namespaces/organization here is welcome!

In an ideal world, these would be the TODOs:
- [ ] add full RGB colours (terminals _do_ support them, we could even parse `#rrggbb` specifiers!)
- [ ] add a better way to format without accidental capture (perhaps use the newer escapes for bold etc. which allow cancelling?)
- [ ] test things out properly?

### DIFF
--- a/libraries/common/string/tty.effekt
+++ b/libraries/common/string/tty.effekt
@@ -1,45 +1,163 @@
 module string/tty
 
-// ANSI escape codes
-namespace ANSI {
-  val CSI = "\u001b["
-
-  def escape(s: String) = CSI ++ s ++ "m"
-
-  val BLACK = escape("30")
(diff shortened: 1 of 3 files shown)


jiribenes:
=> move to `tty`

PR #734, opened 2024-12-10
Add UTF-8 console encoding support for Windows

### LINKED ISSUE #726: Cannot print fancier characters with Node.js on Windows in VSCode
On the JS backend, on Windows, in the VSCode terminal (Powershell):

```scala
println("🚩")
```
gets printed as `?`:
image

However, `echo "🚩"` in the very same shell just works.

---

Needs more investigating by someone with an access to a Windows machine to narrow down the problem.

zaryar:
Fixes #726 

### Problem
When displaying Unicode characters (like chess pieces ♔) on Windows systems, the console shows '?' instead of the actual characters. This works correctly on Linux systems.
I saw the post in the effective programming with effekt course. I had the same problem because I want to create chess with effekt.

### Solution
Added Windows-specific console configuration:
- Set console codepage to 65001 (UTF-8) using `chcp` command
- Configure stdout to use UTF-8 encoding

### Notes
This change only affects Windows systems and has no impact on other platforms.

### DIFF
--- a/libraries/common/io/console.effekt
+++ b/libraries/common/io/console.effekt
@@ -46,7 +46,7 @@ namespace js {
   extern async def readLine(console: JSConsole): String =
     jsNode "$effekt.capture(k => ${console}.once('line', k))"
   extern io def writeLine(console: JSConsole, message: String): Unit =
-    jsNode """${console}.write(${message} + '\\n', 'utf8')"""
+    jsNode "${console}.output.write(${message} + '\\n')"
 }
 
 namespace examples {
--- a/libraries/common/io/console.effekt
+++ b/libraries/common/io/console.effekt
@@ -24,6 +24,11 @@ namespace js {
 
   extern jsNode """
     const readline = require('node:readline');
+    if (process.platform === 'win32') {
+      // Set UTF-8 codepage
+      require('child_process').execSync('chcp 65001');
+      process.stdout.setDefaultEncoding('utf8');
+    }
   """
 
   extern type JSConsole
@@ -32,6 +37,7 @@ namespace js {
     jsNode """readline.createInterface({
         input: process.stdin,
         output: process.stdout,
+        encoding: 'utf8',
       })"""
 
   extern io def close(console: JSConsole): Unit =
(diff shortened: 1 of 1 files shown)


libraries/common/io/console.effekt
```diff
@@ -24,6 +24,11 @@ namespace js {
 
   extern jsNode """
     const readline = require('node:readline');
+    if (process.platform === 'win32') {
+      // Set UTF-8 codepage
+      require('child_process').execSync('chcp 65001');
```
jiribenes, line 29: Just to make sure, is this call truly necessary? (Can you check: does it work without it?)

Is there an alternate way to call this directly via some call that doesn't `exec` a whole new process?

jiribenes (commented):
Hi, thanks for the contribution, this is wonderful!
Could you take a look at my comment? Is there anything we can do to minimise the side-effectful calls?

libraries/common/io/console.effekt
zaryar: If I do it like this:
```
    if (process.platform === 'win32') {
      // Set UTF-8 codepage
      // require('child_process').execSync('chcp 65001');
      process.stdout.setDefaultEncoding('utf8');
    }
```
It sadly does not work for me, maybe I can find another way of doing it tho :)

libraries/common/io/console.effekt
```diff
@@ -24,6 +24,11 @@ namespace js {
...
+      process.stdout.setDefaultEncoding('utf8');
```
jiribenes, line 30: 👀 I _think_ something like this would belong to a `jsNode` block above the definition of `println` in `libraries/common/effekt.effekt` to ensure that the output of `println` itself is UTF-8. Otherwise this line of code only gets interpreted if someone imports `console`...

jiribenes:
Could you perhaps try to set up your cmd process according to this guide? https://stackoverflow.com/a/57134096
Then the `require('child_process').execSync('chcp 65001');` line shouldn't be needed anymore.
If it works, we could add it to the install guide on the website or something 🤔

zaryar:
> Could you perhaps try to set up your cmd process according to this guide? https://stackoverflow.com/a/57134096 Then the `require('child_process').execSync('chcp 65001');` line shouldn't be needed anymore. If it works, we could add it to the install guide on the website or something 🤔


sorry was gone the last few days, will try :)

jiribenes:
Since that works, I think that the next move would be to close the PR and update the Getting Started guide in the Windows section :)

PR #711, opened 2024-11-23
Make 'list::sort' quick by using natural mergesort

jiribenes:
Please see issue #710 for context and analysis.

It would be nice if somebody benchmarked this properly: the old quicksort, this new quicksort on JS and LLVM + with JS native sort, but I can notice the improvement as this version can suddenly quickly sort huge random lists.

Other considerations:
- we could do some hybrid sort for smaller lists (and it would probably speed it up!), but I don't want the mental overhead of _two different_ sorting algorithms
- I'm pretty limited here by Effekt not allowing mutually recursive subfunctions -- this means that I need to pass `compare` everywhere manually.
- Not stacksafe (but works fine for ~5M random elements, so it's good enough for me!)

PR #736, opened 2024-12-10
Fix some memory leaks

phischu:
@marvinborner With these fixes the following does not leak anymore:

```
import array

def main() = {
  array::allocate[Int](1)
  ()
}
```

Does this also fix other tests that are commented out?


marvinborner:
No, the other leaking tests are not fixed by this entirely. My additional pos erasers in the stdlib make Valgrind even angrier.

phischu:
Could you please add your additional pos erasers in the stdlib to this pull request?

marvinborner:
I've added them. I'm still not 100% sure when we are supposed to add erasers and whether my additions are correct. We should really start writing a detailed sharer/eraser guide :)

phischu:
Thank you. I have applied some fixes, please check if we can enable more tests for llvm now.

marvinborner:
This seems to improve things by a lot! There are still some tests left that fail with segfaults (use-after-free most probably). I've pushed the uncommented tests to see some immediate CI response on further fixes.

jiribenes:
Does this fix https://github.com/effekt-lang/effekt/pull/711 ?

phischu:
`examples/stdlib/io/filesystem/files.effekt (llvm)` and `examples/stdlib/io/filesystem/async_file_io.effekt (llvm)` fail with "Syscall param write(buf) points to uninitialised byte(s)", @marvinborner could you investigate? I'll try to hunt down the other ones.

b-studios (approved):

PR #739, opened 2024-12-11
Implement missing math functions in LLVM backend

### LINKED ISSUE #608: Implement missing math functions on the LLVM backend
The following math functions/values in `libraries/common/effekt.effekt` are _not_ implemented in the LLVM backend yet:
- [ ] `cos: Double => Double`
- [ ] `sin: Double => Double`
- [ ] `atan: Double => Double`
- [ ] `tan: Double => Double`
- [ ] `log: Double => Double`
- [ ] `log1p: Double => Double`
- [ ] `exp: Double => Double`
- [ ] `_pi: Double`

See #606 on how to do this, don't forget to add tests (or re-enable tests ignored on the LLVM backend).
Here's the list of all LLVM intrinsics: https://llvm.org/docs/LangRef.html#standard-c-c-library-intrinsics

mattisboeckle:
Implement cos, sin, atan, tan, log, log1p, _pi, exp & pow. All functions use the LLVM intrinsics except tan & atan. They do not exist in the version of LLVM we are using.

Fixes #608


libraries/common/effekt.effekt
```diff
@@ -352,14 +357,20 @@ def abs(n: Double): Double =
...
   chez "(log ${x})"
+  llvm "%z = call %Double @llvm.log.f64(double ${x}) ret %Double %z"
 
 extern pure def log1p(x: Double): Double =
   js "Math.log1p(${x})"
   chez "(log (+ ${x} 1))"
+  llvm """
+    %i = fadd %Double 1.0, ${x}
+    %z = call %Double @llvm.log.f64(double %i) ret %Double %z
+  """
```
jiribenes, line 368: FWIW, as far as I remember, `log1p` is in libc as a separate function for numerical precision reasons. I know [MLIR has it as a separate intrinsic](https://mlir.llvm.org/docs/Dialects/MathOps/#mathlog1p-mathlog1pop), but I don't know about LLVM.
I'd either use either that ^ or just with a `default { log(1.0 + x) }` instead of the Chez and LLVM externs?

examples/pos/math.effekt
```diff
@@ -0,0 +1,10 @@
+def main() = {
+    println(cos(0.0));
+    println(sin(0.0));
+    println(log(1.0));
+    println(log1p(0.0));
+    println(exp(0.0));
+    println(pow(2.0, 5.0));
+    println(tan(0.0));
+    println(atan(0.0));
+}
```
jiribenes, line 10: Nitpick: we tend to use two spaces to format Effekt files. :)

jiribenes (commented):
Hi @mattisboeckle, thanks for the great first contribution!

I have a few maintenance discussion topics below -- even though the contribution is great and already mergeable, it would be nice to make sure that the definitions we want to use are maintainable in the future. I'm interested in your opinion as to how we should act regarding the more complex functions / constants that are not available directly as intrinsics before we proceed :)

libraries/common/effekt.effekt
```diff
@@ -373,11 +384,18 @@ def pow(base: Double, exponent: Int): Double = {
...
 // since we do not have "extern val", yet
 extern pure def _pi(): Double =
   js "Math.PI"
   chez "(* 4 (atan 1))"
+  // We have to use the non-intrinsic version of 'atan' as it was only added in LLVM 19
+  llvm """
+    %i = call %Double @atan(double 1.0)
+    %z = fmul %Double %i, 4.0
+    ret %Double %z
+  """
```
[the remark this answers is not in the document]
mattisboeckle: Just checked in a small test file. LLVM will optimize this function to a constant.

The other options work as well, not sure which I prefer

libraries/common/effekt.effekt
mattisboeckle: `log1p` does not appear in the LLVM docs for intrinsics, so I defaulted to this

examples/pos/math.effekt
mattisboeckle: Nitpick accepted. Although I did copy the `doubles.effekt` file which uses 4 spaces 🤷

libraries/common/effekt.effekt
jiribenes: => use libc's `log1p`?

libraries/common/effekt.effekt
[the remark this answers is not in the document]
mattisboeckle: Now just uses the constant. I used the largest one that still fits in a double

libraries/common/effekt.effekt
```diff
@@ -343,22 +343,38 @@ extern pure def infixDiv(x: Double, y: Double): Double =
 extern pure def cos(x: Double): Double =
   js "Math.cos(${x})"
   chez "(cos ${x})"
+<<<<<<< HEAD
   vm "effekt::cos(Double)"
+=======
+  llvm "%z = call %Double @llvm.cos.f64(double ${x}) ret %Double %z"
+>>>>>>> 43ee27b6 (Implement missing math functions in LLVM backend)
```
jiribenes, line 350: 👀 `:(`

jiribenes (approved):

Issue #732, opened 2024-12-10
Runtime compiler exception on LLVM backend with `--no-optimize`


jiribenes:
The following somewhat minimised example crashes the LLVM backend when used with `--no-optimize`:
```scala
import bytearray

// === inlined `stream` ===
effect read[A](): A / stop
effect stop(): Nothing

def feed[R](bytes: ByteArray) { reader: () => R / read[Byte] }: R = {
  var i = 0
  try {
    reader()
  } with read[Byte] {
    resume {
      if (i < bytes.size) {
        val c = i
        i = c + 1
        bytes.unsafeGet(c)
      } else {
        do stop()
      }
    }
  }
}

def decodeChar(): Char / {read[Byte], stop} = {
  val b = do read().toInt
  b.toChar // simplified!
}

def decodeUTF8[R] { reader: () => R / read[Char] }: R / read[Byte] =
  try {
    reader()
  } with read[Char] {
    resume { decodeChar() }
  }

// === the test case itself ===
// (feed from stdlib, just parametric over the return type)
def feedString[R](input: String) { reader: () => R / read[Char] }: R = {
  with feed(input.fromString)
  with decodeUTF8

  reader()
}

// note that `solve` doesn't even use `read`!
def solve(): Int / read[Char] = 42

def main(): Unit = {
  val sampleInput = "hello world"
  println(feedString(sampleInput) { solve })
}
```

results in (truncated):
```
$ effekt --no-optimize --backend=llvm repro.effekt
Exception in thread "main" java.lang.RuntimeException: Cannot find block info for b_tmp_4172.
HashMap(show -> Definition(List(),List(Variable(value_13,Int()))), read -> Parameter(Interface(read,List(Data(BoxedChar,List())))), read -> Parameter(Interface(read,List(Data(BoxedByte,List())))), feedString -> Definition(List(),List(Variable(input_2755,String()), Variable(reader_2756,Negative()))), g -> Parameter(Interface(stop,List())), infixLt -> Definition(List(),List(Variable(x_173,Int())...
```

However, the program does work on JS (with and without opt) and on LLVM with opt, just the one case with `--no-optimize --backend=llvm` doesn't work.

PR #741, opened 2024-12-11
Make sure unique name is actually fresh

phischu:
This should fix the failing program on effekt plots. The whole symbol, name, id, stuff needs redesign.


jiribenes:
As far as I can tell, this does _not_ fix #732, I still get:
```
$ effekt --no-optimize --backend=llvm repro.effekt
Exception in thread "main" java.lang.RuntimeException: Cannot find block info for b_tmp_3676.
```

PR #743, opened 2024-12-12
Add basic string interpolation / string templates

### LINKED ISSUE #722: Basic string templates / interpolation in normal strings
Currently, we don't parse string templates in normal strings.

As a first step, we could allow at least arguments which are already strings like:
```scala
val domain = "https://api.effekt-lang.org"
val user = "effekt"
val resourceId = 42

"GET ${domain}/users/${user}/resource/${resourceId.show}"
```

This would involve:
1. changing the parser (allowing `${ ... }` to appear in normal strings)
2. allowing string literals to be `Template[...]` in `source.Tree`
3. piping them through Namer
4. in Typer, making sure all of the splices are `String`s
5. and then when transforming into Core, just concatenate them for now into a Core string literal

There are some interesting detours/shortcuts:
- try using `show` on the splices instead (probably horrible to use because of the error messages)
- desugar to string concatenation directly in the parser
- have the _templates_ as a primitive instead of string concatenation; just define `def infixConcat(s1, s2) = "${s1}${s2}"`

dvdvgt:
Fixes #722 and adds basic support for string templates. 

There are a few caveats to keep in mind with the current initial implementation:
- All arguments already have to be strings. It is basically just syntactic sugar so that you can write `"${a.show} and ${b.show}"` instead of `a.show ++ " and " ++ b.show` where `a` and `b` are some 'showable' values. Internally, however, this is just how it is desugared.
- String templates cannot be used to match string literals in pattern matches
- The generated JS code looks a bit weird:
  ```
  const v_r_897 = $effekt.println(((((((((("GET ") + (domain_0)))) + (((("/users/") + (user_0))))))) + (((("/resource/") + (('' + resourceId_0))))))));
  ```
+ If one forgets to call `show` on arguments which are not already of type string, the error messages are hideous and possibly confusing:
<details>
<summary>error</summary>

```
-There are multiple overloads, which all fail to check:
-Possible overload: effekt::println of type Bool => Unit
-  Expected String but got Int.
-  Expected Bool but got String.
-
-Possible overload: list::println of type List[Bool] => Unit
-  Expected String but got Int.
-  Expected List[Bool] but got String.
-
-Possible overload: array::println of type Array[Int] => Unit
-  Expected String but got Int.
-  Expected Array[Int] but got String.
-
-Possible overload: effekt::println of type String => Unit
-  Expected String but got Int.
-
-Possible overload: list::println of type List[Int] => Unit
-  Expected String but got Int.
-  Expected List[Int] but got String.
-
-Possible overload: effekt::println of type Byte => Unit
-  Expected String but got Int.
-  Expected Byte but got String.
-
-Possible overload: list::println of type List[String] => Unit
-  Expected String but got Int.
-  Expected List[String] but got String.
-
-Possible overload: list::println of type List[Double] => Unit
-  Expected String but got Int.
-  Expected List[Double] but got String.
-
-Possible overload: array::println of type Array[String] => Unit
-  Expected String but got Int.
-  Expected Array[String] but got String.
-
-Possible overload: effekt::println of type Ordering => Unit
-  Expected String but got Int.
-  Expected Ordering but got String.
-
-Possible overload: array::println of type Array[Bool] => Unit
-  Expected String but got Int.
-  Expected Array[Bool] but got String.
-
-Possible overload: effekt::println of type Unit => Unit
-  Expected String but got Int.
-  Expected Unit but got String.
-
-Possible overload: option::println of type Option[Int] => Unit
-  Expected String but got Int.
-  Expected Option[Int] but got String.
-
-Possible overload: option::println of type Option[Double] => Unit
-  Expected String but got Int.
-  Expected Option[Double] but got String.
-
-Possible overload: array::println of type Array[Double] => Unit
-  Expected String but got Int.
-  Expected Array[Double] but got String.
-
-Possible overload: effekt::println of type Double => Unit
-  Expected String but got Int.
-  Expected Double but got String.
-
-Possible overload: effekt::println of type Int => Unit
-  Expected String but got Int.
-  Expected Int but got String.
-
-Possible overload: option::println of type Option[Bool] => Unit
-  Expected String but got Int.
-  Expected Option[Bool] but got String.
-
-  println("GET ${domain}/users/${user}/resource/${resourceId}")
```
</details>

Apart from that, the changes are really lightweight because of the early desugaring in the parser.

I am looking forward to your feedback and ideas.


jiribenes:
This is a great first implementation, but I'm not sure that desugaring in parser is the best choice here :/
We talked about it at a meeting when #722 was created, but I don't remember what we agreed on.

Personally, I'd much rather see templates up until the transformation to core so that we can:
- have reasonable error messages in typer
- later on resolve typeclasses or something for the splices
- possibly have a more performant impl (?)

dvdvgt:
> Personally, I'd much rather see templates up until the transformation to core so that we can:
> - have reasonable error messages in typer
> - later on resolve typeclasses or something for the splices
> - possibly have a more performant impl (?)

You raise some very good concerns and I completely agree. I will try to go the 'non-lazy route' and I actually completely thread through templates to core.

jiribenes:
Don't get me wrong, having _a_ version is still super helpful; I truly have no idea how difficult it really is to do the thing I outlined above. But I hope it's worth it, especially wrt technical debt. :)

b-studios:
I am not completely against desugaring in parser, but maybe we could actually desugar

```
"foo ${exp1} bar ${exp2} baz"
```

to something like:

```
splice(["foo ", " bar ", "baz"], [exp1, exp2])
```
this way, as a user I can still overload `splice` for types other than `exp1: String` :)

dvdvgt:
I just pushed a prototype that threads through string templates all the way until code generation.

This has the added benefit that we remain somewhat flexible for later changes, as Jiří mentioned. Furthermore, we can re-use the string interpolation capabilities of each respective backend, e.g. currently the JS string templates are generated as [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

Note: I currently only implemented it fully JS. If we come to an agreement that this is the right way to proceed, I will implement for the remaining backends.

jiribenes:
> I am not completely against desugaring in parser, but maybe we could actually desugar
> 
> ```
> "foo ${exp1} bar ${exp2} baz"
> ```
> 
> to something like:
> 
> ```
> splice(["foo ", " bar ", "baz"], [exp1, exp2])
> ```
> 
> this way, as a user I can still overload `splice` for types other than `exp1 exp2: String` :) Not that don't have enough overloaded functions already...

This relates to some strategy we'll need to come up with in order to represent custom string interpolators (which is very outside of the scope of this PR, I believe...)

Then I'd expect some desugaring like:
```scala
val name = "Joe"
val age = 42
val s: SQL::Stmt = sql"SELECT * FROM Person p WHERE p.last_name = ${name} AND p.age = ${age}"

// ~>

val name = "Joe"
val age = 42
val s: SQL::Stmt = sql { // sql: { prog: () => Unit / { Interpolate[String], Interpolate[Int] } }: String
  do literal("SELECT * FROM Person p WHERE p.last_name = ")
  do splice(name)
  do literal(" AND p.age = ")
  do splice(age)
}
```

dvdvgt:
There are three things keeping this from merging:

- [ ] The test fails on LLVM with error code 139 (segmentation fault)
- [ ] For some reason the test now also fails on JS even though the output is apparently the same as the expected. Maybe some weird non-displayable characters are inserted by accident?
- [ ] I would like to include the `splice` module in the prelude however, I am prompted with `Effekt Compiler Crash: null`

jiribenes:
> I would like to include the `splice` module in the prelude however, I am prompted with `Effekt Compiler Crash: null`
I'll merge #780 ASAP, then it should at least tell you what's wrong :)

libraries/common/stringbuffer.effekt
```diff
@@ -35,7 +36,10 @@ def stringBuffer[A] { prog: => A / StringBuffer }: A = {
       resume(())
     }
     def flush() = {
+      // resize buffer to strip trailing zeros that otherwise would be converted into 0x00 characters
+      buffer = bytearray::resize(buffer, pos)
       val str = buffer.toString()
+      // after flushing, the stringbuffer should be empty again
       buffer = bytearray::allocate(initialCapacity)
```
jiribenes, line 43: I think this could be made somewhat more efficient -- `resize` copies into a newly allocated bytearray, but `toString` doesn't (since on LLVM strings are bytearrays)
So it should be possible to just _clear_ the buffer without allocating it again?
```suggestion
      // resize buffer to strip trailing zeros that otherwise would be converted into 0x00 characters
      val str = bytearray::resize(buffer, pos).toString()
      // after flushing, the stringbuffer should be empty again
      buffer = buffer.fill(0)
```
where `buffer.fill(0)` means `memset(buffer, 0)` (which we don't have yet) :)

libraries/common/stringbuffer.effekt
```diff
@@ -19,7 +19,8 @@ def stringBuffer[A] { prog: => A / StringBuffer }: A = {
     else {
       // Double the capacity while ensuring the required capacity
       val newSize = max(buffer.size * 2, buffer.size + sizeToAdd)
-      buffer = bytearray::resize(buffer, newSize)
+      val newBuffer = bytearray::allocate(newSize)
+      buffer = buffer.resize(newSize)
```
jiribenes, line 23: Similarly here: `resize` already allocates, right?
https://github.com/effekt-lang/effekt/blob/26c642df51bcdbb22186e16069d7e523d4717a85/libraries/common/bytearray.effekt#L40-L41

... also, you never use `newBuffer` anyway? :^)

libraries/common/stringbuffer.effekt
dvdvgt: Yes, you could just clear the buffer, however, then the buffer's size will not be reset to a (potentially) smaller size and instead just buffer's previous size is kept. This probably would not be a big deal, though.

b-studios:
Regarding the prelude, we have an undocumented invariant: all imports of prelude modules have to be in the prelude as well...

That is, the prelude in the JS runner needs to look like this:

```
override def prelude: List[String] = List("effekt", "option", "list", "result", "exception", "array", "string", "ref", "bytearray", "stringbuffer", "splice")
```
and `bytearray`, and `stringbuffer` also need to import `effekt` and dependencies that they need from the prelude manually.

dvdvgt:
For some reason `104.toByte` now fails in the tests.

effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
```diff
@@ -37,6 +37,9 @@ abstract class ChezSchemeTests extends EffektTests {
     examplesDir / "pos" / "unsafe_cont.effekt",
     examplesDir / "pos" / "propagators.effekt",
 
+    // bytearray is not implemented
+    examplesDir / "pos" / "string_interpolation.effekt",
+
```
jiribenes, line 42: With #791 merged, you should be able to remove this after rebasing on `master` :)
```suggestion
```

b-studios:
With #803 merged, maybe this PR is unblocked?

marvinborner:
There don't seem to be any segfaults but the relevant files are failing with overload errors.

---

Interestingly, the `.toByte` error only happens when the file is called `bytearray.effekt` and not when, for example, `bytearray2.effekt`!

`[error] examples/stdlib/bytearray/bytearray.effekt:6:22: Cannot resolve function toByte, called on an expression.`

b-studios:
Sounds like double import or something of the standard library bytearray, which is now part of the prelude.

dvdvgt:
Funnily, when I run `sbt test`, the tests fail for me locally as well. But when I only run `sbt testOnly effekt.JavaScriptTests` for example, no failures occur.

I continue to be at a loss to figure out what problem seems to be here, and I sadly do not have too much time to investigate the cause at the moment. Removing unneeded `import bytearray` does not seem to be the fix.

b-studios:
I'm gonna go ahead and merge now. Thanks @dvdvgt, that's great!

Next up, overloaded list literals:

```
collectList[1, 2, 3, 4]
```
~>
```
collectList { do emit(1); do emit(2); do emit(3); do emit(4) }
```

PR #730, opened 2024-12-09
Cherry-pick inlining fix from #725

jiribenes:
As discussed in https://github.com/effekt-lang/effekt/pull/561#issuecomment-2513199799, there's a missing rewrite.
This fix is necessary for the `map` and `set` examples, as well as some of my AoC solutions using them, so I'd like to upstream it, if possible.

PR #745, opened 2024-12-12
Revert "Cherry-pick inlining fix from #725"

jiribenes:
Reverts effekt-lang/effekt#730 as it seems like an incorrect fix, judging both by the Effekt Working Group meeting, and the graphs after merging it:

image
image


dvdvgt:
Does #732 reappear with this change?

dvdvgt (approved):
LGTM; #732 seems to remain fixed thanks to #741 (tested locally)

PR #755, opened 2024-12-28
CI: Prevent unnecessary release

jiribenes:
Don't release when there are no new changes.
Also has a special check to not release when the only changes were in `.github/`.


jiribenes:
I'll merge this anyways before Monday since because we have no changes on `master` currently so there's nothing much that could go wrong here...

Issue #715, opened 2024-11-25
Fix Valgrind issues in 'stream' examples


jiribenes:
In #705, we disabled two of the tests because of Valgrind problems:
https://github.com/effekt-lang/effekt/blob/c6e22028de6206ac68508c3717ef42226253b046/effekt/jvm/src/test/scala/effekt/StdlibTests.scala#L49-L50

It would be nice to get them working again. :)

marvinborner:
Both tests leak bytearrays. This comes down to a similar issue with array allocations that [we've had for a long time](https://github.com/effekt-lang/effekt/pull/568#issuecomment-2340933313). Several wrapper functions call the semi-internal allocation function (`...array_new`) directly, but never erase the allocated objects.

That's why:

``` scala
import bytearray

// leaks 17 byte
def main() = {
  bytearray::allocate(1)
  ()
}
```

as well as

``` scala
import array

// leaks 40 byte
def main() = {
  val arr: Array[Int] = array::allocate(1)
  ()
}
```

I'm still not sure how we want to fix this specifically. We must make it a lot clearer which (C) functions are internal and could leak, so we don't use them accidentally.

PR #757, opened 2025-01-03
Run no-valgrind tests, except those that segfault

b-studios:
Some of the tests that were ignored because of valgrind actually segfaulted, so I now separate valgrind failures from actual bugs.


b-studios:
Currently, this doesn't work since it is not valgrind that is complaining, but the address sanitizer... 

@phischu argues that we do not need this PR if we merge #715

b-studios:
TBH I am not the best to work on this since on my M1 I can neither run valgrind, nor does my address sanitizer complain in these cases...

b-studios:
Closing for now.

PR #763, opened 2025-01-07
Improve dead code elimination to remove unused cases and operations

b-studios:


b-studios:
This should only affect code size and not performance (very much at least, except it might decrease closure size).

before
```
wc  out/anf.js
    2663    8907   99199 out/anf.js
```
after
```
wc  out/anf.js
    2294    7995   89872 out/anf.js
```

PR #765, opened 2025-01-08
Drop Select from core and add Run to CPS

b-studios:


b-studios:
This is a draft to tries to unblock #711 by translating `Select` away. 

I am also experimenting with `Run` in CPS, but will probably move this to a separate PR to merge this one after cleanup.

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -292,13 +287,12 @@ class VMTests extends munit.FunSuite {
...
+      patternMatches = 455,
       branches = 56,
       pushedFrames = 80,
       poppedFrames = 80,
       allocations = 31,
       closures = 0,
-      fieldLookups = 455,
       variableReads = 34,
       variableWrites = 30,
       resets = 0,
```
b-studios, line 298: Not surprisingly, the fieldLookups now are pattern matches

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -706,13 +677,12 @@ class VMTests extends munit.FunSuite {
     examplesDir / "casestudies" / "inference.effekt.md" -> Some(Summary(
       staticDispatches = 1457444,
       dynamicDispatches = 3201452,
-      patternMatches = 1474376,
+      patternMatches = 1474290,
       branches = 303298,
-      pushedFrames = 7574572,
-      poppedFrames = 6709277,
+      pushedFrames = 7574480,
+      poppedFrames = 6709185,
```
b-studios, line 683: This is interesting: we save a few pushedFrames by commuting match and val.

b-studios:
Ok, this is ready. However, I do not support existential type parameters on constructors.

PR #766, opened 2025-01-08
Experiment: calculate test coverage and send it to Codecov

jiribenes:
I tried setting up Codecov (free for public repos) so that we have some access to test coverage metrics.
Not sure if it even works or if we want this, hence the experiment label. :)

For each new PR after merging this, it should generate a report of how the coverage changed with the incoming patch, thus giving us (ignorable) signal as to what changes are not tested.
See the [PRs in Polarity](https://github.com/polarity-lang/polarity/pulls) for an idea of how this looks like there:
Screenshot 2025-01-08 at 18 12 47

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,8 +90,10 @@ jobs:
     - name: Run tests with coverage
       run: EFFEKT_VALGRIND=1 EFFEKT_DEBUG=1 sbt clean coverage test coverageReport
 
-    - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v4
+    - name: Upload test coverage to Codecov
+      uses: codecov/codecov-action@v5
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
 
     - name: Assemble fully optimized js file
       run: sbt effektJS/fullOptJS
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -87,8 +87,13 @@ jobs:
       with:
         node-version: ${{ env.NODE_VERSION }}
 
-    - name: Run tests
-      run: EFFEKT_VALGRIND=1 EFFEKT_DEBUG=1 sbt clean test
+    - name: Run tests with coverage
+      run: EFFEKT_VALGRIND=1 EFFEKT_DEBUG=1 sbt clean coverage test coverageReport
```
jiribenes, line 91: As a downside, this could make the tests _much_ slower :'(
In that case, we could also running tests with coverage separately to reduce the turnaround.

.github/workflows/ci.yml
```diff
@@ -87,8 +87,13 @@ jobs:
...
+
+    - name: Upload test coverage to Codecov
+      uses: codecov/codecov-action@v5
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
```
jiribenes, line 96: As far as I know, this should _not_ fail builds ever, just upload the coverage report...

jiribenes:
Although this is a nice idea, I'm not planning on debugging the compile errors here, so I'm closing the PR.
We can pick the experiment up at a later date :)

PR #768, opened 2025-01-09
Store sharer and eraser in function prefix

serkm:
This reduces the size of frame headers for faster PushFrame/Return.
Non-linear handlers get slower though, because now there is an additional indirection for sharers/erasers.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala
@@ -12,7 +12,13 @@ object PrettyPrinter {
     definitions.map(show).mkString("\n\n")
 
   def show(definition: Definition)(using C: Context): LLVMString = definition match {
-    case Function(callingConvention, returnType, name, parameters, basicBlocks) =>
+    case Function(callingConvention, returnType, name, parameters, Some(prefix), basicBlocks) =>
+      s"""
+define ${show(callingConvention)} ${show(returnType)} ${globalName(name)}(${commaSeparated(parameters.map(show))}) prefix ${show(prefix)} {
+    ${indentedLines(basicBlocks.map(show).mkString)}
+}
+"""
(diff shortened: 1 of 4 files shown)


serkm:
I also moved vtables to function prefixes.
This removes one indirection for the first operator of an object.

Here are some benchmark results:
image

b-studios:
How does this lead to an additional indirection? I would have assumed it would just be an additional offset computation.

serkm:
Previously it was:
Load stack pointer -> add offset -> load sharer address

Now it is:
Load stack pointer -> add offset -> load return address -> add offset -> load sharer address

b-studios:
Hmm... I expected it to be

```
Load stack pointer -> add offset -> load return address -> add offset
```
but maybe my mental model of how LLVM works is just off here...

serkm:
I didn't see an easy way to place functions relative to each other.
I placed function _pointers_ relative to the return address.

b-studios:
ok, then the additional indirection makes sense, thanks for the explanation

b-studios:
Is there a way to make the linker happy?

serkm:
This should be ready now. It turns out the linker doesn't like pie :)
Do we want to give up on position independence though?

b-studios:
Are the benchmark results from above still up to date?

b-studios:
What are the most important down sides in your opinion of giving up position independence?

serkm:
Updated benchmarks:
image

serkm:
I don't know really know about the downsides. According to claude, there are some security vulnerabilities with position dependence.
There might also be issues if we ever want to compile to shared libraries.

serkm:
Without `-no-pie` the linker complains, but the tests still succeed.
```
/usr/bin/ld: /tmp/lto-llvm-1c61d0.o: warning: relocation in read-only section `.text.returnAddress_1616'
/usr/bin/ld: warning: creating DT_TEXTREL in a PIE
```

PR #769, opened 2025-01-10
Fix finding free variables

phischu:


examples/pos/pr769.effekt
```diff
@@ -0,0 +1,16 @@
...
+  case Cons(x, _) => println(x)
+}
+
+def updateFirst(list: List[Int]): List[Int] =
+  // list.drop(0) === list
+  // but if you replace it, everything works again
+  list.drop(0) match {
+    case Cons(x, right) =>
+      Cons(x + 100, right)
+    case l => l
```
jiribenes, line 12: Originally, the repro was like this:
```suggestion
    case _ => list
```

effekt/shared/src/main/scala/effekt/machine/Analysis.scala
```diff
@@ -24,7 +24,7 @@ def freeVariables(statement: Statement): Set[Variable] =
     case Construct(name, tag, values, rest) =>
       Set.from(values) ++ (freeVariables(rest) -- Set(name))
     case Switch(value, clauses, default: Option[Clause]) =>
-      Set(value) ++ clauses.flatMap { case (tag, branch) => freeVariables(branch) } ++ default.map(freeVariables).getOrElse(Set.empty)
+      Set(value) ++ clauses.flatMap(freeVariables) ++ default.map(freeVariables).getOrElse(Set.empty)
```
jiribenes, line 27: This is just stylistic, right?
(now it just uses the `def freeVariables(taggedClause: (Int, Clause)): Set[Variable] = freeVariables(taggedClause._2)` instance, but it's exactly the same computation)

jiribenes:
On my machine, this does unblock #711.
But it seems like the `repro` now crashes somewhere else -- maybe it needs the `if (defa) { eraseValue(...) }` somewhere?

effekt/shared/src/main/scala/effekt/machine/Analysis.scala
jiribenes: btw, IntelliJ suggests `default.fold(Set.empty)(freeVariables)`, but I'd disregard that 😇

phischu: Please refactor and deduplicate this with the other occurence

jiribenes:
Feel free to adopt this PR, I'm not planning to work on it further for now.

marvinborner:
Variables appearing free in normal as well as default clauses were erased multiple times because we used a list instead of a set for tracking the free variables. I've pushed a fix, let me know if this is a valid solution.

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -110,7 +110,7 @@ object Transformer {
 
       case machine.Switch(value, clauses, default) =>
         emit(Comment(s"switch ${value.name}, ${clauses.length} clauses"))
-        val freeInClauses = clauses.flatMap(freeVariables).toSet ++ default.map(freeVariables).getOrElse(Set.empty)
+        val freeInClauses = clauses.flatMap(freeVariables).toSet ++ default.map(freeVariables).getOrElse(Set.empty) -- Set(value)
```
jiribenes, line 113: the `-- Set(value)` is semantically wrong here, no?
In the repro (that I just committed... again), there's an `else` clause like `case _ => value` where `value` is also the scrutinee. But the whole problem of this PR was that this pattern seemed to leak. 🔍

marvinborner: Yes, I've realized this too. It fixed `tree_explore` though so I wanted to try it on CI.

marvinborner:
I'm a bit lost at fixing `tree_explore`, maybe @mattisboeckle can help? In switch clauses that reference the scrutinee (i.e. `value` appears free in `clause.body`), the scrutinee potentially gets shared more than it gets erased, causing a memory leak. Although it's probably just a single-line fix where we forgot to add an eraser, all my attempts made some other tests fail. My reduced reproduction program for two kinds of leaks:

``` scala
def foo(l: List[Int]): Int =
  l match {
    case Nil() => 42
    case _ => foo(l.drop(1))
  }

record Bar[T](drinks: List[T])

def bar(bar: Bar[Int]): Int = bar match {
  case Bar(_) => foo(bar.drinks)
}

def main() = println(Bar([1, 2, 3]).bar)
```

phischu:
Ah, we have to erase the scrutinee in the default clause!

marvinborner:
Ooh yes, that's much simpler than what I suspected. How should we differentiate between default and normal clause generation? We use `labelClause` for both right now.

jiribenes:
> Ooh yes, that's much simpler than what I suspected. How should we differentiate between default and normal clause generation? We use `labelClause` for both right now.

FWIW when trying this out on Friday, @phischu tried to intuitively do something like:

```diff
+        def labelClause(clause: machine.Clause, isDefault: Boolean): String = {
-        def labelClause(clause: machine.Clause): String = {
          implicit val BC = BlockContext()
          BC.stack = stack

+        if (isDefault) { ... } else { ... }

...

        val defaultLabel = default match {
+          case Some(clause) => labelClause(clause, isDefault = true)
-          case Some(clause) => labelClause(clause)
          case None =>
            val label = freshName("label");
            emit(BasicBlock(label, List(), RetVoid()))
            label
        }
        val labels = clauses.map {
+          case (tag, clause) => (tag, labelClause(clause, isDefault = false))
-          case (tag, clause) => (tag, labelClause(clause))
        }
 ```

jiribenes:
If anybody reading this has the time, I'd love to know if the changes from the PR still fix the leak in #711 :)

PR #771, opened 2025-01-10
Implement compare for bytearray

### LINKED ISSUE #748: Add string comparison to LLVM backend
I'd like to have a function `compareStringBytes: (String, String) => Ordering` for #561.
See #561's changes in `effekt.effekt` for the kind of function we need, then figure out a way to support it on the LLVM backend. :)

~~Should it be native (somewhere in C)? Should it be written in Effekt itself? No clue.~~
I _think_ it should be possible _and_ reasonably easy to write this function directly in Effekt, at least if the comparison is byte-wise.

mattisboeckle:
Fixes #748

### DIFF
--- a/examples/stdlib/bytearray/bytearray_compare.effekt
+++ b/examples/stdlib/bytearray/bytearray_compare.effekt
@@ -1,0 +1,14 @@
+module examples/pos/bytearray_compare
+
+import bytearray
+
+def main() = {
+  val b1 = fromString("Hey")
+  val b2 = fromString("Hey")
+  val b3 = fromString("He")
+  val b4 = fromString("Heys")
+  println(compareByteArray(b1, b1))
+  println(compareByteArray(b1, b2))
+  println(compareByteArray(b1, b3))
+  println(compareByteArray(b1, b4))
+}
--- a/libraries/common/bytearray.effekt
+++ b/libraries/common/bytearray.effekt
@@ -88,4 +88,20 @@ extern js """
     bytes[index] = value;
     return $effekt.unit;
   }
+
+  function bytearray$compare(arr1, arr2) {
+    const len = Math.min(arr1.length, arr2.length);
+
+    for (let i = 0; i < len; i++) {
+      if (arr1[i] !== arr2[i]) {
+        return arr1[i] < arr2[i] ? -1 : 1;
+      }
+    }
+
+    if (arr1.length !== arr2.length) {
+      return arr1.length < arr2.length ? -1 : 1;
+    } else {
+      return 0;
+    }
+  }
 """
--- a/libraries/common/bytearray.effekt
+++ b/libraries/common/bytearray.effekt
@@ -105,3 +105,18 @@ extern js """
     }
   }
 """
+
+extern pure def compareByteArrayImpl(b1: ByteArray, b2: ByteArray): Int =
+  js "bytearray$compare(${b1}, ${b2})"
+  llvm """
+    %x = call %Int @c_bytearray_compare(%Pos ${b1}, %Pos ${b2})
+    ret %Int %x
+  """
+
+def compareByteArray(b1: ByteArray, b2: ByteArray): Ordering =
+  compareByteArrayImpl(b1, b2) match {
+    case -1 => Less()
+    case 0 => Equal()
+    case 1 => Greater()
+    case _ => <{ "Impossible: ByteArray comparison returned invalid value!" }>
+  }
--- a/libraries/llvm/bytearray.c
+++ b/libraries/llvm/bytearray.c
@@ -152,39 +152,38 @@ struct Pos c_bytearray_equal(const struct Pos left, const struct Pos right) {
         erasePositive(right);
         return BooleanFalse;
     }
-    for (uint64_t j = 0; j < left_size; ++j) {
-        if (c_bytearray_data(left)[j] != c_bytearray_data(right)[j]) {
-            erasePositive(left);
-            erasePositive(right);
-            return BooleanFalse;
-        }
-    }
+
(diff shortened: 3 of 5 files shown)


libraries/llvm/bytearray.c
```diff
@@ -164,6 +164,35 @@ struct Pos c_bytearray_equal(const struct Pos left, const struct Pos right) {
     return BooleanTrue;
 }
 
+Int c_bytearray_compare(const struct Pos left, const struct Pos right) {
+    uint64_t left_size = left.tag;
+    uint64_t right_size = right.tag;
+    
+    for (uint64_t j = 0; j < left_size && j < right_size; ++j) {
```
jiribenes, line 171: Would it be worth it here to do a fast path for:
1. are the structs the same, and
2. are they the same size
?

Also, is it worth it to use `memcmp` for the comparison itself (since it should be reasonably optimised even if the compiler doesn't figure out that the for loop is a `memcmp`?)

WDYT?

jiribenes: Note that all of these are also possibly a consideration for `c_bytearray_equal` -- the two functions should definitely always agree (`c_bytearray_compare == 0 <=> c_bytearray_equal == true`)...

mattisboeckle: 1. This sounds like a good idea
2. I was unsure about what the compare should prioritize. It is implemented as "c" > "abcde", so characters being the deciding factor before size. Otherwise i'll just need to switch the order of for and if, which would be the fast path

`memcmp` is probably a good idea

jiribenes: Ah, yes, you're right, `2` makes only sense for the equality, not the comparison, thanks!

jiribenes:
Hey, what are the next steps of this PR? We could either:
1. Review and merge this now, then handle the Effekt bindings (+ a few tests) in a followup PR
2. Add the Effekt bindings (+ tests) first here, then proceed with review and merge

libraries/llvm/bytearray.c
```diff
@@ -152,16 +152,41 @@ struct Pos c_bytearray_equal(const struct Pos left, const struct Pos right) {
...
+
+Int c_bytearray_compare(const struct Pos left, const struct Pos right) {
+    uint64_t left_size = left.tag;
+    uint64_t right_size = right.tag;
+    uint64_t min_size = (left_size < right_size) ? left_size : right_size;
+    
+    uint8_t* left_data = c_bytearray_data(left);
+    uint8_t* right_data = c_bytearray_data(right);
+
+    if (left_data == right_data) return 0;
```
marvinborner, line 170: You forgot to erase `left` and `right` here

b-studios:
rough draft
```
extern def compare(b1: ByteArray, b2: ByteArray): Ordering = 
  llvm { llvm::compare(b1).... }

namespace llvm {

  extern def compare(b1: ByteArray, b2: ByteArray): Int
}
```

mattisboeckle:
Where should I add the JS function? In `effekt_builtins.js`?
There are already some similar functions there which sadly do not work for this case

libraries/common/bytearray.effekt
```diff
@@ -88,4 +88,35 @@ extern js """
...
+        return arr1[i] < arr2[i] ? -1 : 1;
+      }
+    }
+
+    if (arr1.length !== arr2.length) {
+      return arr1.length < arr2.length ? -1 : 1;
+    } else {
+      return 0;
+    }
+  }
```
b-studios, line 106: Why don't we first compare the length? This is supposed to be much faster.

```suggestion
  function bytearray$compare(arr1, arr2) {
    if (arr1.length !== arr2.length) {
      return arr1.length < arr2.length ? -1 : 1;
    }
    
    const len = Math.min(arr1.length, arr2.length);
    for (let i = 0; i < len; i++) {
      if (arr1[i] !== arr2[i]) {
        return arr1[i] < arr2[i] ? -1 : 1;
      }
    }
    
    return 0;
  }
```

b-studios: Am I repeating a mistake again?

examples/stdlib/bytearray/bytearray_compare.effekt
```diff
@@ -0,0 +1,14 @@
...
+
+def main() = {
+  val b1 = fromString("Hey")
+  val b2 = fromString("Hey")
+  val b3 = fromString("He")
+  val b4 = fromString("Heys")
+  println(compareByteArray(b1, b1))
+  println(compareByteArray(b1, b2))
+  println(compareByteArray(b1, b3))
+  println(compareByteArray(b1, b4))
```
b-studios, line 13: Can you make sure that these test cases cover all paths in the implementation? (shorter, longer,  different prefix, etc.)

libraries/llvm/bytearray.c
```diff
@@ -152,16 +152,45 @@ struct Pos c_bytearray_equal(const struct Pos left, const struct Pos right) {
...
+
+    int cmp = memcmp(left_data, right_data, min_size);
+
     erasePositive(left);
     erasePositive(right);
-    return BooleanTrue;
+
+    if (cmp < 0) {
+        return -1;
+    } else if(cmp > 0) {
```
b-studios, line 183:
```suggestion
    } else if (cmp > 0) {
```

examples/stdlib/bytearray/bytearray_compare.effekt
jiribenes: FWIW, I'd love a test comparing [haha!] this with equality. :)

```
compare == 0 <=> equals == true
compare != 0 <=> equals == false
```

b-studios (approved):

PR #775, opened 2025-01-13
Add Examples / Unit Tests for Exceptions

### LINKED ISSUE #480: Tests for `exception` stdlib.
There are currently no tests for the `exception` part of the stdlib.
(As noticed in #479).

JakubSchwenkbeck:
## Motivation
Resolves #480
This module demonstrates exception handling combinators in Effekt using a simple, focused example. By showcasing custom exception handling (`TestException`) with different strategies (default handling, ignoring, reporting, finalizing), it provides a practical guide for understanding these features.

## Changes
- Introduced `TestException` as a custom exception type, independent of other exceptions like `OutOfBounds`.
- Added `generalOperation`, which raises a `TestException` for invalid input or returns a string.
- Included tests for handling exceptions with:
  - Default behavior
  - Ignoring exceptions
  - Reporting exceptions
  - Finalization hooks
  - "Reifying" with Results

## Testing
The examples are self-contained in the `main` function, covering all provided exception-handling strategies. Each approach is tested and outputs expected results for validation (`combinators.check`).


JakubSchwenkbeck:
Not quite sure what causes `Error: fatal: unable to access 'https://github.com/effekt-lang/effekt/': The requested URL returned error: 502` from the `.workflow` ?

jiribenes:
I'm not sure about the value of testing all the different _exceptions_ as in _types_: `Exception[E]` is generic in `E`, so once you test one, you've tested them all by parametricity ;).
I'd guess the more important thing is testing all of the `on[E].xyz` combinators here: https://github.com/effekt-lang/effekt/blob/112bc876bf817d471f9323bdc1155dd7bcd75b28/libraries/common/exception.effekt#L108-L121 / their full versions & perhaps the one more in `result.effekt` too :)

JakubSchwenkbeck:
Thank you for your input and help, I really appreciate it!
Feel free to tell me if the current approach is of more sense:)

examples/stdlib/exception/combinators.effekt
```diff
@@ -0,0 +1,48 @@
...
+    with on[OutOfBounds].finalize { println("Finalizer executed") }
+    println(p().show) 
+    
+  
+  }
+
+  finalizeOutOfBounds { str.charAt(10) } // Finalizer executed \n a
+  finalizeOutOfBounds { str.charAt(1) } // Finalizer will execute even without exception
+  */
+  //!! This somehow causes : Main cannot have effects, but includes effects: { Exception[OutOfBounds] }
```
jiribenes, line 47: Take a look at the type of `finalise` ;)
```scala
def finalize[E, T] { finalizer: => Unit } { prog: => T / Exception[E] }: T / Exception[E]
```

JakubSchwenkbeck: Thanks, got it at the same ;)

examples/stdlib/exception/combinators.effekt
```diff
@@ -0,0 +1,48 @@
...
+def main() = {
+  val str = "Hallo"
+
+  // Test for default handling of OutOfBoundsException
+  def defaultOutOfBounds { p: => Char / Exception[OutOfBounds] }: Unit = {
+    with on[OutOfBounds].default { println("Index Out of Bound") }
+    println(p().show)
+  }
+
+  defaultOutOfBounds { str.charAt(10) } // Index Out of Bound
```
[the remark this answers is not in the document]
JakubSchwenkbeck: Makes sense, do you think staying with `Exception[OutOfBound]` and triggering it by List-accesses is sufficient? :)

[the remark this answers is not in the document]
jiribenes: To be honest, I think it would be worth it here to make your own `TestException` and test solely with it in this file to reduce the dependencies. Again, most of our modules are highly unstable and it's easy to make the tests more coupled to random modules than necessary, thus leading to worse maintainability.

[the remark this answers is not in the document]
JakubSchwenkbeck: Good idea! I'll look into it when I have the time, thanks again for your help!:)

JakubSchwenkbeck:
Note : I observed some strange behavior regarding `raise(...)` and `llvm` as backend. While these calls work with no problem with `js`, using `llvm` returns the error `multiple definitions of local name  'v_r_xxxx_x_xx_xxxx'` which seems odd to me

examples/stdlib/exception/combinators.effekt
```diff
@@ -0,0 +1,75 @@
...
+  finalizeTestException { str.generalOperation(0) } // Test: Finalizer
+  finalizeTestException { str.generalOperation(1) } // Test: Finalizer hello
+  
+  // Test for "reifying" an Exception using Result
+  def resultTestException { p: => String / Exception[TestException] }: Unit = {
+    val res= result[String, TestException](on[TestException]) { p() }
+    res match {
+      case Success(msg) => println("Success: " ++ msg) 
+      case Error(exc, msg) => println(msg)
+    }
```
jiribenes, line 69: You can use the direct variant here, it's probably nicer :)
```suggestion
    val res = result[String, TestException] { p() }
    res match {
      case Success(msg) => println("Success: " ++ msg) 
      case Error(exc, msg) => println(msg)
    }
```

examples/stdlib/exception/combinators.effekt
```diff
@@ -0,0 +1,75 @@
...
+import result
+
+/**
+ * Used as a type for `Exception` purely for independent testing
+ */
+record TestException()
+
+/**
+ * Returns the string if index > 0; otherwise raises a TestException.
+ */
```
jiribenes, line 13: Please use `///` instead: I'm planning to deprecate multiline comments soon. :)
```suggestion
/// Used as a type for `Exception` purely for independent testing
record TestException()

/// Returns the string if index > 0; otherwise raises a TestException.
```

examples/stdlib/exception/combinators.effekt
```diff
@@ -0,0 +1,75 @@
...
+def generalOperation(str: String, index: Int): String / Exception[TestException] = {
+  if (index <= 0)
+    do raise(TestException(), "Error: Invalid index (" ++ show(index) ++ ")")
+  else
+    str
+}
+
+def main() = {
+  val str : String = "hello"
```
jiribenes, line 22: Nitpick, feel free to ignore:
```suggestion
  val str: String = "hello"
```

jiribenes (approved):
I have a few style-related nitpicks, but otherwise LGTM, thanks! :shipit:

jiribenes:
Resolves #480.

examples/stdlib/exception/combinators.effekt
```diff
@@ -60,9 +58,9 @@ def main() = {
   finalizeTestException { str.generalOperation(0) } // Test: Finalizer
   finalizeTestException { str.generalOperation(1) } // Test: Finalizer hello
   
-  // Test for "reifying" an Exception using Result
+  // Test for "reifying" an Exception usix^xng Result
```
jiribenes, line 61:
```suggestion
  // Test for "reifying" an Exception using Result
```
typo :)

JakubSchwenkbeck: Thanks for looking over it!:)

jiribenes (approved):
Thanks, let's ship it! :shipit:

PR #776, opened 2025-01-14
Revise ignored LLVM tests

### LINKED ISSUE #740: Revise LLVM tests that are ignored because of missing features
We have a lot of tests that are ignored on LLVM because of missing features, but a lot of these are -- I think -- already present in LLVM (like arrays!).

It would be valuable to go over these tests and see:
- if they still fail but are fixable quickly (perhaps relying on `inspect` / `genericShow` for little to no reason)
- if they actually work but fail with valgrind (then they should be moved to the correct section)
- if they fail because of missing features (then keep them in the `missingFeatures` section)
- if they Just Work :tm: now (then there's no reason to ignore these tests!)

Here are the tests:
https://github.com/effekt-lang/effekt/blob/92a6e0f4759bbd5785a77bcbaeb4eb04b96164aa/effekt/jvm/src/test/scala/effekt/LLVMTests.scala#L44-L97

marvinborner:
Eventually closes #740


effekt/jvm/src/test/scala/effekt/LLVMTests.scala
```diff
@@ -33,36 +31,17 @@ class LLVMTests extends EffektTests {
...
-    // now show instance for records / datatypes
-    examplesDir / "pos" / "builtins.effekt",
-    examplesDir / "pos" / "namespaces.effekt",
-    examplesDir / "pos" / "triples.effekt",
-    examplesDir / "pos" / "either.effekt",
-
     // inspect
-    examplesDir / "pos" / "probabilistic.effekt",
-    examplesDir / "pos" / "nim.effekt",
     examplesDir / "pos" / "exists.effekt",
```
marvinborner, line 35: I have some difficulties writing an elegant show function for `exists.effekt`, does anyone have a good idea?

b-studios: You can just change the type to store it:

```
type Exists {
  Op[Z](el: Z, combine: (Z, Z) => Z at {}, show: Z => String)
}
```

b-studios (approved):

examples/pos/probabilistic.effekt
```diff
@@ -80,6 +72,17 @@ def choose3[R] { x: => R } { y: => R } { z: => R } : R / fork =
 
 type Weighted[R] { MkWeighted(weight: Double, value: R) }
 
+def showRounded(d: Double): String = show(d.round(2))
```
marvinborner, line 75: I don't understand how this rounding doesn't fix the CI issue. The float inaccuracy is not reproducible on my computer.

b-studios: > Also: since when is time.effekt flaky?

No idea.. I just triggered a re-run of CI.

b-studios: Also, how does this not fail the VMTests?

marvinborner: Ah you're right, I misinterpreted the obtained/expected in CI.. No idea about VMTests

PR #777, opened 2025-01-14
Make local continuations direct style if possible

b-studios:


effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
```diff
@@ -209,13 +209,39 @@ object Normalizer { normal =>
 
     case Stmt.Val(id, tpe, binding, body) =>
 
-      def normalizeVal(id: Id, tpe: ValueType, binding: Stmt, body: Stmt): Stmt = binding match {
+      def barendregdt(stmt: Stmt): Stmt = new Renamer().apply(stmt)
```
jiribenes, line 212:
```suggestion
      def barendregt(stmt: Stmt): Stmt = new Renamer().apply(stmt)
```
:^)
(yeah, I know, it's callsites are commented out rn)

Issue #760, opened 2025-01-07
Compiler crash on a cyclic import


jiribenes:
Cyclic imports make the compiler crash. It would be appropriate to detect cyclic imports and report them to the user in a nicer way :)

(Reported by @TheDying0fLight)

`a.effekt`
```scala
module a

import b
```

`b.effekt`
```scala
module b

import a
```

crashes the compiler:
```
[info] ./a.effekt:3:1: Effekt Compiler Crash: null
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.Compiler.runFrontend(Compiler.scala:138)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.Compiler.runFrontend$(Compiler.scala:95)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.generator.js.JavaScript.runFrontend(JavaScript.scala:12)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.context.ModuleDB.tryModuleOf(ModuleDB.scala:53)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.context.ModuleDB.tryModuleOf$(ModuleDB.scala:14)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.context.Context.tryModuleOf(Context.scala:40)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.context.ModuleDB.moduleOf(ModuleDB.scala:42)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.context.ModuleDB.moduleOf$(ModuleDB.scala:14)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.context.Context.moduleOf(Context.scala:40)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.context.ModuleDB.moduleOf(ModuleDB.scala:36)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.namer.Namer$.effekt$namer$Namer$$$_$importDependency$1(Namer.scala:49)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.namer.Namer$$anon$1.$anonfun$2(Namer.scala:75)
import b
^^^^^^^^
[info] ./a.effekt:3:1:   at effekt.util.messages$ErrorReporter.at(Messages.scala:147)
import b
^^^^^^^^
[...]

PR #731, opened 2024-12-09
Custom compiler crash/panic handler

JakubSchwenkbeck:
## Motivation
Print and log a more verbose compiler crash/panic message with information about the specific user setup
Resolves #576.

## Changes
Implemented the `generateCrashReport` function, which creates a more verbose output if the compiler crashes/panics. 
This Report retrieves **general** and **backend-specific** information to make crashes more reproducable or traceable

### It holds information about :
- Effekt Version
-  Backend 
- OS
-  JVM
- and a full stacktrace. 

Additionally, the report is **written** to a `.md`-file in the .out/ directory

### Terminal Report
```
[info] === Effekt Compiler Crash Report ===

[Error] Internal compiler error: Error Text here bla bla

The compiler unexpectedly panicked. This is a compiler bug.
Please report it: https://github.com/effekt-lang/effekt/issues/new?labels=bug

Effekt Version: 0.13.0

Backend used: js

A detailed crash report has been written to:
/home/jakub/Effekt/repo/effekt/bin/./out

Operating System: Linux

JVM Version: 21.0.5

Full Stack Trace:
PlainTextError(Internal compiler error: Error Text here bla bla,None,Error)
        at effekt.util.messages$CompilerPanic$.apply(Messages.scala:37)
        at effekt.util.messages$.INTERNAL_ERROR(Messages.scala:53)
        at effekt.Driver.compileSource(Driver.scala:54)
        at effekt.Driver.compileSource$(Driver.scala:21)
        at effekt.Server$.compileSource(Server.scala:261)
        at effekt.Server$.compileSource(Server.scala:261)
        at kiama.util.Compiler.compileFile(Compiler.scala:102)
        at kiama.util.Compiler.compileFile$(Compiler.scala:22)
        at effekt.Server$.compileFile(Server.scala:261)
        at effekt.Driver.run$$anonfun$1(Driver.scala:41)
        at scala.runtime.function.JProcedure1.apply(JProcedure1.java:15)
        at scala.runtime.function.JProcedure1.apply(JProcedure1.java:10)
        at scala.collection.immutable.List.foreach(List.scala:333)
        at effekt.Driver.run(Driver.scala:42)
        at effekt.Driver.run$(Driver.scala:21)
        at effekt.Server$.run(Server.scala:261)
        at effekt.Server$.run(Server.scala:261)
        at kiama.util.Compiler.driver(Compiler.scala:86)
        at kiama.util.Compiler.driver$(Compiler.scala:22)
[...]

PR #780, opened 2025-01-15
Fix unknown compiler crash printing 'null'

jiribenes:
Resolves the `Effect Compiler Crash: null` part of #760.
The actual problem there was that `e.getMessage` can return `null` and it does for... _StackOverflowException_ 🥳.
Looking at the docs, it's recommended to use `e.toString`.

While I was at it, I modified stack trace printing, now we do it properly by using `.printStackTrace`.

This is very related to #731, I just want to get this out quickly to unblock other problems.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Compiler.scala
+++ b/effekt/shared/src/main/scala/effekt/Compiler.scala
@@ -115,6 +115,11 @@ trait Compiler[Executable] {
    * - Server / Driver to typecheck and report type errors in VSCode
    */
   def runFrontend(source: Source)(using C: Context): Option[Module] =
+    def getStackTrace(e: Throwable): String =
+      val stringWriter = new java.io.StringWriter()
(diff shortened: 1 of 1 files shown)


effekt/shared/src/main/scala/effekt/Compiler.scala
```diff
@@ -115,6 +115,11 @@ trait Compiler[Executable] {
    * - Server / Driver to typecheck and report type errors in VSCode
    */
   def runFrontend(source: Source)(using C: Context): Option[Module] =
+    def getStackTrace(e: Throwable): String =
+      val stringWriter = new java.io.StringWriter()
+      e.printStackTrace(new java.io.PrintWriter(stringWriter))
+      stringWriter.toString
```
jiribenes, line 121: This part is essentially cherry-picked from the discussion in @JakubSchwenkbeck's #731. Thanks!

dvdvgt (approved):

PR #782, opened 2025-01-15
Try to fix flaky "Text file busy"

### LINKED ISSUE #322: Workflow runs flaky / "Text file busy"
In both the workflow runs
- (#320) https://github.com/effekt-lang/effekt/actions/runs/7034447758/job/19142525276?pr=321  |  https://github.com/effekt-lang/effekt/actions/runs/7034447758/job/19142525276?pr=321#step:9:208
- (#321) https://github.com/effekt-lang/effekt/actions/runs/7019874175/job/19098324052  |  https://github.com/effekt-lang/effekt/actions/runs/7019874175/job/19098324052#step:9:197
one of the tests fails with something like
```
==> X effekt.ChezScheme*Tests.examples/pos/multieffects.effekt (chez-callcc)  2.115s java.io.IOException: Cannot run program "*": error=26, Text file busy
```

This is probably a chance event, and might be related to https://bugs.openjdk.org/browse/JDK-8068370.
Leaving this here because it occured in both of my last PRs, so we might want to look into this if this stays this frequent for more than two runs.

jiribenes:
Resolves #322 (maybe?) by just locking a file while we evaluate it.
This is inspired by me being mad at the flakiness while remembering that locks exist. :) 

The whole thing is completely untested, but maybe we can just run CI in a loop few times and see if we're safe.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Runner.scala
+++ b/effekt/jvm/src/main/scala/effekt/Runner.scala
@@ -1,5 +1,9 @@
 package effekt
 
+import java.nio.channels.FileLock
+import java.nio.file.{Paths, StandardOpenOption}
+import java.nio.channels.FileChannel
+
 import effekt.context.Context
 import effekt.util.messages.FatalPhaseError
 import effekt.util.paths.{File, file}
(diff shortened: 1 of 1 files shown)


jiribenes:
In a fun twist, now _every_ file reports `Text file busy` 😆 
Now I'm not really sure as to where we open the executable further -- maybe it's the `Process` library that needs to acquire a lock?

jiribenes:
I'm closing this as a failed experiment, but I think that the next step would be to try to `strace` and see what locks are acquired when -- since now we clearly see that something (`Process`?) is acquiring a lock.

PR #787, opened 2025-01-22
Refactor/replace environments by substitution

b-studios:


effekt/shared/src/main/scala/effekt/generator/js/TransformerCps.scala
```diff
@@ -350,13 +391,13 @@ object TransformerCps extends Transformer {
...
-      pure(js.Return(Call(RESET, toJS(prog)(using nonrecursive(prog)), toJS(ks), toJS(k))))
+      pure(js.Return(Call(RESET, requiringThunk { toJS(prog)(using nonrecursive(prog)) }, toJS(ks), toJS(k))))
 
     case cps.Stmt.Shift(prompt, body, ks, k) =>
-      pure(js.Return(Call(SHIFT, nameRef(prompt), noThunking { toJS(body)(using nonrecursive(body)) }, toJS(ks), toJS(k))))
+      pure(js.Return(Call(SHIFT, nameRef(prompt), requiringThunk { toJS(body)(using nonrecursive(body)) }, toJS(ks), toJS(k))))
 
     case cps.Stmt.Resume(r, b, ks2, k2) =>
-      pure(js.Return(js.Call(RESUME, nameRef(r), toJS(b)(using nonrecursive(b)), toJS(ks2), toJS(k2))))
+      pure(js.Return(js.Call(RESUME, nameRef(r), toJS(b)(using nonrecursive(b)), toJS(ks2), requiringThunk { toJS(k2) })))
```
b-studios, line 400: This might slow down some benchmarks, but is necessary in the general case to avoid SOs

effekt/shared/src/main/scala/effekt/core/optimizer/Optimizer.scala
```diff
@@ -21,6 +21,8 @@ object Optimizer extends Phase[CoreTransformed, CoreTransformed] {
 
   def optimize(source: Source, mainSymbol: symbols.Symbol, core: ModuleDecl)(using Context): ModuleDecl =
 
+    val isLLVM = Context.config.backend().name == "llvm"
```
b-studios, line 24: This is REALLY hacky. Another solution would be to first optimize and then do polymorphism boxing. However, this alternative currently (I didn't try to find out why) fails a lot of test.

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -761,26 +761,26 @@ class VMTests extends munit.FunSuite {
...
       staticDispatches = 7,
-      dynamicDispatches = 1062912,
+      dynamicDispatches = 797188,
       patternMatches = 0,
       branches = 5,
       pushedFrames = 531467,
       poppedFrames = 531467,
       allocations = 0,
-      closures = 265750,
+      closures = 26,
```
b-studios, line 783: This change is particularly significant...

PR #788, opened 2025-01-22
Add more VM tests

b-studios:
I added very basic support for regexes in the VM so that we can run the case studies.

I also had to move `regex.effekt` to common to do that, even though it is not supported by LLVM.


effekt/shared/src/main/scala/effekt/core/vm/Builtin.scala
```diff
@@ -267,4 +333,17 @@ protected object As {
...
   }
+  object Regex {
+    def unapply(v: Value): Option[regex.Regex] = v match {
+      case Value.Literal(v: regex.Regex) => Some(v)
+      case _ => None
+    }
+  }
+  object RegexMatch {
+    def unapply(v: Value): Option[regex.Regex.Match | Null] = v match {
+      case Value.Literal(null) => Some(null)
```
b-studios, line 344: I am not entirely happy about using `Literal` for arbitrary extern values, but well...

PR #790, opened 2025-01-22
Valgrind fails in CI

phischu:


mattisboeckle:
I did some investigations on these valgrind fails and found that it was complaining about "uninitialized value was created by a heap allocation" when using bytearrays.
If we just initialize the allocated memory to 0 the valgrind errors disappear (at least locally)

b-studios:
That would be great! Is there any performance implications of doing that?

marvinborner:
I'm confused, this shouldn't be a problem, right? Why are we reading from the bytearray when it's not initialized yet (especially in all IO tests)?

phischu:
It is good to know that initializing the array would fix it, but we should not actually initialize them to zero every time. Rather, we should find the underlying out-of-bounds access. I can not reproduce those errors locally, but @mattisboeckle since you can, please take a look.

phischu:
Try to find a way to silence those warnings specifically, perhaps telling valgrind about `uv_fs_read`. Also please remove the line that initializes bytearrays again so we see what CI is complaining about.

jiribenes:
FWIW, for debugging, it might be worth it to memset the bytes of the bytearray to something recognisable like `0xEF` and then trace in `gdb`/`lldb` + `rwatch buf` to see when it gets read from 🤔

mattisboeckle:
I'm pretty sure there is no real problem here. 
Without valgrind the programs work exactly as I would expect them to. Correct results, no errors. Definitely no out of bounds accesses, because that would mean segfaults.

I'll try and see if valgrind can handle libuv better somehow, but otherwise just suppressing the errors should be fine.

marvinborner:
Note that uses of uninitialized memory as well as out-of-bounds accesses not necessarily cause segfaults (or influence anything at all). Valgrind detects even the slightest mistakes that -- even if they don't affect us right now -- could cause larger hard-to-debug problems in the future.

I don't think suppressing the error is a good idea unless we're 100% certain that this use of uninitialized memory could never influence the program state.

examples/benchmarks/input_output/word_count_ascii.effekt
```diff
@@ -59,13 +59,15 @@ def countWords(): Output / read[Byte] = {
   Output(chars, words, lines)
 }
 
+extern def filename(): String =
+  js { "/tmp/word_count_ascii_js.txt" }
+  llvm { "/tmp/word_count_ascii_llvm.txt" }
+
```
phischu, line 65: Check if this is necessary and undo it if not.

examples/benchmarks/input_output/dyck_one.effekt
```diff
@@ -37,17 +37,19 @@ def size(tree: Tree): Int =
     }
   }
 
+extern def filename(): String =
+  js { "/tmp/dyck_one_js.txt" }
+  llvm { "/tmp/dyck_one_llvm.txt" }
+
```
phischu, line 43: Check if this is necessary and undo it if not. Its purpose was to prevent multiple test processes (for the different backends) accessing the same file.

marvinborner: Could we also add some `EFFEKT_BACKEND` constant such that we can write `def filename(): String = "/tmp/dyck_one_${EFFEKT_BACKEND}.txt"` using the new string interpolation? This would reduce code duplication and make it easier to add new backends in the future.

b-studios: We do not have `extern val`s but it is easy to have a function somewhere:

```
extern def EFFEKT_BACKEND(): String =
  js { "js" }
  llvm { "llvm" }
  chez { "chez" }
   ...
```

b-studios: I am not sure we want to use string interpolation in this particular scenario, since this pulls a lot of different dependencies and @phischu would be very unhappy.

PR #791, opened 2025-01-22
ByteArray support for Chez

jiribenes:
Tried to one-shot with Claude, we'll see if it works, eh?
EDIT: Wonderful! 🥳 ✨


libraries/common/bytearray.effekt
```diff
@@ -106,12 +112,29 @@ extern js """
...
+      (let loop ([i 0])
+        (cond
+          [(= i minlen)
+          (cond [(< len1 len2) -1]
+                [(> len1 len2) 1]
+                [else 0])]
+          [(< (bytevector-u8-ref b1 i) (bytevector-u8-ref b2 i)) -1]
+          [(> (bytevector-u8-ref b1 i) (bytevector-u8-ref b2 i)) 1]
+          [else (loop (+ i 1))]))))
+"""
```
jiribenes, line 129: This is probably further signal that we should really define the comparison in Effekt directly.

examples/stdlib/bytearray/bytearray_compare.effekt
```diff
@@ -3,12 +3,51 @@ module examples/pos/bytearray_compare
...
 def main() = {
-  val b1 = fromString("Hey")
-  val b2 = fromString("Hey")
-  val b3 = fromString("He")
-  val b4 = fromString("Heys")
-  println(compareByteArray(b1, b1))
-  println(compareByteArray(b1, b2))
-  println(compareByteArray(b1, b3))
-  println(compareByteArray(b1, b4))
+  val empty: ByteArray = "".fromString
```
jiribenes, line 6: I'm still not 100% happy with the test, but at least all backends always return the same thing, so it's at least consistent :shipit:

jiribenes:
Since nobody really cares about the Chez backend, I'll just merge this :shipit:

PR #792, opened 2025-01-22
Fix primitive string comparison on Chez

jiribenes:
Found when debugging #561 -- string comparison was present on Chez, but it never worked as strings are compared with operations that look like `string<?`.
See section 7.4 here https://www.scheme.com/csug8/objects.html :)

I also added a tiny test to ensure that this doesn't confuse anybody else down the line.


libraries/common/effekt.effekt
```diff
@@ -572,19 +572,19 @@ extern pure def infixGte(x: Double, y: Double): Bool =
 // TODO do we really need those? if yes, move to string.effekt
 extern pure def infixLt(x: String, y: String): Bool =
   js "(${x} < ${y})"
-  chez "(< ${x} ${y})"
+  chez "(string<? ${x} ${y})"
```
jiribenes, line 575: Good one! :)

jiribenes (commented):
:shipit:

PR #795, opened 2025-01-23
Add stdlib test importing every stdlib module + fix stdlib

### LINKED ISSUE #614: Add a test that loads all stdlib files
Add a simple test similar to the existing `examples/chez/libraries.effekt` (but for all/any backend) that just loads all stdlib modules to see whether they still typecheck.

jiribenes:
Resolves #614 by adding a stdlib test that imports every single (common) stdlib module.

Let's add [ACME](https://en.wikipedia.org/wiki/Acme_Corporation) (in our context "All Common Modules in Effekt" :)) to the stdlib tests: just import and compile all modules.

This revealed 3 different bugs in stdlib + codegen bugs on LLVM (moved to #801).


jiribenes:
Ideal follow-up: make CI so that on change in `/libraries/common/`, we check that the file contains every module (just generate it again and diff). If not, fail and spam the PR.

jiribenes:
New LLVM bugs, cc @phischu:
```
$ EFFEKT_DEBUG=1 effekt --backend=llvm --ir-write-all --debug --optimize examples/stdlib/json.effekt
opt: ./out/json.ll:4001:9: error: multiple definition of local value named 'd_24_132_3_22162'
        %d_24_132_3_22162 = insertvalue %Neg %vtable_temporary_865, %Object %closure_861, 1
        ^

$ EFFEKT_DEBUG=1 effekt --backend=llvm --ir-write-all --debug --optimize examples/stdlib/buffer.effekt
Instruction does not dominate all uses!
  %buffer_24_4387 = insertvalue %Neg %vtable_temporary_1257, ptr %closure_1239, 1
  call void @shareNegative(%Neg %buffer_24_4387)
Instruction does not dominate all uses!
  %buffer_24_4387 = insertvalue %Neg %vtable_temporary_1257, ptr %closure_1239, 1
  store %Neg %buffer_24_4387, ptr %buffer_24_4387_pointer_1255, align 8, !noalias !0
Instruction does not dominate all uses!
  %buffer_24_4387 = insertvalue %Neg %vtable_temporary_1257, ptr %closure_1239, 1
  call void @eraseNegative(%Neg %buffer_24_4387)
opt: ./out/buffer.ll: error: input module is broken!
```

libraries/common/buffer.effekt
```diff
@@ -51,15 +51,15 @@ def arrayBuffer[T](initialCapacity: Int): Buffer[T] at {global} = {
     def read() = {
       if (buffer.empty?) None()
       else {
-        val result: T = contents.remove(head).getOrElse { <> };
+        val result: T = contents.unsafeGet(head);
```
jiribenes, line 54: I _think_ this is semantically correct, we use it as a circular buffer, so we don't really need to invalidate the previous value -- we'd have overwritten it.

jiribenes (commented):
I'll merge this since it's a change that I really want in the next release, any nitpicks can be fixed in a separate PR :shipit:

PR #798, opened 2025-01-23
Unification benchmark using 'map' and 'stream'

### LINKED ISSUE #796: Add a benchmark involving 'map' and/or 'set'
@b-studios suggested something like unification.
It would be very fun to have a Prolog/Datalog benchmark/case study. :)

I'm assigning myself, but if somebody's really excited to do this and would do this soon, I'm happy to let this issue go to a better home. :^)

jiribenes:
Resolves #796 

It's an OK (not great, not terrible) benchmark:
1. generates two nested, deep pair types like `(...((XLLL, XLLR), (XLRL, XLRR)), ...` for `X = { A, B }`,
2. then unifies them,
3. and checks that the substitutions are like `ALLL -> BLLL`.

Configurable with a given `N` if some tweaking is needed.

#### My machine

On my computer with `N = 12`:
- JS compilation only: 4.5s
- LLVM compilation only: 5.5s
- then hyperfined:

```
$ hyperfine --warmup 5 --min-runs 20 './out/unify-js' './out/unify-llvm'
Benchmark 1: ./out/unify-js
  Time (mean ± σ):     703.0 ms ±   7.8 ms    [User: 1035.2 ms, System: 77.1 ms]
  Range (min … max):   690.9 ms … 721.1 ms    20 runs

Benchmark 2: ./out/unify-llvm
  Time (mean ± σ):      37.6 ms ±   0.5 ms    [User: 36.1 ms, System: 1.0 ms]
  Range (min … max):    37.1 ms …  39.7 ms    69 runs

Summary
  ./out/unify-llvm ran
   18.68 ± 0.32 times faster than ./out/unify-js
```

On my computer with `N = 16` (16x more work than `N = 12`)
- JS compilation only: 4.7s
- LLVM compilation only: 5.6s
- then hyperfined:

```
Benchmark 1: ./out/unify-js
  Time (mean ± σ):     19.982 s ±  0.608 s    [User: 26.011 s, System: 1.525 s]
  Range (min … max):   19.300 s … 21.267 s    20 runs

Benchmark 2: ./out/unify-llvm
  Time (mean ± σ):     928.2 ms ±  20.8 ms    [User: 906.3 ms, System: 16.1 ms]
  Range (min … max):   913.3 ms … 1006.2 ms    20 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  ./out/unify-llvm ran
   21.53 ± 0.81 times faster than ./out/unify-js
```

#### CI

With `N = 12` in CI (measured imprecisely in two CI rounds):
- 9.2-12.0 seconds on Chez backends
- 8.0-8.9 seconds on JS
- 9.9-12.7 seconds on LLVM

With `N = 16` in CI (measured imprecisely in two CI rounds):
- 21-29 seconds on Chez backends
- 55-57 seconds on JS
- 79-86 seconds on LLVM

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
+++ b/effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
@@ -160,6 +160,9 @@ object Normalizer { normal =>
 
   def normalize(s: Stmt)(using C: Context): Stmt = s match {
 
+     case Stmt.Def(id, block, body) if !C.usage.contains(id) || C.usage.get(id).contains(Usage.Never) => // see #798
+      normalize(body)
+
     case Stmt.Def(id, block, body) =>
       val normalized = active(block).dealiased
       Stmt.Def(id, normalized, normalize(body)(using C.bind(id, normalized)))
--- a/effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
+++ b/effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
@@ -66,6 +66,9 @@ object Normalizer { normal =>
       case None => false
     }
 
+  private def isUnused(id: Id)(using ctx: Context): Boolean =
+    ctx.usage.get(id).forall { u => u == Usage.Never }
+
   def normalize(entrypoints: Set[Id], m: ModuleDecl, maxInlineSize: Int, preserveBoxing: Boolean): ModuleDecl = {
     // usage information is used to detect recursive functions (and not inline them)
     val usage = Reachable(entrypoints, m)
@@ -161,7 +164,7 @@ object Normalizer { normal =>
   def normalize(s: Stmt)(using C: Context): Stmt = s match {
 
     // see #798 for context (led to stack overflow)
-    case Stmt.Def(id, block, body) if !C.usage.contains(id) || C.usage.get(id).contains(Usage.Never) =>
+    case Stmt.Def(id, block, body) if isUnused(id) =>
       normalize(body)
 
     case Stmt.Def(id, block, body) =>
--- a/effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
+++ b/effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
@@ -160,9 +160,8 @@ object Normalizer { normal =>
 
   def normalize(s: Stmt)(using C: Context): Stmt = s match {
 
-     case Stmt.Def(id, block, body) if !C.usage.contains(id) || C.usage.get(id).contains(Usage.Never) => // see #798
+    case Stmt.Def(id, block, body) if !C.usage.contains(id) || C.usage.get(id).contains(Usage.Never) => // see #798
       normalize(body)
-
     case Stmt.Def(id, block, body) =>
       val normalized = active(block).dealiased
       Stmt.Def(id, normalized, normalize(body)(using C.bind(id, normalized)))
--- a/effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
+++ b/effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
@@ -160,8 +160,10 @@ object Normalizer { normal =>
 
   def normalize(s: Stmt)(using C: Context): Stmt = s match {
 
-    case Stmt.Def(id, block, body) if !C.usage.contains(id) || C.usage.get(id).contains(Usage.Never) => // see #798
+    // see #798 for context (led to stack overflow)
+    case Stmt.Def(id, block, body) if !C.usage.contains(id) || C.usage.get(id).contains(Usage.Never) =>
       normalize(body)
+
     case Stmt.Def(id, block, body) =>
       val normalized = active(block).dealiased
       Stmt.Def(id, normalized, normalize(body)(using C.bind(id, normalized)))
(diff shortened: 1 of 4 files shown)


jiribenes:
Here's a slightly minified version:
```scala
/// Robinson-style Unification Algorithm
module examples/benchmarks/unify

import map
import result
import bytearray

type Type {
  Var(name: String)
  Con(name: String, args: List[Type])
}

type Substitution = Unit

type UnificationError {
  UnificationManyFailure(tps1: List[Type], tps2: List[Type])
}

def showType(ty: Type): String = ty match {
  case Var(name) => name
  case Con(name, Nil()) => name
  case Con(name, args) =>
    name ++ "(" ++ args.map { t => showType(t) }.join(", ") ++ ")"
}

def show(err: UnificationError): String = err match {
    case UnificationManyFailure(tps1, tps2) =>
    "Cannot unify " ++ tps2.map { t => showType(t) }.join(", ") ++ " with " ++ tps1.map { t => showType(t) }.join(", ")
}

def reporting { body : => Substitution / Exception[UnificationError] }: Unit = {
  val res = result[Substitution, UnificationError] {
    body()
  }

  res match {
    case Success(subst) => {
      println("Unification successful!")
    }
    case Error(err, msg) =>
      println("Unification failed: " ++ show(err))
      if (msg.length > 0) {
        println(msg)
      }
  }
}

def main() = reporting {
  val intType = Con("Int", [])
  val listType = Con("List", [intType])
  val typeVar = Var("a")

  do raise(UnificationManyFailure([intType], [listType, typeVar]), "")
}
```

examples/benchmarks/other/unify.effekt
```diff
@@ -0,0 +1,173 @@
...
+
+def main() = {
+  ()
+  // reporting {
+  //   val intType = Con("Int", [])
+  //   val listType = Con("List", [intType])
+  //   val typeVar = Var("a")
+
+  //   unify(typeVar, listType)
+  // }
```
jiribenes, line 130: Note: this will never terminate currently, but I want to see what happens :)
```suggestion
  reporting {
     val intType = Con("Int", [])
     val listType = Con("List", [intType])
     val typeVar = Var("a")

     unify(typeVar, listType)
  }
```

examples/benchmarks/other/unify.effekt
```diff
@@ -0,0 +1,172 @@
...
+    body()
+  }
+
+  res match {
+    case Success(subst) => {
+      println("Unification successful!")
+      println("Substitution:")
+      for { subst.each } { case (k, v) =>
+        println("  " ++ k ++ " -> " ++ showType(v))
+      }
```
jiribenes, line 112: Funnily enough, this doesn't seem to be the culprit of non-termination: even if you replace it with a `subst.foreach { (k, v) =>`, it still hangs.

effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
```diff
@@ -160,6 +160,9 @@ object Normalizer { normal =>
 
   def normalize(s: Stmt)(using C: Context): Stmt = s match {
 
+     case Stmt.Def(id, block, body) if !C.usage.contains(id) || C.usage.get(id).contains(Usage.Never) => // see #798
+      normalize(body)
+
```
jiribenes, line 165: Argh, indentation!
```suggestion
    case Stmt.Def(id, block, body) if !C.usage.contains(id) || C.usage.get(id).contains(Usage.Never) => // see #798
      normalize(body)
```

examples/benchmarks/other/unify.effekt
```diff
@@ -0,0 +1,176 @@
...
+      if (occurs(y, substTy1)) {
+        do raise(OccursCheckFailure(y, substTy1), "")
+      } else {
+        subst.put(y, substTy1)
+      }
+
+    // If both are constructors, unify their arguments
+    case (Con(name1, args1), Con(name2, args2)) =>
+      if (name1 != name2 || args1.size != args2.size) {
+        do raise(UnificationFailure(substTy1, substTy2), "Different number of arguments!")
```
jiribenes, line 64: We can even customise the errors here :)
```suggestion
      if (name1 != name2) {
        do raise(UnificationFailure(substTy1, substTy2), "Different constructors!")
      } else if (args1.size != args2.size) {
        do raise(UnificationFailure(substTy1, substTy2), "Different number of arguments!")
```

examples/benchmarks/other/unify.effekt
```diff
@@ -0,0 +1,178 @@
...
+      val intType = Con("Int", [])
+      val listType = Con("List", [intType])
+      val typeVar = Var("a")
+
+      unify(typeVar, listType)
+    }
+  } { printBinding }
+
+  // the actual test
+  val N = 12
```
jiribenes, line 160: `N = 12` means unification of two trees of depth 12, i.e. tree of tuples with `2^12 = 4096` variables in the leaves.
We can tweak this number or even use the `benchmarks/runner`...

examples/benchmarks/other/unify.effekt
```diff
@@ -0,0 +1,183 @@
...
+  case Con(name, args) =>
+    name ++ "(" ++ args.map { t => showType(t) }.join(", ") ++ ")"
+}
+
+def show(err: UnificationError): String = err match {
+  case OccursCheckFailure(variable, ty) =>
+    "Occurs check failed: " ++ variable ++ " occurs in " ++ showType(ty)
+  case UnificationFailure(ty1, ty2) =>
+    "Cannot unify " ++ showType(ty1) ++ " with " ++ showType(ty2)
+    case UnificationManyFailure(tps1, tps2) =>
```
jiribenes, line 99: Fix indentation
```suggestion
  case UnificationManyFailure(tps1, tps2) =>
```

b-studios:
IMO we can merge this now

jiribenes:
Great, let's! :shipit:

PR #800, opened 2025-01-24
Try polymorphism boxing after optmization

b-studios:
@phischu

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
+++ b/effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
@@ -39,7 +39,7 @@ class Deadcode(reachable: Map[Id, Usage]) extends core.Tree.Rewrite {
         declarations.map(rewrite),
         // drop unreachable externs
         m.externs.collect {
-          case e: Extern.Def if used(e.id) => e
+          case e: Extern.Def if used(e.id) || e.id.name.name.startsWith("box") || e.id.name.name.startsWith("unbox") => e
           case e: Extern.Include => e
         },
         // drop unreachable definitions
--- a/effekt/shared/src/main/scala/effekt/generator/llvm/LLVM.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/llvm/LLVM.scala
@@ -39,7 +39,7 @@ class LLVM extends Compiler[String] {
   // The Compilation Pipeline
   // ------------------------
   // Source => Core => Machine => LLVM
-  lazy val Compile = allToCore(Core) andThen Aggregate andThen core.PolymorphismBoxing andThen optimizer.Optimizer andThen Machine map {
+  lazy val Compile = allToCore(Core) andThen Aggregate andThen optimizer.Optimizer andThen core.PolymorphismBoxing andThen Machine map {
     case (mod, main, prog) => (mod, llvm.Transformer.transform(prog))
   }
 
@@ -52,7 +52,7 @@ class LLVM extends Compiler[String] {
   // -----------------------------------
   object steps {
     // intermediate steps for VSCode
-    val afterCore = allToCore(Core) andThen Aggregate andThen core.PolymorphismBoxing andThen optimizer.Optimizer
+    val afterCore = allToCore(Core) andThen Aggregate andThen optimizer.Optimizer andThen core.PolymorphismBoxing
     val afterMachine = afterCore andThen Machine map { case (mod, main, prog) => prog }
     val afterLLVM = afterMachine map {
       case machine.Program(decls, prog) =>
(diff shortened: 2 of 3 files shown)


phischu:
```
def foo[A](x: Int){ bar: Int => A }: A = {
  bar(x)
}

def baz(x: Int): Int = {
  if (x > 0) {
    foo(x - 1){baz}
  } else { 0 }
}

def main() = { 
  println(baz(10))
}
```

effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala
```diff
@@ -522,16 +465,14 @@ object PolymorphismBoxing extends Phase[CoreTransformed, CoreTransformed] {
...
         val bargs = (bcoercers zip bparams).map { case (c, p) => c(Block.BlockVar(p.id, p.tpe, Set.empty)) }
         Block.BlockLit(ftparams, bparams.map(_.id), vparams, bparams,
-          Def(inner, block,
             coerce(Stmt.App(
-              Block.BlockVar(inner, block.tpe, block.capt),
+              block,
               (targs map transformArg) ++ (ftparams map core.ValueType.Var.apply),
               vargs,
-              bargs), tresult)))
+              bargs), tresult))
```
marzipankaiser, line 475: Just to be sure [I'm sure you know.]: Can
a) Machine now handle `App(BlockLit(...), ...)`, or
b) are you ensuring this is not generated here?

I vaguely remember this being a problem (which is why the implementation is as it is).
You probably want to do b (i.e. Beta-reduce) if this runs after optimizations.

marzipankaiser (commented):
Changes to `PolymorphismBoxing.scala` look good to me so far.

effekt/shared/src/main/scala/effekt/core/PolymorphismBoxing.scala
phischu: It can handle it now, thanks for the reminder.

phischu:
As a follow-up it will also allow us to delete the special casing for LLVM in the optimizer [here](https://github.com/effekt-lang/effekt/blob/b115f3ad2e89da907f48295820fb6879ced27877/effekt/shared/src/main/scala/effekt/core/optimizer/Optimizer.scala#L24).

b-studios:
While at it, maybe you can also add tests for #1012 and #1030 so we can immediately close them.

mattisboeckle:
The tests still fail. I get `values are not equal even if they have the same toString()`
So some variable is not getting renamed. I could not figure out which one

marzipankaiser:
@marzipankaiser Pls take a look

effekt/jvm/src/test/scala/effekt/core/PolymorphismBoxingTests.scala
```diff
@@ -18,6 +18,8 @@ abstract class AbstractPolymorphismBoxingTests extends CorePhaseTests(Polymorphi
     "boxInt" -> Id("boxInt"),
     "unboxInt" -> Id("unboxInt"),
     "unboxString" -> Id("unboxInt"),
+    "coerceIntPos" -> Id("@coerceIntPos"),
+    "coercePosInt" -> Id("@coercePosInt"),
```
marzipankaiser: The problem is here. `TestRenamer` *only* renames non-toplevel *bound* names. These names should be the ones generated by PolymorphismBoxing, e.g. `PolymorphismBoxing.TBoxedInt.name` (after casting/type annotating)

mattisboeckle: Thanks that was it

phischu: Puh, thanks!

phischu:
Super, now rebase this onto master (and force push to this branch).

effekt/jvm/src/test/scala/effekt/core/TestRenamer.scala
```diff
@@ -84,6 +84,9 @@ class TestRenamer(names: Names = Names(Map.empty), prefix: String = "") extends
         Block.BlockLit(tparams map rewrite, cparams map rewrite, vparams map rewrite, bparams map rewrite,
           rewrite(body))
       }
+    case Block.BlockVar(id, annotatedTpe, annotatedCapt) => 
+      val resolvedCapts = annotatedCapt map rewrite
+      withBinding(id) { Block.BlockVar(rewrite(id), rewrite(annotatedTpe), resolvedCapts) }
```
marzipankaiser: Why is this using `withBinding`? This is not a binding occurence usually?

mattisboeckle: Yeah, this was incorrect. Fixed in the new commits

PR #802, opened 2025-01-24
More filesystem benchmarks

phischu:
@marvinborner I would like these and the other benchmarks in `input_output` integrated into effekt plots.


examples/benchmarks/input_output/financial_format.effekt
```diff
@@ -0,0 +1,117 @@
...
+  assertTag(44)
+
+  assertTag(10)
+}
+
+
+def run(n: Int) = {
+  with on[IOError].panic;
+
+  val filename = "/tmp/financial_information_exchange.txt"
```
b-studios, line 97: This will be problematic when run in parallel by the different backend tests, like before.

marvinborner:
The current setup runs every benchmark that's listed in `benchmarks/config_${BACKEND}.txt` with their respective argument. Since the IO benchmarks don't use a configurable argument, we can just add the filename without it -- the benchmark will be run regardless.

We can add the lines in this PR, do we want to run them for JS and LLVM?

phischu:
I figured it out: classical buffer overflow

examples/benchmarks/config_js.txt
```diff
@@ -22,3 +22,11 @@ effect_handlers_bench/product_early 32768
 effect_handlers_bench/resume_nontail 1366
 effect_handlers_bench/tree_explore 14
 effect_handlers_bench/triples 256
+input_output/dyck_one 5
```
marvinborner, line 25: Do we want to use larger numbers here as well?

marvinborner (approved):
otherwise this looks good to me. If I'm not mistaken the benchmarks will be tracked automatically in plots after merging, so we don't have to change anything else.

examples/benchmarks/config_js.txt
phischu: Ah, yes, please adjust them such that each run takes between 0.5 and 2 seconds.

PR #785, opened 2025-01-21
Fix byte-related LLVM segfaults

marvinborner:
Closes #783. And closes #778.

I believe the way we currently handle size calculations for stack pushes/pops of non-structural types has some flaws for non-pointer values. Some changes in how we use the GEP instruction will probably fix this, but I find its syntax really confusing. I have a proof-of-concept alternate fix at the end.

To bypass the calculation issues, we can also just change the size of a byte to 8 byte instead. This *hotfix* fixes the segfault in the original issue as well as all of my minified versions.

Some additional explanations of the issue using an example program:

```scala
def foo { action: Byte => Unit }: Unit =
  action(42.toByte)

def main() = {
  foo { b =>
    // does not segfault when printlns are swapped
    println(42)
    println(b)
  }
}
```

<details>

<summary>Pseudo-flow with relevant (for the segfault) lines marked</summary>

```
main(stack)

main(stack):
alloc(3 * 8 = 24)
push returnAddress_1
push sharer_5
push eraser_7
blockLit = vtable[blockLit_3289_clause_13]
foo(blockLit, stack)

foo(blockLit, stack):
alloc(3 * 8 = 24)
push returnAddress_40
push sharer_5
push eraser_7
literal = 42
byte = toByte(literal)
vtable = blockLit[0]
closure = blockLit[1]
fp = *vtable // = blockLit_3289_clause_13
fp(closure, byte, stack)

blockLit_3289_clause_13(closure, byte, stack):
alloc(3 * 8 + 1 = 25) // <---------------------------
push byte
push returnAddress_14
push sharer_26
push eraser_30
literal = 42
printlnInt(literal, stack)

printlnInt(value, stack): // _4
str = show(value)
unit = println(str)
dealloc(3 * 8 = 24)
pop _
pop _
pop returnAddress // _14
returnAddress(unit, stack) // <---------------------------

returnAddress_14(ret, stack):
dealloc(1)
pop byte
erase(ret)
alloc(3 * 8 = 24)
push returnAddress_17
push sharer_5
push eraser_7
printlnByte(byte, stack)

printlnByte(byte, stack):
str = show(byte)
unit = println(str)
dealloc(24)
pop _
pop _
pop returnAddress // _17
returnAddress(unit, stack)

returnAddress_17(unit, stack):
dealloc(24)
pop _
pop _
pop returnAddress // _40
returnAddress(unit, stack)

returnAddress_40(unit, stack)
dealloc(24)
pop _
pop _
pop returnAddress // _1
[...]

PR #803, opened 2025-01-26
Fix LLVM byte alignment for real

### LINKED ISSUE #794: Properly fix alignment of bytes on stack
The following example still segfaults:

```scala
def foo { action: (Byte, Byte) => Unit }: Unit =
  action(42.toByte, 42.toByte)

def main() = {
  foo { (b0, b1) =>
    println(42)
    println(b0)
    println(b1)
  }
}
```

I see three solutions:

1. Undo the optimization and reserve stack space twice when pushing a frame: once for the environment and once for the header.
2. Change the definition of `%Byte` to `i64` in llvm and hope for the best. This will require greate care in extern functions acting on bytes.
3. Insert `getelementptr` instructions to ask llvm to calculate the size of a frame for us.

Unsurprisingly I am for (1).

marvinborner:
Closes #794.

I've also added a test but it's a bit useless since the buggy parts all get optimized away and we currently only run tests *with* optimizations.


effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -580,19 +580,22 @@ object Transformer {
   }
 
   def pushFrameOnto(stack: Operand, environment: machine.Environment, returnAddressName: String, sharer: Operand, eraser: Operand)(using ModuleContext, FunctionContext, BlockContext) = {
+    if (!environment.isEmpty) {
```
marvinborner, line 583: Should we create an additional function for this? (like in `popEnvironmentFrom`)

marvinborner:
After some more thought I'm not sure how my initial change would fix all of the alignment problems.

Would my second solution in which we use *packed* structural types, also be okay (2-byte change)? In general not packing structs has all kinds of potential benefits for performance but they don't apply to us anyway, since such optimizations would segfault right now(?). Because our stack allocation is fundamentally based on knowing the *exact* allocation size, I believe this to be the only correct choice.

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -428,7 +428,7 @@ object Transformer {
       case machine.Type.Prompt()     => 8 // TODO Make fat?
       case machine.Type.Stack()      => 8 // TODO Make fat?
       case machine.Type.Int()        => 8 // TODO Make fat?
-      case machine.Type.Byte()       => 8
+      case machine.Type.Byte()       => 1
```
b-studios, line 431: Didn't @phischu just change this from 1 to 8 to fix something else?

marvinborner: This was an insufficient fix as described in https://github.com/effekt-lang/effekt/pull/785#issuecomment-2607553642

b-studios:
I'll interpret the thumbs up by @phischu as an approve and merge this :)

PR #810, opened 2025-02-03
Panic on prompt duplication

serkm:
Detect an invalid state dynamically and panic (#244).

Todo: print proper error message


b-studios:
@serkm Do you want to print an error message as part of this PR, or should we just merge this?

Could you add one test where this is triggered, so we can make sure it works? (I guess an example is maybe difficult to construct).

b-studios:
The concrete message is not so important, it's more important that it's unique so that users can find this PR and potentially related issues.

PR #816, opened 2025-02-05
Specialize on most common opening flags

### LINKED ISSUE #809: Cannot write files on JS backend
_Originally reported by @flovalle1_

On versions 0.17.0, 0.18.0, 0.19.0 running the following file on the JS backend:

```scala
def main(): Unit = {
  with on[IOError].panic
  with filesystem

  do writeFile("output", "foo bar baz")
}
```
reports `no such file or directory`.

This worked back in 0.16.0 (checked via `effekt-nix`), and it's somewhat easy to alleviate the problem by using `--no-optimize`.

---

Here's the two `main.js` for comparison (I kept only the `main` function), you can clearly see way more inlining:
https://gist.github.com/jiribenes/dc1596855c102df3f66e846b48b2c119

phischu:
Should fix #809 

Please test on Windows and Mac

### DIFF
--- a/examples/stdlib/io/filesystem/write_read_append.check
+++ b/examples/stdlib/io/filesystem/write_read_append.check
@@ -1,0 +1,2 @@
+hello!
+hello! world!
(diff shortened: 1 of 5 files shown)


b-studios:
Looks good to me, but we need to have someone run it on windows

phischu:
Please also add tests for writing and appending to files

PR #822, opened 2025-02-07
Rebalance benchmarks

phischu:
- [ ] Prevent `iterate_increment` from being constant-folded
- [ ] Prevent `factorial_accumulator` from being constant-folded
- [ ] Change input sizes to more evenly distribute time
- [ ] Make input sizes round numbers
- [x] Add generator
- [x] Add handler sieve
- [ ] Add checkpointing


phischu:
@mattisboeckle it would be great if you could figure out why `iterate_increment` and `factorial_accumulator` are constant-folded (in the llvm backend), and how we can cheaply prevent this (perhaps by using different arithmetic operations involving a magical modulus?)

mattisboeckle:
I've adjusted the magic `mod` value to 2^63-1 and added a comment explaining, why we do this.

examples/benchmarks/duality_of_compilation/factorial_accumulator.effekt
```diff
@@ -4,7 +4,8 @@ def factorial(a: Int, i: Int): Int =
   if (i == 0) {
     a
   } else {
-    factorial(i * a, i.mod(1000000007) - 1)
+    // We are using `i.mod(2^63-1)` to prevent LLVM from constant folding this whole program
+    factorial(i * a, i.mod(9223372036854775807) - 1)
```
jiribenes, line 8: 👀 What I'm saying might be obvious, but 2^63 - 1 is **not** a prime, right? (unlike, say 2^61 - 1 or 1000000007, which cannot be factored)
Also, I'm not really sure what will this do on JS where the usable number range is bounded by something like 2^53-1...

mattisboeckle: It's not prime, but does it really *need* to be for this use case?

I do agree with the JS part, this should probably just be 2^53-1

mattisboeckle: I tested JS and at least locally the modulo value does not seem to make any difference to performance or compilation, but it might still be better for stability

jiribenes:
In #1472, I added the benchmarks themselves (+ checkpointing), but not the configs themselves, since I don't want to balance it (for now).

phischu:
Perhaps we can close this then.

PR #826, opened 2025-02-14
Transform machine to direct style, if possible

b-studios:
Machine doesn't receive as much attention as it should. One reason is that I cannot read it.

This PR tries to fix this. On top, as a first improvement it also adds a phase that performs contification (transforming functions to continuations)


b-studios:
I don't understand why `word_count_ascii.effekt` does consistently fail for me on my machine:

```
run --backend llvm examples/benchmarks/input_output/word_count_ascii.effekt
[info] running effekt.Server --no-exit-on-error --backend llvm examples/benchmarks/input_output/word_count_ascii.effekt
[error] Process exited with non-zero exit code 139.
```
but the tests here in CI pass...

b-studios:
I opened a separate issue for the test that fails locally for me: #827

Issue #524, opened 2024-07-24
LSP inlay hints


jiribenes:
We could use _inlay hints_ from [LSP v3.17](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_inlayHint) for:
1. showing the inferred capabilities in a principled way (`[{io}] def ...`)
2. showing the inferred return types and effects of functions (`def bar(x: Int) [: Int / Log] = ...`)
3. showing the inferred types of bindings (`val foo [: Int] = ...`)
4. showing the expected & actual type of a typed hole (`"answer is: " ++ <{ 42 [: Int] }> [: String]`)
5. showing the inferred boxes/unboxes?
6. showing the chosen overloads?

Of course, these would probably need to be configurable so that the code is still somewhat navigable. :)

As far as I can tell, we're already doing 1. in a somewhat ad-hoc way:
https://github.com/effekt-lang/effekt/blob/7b7629c942e2809cab08b0122e244ba3bcbb9410/effekt/js/src/main/scala/effekt/LanguageServer.scala#L138-L143
but it would be nice to use the LSP protocol instead.

There are also a lot of other quality of life related features like "if we already have an inlay, we can just double-click it to apply" like in [rust-analyzer](https://rust-analyzer.github.io/thisweek/2023/04/17/changelog-177.html) (see first video there for a demo).

jiribenes:
#### How to start with this PR:

1. bump LSP4J in Kiama:

https://github.com/effekt-lang/kiama/blob/ab6aef78fc4ce99a9ad2db8555718d00b03f7b8a/build.sbt#L11

2. catch the `textDocument/inlayHint` event

https://github.com/effekt-lang/kiama/blob/ab6aef78fc4ce99a9ad2db8555718d00b03f7b8a/jvm/src/main/scala/kiama/util/Server.scala#L518-L519

jiribenes:
steps 2-6 are in https://github.com/effekt-lang/kiama/pull/7
needs a new VSCode plugin version for testing (needs VSCode 8.0+ API)

PR #828, opened 2025-02-15
Inlay hints: scaffolding and captures

jiribenes:
~~Requires https://github.com/effekt-lang/kiama/pull/7 (blocked)~~
Requires the newest version of the VSCode extension (v0.2.4)
Requires that the VSCode user enables inlay hints (see a separate comment below on how to set them up).

As part of the scaffolding, I also replaced Scala 2 `implicit`s with Scala 3 `using`

Instead of using our hand-rolled `inferredCaptures` command, we can _just_ use inlay hints 🥳 
(See #524 for more details)

Screenshots:
Screenshot 2025-02-15 at 10 49 38
Screenshot 2025-02-15 at 10 54 19

Bugs:
- see comment below.

### DIFF
--- a/effekt/js/src/main/scala/effekt/LanguageServer.scala
+++ b/effekt/js/src/main/scala/effekt/LanguageServer.scala
@@ -122,7 +122,9 @@ class LanguageServer extends Intelligence {
   @JSExport
   def inferredCaptures(path: String): js.Array[lsp.CaptureInfo] = {
     typecheck(path)
-    getInferredCaptures(VirtualFileSource(path)).map {
+    val source = VirtualFileSource(path)
+    val range = kiama.util.Range(source.offsetToPosition(0), source.offsetToPosition(source.charCount))
+    getInferredCaptures(range).map {
       case (p, c) => new lsp.CaptureInfo(toLSPPosition(p), c.toString)
     }.toJSArray
   }
--- a/effekt/jvm/src/main/scala/effekt/Server.scala
+++ b/effekt/jvm/src/main/scala/effekt/Server.scala
@@ -221,20 +221,6 @@ trait LSPServer extends kiama.util.Server[Tree, EffektConfig, EffektError] with 
     tpe1 != tpe2 || effs1 != effs2
   }
 
-  case class CaptureInfo(location: Location, captureText: String)
-
-  override def executeCommand(src: Source, params: ExecuteCommandParams): Option[Any] =
-    None
-    /*
-    if (params.getCommand == "inferredCaptures") {
-      val captures = getInferredCaptures(src)(using context).map {
-        case (p, c) => CaptureInfo(positionToLocation(p), TypePrinter.show(c))
-      }
-      if (captures.isEmpty) None else Some(captures.toArray)
-    } else {
-      None
-    }*/
-
   override def createServices(config: EffektConfig) = new LSPServices(this, config)
 
   // Class to easily test custom LSP services not (yet) meant to go into kiama.Services
--- a/effekt/jvm/src/main/scala/effekt/Server.scala
+++ b/effekt/jvm/src/main/scala/effekt/Server.scala
@@ -30,7 +30,7 @@ trait LSPServer extends kiama.util.Server[Tree, EffektConfig, EffektError] with 
     diagnostic(message.range, lspMessaging.formatContent(message), message.severity)
 
   override def getDefinition(position: Position): Option[Tree] =
-    getDefinitionAt(position)(context)
+    getDefinitionAt(position)(using context)
 
   /**
    * Overriding hook to also publish core and target for LSP server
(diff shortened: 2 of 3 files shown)


jiribenes:
If you change a file with captures too quickly, the server sends `null` as a response (since we don't have any captures yet!) _and either the client or the server caches it_, so it's no longer possible to ever get captures for that particular state of the file without restarting the LSP.

jiribenes:
Here's how you can set up how your inlay hints look in VSCode (+ then you can customise them in your theme):
Screenshot 2024-11-07 at 22 50 52
Screenshot 2024-11-07 at 22 50 46

jiribenes:
Excuse the bad location, but we can even print types and effects for functions :)
Screenshot 2025-02-15 at 13 21 12

effekt/jvm/src/main/scala/effekt/Server.scala
```diff
@@ -115,17 +115,41 @@ trait LSPServer extends kiama.util.Server[Tree, EffektConfig, EffektError] with
...
+
+    // TODO: Also do this for un-annotated `val`s and `var`s.
+    val unannotated = getTreesInRange(range)(using context).map { trees =>
+      trees.collect {
+        // Functions without an annotated type:
+        case fun: FunDef if fun.ret.isEmpty => for {
+          // TODO(question): `context.inferredTypeAndEffectOption(fun)` didn't work here. Why?
+          sym <- context.symbolOption(fun.id)
+          tpe <- context.functionTypeOption(sym)
+          pos <- positions.getFinish(fun.bparams) // TODO(bug): This position doesn't always work. Ideally, I'd use the position of `fun.ret`, but that doesn't exist!
```
jiribenes, line 144: cc @b-studios
- `context.inferredTypeAndEffectOption(fun)` didn't work here, but I don't know why [?]
- `pos <- positions.getFinish(fun.bparams)` sometimes returns bogus positions like here for `hello`:
image

jiribenes: The latter here is because `hello` doesn't have block params. Sigh.
I'd like us to do better here, see https://github.com/effekt-lang/effekt/issues/833.

b-studios:
Regarding the bogus positions, in the meeting we discussed to maybe have our own `Option` type to improve position info (since `None` is pooled).

```scala
enum Optional[+T] {
  case None() // we need to make sure this is not pooled and we get a fresh instance
  case Some(v: T)
}
```

another option (haha!) would be to add one layer of indirection via a 'box'

```scala
class Positioned[+T](value: T)
implicit def autoUnwrap[T](p: Positioned[T]): T = p.value
// WARNING: this creates a new instance and would required copying them... this is very annoying!
implicit def autoWrap[T](v: T): Positioned[T] = new Positioned(v)
```

and use this for optional values in the AST. The problem actually might also exist for Lists, where `Nil` could be pooled. Here is how this could look like:

```scala
type Maybe[+T] = Positioned[Option[T]]
type Many[+T] = Positioned[List[T]]

case class OpClause(id: IdRef,  tparams: Many[Id], vparams: Many[ValueParam], bparams: Many[BlockParam], ret: Maybe[Effectful], body: Stmt, resume: IdDef) extends Reference
```

jiribenes:
Blocked on #847, I guess?

effekt/jvm/src/main/scala/effekt/Server.scala
jiribenes: Since this concerns only the function annotations, let's put this into a follow-up issue.

Here's the code if needed. :)
```scala
    // TODO: Also do this for un-annotated `val`s and `var`s.
    val unannotated = getTreesInRange(range)(using context).map { trees =>
      trees.collect {
        // Functions without an annotated type:
        case fun: FunDef if fun.ret.isEmpty => for {
          // TODO(question): `context.inferredTypeAndEffectOption(fun)` didn't work here. Why?
          sym <- context.symbolOption(fun.id)
          tpe <- context.functionTypeOption(sym)
          pos <- positions.getFinish(fun.bparams) // TODO(bug): This position doesn't always work. Ideally, I'd use the position of `fun.ret`, but that doesn't exist!
        res = InlayHint(InlayHintKind.Type, pos, pp": ${tpe.result} / ${tpe.effects}", resolveToLabel = true, paddingLeft = true, effektKind = "missing-type-annotation")
        } yield res
      }.flatten
    }.getOrElse(Vector())

    val allHints = captures ++ unannotated
    if allHints.isEmpty then None else Some(allHints)
```

jiribenes: ~> moved to #869

effekt/js/src/main/scala/effekt/LanguageServer.scala
```diff
@@ -122,7 +122,9 @@ class LanguageServer extends Intelligence {
   @JSExport
   def inferredCaptures(path: String): js.Array[lsp.CaptureInfo] = {
     typecheck(path)
-    getInferredCaptures(VirtualFileSource(path)).map {
+    val source = VirtualFileSource(path)
+    val range = kiama.util.Range(source.offsetToPosition(0), source.offsetToPosition(source.charCount))
+    getInferredCaptures(range).map {
       case (p, c) => new lsp.CaptureInfo(toLSPPosition(p), c.toString)
     }.toJSArray
   }
```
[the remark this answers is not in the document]
b-studios: I think this was used in the docs on the website somewhere, but maybe that's not the case anymore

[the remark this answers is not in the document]
b-studios: Maybe I was also just preparing to translate the examples from the oopsla artifact to the website, but never did.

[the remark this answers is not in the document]
b-studios: I cannot find it anywhere in the website repo, so I think this was only in preparation to add it to the website Monaco editor.

b-studios (approved):

b-studios:
Here is design proposal on how to add position info in the parser: #875

PR #830, opened 2025-02-16
Move direct style transformation from machine to core

b-studios:
Do not use old transformation


b-studios:
This is not ready to merge since it slows down the JS backend at the moment!

libraries/common/list.effekt
```diff
@@ -311,8 +311,10 @@ def reverseOnto[A](l: List[A], other: List[A]): List[A] = l match {
 /// ```
 ///
 /// O(N)
-def append[A](l: List[A], other: List[A]): List[A] =
-  l.reverse.reverseOnto(other)
+def append[A](l: List[A], other: List[A]): List[A] = l match {
+  case Nil() => other
+  case Cons(a, rest) => Cons(a, append(rest, other))
+}
```
b-studios, line 317: Lol, thanks. It would be interesting to see the performance implications on JS.

b-studios: Update: on JS this slows things down...

```
// 140-150ms before
// 240ms after
def main() = {
  var i = 0;
  val l1 = fill(100000, 1)
  val l2 = l1.append(l1).append(l1).append(l1).append(l1)
  println(l2.sum)
}
```

b-studios:
Once the CI passes, we are ready to go...

Issue #827, opened 2025-02-15
Wordcount ASCII segfaults on MacOS


b-studios:
The following test locally segfaults on my machine:

https://github.com/effekt-lang/effekt/blob/master/examples/benchmarks/input_output/word_count_ascii.effekt

Concretely:

```
running effekt.Server --backend llvm examples/benchmarks/input_output/word_count_ascii.effekt
[error] Process exited with non-zero exit code 139.
```

I bisected and found the problem was introduced in PR #705, specifically in commit:

https://github.com/effekt-lang/effekt/pull/705/commits/b541ce7a43bc4fc8b9b527eb90ab74786c8bf8ab

The one before (32d06bc2) does return `6170`.

This will be very tricky to debug, since our LLVM experts run Linux Machines. I am assigning @phischu anyway since he might have a first idea why this could be.

---

If you also want to bisect, here is the script (bisect.sh) that I used:

```
#!/bin/bash
sbt "effektJVM/run --backend llvm examples/benchmarks/input_output/word_count_ascii.effekt"
exit $?
```

b-studios:
Note that the change that introduced the problem is an effekt-only change. No compiler implementation changed, only the source program.

It might be a good start to inspect the resulting core / machine for the two versions to see what's the significant difference and how that could result in problems on LLVM.

phischu:
There is a slim chance that this is fixed in #802

b-studios:
@phischu  No, sadly not.

mattisboeckle:
I can reproduce this on Ubuntu, definitely not a MacOS only issue.
The program does run for me if using `--no-optimize`

PR #831, opened 2025-02-17
Finer suppression for uv valgrind

mattisboeckle:
This PR adds some finer granularity for Valgrind suppression.

We can check for specific errors that originate from libuv instead of suppressing all undefined value errors.

Currently blocked on #827 because I can't test if this is all the suppressions we need.


mattisboeckle:
Let me know if there is a better location for the suppression file or names of the suppressions themselves.

mattisboeckle:
After trying this for a while i've come to that conclusion that is not possible to do this much finer.
For the time being we are stuck with suppressing all undefined value errors

PR #838, opened 2025-02-20
Fixed ResizableArray bounds check including a test

airzocker:
Fixed a bug where the bounds check of ResizableArray used `&&` instead of `||` for the bounds check condition, leading to it never `raise()`ing OutOfBounds.


b-studios (approved):

b-studios:
Great, thanks for your contribution!

PR #839, opened 2025-02-21
Extend array stdlib

### LINKED ISSUE #804: Expand stdlib 'array' module
Add more functions that work on `array`, ideally all of the ones from `list`, but we really ought to prioritise the most important ones.

Every new function should have a test in `examples/stdlib/array/<functionname>.effekt`.

dvdvgt:
closes #804

This PR aims to extend the stdlib of `array`. Most of the added functions are in-place compared to their counterparts for `list`.

Feel free to bike-shed. I am looking forward to your feedback.

## TODO

- [x] Add complete test suit for all functions
- [x] There is `indexOf`, `lastIndexOf` and `find`, probably not all of them are needed.
- [x] It would be nice if someone could check the implementation `c_array_fill` as I am not too familiar with LLVM and the memory management.
- [x] ~~Decide on naming convention: ruby- vs scala-style (e.g. `map!` vs `mapped`)?~~ Change everything to Ruby style
- [x] Measure performance of the extern definition of `fill` compared to a definition in effekt


libraries/common/array.effekt
```diff
@@ -185,6 +209,282 @@ def sum(list: Array[Int]): Int = {
...
+
+/// Check if predicate is true for all elements.
+///
+/// O(N)
+def all[A](arr: Array[A]) { predicate: A => Bool }: Bool = {
+  var result = true
+  arr.foreach { x =>
+    if (not(predicate(x))) result = false
+  }
+  result
```
jiribenes, line 307: optimisation 🏃 
```suggestion
  var result = true
  arr.foreach { (x) {label} =>
    if (not(predicate(x))) {
      result = false
      label.break()
    }
  }
  result
```

libraries/common/array.effekt
```diff
@@ -81,7 +82,14 @@ extern global def unsafeSet[T](arr: Array[T], index: Int, value: T): Unit =
...
 
-
+/// Fills a given array in-place.
+extern def fill[T](arr: Array[T], filler: T): Unit = 
+  js "${arr}.fill(${filler})"
+  chez "(vector-fill! ${arr} ${filler})"
+  llvm """
+    %z = call %Pos @c_array_fill(%Pos ${arr}, %Pos ${filler})
+    ret %Pos %z
+  """
```
jiribenes, line 92: Nice!
We should be able to use this function in the `array::array` constructor now, right?

jiribenes: It would even be nice to look at other uses of the `each(0, arr.size) { i => arr.unsafeSet(i, x) }` idiom in the codebase, I think there are many places in `io` where we use that idiom to fill an array with a value `x` (which should now be replaceable with just `arr.fill(x)`...

libraries/common/array.effekt
```diff
@@ -185,6 +209,282 @@ def sum(list: Array[Int]): Int = {
...
+  quickSort(0, arr.size - 1)
+}
+
+/// swap the value at index i with that at index j
+def swap[A](arr: Array[A], i: Int, j: Int): Unit / Exception[OutOfBounds] = {
+  if (i < 0 || i >= arr.size || j < 0 || j >= arr.size) do raise(OutOfBounds(), "Index out of bounds")
+  val tmp = arr.unsafeGet(i)
+  arr.unsafeSet(i, arr.unsafeGet(j))
+  arr.unsafeSet(j, tmp)
+}
```
jiribenes, line 486: 👀 Somewhat duplicated with the `unsafeSwap` defined locally in `sort`.
Would it be worth it to extract that one out to the outer scope and then define `swap` similarly to how `set` and `get` are defined via their unsafe variants?

libraries/common/array.effekt
```diff
@@ -185,6 +209,282 @@ def sum(list: Array[Int]): Int = {
...
+      lastFound = Some(idx)
+    }
+  }
+  lastFound
+}
+
+/// Perform binary search to find the index of a given element.
+/// 
+/// O(log N)
+def find[A](arr: Array[A], toFind: A) { ord: (A, A) => Ordering }: Option[Int] = {
```
jiribenes, line 425: There are a few uses where `lowerBound` and `upperBound` are also helpful (for algorithmic stuff or for stuff like "if I have a list of offsets of newlines in a file and an offset, get me the offset of the start of the line)-- it's again binary search. But I don't think it's _too_ important.

libraries/common/array.effekt
```diff
@@ -185,6 +209,282 @@ def sum(list: Array[Int]): Int = {
...
+    result.unsafeSet(i, arr.unsafeGet(i + start))
+  }
+  result
+}
+
+/// Return a slice from start (inclusive) to end (exclusive).
+/// Returns a new array.
+///
+/// O(N)
+def slice[A](arr: Array[A], start: Int, end: Int): Array[A] = {
```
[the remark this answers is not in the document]
dvdvgt: Hm... `take` -> `taked`? 🤔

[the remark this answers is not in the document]
jiribenes: Funnily enough, here the Ruby-style convention of `take!` and `drop!` would work well.
But I've checked and Scala just calls these `take` and `drop` 🤷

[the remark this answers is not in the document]
dvdvgt: We can also use ruby-style. We can discuss on the next Effekt meeting. I will just continue with the Scala style for now.

jiribenes:
I _think_ the `fill` now needs to be added here?

https://github.com/effekt-lang/effekt/blob/5599ca1101e0d52538ba84adaa9e1bfdc1c716be/effekt/shared/src/main/scala/effekt/core/vm/Builtin.scala#L250-L263

dvdvgt:
The differences for the `are_we_fast_yet/sieve` test sound concerning. That is a 30% increase for the static dispatches and 23% for branches.

```
 Summary(
-  staticDispatches = 16737,
+  staticDispatches = 21738,
   dynamicDispatches = 0,
   patternMatches = 0,
-  branches = 21735,
-  pushedFrames = 4999,
-  poppedFrames = 4999,
+  branches = 26736,
+  pushedFrames = 5000,
+  poppedFrames = 5000,
   allocations = 0,
```

libraries/common/array.effekt
dvdvgt: Hm looking through the result of searching for the regex `each\(0,[\s]*[\w]*.size\)` shows no other occurrence of this idiom using arrays. There are some occurrences where this is done for `bytearray`s.

libraries/common/array.effekt
```diff
@@ -185,6 +228,340 @@ def sum(list: Array[Int]): Int = {
...
+  val res = allocate(nMatches)
+  var idx = 0
+  arr.foreach { x =>
+    if (pred(x)) {
+      res.unsafeSet(idx, x)
+      idx = idx + 1
+    }
+  }
+  res
+}
```
jiribenes, line 335: The semantics here can be a bit surprising: `pred` will get called twice on each element -- but it can have side effects which will be then executed multiple times. 👀

(similarly for `partitioned`)

jiribenes: Since I'm nitpicking here already 😇 

```suggestion
/// Filters a given array with respect to a given predicate such that a
/// new array containing all the elements for which the predicate
/// holds is allocated.
/// 
/// O(N)
def filtered[A](arr: Array[A]) { shouldKeep: A => Bool }: Array[A] = {
  val nMatches = arr.count { shouldKeep }
  val res = allocate(nMatches)
  var idx = 0
  arr.foreach { x =>
    if (shouldKeep(x)) {
      res.unsafeSet(idx, x)
      idx = idx + 1
    }
  }
  res
}
```

b-studios: I agree that shrinking or dynamically resizing is better than calling it twice.

dvdvgt:
# Benchmark Results

*Methodology*:

Running the following twice:

```bash
for file in "sieve" "queens"                                                                                                                                                                          
do 
  for backend in "js" "llvm"
  do
    effekt --backend "$backend" "examples/benchmarks/are_we_fast_yet/$file.effekt"
    hyperfine --warmup 5 -m 10 "./out/$file"
  done
done
```

Once with

```scala
def fill[T](arr: Array[T], filler: T): Unit = {
  arr.foreachIndex { (idx, _) =>
    arr.unsafeSet(idx, filler)
  }
}
```

and once with

```
extern def fill[T](arr: Array[T], filler: T): Unit = 
  js "${arr}.fill(${filler})"
  chez "(vector-fill! ${arr} ${filler})"
  llvm """
    %z = call %Pos @c_array_fill(%Pos ${arr}, %Pos ${filler})
    ret %Pos %z
  """
  vm "array::fill[T](Array[T], T)"
```

The input for `sieve.effekt` was changed from `10000` to `1000000` to obtain conclusive results. The same for `queens.effekt`, were the input was changed from `8` to `24`.

# `extern`

## sieve
### js (`extern`)
```
Benchmark 1: ./out/sieve
  Time (mean ± σ):      80.0 ms ±   3.2 ms    [User: 95.8 ms, System: 17.2 ms]
  Range (min … max):    77.9 ms …  96.7 ms    36 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
```

### js

```
Benchmark 1: ./out/sieve
  Time (mean ± σ):      80.1 ms ±   1.1 ms    [User: 97.4 ms, System: 16.9 ms]
  Range (min … max):    78.7 ms …  83.6 ms    36 runs
```

### llvm (`extern`)

```
Benchmark 1: ./out/sieve
  Time (mean ± σ):      45.6 ms ±   2.1 ms    [User: 42.9 ms, System: 1.9 ms]
  Range (min … max):    42.9 ms …  53.4 ms    59 runs
```

## llvm

```
Benchmark 1: ./out/sieve
  Time (mean ± σ):      51.8 ms ±   2.0 ms    [User: 48.9 ms, System: 2.0 ms]
  Range (min … max):    49.0 ms …  57.2 ms    49 runs
```

## queens
### js (`extern`)

```
Benchmark 1: ./out/queens
  Time (mean ± σ):     126.5 ms ±   0.8 ms    [User: 136.9 ms, System: 8.7 ms]
  Range (min … max):   125.4 ms … 129.0 ms    23 runs
```

## js

```
Benchmark 1: ./out/queens
  Time (mean ± σ):     143.1 ms ±   3.6 ms    [User: 152.9 ms, System: 9.0 ms]
  Range (min … max):   139.8 ms … 157.1 ms    20 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
```

### llvm (`extern`)

```
Benchmark 1: ./out/queens
  Time (mean ± σ):     195.9 ms ±   1.6 ms    [User: 194.0 ms, System: 1.2 ms]
  Range (min … max):   191.7 ms … 198.8 ms    15 runs
```

## llvm

```
Benchmark 1: ./out/queens
  Time (mean ± σ):     144.4 ms ±   0.9 ms    [User: 142.4 ms, System: 1.2 ms]
  Range (min … max):   143.0 ms … 147.2 ms    19 runs
```

dvdvgt:
One more thing regarding the naming convention: If we choose `...!`-style for unsafe functions instead of 'in-place' functions, we may do so more consistently across different domains and not only for collections where the notion of 'in-place' is applicable.

PR #841, opened 2025-02-21
Use given pattern names when compiling patterns

### LINKED ISSUE #793: Use source field names to name new scrutinees
> [!WARNING]
> This is a super low-priority nice-to-have.

In every type declaration, we force the user to specify field names.
It would be therefore lovely to use them when coming up with names for new scrutinees -- currently we just use `y` everywhere,
but we could use, let's say, `head` and `tail` for a `Cons` when it's defined as `Cons(head: A, tail: List[A])`...

---

Here's where one could start:

1. this is where the `y` gets created:

https://github.com/effekt-lang/effekt/blob/e4be2f9c8e85f7f8a5ad73464701e07b2f202539/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala#L181-L183

2. this is where we create a `Pattern.Tag` and where we have some type information:

https://github.com/effekt-lang/effekt/blob/e4be2f9c8e85f7f8a5ad73464701e07b2f202539/effekt/shared/src/main/scala/effekt/core/Transformer.scala#L690-L691

jiribenes:
~~Addresses one part of #793:~~
Resolves #793 completely.

> I also think we could sometimes use the original pattern names, but I'm not 100% certain of that, would need to investigate.

I'm doing this since it helps me to debug problems just a little bit more easily :)

Note that this does not actually use the names of the fields as declared.

---

```scala
def isShorterThan(x: List[Int], y: List[Int]): Bool =
  (x, y) match {
    case (_head, Nil()) => false
    case (Nil(), _tail) => true
    case (Cons(_x, xs),Cons(_y, ys)) => isShorterThan(xs, ys)
  }
```

Here's a comparison before (red) vs after (green).
Just to drive my point home further, I removed the numbers after the identifier :)

Screenshot 2025-02-21 at 22 37 44

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
+++ b/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
@@ -4,7 +4,6 @@ package core
 import effekt.context.Context
 import effekt.core.substitutions.Substitution
 import effekt.symbols.TmpValue
-import effekt.util.messages.INTERNAL_ERROR
 
 import scala.collection.mutable
 
--- a/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
+++ b/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
@@ -192,6 +192,7 @@ object PatternMatchingCompiler {
 
       normalized.foreach {
         case Clause(Split(Pattern.Tag(constructor, patternsAndTypes), restPatterns, restConds), label, args) =>
+          // NOTE: Ideally, we would use a `DeclarationContext` here, but we cannot: we're currently in the Source->Core transformer, so we do not have all of the details yet.
           val fieldNames: List[String] = constructor match {
             case c: symbols.Constructor => c.fields.map(_.name.name)
             case _ => List.fill(patternsAndTypes.size) { "y" } // NOTE: Only reached in PatternMatchingTests
--- a/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
+++ b/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
@@ -179,12 +179,20 @@ object PatternMatchingCompiler {
 
       // used to make up new scrutinees
       val varsFor = mutable.Map.empty[Id, List[ValueVar]]
-      def fieldVarsFor(constructor: Id, fieldTypes: List[ValueType]): List[ValueVar] =
-        varsFor.getOrElseUpdate(constructor, fieldTypes.map { tpe => ValueVar(TmpValue("y"), tpe) })
+      def fieldVarsFor(constructor: Id, fieldPatternsAndTypes: List[(Pattern, ValueType)]): List[ValueVar] =
+        varsFor.getOrElseUpdate(
+          constructor,
+          fieldPatternsAndTypes.map { case (pat, tpe) =>
+            val idHint = pat match
+              case Pattern.Any(id) => id.name.name // if we have a pattern name, use it as a hint!
+              case _ => "y"
+            ValueVar(TmpValue(idHint), tpe)
+          }
+        )
 
       normalized.foreach {
         case Clause(Split(Pattern.Tag(constructor, patternsAndTypes), restPatterns, restConds), label, args) =>
-          val fieldVars = fieldVarsFor(constructor, patternsAndTypes.map { case (pat, tpe) => tpe })
+          val fieldVars = fieldVarsFor(constructor, patternsAndTypes)
           val nestedMatches = fieldVars.zip(patternsAndTypes.map { case (pat, tpe) => pat }).toMap
           addClause(constructor,
             // it is important to add nested matches first, since they might include substitutions for the rest.
--- a/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
+++ b/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
@@ -184,9 +184,9 @@ object PatternMatchingCompiler {
           constructor,
           fieldInfo.map {
             // if it's a pattern named by the user in the program, use the supplied name
-            case ((Pattern.Any(id), tpe),         _) => ValueVar(TmpValue(id.name.name), tpe)
+            case ((Pattern.Any(id), tpe),         _) => ValueVar(Id(id.name.name), tpe)
             // otherwise, use the field name of the given field
-            case ((_,               tpe), fieldName) => ValueVar(TmpValue(fieldName),    tpe)
+            case ((_,               tpe), fieldName) => ValueVar(Id(fieldName),    tpe)
           }
         )
 
--- a/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
+++ b/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
@@ -184,14 +184,16 @@ object PatternMatchingCompiler {
         varsFor.getOrElseUpdate(
           constructor,
           fieldInfo.map {
+            // if it's a pattern named by the user in the program, use the supplied name
             case ((Pattern.Any(id), tpe),         _) => ValueVar(TmpValue(id.name.name), tpe)
+            // otherwise, use the field name of the given field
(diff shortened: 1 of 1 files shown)


jiribenes:
Just checked, it works correctly when a discard/wildcard pattern (`_`) is used. In that case, we use `v_y_192038120398` as usual.

effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
```diff
@@ -179,12 +179,20 @@ object PatternMatchingCompiler {
...
-      def fieldVarsFor(constructor: Id, fieldTypes: List[ValueType]): List[ValueVar] =
-        varsFor.getOrElseUpdate(constructor, fieldTypes.map { tpe => ValueVar(TmpValue("y"), tpe) })
+      def fieldVarsFor(constructor: Id, fieldPatternsAndTypes: List[(Pattern, ValueType)]): List[ValueVar] =
+        varsFor.getOrElseUpdate(
+          constructor,
+          fieldPatternsAndTypes.map { case (pat, tpe) =>
+            val idHint = pat match
+              case Pattern.Any(id) => id.name.name // if we have a pattern name, use it as a hint!
+              case _ => "y"
+            ValueVar(TmpValue(idHint), tpe)
```
jiribenes, line 189: Here I could also use the previous `id` directly if we have one, i.e.:
 ```suggestion
          fieldPatternsAndTypes.map { case (pat, tpe) =>
            val patId = pat match
              case Pattern.Any(id) => id // if we have a pattern id, use it!
              case _ => TmpValue("y")
            ValueVar(patId, tpe)
```

The downside is that then variables are not globally unique, only locally. :(

Is there some other option?

**Comment by jiribenes** [comment]
After the latest commit, `v_y` is no more -- we always have a proper name: either by the pattern being named (`xs` and `ys` below) or by the field name!
(if it works, that is)
```scala
  val v_r_27913746: Bool396 = v_r_27743741 match {
    case Tuple2337 { (v_first_27792782: List921[Int405], v_second_27802781: List921[Int405]) => 
      v_second_27802781 match {
        case Nil1158 { () => 
          b_k_27753742()
        }
        case Cons1159 { (v__y_27832786: Int405, v_ys_27842785: List921[Int405]) => 
          v_first_27792782 match {
            case Nil1158 { () => 
              b_k_27763743()
            }
            case Cons1159 { (v__x_27872790: Int405, v_xs_27882789: List921[Int405]) => 
              b_k_27783744(v__x_27872790, v_xs_27882789, v__y_27832786, v_ys_27842785)
            }
          }
        }
      } else {
        v_first_27792782 match {
          case Nil1158 { () => 
            b_k_27763743()
          }
        }}
    }
  };
```

effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
```diff
@@ -4,6 +4,7 @@ package core
 import effekt.context.Context
 import effekt.core.substitutions.Substitution
 import effekt.symbols.TmpValue
+import effekt.util.messages.INTERNAL_ERROR
```
jiribenes, line 7: Stale import
```suggestion
```

jiribenes:
Fascinating. This used to work before I removed a spurious import, but doesn't anymore -- now it stack overflows the Machine transformer -- why? cc @phischu

effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
```diff
@@ -179,12 +179,24 @@ object PatternMatchingCompiler {
...
-      def fieldVarsFor(constructor: Id, fieldTypes: List[ValueType]): List[ValueVar] =
-        varsFor.getOrElseUpdate(constructor, fieldTypes.map { tpe => ValueVar(TmpValue("y"), tpe) })
+      def fieldVarsFor(constructor: Id, fieldInfo: List[((Pattern, ValueType), String)]): List[ValueVar] =
+        varsFor.getOrElseUpdate(
+          constructor,
+          fieldInfo.map {
+            // if it's a pattern named by the user in the program, use the supplied name
+            case ((Pattern.Any(id), tpe),         _) => ValueVar(TmpValue(id.name.name), tpe)
+            // otherwise, use the field name of the given field
+            case ((_,               tpe), fieldName) => ValueVar(TmpValue(fieldName),    tpe)
```
jiribenes, line 189:
```suggestion
            // if it's a pattern named by the user in the program, use the supplied name
            case ((Pattern.Any(id), tpe),         _) => ValueVar(Id(id.name.name), tpe)
            // otherwise, use the field name of the given field
            case ((_,               tpe), fieldName) => ValueVar(Id(fieldName),    tpe)
```

effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
```diff
@@ -179,12 +179,24 @@ object PatternMatchingCompiler {
...
+            case ((Pattern.Any(id), tpe),         _) => ValueVar(Id(id.name.name), tpe)
+            // otherwise, use the field name of the given field
+            case ((_,               tpe), fieldName) => ValueVar(Id(fieldName),    tpe)
+          }
+        )
 
       normalized.foreach {
         case Clause(Split(Pattern.Tag(constructor, patternsAndTypes), restPatterns, restConds), label, args) =>
-          val fieldVars = fieldVarsFor(constructor, patternsAndTypes.map { case (pat, tpe) => tpe })
+          val fieldNames: List[String] = constructor match {
```
jiribenes, line 195: @b-studios ?
```suggestion
          // NOTE: Ideally, we would use a `DeclarationContext` here, but we cannot: we're currently in the Source->Core transformer, so we do not have all of the details yet.
          val fieldNames: List[String] = constructor match {
```

jiribenes:
Currently,
```scala
def isShorterThan(x: List[Int], y: List[Int]): Bool =
  (x, y) match {
    case (_, Nil()) => false
    case (Nil(), _) => true
    case (Cons(_, xs),Cons(_, ys)) => isShorterThan(xs, ys)
  }
```

gets translated to [w/o opts]:
```scala
def isShorterThan2748(x2746: List921[Int405], y2747: List921[Int405]) = {
  let v_r_34564312 = make Tuple2252[List921[Int405], List921[Int405]] Tuple2337(x2746, y2747)
  def b_k_34574313() = return false
  def b_k_34584314() = return true
  def b_k_34604315(xs2750: List921[Int405], ys2751: List921[Int405]) = {
    val v_r_34594316: Bool396 = isShorterThan2748(xs2750, ys2751);
    return v_r_34594316
  }
  val v_r_34674317: Bool396 = v_r_34564312 match {
    case Tuple2337 { (first3462: List921[Int405], second3461: List921[Int405]) => 
      second3461 match {
        case Nil1158 { () => 
          b_k_34574313()
        }
        case Cons1159 { (head3464: Int405, ys3463: List921[Int405]) => 
          first3462 match {
            case Nil1158 { () => 
              b_k_34584314()
            }
            case Cons1159 { (head3466: Int405, xs3465: List921[Int405]) => 
              b_k_34604315(xs3465, ys3463)
            }
          }
        }
      } else {
        first3462 match {
          case Nil1158 { () => 
            b_k_34584314()
          }
        }}
    }
  };
  return v_r_34674317
};
```

and produces the following JS [w/ opts]
```js
function isShorterThan_0(x_0, y_0, ks_9, k_2) {
  isShorterThan_2: while (true) {
    switch (y_0.__tag) {
      case 0:  return () => k_2(false, ks_9);
      case 1: 
        const ys_1 = y_0.tail_0;
        switch (x_0.__tag) {
          case 0:  return () => k_2(true, ks_9);
          case 1: 
            const xs_1 = x_0.tail_0;
            /* prepare call */
            x_0 = xs_1;
            y_0 = ys_1;
            continue isShorterThan_2;
        }
        break;
      default:  return () => k_2(true, ks_9);
    }
  }
}
```

effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
b-studios: The suggested comment looks good to me

jiribenes:
Blocked on #847 (again).

Issue #842, opened 2025-02-21
JS CPS transformer crash: no statements in binding of a match


jiribenes:
_Context: I was testing #841 with fun examples on my drive :)_

The following program (the repro from #770!) _when compiled with `--no-optimize`_:
```scala
type Answer { Yes(); No() }

def println(ans: Answer): Unit = ans match {
  case Yes() => println("y")
  case No()  => println("n")
}

def join(left: Answer, right: Answer): Answer = (left, right) match {
  case (Yes(), Yes()) => Yes()
  case (No() ,     _) => No()
  case (_    , No() ) => No()
}

def main() = println(join(Yes(), No()))
```

returns an error in `generator.js.TransformerCps`:
```
Exception in thread "main" java.util.NoSuchElementException: last of empty list
        at scala.collection.immutable.Nil$.last(List.scala:665)
        at scala.collection.immutable.Nil$.last(List.scala:661)
        at effekt.generator.js.TransformerCps$.$anonfun$7(TransformerCps.scala:253)
        at scala.collection.immutable.List.map(List.scala:250)
        at effekt.generator.js.TransformerCps$.toJS(TransformerCps.scala:257)
        at effekt.generator.js.TransformerCps$.toJS$$anonfun$17(TransformerCps.scala:226)
        at effekt.generator.js.Tree$package$.stmts(Tree.scala:253)
        at effekt.generator.js.TransformerCps$.toJS$$anonfun$17(TransformerCps.scala:226)
        at effekt.generator.js.Tree$package$.stmts(Tree.scala:253)
        at effekt.generator.js.TransformerCps$.toJS$$anonfun$17(TransformerCps.scala:226)
        at effekt.generator.js.Tree$package$.stmts(Tree.scala:253)
        at effekt.generator.js.TransformerCps$.toJS$$anonfun$17(TransformerCps.scala:226)
        at effekt.generator.js.Tree$package$.stmts(Tree.scala:253)
```

The error comes from the `stmts.last` here:
https://github.com/effekt-lang/effekt/blob/5599ca1101e0d52538ba84adaa9e1bfdc1c716be/effekt/shared/src/main/scala/effekt/generator/js/TransformerCps.scala#L243-L258

b-studios:
Closed by #871

PR #843, opened 2025-02-22
Direct-style continuations on unoptimized core

b-studios:
When not opimizing core, continuations are not inlined, complicating the direct-style analysis.

For the example in #842, this currently yields:

```
repro.js:365
  let v_r_0 = undefined;
      ^

SyntaxError: Identifier 'v_r_0' has already been declared
```

since the generated code looks like this:

```js
function println_0(ans_0, ks_0, k_0) {
  let v_r_0 = undefined;
  let v_r_0 = undefined;
  switch (ans_0.__tag) {
    case 0:  return () => b_k_0(ks_0);
    case 1:  break;
  }
  const v_r_1 = $effekt.println("n");
  v_r_0 = v_r_1;
  return () => k_0(v_r_0, ks_0);
  const v_r_2 = $effekt.println("y");
  v_r_0 = v_r_2;
  return () => k_0(v_r_0, ks_0);
}
```

Disabling the direct style optimization by commenting out this

https://github.com/effekt-lang/effekt/blob/5599ca1101e0d52538ba84adaa9e1bfdc1c716be/effekt/shared/src/main/scala/effekt/generator/js/TransformerCps.scala#L223-L228

resolves the issue, but is slow.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/generator/js/TransformerCps.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/js/TransformerCps.scala
@@ -232,9 +232,11 @@ object TransformerCps extends Transformer {
         js.Const(nameDef(id), toJS(binding)(using nonrecursive(ks2))) :: requiringThunk { toJS(body) }.run(k)
       }
 
+    // no clauses
     case cps.Stmt.Match(sc, Nil, None) =>
       pure(js.Return($effekt.call("emptyMatch")) :: Nil)
 
+    // only one match clause
     case cps.Stmt.Match(sc, List((tag, clause)), None) =>
       val scrutinee = toJS(sc)
       val (_, stmts) = toJS(scrutinee, tag, clause)
@@ -250,8 +252,8 @@ object TransformerCps extends Transformer {
 
           val stmts = binding.stmts
 
-          stmts.last match {
-            case terminator : (js.Stmt.Return | js.Stmt.Break | js.Stmt.Continue) => (e, stmts)
+          stmts.lastOption match {
+            case Some(terminator : (js.Stmt.Return | js.Stmt.Break | js.Stmt.Continue)) => (e, stmts)
             case other => (e, stmts :+ js.Break())
           }
         },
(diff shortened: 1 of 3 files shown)


b-studios:
This is the CPS IR for println

```
let k3755 = { (v_r_27793745) | ks3756 =>
  jump k3754(v_r_27793745) @ ks3756
}
let b_k_27763741 = { () | ks3757 =>
  let v_r_27753743 = println1("y")
  jump k3755(v_r_27753743) @ ks3757
}
let k3753 = { (v_r_27793745) | ks3758 =>
  jump k3754(v_r_27793745) @ ks3758
}
let b_k_27783742 = { () | ks3759 =>
  let v_r_27773744 = println1("n")
  jump k3753(v_r_27773744) @ ks3759
}
ans2747 match {
  case Yes2753 () => {
    jump b_k_27763741() @ ks3760
  }
  case No2754 () => {
    jump b_k_27783742() @ ks3760
  }
}
```

which it believes is direct style in `k3755`.

b-studios:
It also suffices to make the rhs of this

https://github.com/effekt-lang/effekt/blob/5599ca1101e0d52538ba84adaa9e1bfdc1c716be/effekt/shared/src/main/scala/effekt/generator/js/TransformerCps.scala#L514

to always return `false`, which effectively requires all branches to return to the same continuation. This again might be too conservative since under certain conditions we can convert a program to direct style (for the fast path), but still keep non-local jumps to other continuations.

effekt/shared/src/main/scala/effekt/generator/js/TransformerCps.scala
```diff
@@ -250,8 +252,8 @@ object TransformerCps extends Transformer {
 
           val stmts = binding.stmts
 
-          stmts.last match {
-            case terminator : (js.Stmt.Return | js.Stmt.Break | js.Stmt.Continue) => (e, stmts)
+          stmts.lastOption match {
+            case Some(terminator : (js.Stmt.Return | js.Stmt.Break | js.Stmt.Continue)) => (e, stmts)
```
b-studios, line 256: Oh this already fixed #842...

PR #844, opened 2025-02-24
Keep stack pointer and limit in registers

serkm:
This avoids unnecessary loads/stores.
Makes `fibonacci_recursive` faster and `resume_nontail` slower.


serkm:
I am now convinced that LLVM's `musttail` has bugs that make this fail, possibly due to larger struct arguments.
Maybe revisit with newer LLVM versions.

PR #848, opened 2025-02-26
Name join points introduced when compiling clauses

jiribenes:
Quick drive-by from #841 that deserves to be considered separately.

Instead of naming each join point created for the RHS of a clause as `b_k_123`, we can give a slightly more specific name based on the origin of the clause.

Feel free to bikeshed the names chosen or close this PR altogether if it's useless.
FWIW, it helps me ever-so-slightly in the `b_k${index of the case in the source}_123` case when debugging.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/core/Transformer.scala
@@ -386,8 +386,8 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
       Context.bind(If(c, transform(thn), transform(els)))
 
     case source.If(guards, thn, els) =>
-      val thnClause = preprocess(Nil, guards, transform(thn))
-      val elsClause = preprocess(Nil, Nil, transform(els))
+      val thnClause = preprocess("thn", Nil, guards, transform(thn))
+      val elsClause = preprocess("els", Nil, Nil, transform(els))
       Context.bind(PatternMatchingCompiler.compile(List(thnClause, elsClause)))
 
     //    case i @ source.If(guards, thn, els) =>
@@ -416,8 +416,8 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
           insertBindings { core.If(transformAsPure(cond), thenBranch, elseBranch) }
         case _ =>
           insertBindings {
-            val thenClause = preprocess(Nil, guards, thenBranch)
-            val elseClause = preprocess(Nil, Nil, elseBranch)
+            val thenClause = preprocess("guard_thn", Nil, guards, thenBranch)
+            val elseClause = preprocess("guard_els", Nil, Nil, elseBranch)
             PatternMatchingCompiler.compile(List(thenClause, elseClause))
           }
       }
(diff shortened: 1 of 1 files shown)


jiribenes:
```scala
def isShorterThan(x: List[Int], y: List[Int]): Bool =
  (x, y) match {
    case (_, Nil()) => false // case 1
    case (Nil(), _) => true // case 2
    case (Cons(_, xs),Cons(_, ys)) => isShorterThan(xs, ys) // case 3
  }
```

~>

```scala
def isShorterThan2748(x2746: List921[Int405], y2747: List921[Int405]) = {
  let v_r_34564432 = make Tuple2252[List921[Int405], List921[Int405]] Tuple2337(x2746, y2747)
  def b_k0_34574433() = return false
  def b_k1_34584434() = return true
  def b_k2_34604435(xs2750: List921[Int405], ys2751: List921[Int405]) = {
    val v_r_34594436: Bool396 = isShorterThan2748(xs2750, ys2751);
    return v_r_34594436
  }
  val v_r_34734437: Bool396 = v_r_34564432 match {
    case Tuple2337 { (v_y_34613464: List921[Int405], v_y_34623463: List921[Int405]) => 
      v_y_34623463 match {
        case Nil1158 { () => 
          b_k0_34574433()
        }
        case Cons1159 { (v_y_34653468: Int405, v_y_34663467: List921[Int405]) => 
          v_y_34613464 match {
            case Nil1158 { () => 
              b_k1_34584434()
            }
            case Cons1159 { (v_y_34693472: Int405, v_y_34703471: List921[Int405]) => 
              b_k2_34604435(v_y_34703471, v_y_34663467)
            }
          }
        }
      }
    }
  };
  return v_r_34734437
};
```

---

```scala
def main() = {
  var curr = Some(10)
  while (curr is Some(i)) {
    println(i);
    if (i > 0)
      curr = Some(i - 1)
    else
      curr = None()
  }
}
```

~>
```scala
def main2746() = {
  val v_r_27623720: Option754[Int405] = return make Option754[Int405] Some809(10);
  var curr2747 = v_r_27623720;
  def b_whileLoop_27633721() = {
    def b_guard_thn_27703722(i2748: Int405) = {
      val v_whileThen_27693732: Unit401 = {
        val v_r_27653731: Unit401 = println4(i2748);
        let _3727 = v_r_27653731
        val v_r_27683728: Unit401 = if (infixGt184(i2748, 0)) {
          val v_r_27663733: Unit401 = {
            curr2747 := make Option754[Int405] Some809(infixSub105(i2748, 1));
            return ()
          };
          return ()
        } else {
          val v_r_27673734: Unit401 = {
            curr2747 := make Option754[Int405] None808();
            return ()
          };
          return ()
        };
        return v_r_27683728
      };
      b_whileLoop_27633721()
    }
    def b_guard_els_27723723() = return ()
    val v_r_27713725: Option754[Int405] = {
      let s3724 = !curr2747;
      return s3724
    };
    v_r_27713725 match {
      case Some809 { (v_y_27733729: Int405) => 
        b_guard_thn_27703722(v_y_27733729)
      }
    } else {
      b_guard_els_27723723()}
  }
  val v_r_27743730: Unit401 = b_whileLoop_27633721();
  return v_r_27743730
}
```

effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -416,8 +416,8 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
           insertBindings { core.If(transformAsPure(cond), thenBranch, elseBranch) }
         case _ =>
           insertBindings {
-            val thenClause = preprocess(Nil, guards, thenBranch)
-            val elseClause = preprocess(Nil, Nil, elseBranch)
+            val thenClause = preprocess("guard_thn", Nil, guards, thenBranch)
+            val elseClause = preprocess("guard_els", Nil, Nil, elseBranch)
```
jiribenes, line 420: maybe to be consistent with `whileThen`, do this instead?
```suggestion
            val thenClause = preprocess("guardThn", Nil, guards, thenBranch)
            val elseClause = preprocess("guardEls", Nil, Nil, elseBranch)
```

PR #851, opened 2025-03-04
Run tests without optimization

### LINKED ISSUE #846: Run some tests without optimisations
Related to https://github.com/effekt-lang/effekt/pull/843

We should run [at least some of the] tests on LLVM and JS _without optimizations_ (in addition to running them _with_ optimisation as we do already).

Alternatively, we could instead run all of the tests without opts (again, in addition) on `master` and just note the regressions later.

marvinborner:
Closes #846 

This runs every test without optimization *in addition* to running them with optimization. While this is good for testing, we should change this once it's mergable so our CI doesn't take ~~twice as long~~ much longer.

There are many new errors now. I've spent some time investigating and they generally fall into these two categories:

- no block info (unsoundness in the optimizer)
- unsupported LLVM feature (toplevel object definitions, reached hole (no LLVM FFI for extern definition)), which were somehow optimized away
- valgrind error

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/EffektTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/EffektTests.scala
@@ -36,10 +36,12 @@ trait EffektTests extends munit.FunSuite {
 
   def negatives: List[File] = List()
 
+  // Test files that should be run with optimizations disabled
+  def withoutOptimizations: List[File] = List()
+
   def runTestFor(input: File, expected: String): Unit =
     test(input.getPath + s" (${backendName})") {
       assertNoDiff(run(input, true), expected)
-      assertNoDiff(run(input, false), expected)
     }
 
   // one shared driver for all tests in this test runner
@@ -98,6 +100,16 @@ trait EffektTests extends munit.FunSuite {
       case Right(value) =>
         negatives.foreach(runNegativeTestsIn)
         positives.foreach(runPositiveTestsIn)
+        withoutOptimizations.foreach(runWithoutOptimizations)
+    }
+
+  def runWithoutOptimizations(dir: File): Unit =
+    foreachFileIn(dir) {
+      case (f, None) => sys error s"Missing checkfile for ${f.getPath}"
+      case (f, Some(expected)) =>
+        test(s"${f.getPath} (${backendName})") {
+          assertNoDiff(run(f, false), expected)
+        }
     }
 
   def runPositiveTestsIn(dir: File): Unit =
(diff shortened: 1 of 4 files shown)


marvinborner:
Some notes I took while summarizing/minimizing the failing tests:

# Regions

from permute.effekt: heap-use-after-free

``` scala
def run() = {
  var count in global = 0;
  42
}

def main() = Nil[Int]() match {
  case Nil() => println(run())
  case Cons(_, _) => println(42)
}
```

from generator.effekt: "No block info for r..."

``` scala
interface Foo {
  def v(): Int
}

def generator {r: Region}: Foo at {r} = {
  var v: Int in r = 42;

  new Foo {
    def v() = v
  }
}

def run(): Int = region r {
  def gen = generator {r}
  42
}

def main() = println(run())
```

# Multiple definitions

(only in unoptimized JS)

From `ascii_isalphanumeric`/`ascii_iswhitespace`/`json`/`fibonacci`/`characters` (Char):

``` scala
def foo(): Option[Int] =
  if (true) {
    Some(42)
  } else {
    None()
  }

def main() = println(foo().isDefined)
```

parser.effekt

``` scala
effect fail(msg: String): Nothing
effect error(msg: String): Nothing

type ParseResult[R] {
  Success(t: R);
  Failure(msg: String);
  ParseError(msg: String)
}

def eager[R] { p: => R / { fail, error } }: ParseResult[R] = try {
  Success(p())
} with fail { (msg) =>
  Failure(msg)
} with error { (msg) =>
  ParseError(msg)
}

def parse[R]() =
  eager[R] {
    do fail("")
  }

def main() =
  parse[Int]() match {
    case Success(n) => println(n)
    case Failure(msg) => ()
    case ParseError(msg) => ()
  }
```

sortBy, flatmap: `b_k_...` not defined

``` scala
def main() = {
  val list = [1, 2, 3, 4]
  println(list)
}
```

# V8 OOM

from `list/deleteat.effekt`

``` scala
import list

def main() = {
  val list = [1, 2, 3, 4]
  val left = list.slice(0, 1)
  println(left)
}
```

# Holes

no idea how this works in LLVM with optimizations: `bidirectional/typeparametric.effekt`

``` scala
...

extern pure def cap[U, V](): Cap[U, V] at {} =
  js   "42"
  chez "42"

...
```

# Top-level object definitions

- `pos/object/if_control_effect.effekt`
- `pos/object/higher_order_compose_op.effekt`
- `pos/lambdas/toplevel_objects.effekt`
- `pos/type_omission_op.effekt`
- `pos/bidirectional/higherorderobject.effekt`
- `pos/bidirectional/res_obj_boxed.effekt`
- `pos/bidirectional/effectfulobject.effekt`

# Other

- `pos/issue684`, `stdlib/stream/zip.effekt`, `stdlib/json.effekt` not supported in Chez (resume)

jiribenes (approved):

PR #852, opened 2025-03-04
Add function to abort phase non-fatally

### LINKED ISSUE #850: Compiling on js-web backend ignores '--ir-write-all'
Whenever I try to compile with `--ir-write-all` _and_ `--backend=js-web`, the IR files don't show up in the `out/` folder :(

marvinborner:
Closes #850, related to #747.

We can now abort a phase without printing an error while also not printing an error message. This is useful for notifying the user of opening a file in another program (`js-web`, browser) while producing no actual executable file.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Runner.scala
+++ b/effekt/jvm/src/main/scala/effekt/Runner.scala
@@ -92,7 +92,7 @@ trait Runner[Executable] {
   /**
    * Runs the executable (e.g. the main file) by calling the build function.
    */
-  def eval(executable: Executable)(using C: Context): Unit = build(executable).map { execFile =>
+  def eval(executable: Executable)(using C: Context): Unit = build(executable).foreach { execFile =>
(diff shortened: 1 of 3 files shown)


effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -87,13 +87,12 @@ trait Runner[Executable] {
...
    */
-  def build(executable: Executable)(using Context): String
+  def build(executable: Executable)(using Context): Option[String]
 
   /**
    * Runs the executable (e.g. the main file) by calling the build function.
    */
-  def eval(executable: Executable)(using C: Context): Unit = {
-    val execFile = build(executable)
+  def eval(executable: Executable)(using C: Context): Unit = build(executable).map { execFile =>
```
jiribenes, line 95: Nitpick: you can also use 
```suggestion
  def eval(executable: Executable)(using C: Context): Unit = build(executable).foreach { execFile =>
```
and then avoid the `.getOrElse(())`

(We use it elsewhere in the codebase too, I think?)

b-studios (approved):

PR #857, opened 2025-03-05
Make generated shebang in jar file platform-dependent

### LINKED ISSUE #16: Binary on Ubuntu 16.04

timsueberkrueb:
Fixes #16 - hopefully at least!

I verified this works on Ubuntu 24.04 and Windows 11. I did not verify it works on WSL or macOS, but I would assume so.


jiribenes (approved):
LGTM, but I'd prefer if someone with a Windows computer w/ WSL (cc @mattisboeckle?) could try it out.

Also, does this mean that we could finally retire the `effekt.sh` file? 🙏

OriginalCpp:
This also works on WSL :)

Issue #849, opened 2025-02-27
Wrong order of passed block arguments when unused function is defined


jiribenes:
The following (already heavily minimised) code produces different order of `Components` capabilities being passed to `addSystem` based on if the _definition_ of `addTransform` is commented in or out.

If I had to guess, this has something to do with explicit capability passing...

### Repro

```
record Transform()
record Prefab()

interface Component[T] { }
interface Components[T] { }

def component[T] { prog: => Unit / { Component[T], Components[T] } }: Unit / {} = {
  try {
    prog()
  } with Component[T] {
  } with Components[T] {
  }
}

// !!!
// Comment me out to see the change!
def addTransform { prog: => Unit / { Component[Transform], Components[Transform] } }: Unit / {} = {
  with component[Transform];
  prog()
}

def world { prog: => Unit / { Component[Prefab], Component[Transform], Components[Unit], Components[Prefab], Components[Transform] } }: Unit / {} = {
  try {
    prog()
  } with Component[Prefab] {
  } with Component[Transform] {
  } with Components[Unit] {
  } with Components[Prefab] {
  } with Components[Transform] {
  }
}

def addSystem[T, With, Without]()
{ body: () => Unit }
{ prog: => Unit }: Unit / {
  Components[T], Components[With], Components[Without]
} = {
  prog()
  body()
}


def main() = {
  with world;

  with addSystem[Transform, Unit, Prefab] { () };

  ()
}
```

### IR (compiled w/o opts)

When `addTransform` is defined (= not commented out), `main` looks like:
```scala
def main2769() = {
  val v_r_28633821: Unit401 = world2762{ {Component3834}{Component3835}{Components3818}{Components3817}{Components3819} => 
    val v_r_28623820: Unit401 = addSystem2768[Transform2750, Unit401, Prefab2751]{ () => 
      return ()
    }{ () => 
      return ()
    }{Components3817}{Components3818}{Components3819};
    return v_r_28623820
  };
  return v_r_28633821
}
```

When `addTransform` is not defined (= commented out), `main` looks like:
```scala
def main2767() = {
  val v_r_28473805: Unit401 = world2760{ {Component3818}{Component3819}{Components3801}{Components3802}{Components3803} => 
    val v_r_28463804: Unit401 = addSystem2766[Transform2750, Unit401, Prefab2751]{ () => 
      return ()
    }{ () => 
      return ()
    }{Components3801}{Components3802}{Components3803};
    return v_r_28463804
  };
[...]

PR #859, opened 2025-03-05
Fix canonical ordering by using the ordering at the declaration site

b-studios:
This tries to fix issue #849 by NOT canonicalizing AFTER unification.

Instead, we canonicalize FIRST on the function type and then MAINTAIN the order at the callsite.

This requires us to NOT go into type `Effect`, which is a set and thus removes duplicates and destroys the ordering...

Sadly, it is very difficult to write a test for this :(


jiribenes:
I can confirm that this fixes the problem in all of the repros I had. :)

PR #860, opened 2025-03-06
Add a regression test on capability ordering

jiribenes:
Discovered by @timsueberkrueb: on ffde2091956d25dd3522f316ecb1dfc19a33a93e (latest commit made on March 4), this prints:
```
what
```
but on `master` (after #859), it prints the correct solution:
```
Cons(1, Cons(2, Cons(3, Nil())))
Cons(2, Cons(3, Nil()))
Cons(1, Cons(3, Nil()))
Cons(3, Nil())
Cons(1, Cons(2, Nil()))
Cons(2, Nil())
Cons(1, Nil())
Nil()
```

Curiously, this seems to work fine on v0.21 and v0.22, so I'm not sure where the regression happened. 🤷‍♂️ 
Nevertheless, we really should have this as a regression test if it sometimes magically fails :)

### DIFF
--- a/examples/pos/cap_ordering.check
+++ b/examples/pos/cap_ordering.check
@@ -1,0 +1,8 @@
+Cons(1, Cons(2, Cons(3, Nil())))
+Cons(2, Cons(3, Nil()))
+Cons(1, Cons(3, Nil()))
+Cons(3, Nil())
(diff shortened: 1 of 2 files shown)


b-studios (approved):

b-studios:
Great thanks!

PR #864, opened 2025-03-06
More alias information for llvm

serkm:
This adds separate scopes for stacks, stack pointers, prompts, objects and vtables.


b-studios:
\cc @mattisboeckle

serkm:
This doesn't give any speedup on our benchmarks, but it is a more complete version of what we already have

phischu (approved):
Looks good

PR #867, opened 2025-03-07
Add 'stream::fix' fixpoint operator

jiribenes:
Added together with a test.

Not sure if the semantics are what we desire wrt stack, cc @b-studios.


examples/stdlib/stream/fix.effekt
```diff
@@ -0,0 +1,28 @@
...
+type Message {
+  Primitive(n: Int)
+  Complex(msgs: List[Message])
+}
+
+def solveOne(m: Message): Unit / { emit[Int], emit[Message] } = m match {
+  case Primitive(n) => do emit[Int](n)
+  case Complex(msgs) => for { msgs.each } { msg => do emit[Message](msg) }
+}
+
```
jiribenes, line 12: I'd appreciate a simpler example, but I really don't have any myself :(

libraries/common/stream.effekt
```diff
@@ -30,6 +30,17 @@ effect stop(): Nothing
...
+/// Runs the stream as a message queue,
+/// handling each of its elements that can produce new elements ("messages"),
+/// until a fixed point is reached.
+def fix[T] { one: T => Unit / emit[T] } { stream: => Unit / emit[T] }: Unit =
+  try {
+    stream()
+  } with emit[T] { value =>
+    fix[T] {one} { one(value) }
+    resume(())
+  }
```
jiribenes, line 42: Does this do the correct thing wrt stack? (i.e. not building up extra unnecessary frames?)

b-studios: We could use the VMTests to actually measure and test this :)

jiribenes: I added a VM test, but I cannot really tell from the output if it does the correct thing 🤔

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -803,6 +803,22 @@ class VMTests extends munit.FunSuite {
...
+      patternMatches = 37,
+      branches = 0,
+      pushedFrames = 34,
+      poppedFrames = 34,
+      allocations = 37,
+      closures = 37,
+      variableReads = 10,
+      variableWrites = 9,
+      resets = 0,
+      shifts = 0,
```
b-studios, line 819: At least it looks like there are no pushed handlers.

PR #872, opened 2025-03-08
Do not contify under reset

### LINKED ISSUE #861: Multiple resumption misoptimization
The following program discovered by @timsueberkrueb prints something different when ran with optimisations than without:
```scala
effect flip(): Bool
effect choice(): Bool

def main() = {
  try {
    val x = do choice()
    val y = do choice()
    println(x.show ++ " " ++ y.show)
  } with choice {
    try {
      resume(do flip())
    } with flip {
      resume(true)
      resume(false)
    }
  }
}
```

~>

```sh
$ effekt --backend=js --no-optimize hello.effekt
true true
true false
false true
false false

$ effekt --backend=js hello.effekt
false false
```

b-studios:
This PR fixes #861 by not contifying under reset. 

It also adds a pretty printer as a drive-by for easier debugging (the one from #843).


effekt/shared/src/main/scala/effekt/cps/Contify.scala
```diff
@@ -85,12 +85,13 @@ object Contify {
...
-          LetCont(k2, cont,
-            Stmt.LetCont(id, ContLam(vparams, ks, substitute(rewrittenBody)),
-              contify(id, rewrittenRest)))
+        // leads to `k_6 is not defined` when disabling optimizations on issue861.effekt
+        //        case Some(cont: ContLam) if returnsUnique && !isRecursive =>
+        //          val k2 = Id("k")
+        //          given Substitution = Substitution(conts = Map(k -> ContVar(k2)))
+        //          LetCont(k2, cont,
+        //            Stmt.LetCont(id, ContLam(vparams, ks, substitute(rewrittenBody)),
+        //              contify(id, rewrittenRest)))
```
b-studios, line 94: Sadly, we do not have a regression test since we need to disable optimizations.

Maybe a variant of #851 where we selectively run tests without optimizations would be useful (here JS backend so a lot of the reported segfaults etc. do not matter here.)

b-studios: In the present PR I comment this out, however this "might" have performance implications...

jiribenes: I feel like we're hitting the limits of the way we currently do example-based regression tests...
We have multiple axes now, but we really only cleanly account for two of them (1 & 3):

1. by backend:
    - JS
    - LLVM
    - ...
2. by opts:
    - with opts
    - no opts
3. stdlib vs compiler:
    - stdlib
    - compiler
4. by backend-specific details:
    - with valgrind
    - without valgrind

PR #874, opened 2025-03-10
Basic draft for `search` module

phischu:


jiribenes:
One nice way to test this library would be to add n-queens in this style
and then check if the perf isn't too bad compared to the one in `effect_handlers_bench` 👀 
(then it effectively tracks if the extra stream-related stuff isn't too bad for the optimiser)

Issue #744, opened 2024-12-12
Change Position to a cached byte offset, not an always-recomputed line & column pair


jiribenes:
We currently use `Position` which is a `(line, column)` pair.
This means we spend _a lot_ of time on `Source.offsetToPosition` when parsing.

Instead, we should just use byte `offset`s to mark positions, computing the line & column on demand (and caching it).

I tried this a long time ago (almost 6 months), but I'm attaching the code that somewhat worked back then, it just needs a clean up (refactoring :)) and some benchmarking (used to be something like 20-30% faster parsing!). It also needs some proper LSP testing: it worked on my machine for at least a few minutes, but that's probably not good enough -- I'm a bit scared that there could be bugs that make the LSP much slower to use.
Note that this change is across Effekt _and_ [our fork of Kiama](https://github.com/effekt-lang/kiama).

Here's the code I used for this change in the Effekt codebase:
```diff
diff --git i/effekt/js/src/main/scala/effekt/LanguageServer.scala w/effekt/js/src/main/scala/effekt/LanguageServer.scala
index ecbcbd73..9cc157fc 100644
--- i/effekt/js/src/main/scala/effekt/LanguageServer.scala
+++ w/effekt/js/src/main/scala/effekt/LanguageServer.scala
@@ -175,7 +175,7 @@ class LanguageServer extends Intelligence {
     new lsp.Position(p.line - 1, p.column - 1)
 
   private def fromLSPPosition(p: lsp.Position, source: Source): Position =
-    Position(p.line + 1, p.character + 1, source)
+    source.lineAndColumnToPosition(p.line, p.character)
 
   private def toLSPLocation(from: Position, to: Position): Option[lsp.Location] = from.source match {
     case VirtualFileSource(path) => Some(new lsp.Location(from.source.name, new lsp.Range(toLSPPosition(from), toLSPPosition(to))))
diff --git i/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala w/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
index 4c117084..46b052d0 100644
--- i/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
+++ w/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
@@ -39,7 +39,7 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
             val to = source.offsetToPosition(value.end + 1)
             Some(Range(from, to))
           case None =>
[...]

PR #875, opened 2025-03-10
Treat span() as an effect and nonterminal() as handler

b-studios:


jiribenes:
cc @timsueberkrueb

jiribenes:
If we do this, it might be worth bundling it with [#744](https://github.com/effekt-lang/effekt/issues/744)

b-studios:
Sorry if I wasn't clear, it's `empty`, but `Nil` in Scala

PR #877, opened 2025-03-11
Add function attributes to extern functions

serkm:
These function attributes are generated by
`clang -S -emit-llvm -O2 libraries/llvm/main.c`


jiribenes:
I don't think that `sspstrong` and `uwtable` are very helpful when looking at their docs 🤔

PR #879, opened 2025-03-11
Use trampolining Control[A] for bindings in machine transformer

### LINKED ISSUE #855: Change Machine.Bindings monad to alleviate OOM
Follow-up from #853

https://github.com/effekt-lang/effekt/blob/d57b530c328437c62b6c4a4b1e0a0651785aab3c/effekt/shared/src/main/scala/effekt/machine/Transformer.scala#L554-L567

change it into a Writer, similarly to what `Core.Bind` does.

marvinborner:
Closes #855.

This mostly works, but I have some trouble rewriting code like this:

``` scala
Binding { k =>
  LiteralDouble(literal_binding, v, k(literal_binding))
}
```

For now I did

``` scala
binding(literal_binding, l =>
  LiteralDouble(literal_binding, v, l)
)

// where binding is
def binding[A](c: A, run: (Statement => Statement)) = {
  Computation { k =>
    Impure(pure(c), ReturnCont(v => run(Result.trampoline(k(v)))))
  }
```

Which is of course wrong, but I find most definitions in `Control[A]` deeply confusing, so I'd welcome some help!

I also had to make some private functions in utils.control public, not sure if this is okay (we don't use the library anywhere else anyway)

### DIFF
--- a/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
@@ -8,7 +8,6 @@ import effekt.symbols.builtins.TState
 import effekt.util.messages.ErrorReporter
 import effekt.symbols.ErrorMessageInterpolator
 
-
 object Transformer {
 
   private def ErrorReporter(using E: ErrorReporter): ErrorReporter = E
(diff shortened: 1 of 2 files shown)


b-studios:
@phischu

marvinborner:
Superseded by #917

PR #880, opened 2025-03-11
Properly represent existentials in core

### LINKED ISSUE #863: Existential type variables should be substituted during optimization
When we write a program using existential types like the following:

```
type Exists {
  Op[Z](el: Z, show: Z => String at {})
}

def main() = {
  val x = Op[Int](123, box { x => x.show })
  x match {
    case Op(e, f) => println(f(e))
  }
}
```

The `Normalizer` helpfully optimizes it to the following in Core:

```
def main3412() = {
  let tmp4510 = println1(show14(123))
  return tmp4510
}
```

However, when [doing so](https://github.com/effekt-lang/effekt/blob/a1d13cd699f8429a67b4757e89fa3642d7311a27/effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala#L206), we forget to substitute type arguments of the constructor (it doesn't even have any in the current representation), for type parameters of the clause (it doesn't even abstract over any in the current representation).

This slightly blocks my work on #800

phischu:


b-studios:
Great thanks, since this fixed #863 could you add it as a test?

effekt/shared/src/main/scala/effekt/core/Type.scala
```diff
@@ -252,7 +252,7 @@ object Type {
     case Pure.ValueVar(id, tpe) => tpe
     case Pure.Literal(value, tpe) => tpe
     case Pure.PureApp(callee, targs, args) => instantiate(callee.functionType, targs, Nil).result
-    case Pure.Make(tpe, tag, args) => tpe
+    case Pure.Make(tpe, tag, targs, args) => tpe // TODO instantiate?
```
b-studios, line 255: Yes, this seems crucial.

b-studios: Don't you also need to do something on match?

b-studios:
What are the type arguments on constructors? Are those all type parameters, or only the existential ones?

phischu:
This is still not quite right. The resulting core after normalization does not type check, since there are free type variables left.

PR #883, opened 2025-03-13
Add a clear Main entrypoint for the JVM binary

timsueberkrueb:
As a newcomer to this project, I found it hard to find the entrypoint to the JVM binary.
This PR adds a file `Main.scala` under `jvm/src/main/scala/effekt` which should be easy to find and understand.
This could also be one of the first baby steps to disentangle some parts of this project from Kiama such as the LSP Server.
The logic *should* reproduce the current behavior as far as I understand it.


b-studios (commented):
I still need to test it, but from the code this looks great!

PR #884, opened 2025-03-13
Inline kiama

timsueberkrueb:
This PR inlines the kiama submodule into the main repo which makes it easier to modify it in-sync with the rest of the Effekt compiler.

This may be controversial. Happy to close this if you think this is too early.


timsueberkrueb:
This seems to be possible using `git filter-repo --to-subdirectory-filter`. Note that [the kiama fork](https://github.com/effekt-lang/kiama/) did not preserve the history of the [original project](https://github.com/inkytonik/kiama) so we would only get these commits which are probably the ones we care about anyway.
Let me try how that option would turn out.

kiama/.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
+# This workflow uses actions that are not certified by GitHub.
```
jiribenes, line 1: This file can now be removed, right?

.github/workflows/ci.yml
```diff
@@ -98,3 +98,22 @@ jobs:
 
     - name: Run effekt binary
       run: effekt.sh --help
+
+  build-and-test-kiama:
+    runs-on: ubuntu-latest
```
jiribenes, line 103: Nitpicks:
1. Do we even want a separate CI pass for this? (Compiling Effekt forces compiling Kiama anyway, no?)
2. If so, could we have a human-friendly name for the CI pass (so that it's nicer when reading it), something like:
```suggestion
  build-and-test-kiama:
    name: Build and test Kiama
    runs-on: ubuntu-latest
```

jiribenes (commented):
nitpicks

Issue #876, opened 2025-03-11
Inlay hints sometimes don't reappear after file edits


timsueberkrueb:
Consider the following program:

```scala
def main() = println("hello world")
```

After the LSP server loads initially, the following inlay hint gets shown for the inferred `io` capture:

```scala
{io} def main() = println("hello world")
```

Now enter a blank line above the `main` function. The hint disappears.

Relevant LSP trace:

```
[Trace - 11:03:56 AM] Received request 'textDocument/inlayHint - (4)'
[Trace - 11:03:56 AM] Sending response 'textDocument/inlayHint - (4)'. Processing request took 1ms
Result: [
  {
    "position": {
      "line": 0,
      "character": 0
    },
    "label": "{io}",
    "kind": 1,
    "tooltip": {
      "kind": "markdown",
      "value": "captures: `{io}`"
    },
    "paddingLeft": false,
    "paddingRight": true,
    "data": "capture"
  }
[Trace - 11:03:57 AM] Received request 'textDocument/inlayHint - (6)'
[Trace - 11:03:57 AM] Sending response 'textDocument/inlayHint - (6)'. Processing request took 0ms
Result: null
```

marvinborner:
Reminds me of #366 :/

timsueberkrueb:
After some surprisingly painful debugging, I got a step closer to finding the cause of this.

So first of all, you need to use https://github.com/effekt-lang/effekt-vscode/pull/16, otherwise there is a timing race condition where the inlay hints are requested before the document has been reprocessed.

 With reprocess-on-change and adding a `Thread.sleep` to `getInlayHints`, like so

```scala
  override def getInlayHints(range: kiama.util.Range): Option[Vector[InlayHint]] =
    Thread.sleep(10000)
    val captures = getInferredCaptures(range)(using context).map {
    ...
```

I can verify using breakpoints that it is indeed reprocessing the file first before continuing to compute the inlay hints.

Still, the same symptoms! How could this be?

We can verify that we indeed receive the updated source code on `didChange` (with the leading `\n`):

![Image](https://github.com/user-attachments/assets/8a028c83-206b-44d0-aa73-2ab054c9b49a)

However, and I was not aware of that, when transforming source code positions into offsets to compute the inlay hints for the requested position, we **get the source text by querying the kiama range** as you can witness here:

![Image](https://github.com/user-attachments/assets/0bb7428e-939c-47c5-8595-c9d4ffaa5efd)

[...]

Issue #549, opened 2024-08-21
Hovering over a typed hole doesn't show details


jiribenes:
We have specialised code for showing the type inside & outside a typed hole:
https://github.com/effekt-lang/effekt/blob/68d40378fea59cb66074de3eb61df7048578a4e1/effekt/shared/src/main/scala/effekt/Intelligence.scala#L89-L95
called via:
https://github.com/effekt-lang/effekt/blob/68d40378fea59cb66074de3eb61df7048578a4e1/effekt/jvm/src/main/scala/effekt/Server.scala#L86-L87

But I haven't been able to trigger it on the website or in VSCode for any kind of hole (neither `<>` nor `<{ ... }>`) even though I remember it working a long time ago.

b-studios:
I think this is fixed in the WIP branch https://github.com/effekt-lang/effekt/tree/feature/llm

jiribenes:
I tried this again, but I don't understand how the annotations work...

There is a traceable call to:
```
      Context.annotateInferredType(t, got)
      Context.annotateInferredEffects(t, effs.toEffects)
```
but then there's nothing if you try to retrieve it (both immediately after the assigning, and in the LSP).

timsueberkrueb:
Fixed.

Issue #366, opened 2024-01-20
Flaky LSP HoverRequest response


marvinborner:
While working on the LSP testing suite (#351), I had several problems with the [textDocument/hover](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover) request. I initially believed the problem to be in my LSP client implementation (and I was correct, to some extent) but now I think the problem also lies in Effekt/Kiama.

A reproducible communication flow that sometimes returns the wrong result is as follows:

1. `InitializeRequest` -> ...
2. `InitializedNotification`
3. `DidOpenTextDocumentNotification` `effect Exc(msg: String): Unit`
4. `HoverRequest` (e.g. line 0, char 8) -> `null` / { ... }

Where the last `HoverRequest` returns `null` instead of the correct result around 1/2 of the times.

You can try it yourself using [this JS script](https://gist.github.com/marvinborner/ddc72b214352a485d308a7b1c157230b).

I spent a lot of time trying to debug this but ultimately did not get much further than that this function call sometimes returns `None` instead of `Some(...)`:

https://github.com/effekt-lang/effekt/blob/e0d28ef099d3004ddfdde63d1e07826c7983b265/effekt/jvm/src/main/scala/effekt/Server.scala#L91

I was not able to reproduce this issue with other similar request types.

b-studios:
Maybe it is some kind of timing issue? It looks like `process` (which in turn will compile the file) will only be called `onsave`

https://github.com/effekt-lang/kiama/blob/3401470a4cbbe294d2df0c6923e919b8a45ca1f5/jvm/src/main/scala/kiama/util/Server.scala#L498

so maybe this call is not fully processed, when the onhover is triggered?

b-studios:
I don't know what's the LSP best practices here. Should there be _some_ form of synchronization? Should we block until the first `process` is completed? Or is this completely against how LSP usually works?

marvinborner:
Don't we already process on `textDocument/didOpen` [here](https://github.com/effekt-lang/kiama/blob/3401470a4cbbe294d2df0c6923e919b8a45ca1f5/jvm/src/main/scala/kiama/util/Server.scala#L486) and, in the reproduction, [here](https://gist.github.com/marvinborner/ddc72b214352a485d308a7b1c157230b#file-repro-js-L41)?

I assumed `process` to be blocking already, otherwise it should probably return a future or something, right? I think this is probably the best practice, I have not come across synchronization yet.

[...]

PR #885, opened 2025-03-13
Standalone language server

timsueberkrueb:
Free the language server from Kiama and implement a test suite.
Found some bugs while reimplementing, and I've confirmed all of them also exist in the old server.
It should be on-par with the old implementation in terms of features and bugs.

Depends on https://github.com/effekt-lang/kiama/pull/9.

**Differences to the old Server:**

* Added unit test suite
* New server runs single threaded and sequentially
* Compile on `didChange`, by default
* No Monto anymore, added a custom `$/effekt/publishIR` notification instead.

**Implementation status:**

* [x] `initialize`
* [x] `shutdown` - not covered by unit tests - they cannot handle the process exit
* [x] `exit` - not covered by unit tests - they cannot handle the process exit
* [x] setTrace
* [x] Diagnostics `afterCompilation`
* [x] `didChange`
* [x] `didOpen`
* [x] `didSave`
* [x] `didClose`
* [ ] `hover`
  * [x] for symbols
  * [ ] :bug: for holes: https://github.com/effekt-lang/effekt/issues/549
* [ ] `documentSymbol`
  * [x] impl & test
  * [ ] :bug: Fix spurious symbols at `(0, 0)`: https://github.com/effekt-lang/effekt/issues/895
* [ ] `references`
  * [x] impl & test
  * [ ] :bug: Fix not all references being returned (see test case): https://github.com/effekt-lang/effekt/issues/896
* [x] `inlayHint`
  * [x] impl & test
  * [x] :bug: inlayHints and hoverInfo sometimes return null (as in old implementation):
    * [x] inlayHints: https://github.com/effekt-lang/effekt/issues/876 (fixed in #894)
    * [x] hoverInfo: https://github.com/effekt-lang/effekt/issues/366 (**can't reproduce**)
       Note that the timing issue should be gone as we now run on `newSingleThreadExecutor` (which claims "Tasks are guaranteed to execute sequentially"), the caching issues have been fixed in #894.
* [ ] `getCodeActions`
  * [x] port impl from old server
  * [ ] :bug: impl doesn't even work in old server: https://github.com/effekt-lang/effekt/issues/897
* [x] `definition`
* [ ] `notebookDocument/*`
  * not implemented, student works on another branch
* [ ] publish IR (`$/effekt/publishIR`), replacing Monto
  * [x] impl server and test (this PR)
  * [x] impl client (https://github.com/effekt-lang/effekt-vscode/pull/63)
  * [ ] :bug: many options (e.g. `target`) don't work because the server only runs the compiler frontend - same issue with old server

Important questions to reviewers:

* [x] `effekt/jvm/src/main/scala/effekt/KiamaUtils.scala` is MPL-licensed code from Kiama - what do we do with this?
  * I decided to move it to Kiama to be safe.


timsueberkrueb:
Merge [Kiama PR](https://github.com/effekt-lang/kiama/pull/9) first.

jiribenes (approved):
I went through this a few times while on a bus, it seems pretty good.
My only caveat is that it leaks a lot of Java types out instead of their Scala equivalents (since it's hard to work with them), but we can also fix that later.

PR #887, opened 2025-03-14
Bump to llvm version 18 in CI

mattisboeckle:
We are running into problems where the versions used for development and for the CI are differing so much that we sometimes write code in development that is invalid in the CI. (See #877)

I propose upping the Version to 18, which seems to be the Ubuntu standard

Edit: the latest release is 20.1.0, so we are also not in unstable territory

### DIFF
--- a/.github/actions/setup-effekt/action.yml
+++ b/.github/actions/setup-effekt/action.yml
@@ -13,7 +13,7 @@ inputs:
   llvm-version:
     description: 'LLVM version to install'
     required: false
-    default: '15'
+    default: '18'
   install-dependencies:
     description: 'Whether to install system dependencies (Linux only)'
     required: false
--- a/effekt/jvm/src/main/scala/effekt/EffektConfig.scala
+++ b/effekt/jvm/src/main/scala/effekt/EffektConfig.scala
@@ -69,7 +69,7 @@ class EffektConfig(args: Seq[String]) extends REPLConfig(args.takeWhile(_ != "--
   val llvmVersion: ScallopOption[String] = opt[String](
     "llvm-version",
     descr = "the llvm version that should be used to compile the generated programs (only necessary if backend is llvm, defaults to 15)",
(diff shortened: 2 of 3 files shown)


jiribenes:
Note that this needs to be coordinated as it also requires changes on the website, in the autorelease CI, config
https://github.com/effekt-lang/effekt/blob/19fc826eb9e6e5dc3ed1e9d15fa40ee86a8832e2/effekt/jvm/src/main/scala/effekt/EffektConfig.scala#L72

the runner
https://github.com/effekt-lang/effekt/blob/19fc826eb9e6e5dc3ed1e9d15fa40ee86a8832e2/effekt/jvm/src/main/scala/effekt/Runner.scala#L270

+ in the packaging in effekt-nix.

marvinborner:
Are we sure we want to make version 18 a requirement? The default version in current Debian (which is probably which most people install) is still below 18 (or even 15, see [here](https://packages.debian.org/search?keywords=llvm)).

Either way, we'd also need to sync this change with [the plots ci](https://github.com/effekt-lang/effekt-plots/blob/main/.github/workflows/sync.yml#L62) (which uses the default Ubuntu version)

jiribenes:
FWIW, I'm fine with pushing Debian users into using the official apt packages via https://apt.llvm.org/

mattisboeckle:
Changed the remaining locations I could find. I think we should rebase this on top of #888.

jiribenes:
We still need to update the website in sync as it mentions the version requirements: https://github.com/effekt-lang/effekt-website/blob/d2f0fee403ef48777465f48f0998a42cd35ed0e9/contributing.md?plain=1#L34

mattisboeckle:
effekt-website also has a PR now.

[effekt-lang/effekt-website#84](https://github.com/effekt-lang/effekt-website/pull/84)

jiribenes (approved):
LGTM :shipit:

PR #888, opened 2025-03-15
CI: Extract reusable 'test' and 'build' actions

jiribenes:
I was sitting on this idea for a while: the CI is somewhat difficult to navigate since we keep building the env anew every single time...

GitHub Actions allows us to make reusable actions which we can use like subroutines for the actual workflows.
Here, we have:
- `run-effekt-tests` which just runs tests (we can choose whether full tests or partial ones or if it should retry)
- `setup-effekt` which prepares the devenv

See docs [here](https://docs.github.com/en/actions/sharing-automations/avoiding-duplication).

Now it's much easier to, say, change the LLVM version in _all_ CI actions! :^)

---

No idea if this works, most of this was vibe-coded with Sonnet 3.7 and I have no way of testing this locally anyways... 🤷‍♂️ 😎

### DIFF
--- a/.github/actions/run-effekt-tests/action.yml
+++ b/.github/actions/run-effekt-tests/action.yml
@@ -1,0 +1,58 @@
+name: 'Run Effekt Tests'
+description: 'Runs Effekt tests with configurable options'
+
+inputs:
+  full-test:
+    description: 'Whether to run full test suite'
+    required: false
+    default: 'true'
+  use-retry:
+    description: 'Whether to use retry mechanism for tests'
+    required: false
+    default: 'true'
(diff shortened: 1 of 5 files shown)


jiribenes:
cc @marvinborner

If the [docs](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#calling-a-reusable-workflow) aren't lying, this would also mean that you could use the following in your CI in effekt-plots and then it would be always magically up-to-date (same LLVM, NodeJS, npm, JVM, etc. versions):
```yml
- uses: effekt-lang/effekt/.github/actions/setup-effekt
```

Issue #605, opened 2024-09-24
Rethink / re-design "object-system"


marzipankaiser:
In general, we need to rethink / redesign the "object-system" anyways. Some aspects are:

- [ ] #600
- [ ] #274
- [ ] toplevel objects (can those be mutually recursive?)
- [ ] classes
- [ ] implementing multiple interfaces
- [ ] objects as members
- [ ] private methods
- [ ] "extension" methods
- [ ] fields

# See also
- This is lightly edited from @b-studios in https://github.com/effekt-lang/effekt/issues/600#issuecomment-2370616782

marzipankaiser:
(Of course, feel free to add/edit/make sub-issues etc)

b-studios:
I'd like to be able to allocate an object into a region (if it has mutable state).

jiribenes:
> I'd like to be able to allocate an object into a region (if it has mutable state).

Would it also make sense to allocate a _capability_ into a region? 🤔

```scala
def makeCounter {r: Region} = {
  var count in r = 0
  try {
    println(c.get())
    c.increment()
    c.increment()
    println(c.get())
  
    c
  } with c: Counter in r {
//                  ~~~^
    def get() = resume(count)
    def increment() = { count = count + 1; resume(()) }
  }
}
```

PR #890, opened 2025-03-18
Fix failing tests without optimization

marvinborner:
This is a *continuation* of #851. We aim to fix the previously ignored tests that currently fail when being run without optimization.

- [x] permute.effekt: segfault in resume->uniqueStack->copyStack since resume is called with an erased resumption stack
  - apparently from growing the stack via checkLimit
    - fixed e.g. by initial size = `shl 1, 8` instead of 7
- [x] multiple declarations (in JS)
  - `ascii_isalphanumeric.effekt`, `ascii_iswhitespace.effekt`, `parser.effekt`, `probabilistic.effekt`
  - fixed by "Do not contify under reset" (found out via bisect) 
    - is still a problem though
- [x] missing block info
  - [x] generator.effekt: by noting parameters for regions
  - [x] regions.effekt
  - [x] selfregion.effekt
- [x] typeparametric.effekt: by returning garbage value (`undef`)
- [x] issue842.effekt: by #872
- [x] issue861.effekt: by #872
- [x] top-level object definititions
  - if_control_effect.effekt, toplevel_objects.effekt, type_omission_op.effekt, higherorderobject.effekt, res_obj_boxed.effekt, effectfulobject.effekt


marvinborner:
Top-level object definitions are still not supported, we should continue this somewhere else -- maybe #605?

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -357,8 +357,8 @@ class VMTests extends munit.FunSuite {
       poppedFrames = 8661,
       allocations = 0,
       closures = 0,
-      variableReads = 32437,
-      variableWrites = 13699,
+      variableReads = 23776,
+      variableWrites = 5039,
```
b-studios, line 361: We should instrument the refs at some point (can you open an issue?)

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -378,8 +378,6 @@ object Transformer {
 
   def transform(value: machine.Variable)(using FunctionContext): Operand =
     substitute(value) match {
-      // TODO rethink existence of global
-      case machine.Variable("global", machine.Type.Prompt()) => ConstantGlobal("global")
```
b-studios, line 382: :D

b-studios (approved):
Great job!

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
marvinborner: Done, #916

PR #893, opened 2025-03-26
Add support for property-based testing

lenakaeufel:
This pull request is the result of my bachelor thesis, ["Property Testing in Effekt"](https://github.com/lenakaeufel/pbt-effekt). It extends the existing testing module by adding support for **property-based testing (PBT)**, while maintaining full compatibility with the existing functionality.

### 🔧 Changes to `test.effekt`:
- Introduced a new `Generator` effect to model stream-based generators for PBT
- Added predefined generators for primitive types (e.g., `Int`, `Bool`)
- Extended the `Test` effect with four new operations to capture PBT results:
  - Success and failure for both `forall` and `exists` quantifiers
- Implemented quantifiers `forall` and `exists` in various forms:
  - Support for one, two, and three input generators
  - Two versions of the single-input `forall`: one with implicit generator passing, and one with explicit

### 🧪 Added PBT Example Files:
- `list_examples_pbt.effekt`: A mixed test suite with unit and property-based tests, showcasing properties of some standard library list functions
- `tree_examples_pbt.effekt`: Demonstrates custom generators and property tests for the standard library’s tree-based map

### 📌 Note on Shrinking:
Shrinking is **not yet implemented**, as it requires further research and design.  
However, the [thesis repository](https://github.com/lenakaeufel/pbt-effekt) contains a prototype exploring a potential approach to shrinking.

### DIFF
--- a/examples/stdlib/test/list_examples_pbt.effekt
+++ b/examples/stdlib/test/list_examples_pbt.effekt
@@ -4,8 +4,9 @@ import random
 
 def main() = {
   with minstd(1337);
-  
-  println(suite("PBT: List Functions and Other Simple Examples") {
+
+  // Don't print out times in CI.
+  suite("PBT: List Functions and Other Simple Examples", false) {
 
     // example unit tests to show that mixed test suites work
     test("1 + 1 == 2") {
@@ -165,6 +166,8 @@ def main() = {
         assertTrue(n.mod(2) == 0)
         assertTrue(n <= 4)
     }
-  })
+  };
+
+  ()
 }
     
--- a/examples/stdlib/test/tree_examples_pbt.effekt
+++ b/examples/stdlib/test/tree_examples_pbt.effekt
@@ -59,7 +59,8 @@ def main()= {
   with uniqueInt; 
   with arbitraryBinTree[Int, String](8);
 
-  println(suite("PBT: Tree Map Examples") {
+  // Don't print out times in CI.
+  suite("PBT: Tree Map Examples", false) {
 
     // get-put law: get(put(M, K, V), K) = V
     forall[String, internal::Tree[Int, String], Int]("get-put law", 100)
@@ -106,5 +107,7 @@ def main()= {
       val tMap = internal::map(t){ (_k, _v) => ()}
       assertEqual(tForget, tMap)
     }
-  })
+  };
+
+  ()
 }
--- a/libraries/common/test.effekt
+++ b/libraries/common/test.effekt
@@ -510,7 +510,6 @@ def exists[A, B](name: String, n: Int)
 /// }
 /// ```
 def suite(name: String, printTimes: Bool) { body: => Unit / { Test, Formatted } }: Bool / {} = {
-  println("here")
   with Formatted::formatting;
 
   def ms(duration: Int): String / Formatted =
--- a/examples/stdlib/test/list_examples_pbt.check
+++ b/examples/stdlib/test/list_examples_pbt.check
@@ -1,0 +1,50 @@
+PBT: List Functions and Other Simple Examples
+✓ 1 + 1 == 2 [0.87ms]
+✕ 2 + 2 == 3 [0.14ms]
+  Expected: 3
+  Obtained: 4
+✓ reverse-singleton, passed 100 tests  [3.16ms]
+✓ reverse-singleton, explicitely passing the generator, passed 100 tests  [1.59ms]
+✕ reverse-singleton mistake [0.15ms]
+  ! Falsified after 0 passed tests: [0.15ms]
+  Expected: Cons(6, Nil())
+  Obtained: Cons(47, Nil())
+  failed on input:
+    1. 47
+✓ Is the Integer 2 among the generated values? [0.40ms]
+  Verified after 18 tried inputs
+  example value:
+    1. 2
+✓ reverse: distributivity over append, passed 100 tests  [4.36ms]
+✕ reverse: distributivity mistake - swapped order [0.72ms]
+  ! Falsified after 0 passed tests: [0.72ms]
+  Expected: Cons(-45, Cons(43, Cons(15, Cons(0, Nil()))))
+  Obtained: Cons(15, Cons(0, Cons(-45, Cons(43, Nil()))))
+  failed on inputs:
+    1. Cons(43, Cons(-45, Nil()))
+    2. Cons(0, Cons(15, Nil()))
+✓ |zip(xs,ys)| === min(|xs|,|ys|), passed 10 tests  [0.83ms]
+✕ intended mistake: |zip(xs,ys)| != max(|xs|,|ys|) [0.39ms]
+  ! Falsified after 0 passed tests: [0.39ms]
+  Expected: 5
+  Obtained: 2
+  failed on inputs:
+    1. Cons(102, Cons(97, Cons(41, Cons(123, Cons(122, Nil())))))
+    2. Cons(42, Cons(-46, Nil()))
+✓ unzip-zip-take relation, passed 10 tests  [3.92ms]
+✓ drop: concatenation , passed 10 tests  [1.94ms]
+✓ drop-slice relation, passed 10 tests  [0.64ms]
+✓ reverseOnto-reverse-append relation , passed 10 tests  [0.84ms]
+✓ size-foldLeft relation, passed 20 tests  [1.14ms]
+✓ take-drop: inversion over concatenation, passed 20 tests  [0.77ms]
+✕ even numbers are even and smaller than 4 [0.11ms]
+  ! Falsified after 1 passed test: [0.11ms]
+  Assertion failed
+  failed on input:
+    1. 12
+
+ 12 pass
+ 5 fail
+ 17 test(s) total [41.5ms]
+
+false
--- a/examples/stdlib/test/tree_examples_pbt.check
+++ b/examples/stdlib/test/tree_examples_pbt.check
@@ -1,0 +1,9 @@
+PBT: Tree Map Examples
+✓ get-put law, passed 100 tests 
+✓ put-put law, passed 100 tests
+✓ put-get law, passed 100 tests 
+✓ forget-map relation, passed 100 tests 
+ 
+ 4 pass
+ 0 fail
+ 4 test(s) total


examples/stdlib/test/list_examples_pbt.effekt
```diff
@@ -0,0 +1,166 @@
+import test
+
+def main() = {
+  println(suite("PBT: List Functions and Other Simple Examples") {
```
jiribenes, line 4: We're using golden tests here, so there needs to be a check file with the existing output.
This lets us know if the test fails. See [the other tests in this folder](https://github.com/effekt-lang/effekt/tree/master/examples/stdlib/test)


Also, since we're comparing the outputs, we can't really print times in CI:
```suggestion
  // Don't print out times in CI.
  suite("PBT: List Functions and Other Simple Examples", false) {
```
`+` remove the last paren.

examples/stdlib/test/tree_examples_pbt.effekt
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+def main()= {
+  with arbitraryChar;
+  with arbitraryString(5);
+  with uniqueInt; 
+  with arbitraryBinTree[Int, String](8);
+
+  println(suite("PBT: Tree Map Examples") {
```
jiribenes, line 59:
```suggestion
  // Don't print out times in CI.
  suite("PBT: Tree Map Examples", false) {
```
`+` remove the last paren from the println :)

libraries/common/test.effekt
```diff
@@ -80,6 +323,185 @@ def test(name: String) { body: => Unit / Assertion } = {
...
+      do successForall(name, n, duration) 
+    }
+  } with Assertion { 
+    def assert(condition, msg) = 
+      if (condition) {
+        successCounter = successCounter + 1 
+        resume(())}
+      else {
+        val duration = Duration::diff(startTime, bench::relativeTimestamp())
+        do failureForall(name, successCounter, msg ++ "\n  failed on input:\n    1. " ++ genericShow(x), duration)
```
jiribenes, line 349: This won't work on the LLVM backend because of the `genericShow`.
Instead, we need to have even more overloads of `forall` that take a `{ show: A => String }` block parameter, which is later applied at this exact spot.

libraries/common/test.effekt
```diff
@@ -91,9 +513,17 @@ def test(name: String) { body: => Unit / Assertion } = {
...
+///   forall[Int]("reverse-singleton", 100){ x =>
+///     assertEqual(
+///       reverse([x]),
+///       [x]
+///     )
+///   }
 /// }
 /// ```
 def suite(name: String, printTimes: Bool) { body: => Unit / { Test, Formatted } }: Bool / {} = {
+  println("here")
```
jiribenes, line 526: Leftover debug print:
```suggestion
```

libraries/common/test.effekt
```diff
@@ -80,6 +323,185 @@ def test(name: String) { body: => Unit / Assertion } = {
...
+    body(x)
+    if(triedCounter == n) {
+      val duration = Duration::diff(startTime, bench::relativeTimestamp())
+      do failureExists(name, triedCounter, duration) 
+    }
+  } with Assertion { 
+    def assert(condition, msg) = 
+      if (condition) {
+        val duration = Duration::diff(startTime, bench::relativeTimestamp())
+        do successExists(name, triedCounter, " example value:\n    1. " ++ genericShow(x), duration)
```
jiribenes, line 466: Same as in `forall`, in general, this needs to take a `show` parameter which is then applied on the tested value.

examples/stdlib/test/list_examples_pbt.effekt
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    // reverse[x] === [x]
+    with arbitraryInt;
+    forall[Int]("reverse-singleton", 100){ x => 
+      assertEqual(
+        reverse([x]),
+        [x]
+      )
+    }
```
jiribenes, line 22: When you add the checkfiles, this will probably fail for two reasons on LLVM:
1. we need to use the variant of `forall` that knows how to print the tested value (see previous comments on `forall` defn)
2. `assertEqual` works only on JS; if we want it to work on LLVM, you have to use a more general form of `assertEqual` (defn below), so something like `assertEqual(reverse[x], [x]) { list => <{ /* equality on lists */ }> } { list => show(list) }`
https://github.com/effekt-lang/effekt/blob/55e373b537fcf104419b5ec3685a09e1faa2a5e6/libraries/common/test.effekt#L54 

where you might need to define equality on lists in `libraries/common/list.effekt` if we don't have it already.

libraries/common/test.effekt
```diff
@@ -56,10 +69,240 @@ def assertEqual[A](obtained: A, expected: A) { equals: (A, A) => Bool } { show:
...
+  def generate(): Unit/emit[T]
+  
+  /// Produces a puh stream of simplified (shrunk) versions of a given value of type `T`
+  def shrink(v: T): Unit/emit[T]
+}
+
+def arbitraryInt { body: => Unit / Generator[Int] }: Unit =
+  try body() with Generator[Int] {
+    def generate() = resume {
+      //[0, -1, 1, int.minVal, int.maxVal, -2, 2].each // This can reduce the need for shrinking by emitting common edge cases first
```
jiribenes, line 84: Do we want this then?

libraries/common/test.effekt
```diff
@@ -56,10 +69,240 @@ def assertEqual[A](obtained: A, expected: A) { equals: (A, A) => Bool } { show:
...
+      while (true) {
+        do emit(randomInt(-50, 50))
```
jiribenes, line 86: This seems too small for "real" uses, can we bump this up to a lot more?

libraries/common/test.effekt
```diff
@@ -56,10 +69,240 @@ def assertEqual[A](obtained: A, expected: A) { equals: (A, A) => Bool } { show:
...
+
+    def shrink(v: Bool) = resume {
+      var out = v
+      if(v){
+        out = false
+      }else {
+        out = true
+      }
+      do emit(out)
+    }
```
jiribenes, line 157: I don't get this. Is it just
```suggestion
    def shrink(v: Bool) = resume {
      do emit(not(out))
    }
```
?

libraries/common/test.effekt
```diff
@@ -56,10 +69,240 @@ def assertEqual[A](obtained: A, expected: A) { equals: (A, A) => Bool } { show:
...
+    def shrink(v: Char) = resume { <> }
+  }
+
+def arbitraryString { body: => Unit / Generator[String] }: Unit / Generator[Char] = {
+  try body() with Generator[String] {
+    def generate() = resume {
+      while (true) {
+        val string: String = collectString { 
+          with boundary;
+          with limit[Char](randomInt(1, 20))
```
jiribenes, line 200: Does this mean it never generates an empty string?
Also, this again seems like it generates outputs which are too small?

libraries/common/test.effekt
```diff
@@ -56,10 +69,240 @@ def assertEqual[A](obtained: A, expected: A) { equals: (A, A) => Bool } { show:
...
+  }
+}
+
+def arbitraryList[T] { body: => Unit / Generator[List[T]]} : Unit / Generator[T] = {
+  try body() with Generator[List[T]]{ 
+    def generate() = resume {
+      while (true) {
+        val list = collectList[T]{ 
+          with boundary;
+          with limit[T](randomInt(1, 10))
```
jiribenes, line 234: see my notes for the string: does this ever generate an empty list? isn't the limit too small?

libraries/common/test.effekt
```diff
@@ -56,10 +69,240 @@ def assertEqual[A](obtained: A, expected: A) { equals: (A, A) => Bool } { show:
...
+          with limit[T](len)
+          do generate[T]()
+        }
+        do emit(list)
+      }
+    }
+
+    def shrink(v: List[T]) = resume { <> }
+  }
+}
```
jiribenes, line 260: I like the parametric version, but should it take a fixed length `len: Int` for each generated list or a _maximum_ length `max: Int` which it then uses to generate a random list of size 0..max?

jiribenes (changes requested):
Thanks for upstreaming, I'm very much looking forward to using the library :)

I have a few comments, mostly about how we can make the stdlib tests pass and how we enable LLVM to also use your testing library.
And I have a few questions — it's completely OK if you aren't sure about your answers to the open-ended questions, but I'm still interested in knowing your thoughts.

Let me know if you have any questions.

jiribenes:
Bump @lenakaeufel: thanks again for trying to upstream the code.
Are you planning to continue working on it? 

If needed, we can also take the PR over in order to ensure that it gets merged :)

lenakaeufel:
> Bump @lenakaeufel: thanks again for trying to upstream the code. Are you planning to continue working on it?
> 
> If needed, we can also take the PR over in order to ensure that it gets merged :)

Hey, sorry for the silence, I've been quite busy over the past few weeks. I am planning to continue working on it, but sadly haven’t had the time yet :( 
I’ll do my best to make progress by next Wednesday. If I’m not able to get to it by then, it’s probably best if you take it over, as I won’t have more time after that. 
I hope that works for you!

examples/stdlib/test/list_examples_pbt.effekt
lenakaeufel: Since the generators used for the property tests produce random data, the result of some of the tests (for failing `forall` tests the counter examples and number of passed tests, for passing `exists` tests the number of tried test cases and the example that works,...) will differ with each execution
This is problematic for the check files, as we can't predict the exact output, right? 
Do you have any ideas on how we could still create check files that includes these kinds of property tests?
Otherwise, I could remove them, but it would be less than ideal to not have examples for failing `forall` tests, etc.

jiribenes: Good point, we probably need a seeded generator then.
Long term, we can address this by having our own `randomInt` (https://github.com/effekt-lang/effekt/pull/966).

For now, you could change your `randomInt` to:
```
extern js"""
  function prng(seed) {
    const a = 1664525;
    const c = 1013904223;
    const m = Math.pow(2, 32);

    let currentSeed = seed;

    // Return a function that generates a random integer between min and max (inclusive)
    return function(min, max) {
      // Update the seed using the LCG formula
      currentSeed = (a * currentSeed + c) % m;

      return Math.floor(min + (currentSeed / m) * (max - min + 1));
    };
  }

  const getRandomInt = prng(1337);
"""

extern def randomInt(minInclusive: Int, maxExclusive: Int): Int =
  js"getRandomInt(${minInclusive}, ${maxExclusive})"
```

This means that the randomness will, for now, be actually deterministic, which is fine (until #966 gets merged)

lenakaeufel: perfect, thank you very much!

libraries/common/test.effekt
lenakaeufel: sure, what would be a good range in your opinion? Is `(-10000,10000)` okay or still too small?

libraries/common/test.effekt
lenakaeufel: yes it is, I'll change it :D It seems like some leftover code from an experiment, thanks for noticing!

libraries/common/test.effekt
lenakaeufel: That's right, it definitely needs improvements

I think most of the generators do, I'll change all of them with better ranges. Again, what do you think would be good ranges for real uses?

When allowing much bigger inputs it might be even more important to output common edge cases first and continue with the random values afterwards, wdyt?
To realize this it would be great to have an `Int` max/min value, but Effekt doesn't provide it yet, does it?

libraries/common/test.effekt
lenakaeufel: It probably depends on the use case, but since the maximum length version provides more randomness I think it's better for the standard version, I'll change it
If users need the fixed length version, they can still make a custom generator

libraries/common/test.effekt
lenakaeufel: I can try to change it, `show` would be just like in the lowermost `assertEqual` version, right?
```
def assertEqual[A](obtained: A, expected: A) { equals: (A, A) => Bool } { show: A => String }: Unit / { Assertion, Formatted } =
  do assert(equals(obtained, expected), Formatted::tryEmit(Escape::RESET) ++ "Expected: ".dim ++ show(expected).green ++ "\n  Obtained: ".dim ++ show(obtained).red) 
```
  
Sadly I can only use the JS backend on my laptop, would it be okay for you to check it for the other backends once I committed the changes?

PR #894, opened 2025-03-27
Fix lsp caching issues

### LINKED ISSUE #876: Inlay hints sometimes don't reappear after file edits

timsueberkrueb:
Depends on #885 (mostly so we can actually express the unit test).

I split `AnnotationsDB` into three separate databases which each tracks its own map. This allows us to properly distinguish types that should be annotated by their object identity (symbols, trees) and those that should be annotated by their value identity (sources).

The new types are:

* `TreeAnnotations`: key: `source.Tree`, object identity
* `SourceAnnotations`: key: `kiama.util.Source`, value identity
* `SymbolAnnotations`: key: `symbols.Symbol`, object identity

Fixes #876.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/context/Annotations.scala
+++ b/effekt/shared/src/main/scala/effekt/context/Annotations.scala
@@ -26,7 +26,7 @@ class Annotations private(
   /**
    * Local Annotations are organized differently to allow simple access.
    */
-  private var annotations: Map[Annotation[_, _], Map[Key[Any], Any]]
+  private var annotations: Map[Annotation[_, _], Map[Annotations.Key[Any], Any]]
 ) {
   import Annotations._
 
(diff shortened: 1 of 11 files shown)


effekt/shared/src/main/scala/effekt/context/Annotations.scala
```diff
@@ -6,8 +6,28 @@ import effekt.util.messages.ErrorReporter
...
+  type Value = V
+
+  override def toString = name
+}
+
+case class TreeAnnotation[K <: source.Tree, V](name: String, description: String) extends Annotation[K, V] {
+  type Value = V
+
   override def toString = name
 }
```
jiribenes, line 32: Beautiful :)

PR #900, opened 2025-03-28
Allow trailing comma in list literals

### LINKED ISSUE #873: Allow trailing commas in list literals
```scala
val foo = [ 
  1,
  2,
]
```
should be parseable and not cause a crash in the parser :)

The parser code is here:
https://github.com/effekt-lang/effekt/blob/a9827f5e76522a116802d356b3d8b5d129fa3b5b/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala#L990-L996

... and it uses `many` which is a nice candidate for the change (maybe with `manyTrailing` to begin with?) ...
https://github.com/effekt-lang/effekt/blob/a9827f5e76522a116802d356b3d8b5d129fa3b5b/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala#L1394-L1408

... then feel free to add a few tests for the new behaviour in [`RecursiveDescentTests.scala`](https://github.com/effekt-lang/effekt/blob/master/effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala) :)

mattisboeckle:
Fixes #873

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
@@ -128,6 +128,9 @@ class RecursiveDescentTests extends munit.FunSuite {
 
     parseExpr("[1,2,3]")
     parseExpr("[3,2,1,]")
+    parseExpr("[]")
+    parseExpr("[,]")
+    intercept[Throwable] { parseExpr("[,1]") }
   }
 
   test("Boxing") {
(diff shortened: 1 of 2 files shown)


effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
```diff
@@ -125,6 +125,9 @@ class RecursiveDescentTests extends munit.FunSuite {
     parseExpr("fun() { foo(()) }")
 
     parseExpr("10.seconds")
+
+    parseExpr("[1,2,3]")
+    parseExpr("[3,2,1,]")
```
jiribenes, line 130: Is it also possible to add some tests for an empty and a singleton list with trailing commas?
Just to make sure that even the trivial cases work :)

dvdvgt (approved):

PR #901, opened 2025-03-28
Fix: Splices with non-String type break if there is no splice

### LINKED ISSUE #892: Splices with non-`String` type break if there is no splice.
When using #743 with a splice handler that does not return a `String`, like e.g.:
```
def mylit{ body: => Unit / { literal } }: Int = <>
```
we will get a type error when using it *without actually splicing anything*:
```
def main() = {
  println(show(mylit"foo" + 2)) // Cannot typecheck call.
}
```

## Runtime behaviour
Furthermore, if we do a transformation on the `String`, e.g.:
```
def mylit{ body: => Unit / { literal } }: String =
  try { body(); "1" } with literal { l => "0" }
```
it will silently drop the `mylit` for `Strings` without splices:
```
def main() = {
  println(mylit"foo") // prints "foo", not "0"
}
```

dvdvgt:
Fixes #892

### DIFF
--- a/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
+++ b/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
@@ -1030,10 +1030,9 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
   def templateString(): Term =
     nonterminal:
       backtrack(idRef()) ~ template() match {
-        // We do not need to apply any transformation if there are no splices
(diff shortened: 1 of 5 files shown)


jiribenes:
Could you also add the repro at the top of https://github.com/effekt-lang/effekt/issues/892 by @marzipankaiser?
It represents a common use, returning a specific type different than String, which we don't test for.

jiribenes (approved):
LGTM :shipit:

PR #902, opened 2025-03-28
Fix inferring effectful while header

### LINKED ISSUE #601: Wrong effect inferred for effectful `while` header
The following program compiles without typer errors:
```effekt
interface Foo { def foo(): Unit }

def bar(): Bool / Foo = {
  do foo()
  false
}

def fooo() = {
  while(bar() is true) {
      ()
    }
}

def main() = {
  fooo()
}
```
but crashes at runtime (in JS), because there is no handler for the `do foo` in the loop header.
The type of `fooo` is shown (at least via LSP) as `def fooo(): Unit / {}`.

Adding a handler in main generates a warning that the handler is unused (but runs fine).

PhictionalOne:
Should fix #601


jiribenes:
Thanks! 🥳

Issue #667, opened 2024-11-05
Confusing parser error messages on wrong kind


jiribenes:
There are a lot of confusing error messages from the parser when writing things like:
```scala
record Fun(f: (Int, Int) => Bool)
//                       ^^
// Expected ) but got =>

def foo { f: (Int, Int) => Bool at {} } = <>
//                         ^^^^
// Expected an interface type.

def bar(f: (Int, Int) => Bool) = <>
//                    ^^
// Expected ) but got =>
```

They all stem from the fact that we distinguish between computation types and value types in the parser.
Instead, we should parse some generic, kind-less type and only resolve it properly some time later (ideally in Typer?)

My preferred error messages would be something like:
```scala
record Fun(f: (Int, Int) => Bool)
//            ~~~~~~~~~~~~~~~~~~
//
// Expected a value type, got a computation type.
// Did you mean: `(Int, Int) => Bool at {}`?

def foo { f: (Int, Int) => Bool at {} } = <>
//            ~~~~~~~~~~~~~~~~~~~~~~~
//
// Expected a computation type, got a value type.
// Did you mean:
// 1. `{ f: (Int, Int) => Bool }`?
// 2. `(f: (Int, Int) => Bool at {})`?

def bar(f: (Int, Int) => Bool) = <>
```

PR #906, opened 2025-03-28
Fix: make type pretty printer less confusing for boxed types

### LINKED ISSUE #659: Missing parentheses in a pretty-printed type
The suggested type (`Int => Exception at {exc} at {io} / {}`) cannot be annotated, one can only write the parenthesised version: `(Int => Exception at {exc}) at {io}`.
Screenshot 2024-10-31 at 15 28 32

---

#### Parser
https://github.com/effekt-lang/effekt/blob/6e71bcd59bf419f2ba5923d9eca519b87d3804b5/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala#L1080

#### Pretty printer
https://github.com/effekt-lang/effekt/blob/6e71bcd59bf419f2ba5923d9eca519b87d3804b5/effekt/shared/src/main/scala/effekt/symbols/TypePrinter.scala#L53

dvdvgt:
Fixes #659, but not #667. I propose deferring fixing #667 to a later point in time, as it requires more design.

I also noticed that both

```
interface Exc {
	def throw(msg: String): Nothing
}

def divide(n: Int, m: Int) {exc: Exc}: (Int => Exc at {exc}) at {io} = {
	def res(x: Int) = { println(x); exc }
	res
}
```

and

```
interface Exc {
	def throw(msg: String): Nothing
}

def divide(n: Int, m: Int) {exc: Exc}: Int => (Exc at {exc}) at {io} = {
	def res(x: Int) = { println(x); exc }
	res
}
```

parse, yet the former is parsed as the latter regardless. This is very surprising to me, since these types clearly mean different things. In fact, boxed value types (first examples) should fail during type checking and not be accidentally transformed to a correct type.

What this PR does, for now, is to parenthesize boxed types when occurring in parameter or return type positions. For example `(Int => Int at {}) => Int` instead of `Int => Int at {} => Int`, which, in my opinion, is utterly unreadable.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/symbols/TypePrinter.scala
+++ b/effekt/shared/src/main/scala/effekt/symbols/TypePrinter.scala
@@ -66,13 +66,17 @@ object TypePrinter extends ParenPrettyPrinter {
       val tps = if (tparams.isEmpty) emptyDoc else typeParams(tparams)
       val ps: Doc = (vparams, bparams) match {
         case (Nil, Nil)       => "()"
+        case (List(tpe: BoxedType), Nil) => parens(toDoc(tpe))
         case (List(tpe), Nil) => if (tparams.isEmpty) toDoc(tpe) else parens(toDoc(tpe))
         case (_, _) =>
           val vps = if (vparams.isEmpty) emptyDoc else parens(hsep(vparams.map(toDoc), comma))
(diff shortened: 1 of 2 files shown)


jiribenes (approved):
LGTM :shipit:

PR #909, opened 2025-03-28
Remove `var x in global`

marvinborner:
Probably fixes `permute.effekt` and `selfregion.effekt` from #890


marvinborner:
Merged into #851

PR #910, opened 2025-03-28
Reimplement C utilities in llvm

mattisboeckle:
Reimplement various C utilities in LLVM and inline the functions inside of common effekt.

This should allow LLVM to better optimize code containing these funtions, as the optimizer can see the actual implementation instead of guessing.


mattisboeckle:
Benchmarked this version against the current master with `are_we_fast_yet`. Got ~40% speed increase on average.
part1
part2

With `list_tail` being the only program running slightly slower than before.

mattisboeckle:
I inlined parts of bytearray as well.
Because the functions are sometimes used in a c context (main.c for example) the old c functions still exist, but code written in Effekt should use the optimized versions.
This had no noticeable effect on performance

b-studios:
Are there benchmarks that use byte arrays?

libraries/common/array.effekt
```diff
@@ -8,14 +8,46 @@ import option
...
+    %data_pointer = getelementptr inbounds i64, ptr %array_pointer, i64 1
+    %size = load i64, ptr %array_pointer, align 8
+    %size_eq_0 = icmp eq i64 %size, 0
+    br i1 %size_eq_0, label %exit, label %loop
+loop:
+    %loop_phi = phi i64 [ %inc, %loop ], [ 0, %entry ]
+    %element_pointer = getelementptr inbounds %Pos, ptr %data_pointer, i64 %loop_phi
+    %elementTag = load i64, ptr %element_pointer
+    %elementPtr_pointer = getelementptr ptr, ptr %element_pointer, i64 1
+    %elementPtr = load ptr, ptr %elementPtr_pointer
```
phischu, line 25: Use something like

```
%tag_pointer = getelementptr inbounds %Pos, ptr %data_pointer, i64 %loop_phi, i32 0
%object_pointer = getelementptr inbounds %Pos, ptr %data_pointer, i64 %loop_phi, i32 1
...
```
to get the respective pointers.

libraries/common/array.effekt
```diff
@@ -8,14 +8,46 @@ import option
...
-    ret %Pos %z
+    %size0 = shl i64 ${size}, 4
+    %size1 = add i64 %size0, 24
+    %calloc = tail call noalias ptr @calloc(i64 %size1, i64 1)
+    %eraser_pointer = getelementptr ptr, ptr %calloc, i64 1
+    %size_pointer = getelementptr ptr, ptr %calloc, i64 2
+    store i64 0, ptr %calloc
+    store ptr @c_array_erase_fields, ptr %eraser_pointer
+    store i64 ${size}, ptr %size_pointer
+    %ret_pos = insertvalue %Pos { i64 0, ptr poison }, ptr %calloc, 1
```
phischu, line 49: Good job, poison seems to be the modern way!

libraries/common/array.effekt
```diff
@@ -49,8 +81,12 @@ extern pure def size[T](arr: Array[T]): Int =
   js "${arr}.length"
   chez "(vector-length ${arr})"
   llvm """
-    %z = call %Int @c_array_size(%Pos ${arr})
-    ret %Int %z
+    %pos_tag = extractvalue %Pos ${arr}, 0
+    %pos_data = extractvalue %Pos ${arr}, 1
+    %size_pointer = getelementptr inbounds i64, ptr %pos_data, i64 2
+    %size = load i64, ptr %size_pointer
+    tail call void @erasePositive(i64 %pos_tag, ptr %pos_data)
```
phischu, line 88: Why does `@erasePositive` (here and elsewhere) take to separate arguments? In `rts.ll` we have:

```
define void @erasePositive(%Pos %val) alwaysinline {
```

libraries/common/array.effekt
```diff
@@ -62,8 +98,19 @@ extern global def unsafeGet[T](arr: Array[T], index: Int): T =
...
+    %pos_tag = extractvalue %Pos ${arr}, 0
+    %pos_data = extractvalue %Pos ${arr}, 1
+
+    %data_pointer = getelementptr inbounds ptr, ptr %pos_data, i64 3
+    %element_pointer = getelementptr inbounds %Pos, ptr %data_pointer, i64 ${index}
+    %element_tag = load i64, ptr %element_pointer, align 8
+    %element_data_pointer = getelementptr inbounds ptr, ptr %element_pointer, i64 1
+    %element_data = load ptr, ptr %element_data_pointer, align 8
+    tail call void @sharePositive(i64 %element_tag, ptr %element_data)
+    tail call void @erasePositive(i64 %pos_tag, ptr %pos_data)
```
phischu, line 110: Related to a previous comment on the parameters of `@sharePositive` and `@erasePositive`, shouldn't it be possible to load the entire `%Pos` in one load instruction? (Which will probably be lowered to two actual load instructions, but still....

libraries/common/bytearray.effekt
```diff
@@ -12,32 +12,47 @@ extern type ByteArray
 extern global def allocate(size: Int): ByteArray =
   js "(new Uint8Array(${size}))"
   llvm """
-    %arr = call %Pos @c_bytearray_new(%Int ${size})
-    ret %Pos %arr
+    %object_size = add i64 ${size}, 16
+    %object_alloc = tail call noalias ptr @malloc(i64 noundef %object_size)
+    store i64 0, ptr %object_alloc, align 8
+    %object_data_ptr = getelementptr inbounds i8, ptr %object_alloc, i64 8
+    store ptr @c_bytearray_erase_noop, ptr %object_data_ptr, align 8
```
phischu, line 19: Define this `@c_bytearray_erase_noop` here in Effekt instead of in C (In analogy to the eraser for arrays).

libraries/common/bytearray.effekt
```diff
@@ -12,32 +12,47 @@ extern type ByteArray
...
   """
   chez "(bytevector-length ${arr})"
 
 extern global def unsafeGet(arr: ByteArray, index: Int): Byte =
   js "(${arr})[${index}]"
   llvm """
-    %byte = call %Byte @c_bytearray_get(%Pos ${arr}, %Int ${index})
-    ret %Byte %byte
+    %arr_ptr = extractvalue %Pos ${arr}, 1
+    %arr_data_ptr = getelementptr inbounds i8, ptr %arr_ptr, i64 16
```
phischu, line 39: We could do this and other offset calculations for the beginning of a object using the `%Header` struct. But this is ok...

libraries/common/ref.effekt
```diff
@@ -9,26 +9,69 @@ extern js """
   }
 """
 
+extern llvm """
+define void @c_ref_erase_field(ptr %0) {
+  %field_pointer = getelementptr inbounds ptr, ptr %0, i64 1
+  %tag = load i64, ptr %0, align 8
+  %field = load ptr, ptr %field_pointer, align 8
+  tail call void @erasePositive(i64 %tag, ptr %field)
+  ret void
```
phischu, line 18: This should be possible with a single loading of a `%Pos` directly at `%0`.

libraries/common/ref.effekt
```diff
@@ -9,26 +9,69 @@ extern js """
...
 extern type Ref[T]
 
+/** We represent references like positive types in LLVM
+ *  The tag is 0 and the obj points to memory with the following layout:
+ *
+ *   +--[ Header ]--------------+------------+
+ *   | ReferenceCount  | Eraser | Field      |
+ *   +--------------------------+------------+
+ */
+
```
phischu, line 32: Nice! Please add something like this to `array.effekt` and `bytearray.effekt`

libraries/common/ref.effekt
```diff
@@ -9,26 +9,69 @@ extern js """
...
 /// Allocates a new reference, keeping its value _undefined_.
 /// Prefer using `ref` constructor instead to ensure that the value is defined.
 extern global def allocate[T](): Ref[T] =
   js "{ value: undefined }"
   chez "(box #f)"
   llvm """
-    %z = call %Pos @c_ref_fresh(%Pos zeroinitializer)
-    ret %Pos %z
+    ; sizeof Header + sizeof Pos = 32
+    %ref = tail call noalias ptr @malloc(i64 noundef 32)
```
phischu, line 40: This must be `calloc`, because we might erase the reference before filling it.

EDIT: nevermind, I see you are storing `0` and `null` in the code below...

libraries/common/ref.effekt
```diff
@@ -9,26 +9,69 @@ extern js """
...
   llvm """
-    %z = call %Pos @c_ref_fresh(%Pos ${init})
-    ret %Pos %z
+    %initTag = extractvalue %Pos ${init}, 0
+    %initObject_pointer = extractvalue %Pos ${init}, 1
+    ; sizeof Header + sizeof Pos = 32
+    %ref = tail call noalias ptr @malloc(i64 noundef 32)
+    %refEraser = getelementptr ptr, ptr %ref, i64 1
+    %fieldTag = getelementptr ptr, ptr %ref, i64 2
+    %fieldData_pointer = getelementptr ptr, ptr %ref, i64 3
```
phischu, line 65: Here too I believe these could be 1 `getelementptr` and later 1 `store` without extracting the fields of `init`.

phischu (changes requested):
Looks good already. Generally, if possible keep working with entire `%Pos` structs instead of their individual fields.

libraries/common/array.effekt
mattisboeckle: Struct types as arguments are desugared into passing the seperate fields and the other way around.
So in this case 
```
@erasePositive(i64 %pos_tag, ptr %pos_data)
```
is the same as passing the entire `%Pos` struct

libraries/common/array.effekt
mattisboeckle: I will try

libraries/common/array.effekt
```diff
@@ -8,14 +8,46 @@ import option
...
+    %size = load i64, ptr %array_pointer, align 8
+    %size_eq_0 = icmp eq i64 %size, 0
+    br i1 %size_eq_0, label %exit, label %loop
+loop:
+    %loop_phi = phi i64 [ %inc, %loop ], [ 0, %entry ]
+    %element_pointer = getelementptr inbounds %Pos, ptr %data_pointer, i64 %loop_phi
+    %elementTag = load i64, ptr %element_pointer
+    %elementPtr_pointer = getelementptr ptr, ptr %element_pointer, i64 1
+    %elementPtr = load ptr, ptr %elementPtr_pointer
+    tail call void @erasePositive(i64 %elementTag, ptr %elementPtr)
```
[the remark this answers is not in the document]
mattisboeckle: `tail` is just a marker for LLVM that indicate tail call optimization should be used but it can also be ignored by the optimizer. In this case we can probably remove it without any difference in the generated code

libraries/common/array.effekt
mattisboeckle: Included in the newest version. This was possible in a lot of places and makes the code way cleaner

PR #912, opened 2025-03-28
Fix: ufcs integer lexing

### LINKED ISSUE #911: Longer integer literals don't work when UFCS'd
This is a lexer/parser bug:

```scala
// Exception in thread "main" java.lang.NumberFormatException: For input string: "1234567890123"
//        at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
def main() = 1234567890123.println
```

```scala
// .... but
def main() = println(1234567890123)
// ~> 1234567890123
// ... just works :)
```

dvdvgt:
fixes #911


jiribenes (changes requested):
Could we quickly add a test for this?
Just:
```scala
def main() = {
  1234567890123.println
  println(1234567890123)
}
```

jiribenes (approved):
Perfect, thanks! :shipit:

Issue #862, opened 2025-03-06
Errors with wrong or misleading position showing up at `/`


jiribenes:
There are a lot of little errors that don't have a position attached, like the code below (thanks @timsueberkrueb!):

```scala
def run(n: Int) { callback: Int => Unit }: Unit = callback(n)

def main() = {
  with run(42);     // should be `with val n = run(42)`
  ()
}
```

which produces:
```sh
$ effekt repro.effekt
[error] Wrong number of value arguments, given 0, but function expects 1.
```

Since it has no location, it shows up under `/` on line 1, column 1 in VSCode (via LSP),
so the user only sees the error in the "Problems" tab:

<img width="597" alt="Image" src="https://github.com/user-attachments/assets/5ffb6f5d-c631-499a-829d-738bec98f236" />

---

To address this, we should:
1. reduce the number of such errors as much as possible
2. if we don't have a location, we should report the error from somewhere else than `/`, maybe the name of the first file given by the user? 🤔

jiribenes:
For the example error in question, there are missing calls to `withPositionOf` here:
https://github.com/effekt-lang/effekt/blob/a1d13cd699f8429a67b4757e89fa3642d7311a27/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala#L195-L214

See here how `withPositionOf` is used elsewhere when desugaring:
https://github.com/effekt-lang/effekt/blob/a1d13cd699f8429a67b4757e89fa3642d7311a27/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala#L471-L477

jiribenes:
I think this is now resolved, cc @timsueberkrueb?

timsueberkrueb:
Yes, we still need to tear down the old positions infrastructure, but generally we should be in a good shape now wrt. spans on nearly every syntax node. There are still some missing spans in some synthesized cases, but they can be added on-demand.

PR #914, opened 2025-03-28
Allow lambda case patterns for multiple parameters

### LINKED ISSUE #761: Nested patterns in lambda case
Originally reported by @TheDying0fLight -- I thought there already was an issue for this, but I couldn't find it, sorry!

The following nested pattern matching of a tuple in lambda case is not supported:
```scala
[(1, 2), (3, 4)].foldLeft(0) { (acc, (x, y)) => acc + x + y }
```
as it returns a parser error `Expected ;` on the `=>` token.

What also doesn't work is using `case` like Scala would (see comment below):
```scala
[(1, 2), (3, 4)].foldLeft(0) { case (acc, (x, y)) => acc + x + y }
```
as that reports `Wrong number of value arguments, given 1, but function expects 2.`

A workaround that does actually work is, for example:
```scala
[(1, 2), (3, 4)].foldLeft(0) { (acc, tup) => 
  val (x, y) = tup;
  acc + x + y
}
```
(or just using `tup.first`, `tup.second`)

This doesn't show up only for tuples, but for anything nested in general:
```
record Person(name: String, age: Int)
def main() = 
  [Person("Joe", 42), Person("Jolene", 24)].foldLeft(0) { 
    (acc, Person(name, age)) =>
       acc + age
   }
```
again returns `Expected ;` on the `=>`.

This is somewhat related to https://github.com/effekt-lang/effekt/issues/707.

marzipankaiser:
Resolves #761 by allowing multiple patterns in a lambda case seperated by `,` like:
```
def foo[A,B,C](){ fn: (A, B) => C }: C = ...
//...

foo(){
  case true, y => ...
  case x, y => ...
}
```
## Implementation
This becomes a `Match` with multiple scrutinees, which is resolved in the pattern matching compiler.
Typer checks that the clauses have the correct number of patterns (which will be assumed later).


marzipankaiser:
Let the syntax bikeshedding begin. 🙂

effekt/shared/src/main/scala/effekt/typer/ExhaustivityChecker.scala
```diff
@@ -152,7 +160,8 @@ object ExhaustivityChecker {
     def reportNonExhaustive()(using C: ErrorReporter): Unit = {
       @tailrec
       def traceToCase(at: Trace, acc: String): String = at match {
-        case Trace.Root(_) => acc
+        case Trace.Root(_) if originalScrutinees.length == 1 => acc
+        case Trace.Root(e) => originalScrutinees.map { f => if e == f then acc else "_" }.mkString(", ")
```
jiribenes, line 164: Could you also add a neg test testing this new message?

marzipankaiser: Done.

marzipankaiser:
Additional nicety: `nonterminal` now recurses into the result to annotate positions for stuff (unless there are some already).
So this would also improve #862 for `with`-statements.

jiribenes (approved):
Wonderful, LGTM :shipit:

PR #921, opened 2025-04-01
Add 'panik' handler

jiribenes:
_Of course, here's a detailed explanation of how adding a 'panik' handler might help drive business value to the Effekt language. 😊_

Delving deeper into the topic of panicking, well, we all know that mere panicking sometimes 👏 _just_ 👏 _won't_ 👏 _do_ 👏. 🗣️
When a program in Effekt panics, it's so easy to overlook the problem!
Instead, let's use `panik`, a 21st century handler for exceptions which makes it obvious that a panic happened!
```scala
def divide(n: Int, m: Int): Int / Exception[DivisionByZero] =
  if (m == 0) {
    raise(DivisionByZero(), "Dividing by zero!")
  } else {
    n / m
  }

def main() = {
  with on[DivisionByZero].panik

  println(divide(10, 0))
}
```

Here's a production-ready demo (padded for audience retention, not because I don't know how to convert recordings to GIFs without them coming out slowed down) 
Screen Recording 2025-04-01 at 11 59 58

---

Enjoy the Assyrian New Year!


jiribenes:
> java.io.UTFDataFormatException: encoded string too long: 73877 bytes

Fun!

libraries/common/tty.effekt
```diff
@@ -136,7 +136,8 @@ def applyBackgroundColor(color: Color): String =
 /// See https://iterm2.com/documentation-images.html for protocol description
 /// See [imgcat](https://iterm2.com/utilities/imgcat) for more usage details
 def inlineImage(size: Int, base64Image: String) =
-  escape("1337;File=size=" ++ size.show ++ ";inline=1:" ++ base64Image ++ "\u0007")
+  "\u001b]" ++ ("1337;File=size=" ++ size.show ++ ";inline=1:" ++ base64Image) ++ "\u0007"
+  // TODO: Bug! EscapeOSC does not work here, why?
```
jiribenes, line 140: `escape` should never work (semantically bad, I think).
What should work is `escapeOSC` defined as `OSC := "\u001b]"; escapeOSC := OSC ++ _`.
But it doesn't [!] -- however, when I inline it (as in the current version), everything works. I think it has something to do with `OSC` being a top-level constant?
cc @b-studios

b-studios:
Closing this until next April ;)

jiribenes:
This was open mainly because of the bug mentioned above with `escape` ^

Issue #919, opened 2025-03-31
Capture can escape through variable in region


marzipankaiser:
The following program:
```effekt
effect tick(): Unit

def main() = {
  region r {
    var k in r = box { () }
    try {
      k = box { do tick() }
    } with tick { () => () }
    k() // This shouldn't be accepted
  }
}
```
is compiled successfully but shouldn't (and crashes at runtime), since it can't find a handler for `tick` (since there is none any longer)

This does not seem to be a regression from what I can tell.

marzipankaiser:
Note: Capture computation seems to be correct:

<img width="488" alt="Image" src="https://github.com/user-attachments/assets/15780cdd-1be5-497b-833b-0942b75e87e3" />

b-studios:
Trying, but failing, to create a counter example

```
effect tick(): Unit

def main() = {
  region r {    
    val g = try {
      var k in r = box { t.tick() }
      box { () => k() }
    } with t: tick { () => <> }
    g() // This shouldn't be accepted
  }
}
```

b-studios:
Strategy: just check wellformedness of the type at the region allocated `var`. This would rule out the example here.

jiribenes:
> I can't see the type of `k` when hovering over it... Is this related to the issue? 🤔

It's possible to find the type using a tmp val:
<img width="241" alt="Image" src="https://github.com/user-attachments/assets/1e4142bd-a663-420e-96e8-120e66335cb8" />

---

cc @timsueberkrueb: could we also show type on hover for variables allocated in a region?
Currently we only do it for normal variables, values, and computations, but _not_ for vars in a region!
(see screenshots below)

<img width="866" alt="Image" src="https://github.com/user-attachments/assets/72ed8b4c-6c3a-40e0-8bc2-8aebe4714602" />
<img width="862" alt="Image" src="https://github.com/user-attachments/assets/65b87f04-ba49-480d-b0b1-83aa77eef73d" />

PR #922, opened 2025-04-01
Stdlib: `tee` for `emit`-streams

marzipankaiser:
This adds an implementation of `tee`-like functions for `emit` streams.

There is:
- `teeing` which is controlled by the consumer on the outside (i.e. if the outside consumer stops, it stops the inside one too)
- `tee` which runs two consumers with one producer, stopping once both are done
- `manyTee` which allows to implement `tee` for more than 2 consumers easily

<del>Note that my initial implementation of this depended on #919. I don't think this one does, but would probably still want to hold off on merging this for either solving it or someone else being sufficiently certain it doesn't.</del>

### DIFF
--- a/libraries/common/stream.effekt
+++ b/libraries/common/stream.effekt
@@ -416,6 +416,11 @@ effect snapshot(): Unit
 /// 
 /// Starts by executing cons.
 /// 
+/// Example, printing all values consumed by the outside:
+///
+///     with teeing{ s => for{s}{ e => println(e) } }
+///     prod() // some producer
+///
 /// Laws-ish (hopefully):
 /// - hnd{ prd() } === hnd{ teeing{t}{ prd() } } for all hnd and t that calls its argument at most once (and has no other captures)
 def teeing[A]{ cons: { => Unit / emit[A] } => Unit }{ prod: => Unit / emit[A] }: Unit / emit[A] = region r {
--- a/libraries/common/stream.effekt
+++ b/libraries/common/stream.effekt
@@ -416,11 +416,11 @@ namespace internal {
 /// Use cons to handle prod, but also emit to the outside.
 /// The outside controls iteration, i.e., if cons aborts, the rest of prod will still be emitted.
 /// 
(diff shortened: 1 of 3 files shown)


libraries/common/stream.effekt
```diff
@@ -409,6 +409,74 @@ def each(string: String): Unit / emit[Char] =
...
+      teeing[A]{ {b} => hnd{b}; running = running - 1 }{prod}
+    }
+  } with emit[A] { _ =>
+    if(running > 0) resume(())
+  }
+}
+
+/// Streams prod to both cons1 and cons2. Stops once both cons1 and cons2 stopped.
+/// Only runs prod once, resuming at most once at each emit.
+def tee[A]{ cons1: { => Unit / emit[A] } => Unit }{ cons2: { => Unit / emit[A] } => Unit }{ prod: => Unit / emit[A] }: Unit = {
```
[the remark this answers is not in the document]
jiribenes: Agreed.

For completeness, here's a similar example for `teeing`:
```scala
  var sumRes = 0
  val prodRes = product {
    with teeing[Int]{{s} => sumRes = sum{s}};
    range(1, 10)
  }
  println(sumRes.show ++ ", " ++ prodRes.show)
```

libraries/common/stream.effekt
```diff
@@ -409,6 +409,86 @@ def each(string: String): Unit / emit[Char] =
 def collectString { stream: () => Unit / emit[Char] }: String =
   returning::collectString[Unit]{stream}.second
 
+effect snapshot(): Unit
```
jiribenes, line 412: Nitpick: could we hide `snapshot` in some `internal` [`teeing` ?] namespace so that it doesn't look like a part of the API?

jiribenes (approved):
LGTM :shipit: 

I don't like the names of the operations, but I really don't have better ones. 😭

libraries/common/stream.effekt
marzipankaiser: Done.

PR #870, opened 2025-03-08
WIP: Colors, paths and drawing library

marzipankaiser:
Not yet done. Basic library for paths, colors, and drawing onto canvas or into a svg.

PR #923, opened 2025-04-01
Stdlib: Binary/Byte stream utilities

marzipankaiser:
This adds a `binstream` file to the stdlib that can be used to easily create streams of bytes.
See #870 for a usage example (this was mostly factored out from there).

- [ ] Replace `bitwiseNot` as implemented here with an extern
- [ ] Move some of the general utilities into `streams`? (esp. `tracking`)

### DIFF
--- a/libraries/common/binstream.effekt
+++ b/libraries/common/binstream.effekt
@@ -24,6 +24,12 @@ record Signed[A](raw: A)
 /// Bits
 type Bit { B0(); B1() }
 
+/// not on Bits
+def not(b: Bit): Bit = b match {
+  case B0() => B1()
+  case B1() => B0()
+}
+
 /// Splices allowed in hex/byte stream literals
 effect HexSplices = {
   splice[Char], splice[String], 
@@ -89,7 +95,10 @@ def signedBytesLE(int: Int): Unit / emit[Byte] = signedBytesLE(int, 4)
 /// emit bytes of the given int as 4 bytes (in 2s-complement) in big-endian byte order
 def signedBytesBE(int: Int): Unit / emit[Byte] = signedBytesBE(int, 4)
 
+/// emit bits of the given int as 32 Bits in big-endian bit order
 def bitsBE(int: Int): Unit / emit[Bit] = bitsBE(int, 32)
+
+/// collect bits in big-endian bit order into an Int
 def collectBitsBE{ body: => Unit / emit[Bit] }: Int = {
   var res = 0
   try body() with emit[Bit] { b =>
@@ -101,11 +110,10 @@ def collectBitsBE{ body: => Unit / emit[Bit] }: Int = {
   }
   res
 }
-def not(b: Bit): Bit = b match {
-  case B0() => B1()
-  case B1() => B0()
-}
+
+/// bitwise not on integers
 def bitwiseNot(n: Int): Int = {
+  // TODO currently implemented in Effekt, should use extern ?
   collectBitsBE{
     try bitsBE(n) with emit[Bit]{ b => resume(do emit(not(b))) }
   }
@@ -114,6 +122,8 @@ def bitwiseNot(n: Int): Int = {
 // Splicers
 // --------
 
+/// Splicer to emit the bytes in hex notation given, plus evenutal splices
+/// Ignores whitespace
 def hex{ body: => Unit / { literal, HexSplices } }: Unit / emit[Byte] = {
   try {
     try {
@@ -153,6 +163,7 @@ def hex{ body: => Unit / { literal, HexSplices } }: Unit / emit[Byte] = {
   with splice[OfWidth[BE[Signed[Int]]]] { w => signedBytesBE(w.raw.raw.raw, w.width); resume(()) }
 }
 
+/// convert the given hex notation to an integer (big-endian)
 def x{ body: => Unit / { literal, HexSplices } }: Int = {
   var res = 0
   for[Byte]{ hex{body} }{ v => res = res * 256 + v.toInt }
@@ -161,8 +172,16 @@ def x{ body: => Unit / { literal, HexSplices } }: Int = {
 
 // Counting and padding
 // --------------------
+// TODO could also be part of stream library
+
+/// Request padding to a multiple of fac, by calling gen for each
 effect pad[A](fac: Int){ gen: => A }: Unit
+
+/// get current position in stream
 effect getPos(): Int
+
+/// Track position in stream, starting with init
+/// Handles getPos and pad
 def tracking[A](init: Int){ body: => Unit / { emit[A], getPos, pad[A] } }: Unit / emit[A] = {
   var n = init
   try body()
@@ -177,6 +196,9 @@ def tracking[A](init: Int){ body: => Unit / { emit[A], getPos, pad[A] } }: Unit 
     }
   }
 }
+
+/// Track how many bytes were emitted
+/// Handles getPos and pad
 def tracking[A]{ body: => Unit / { emit[A], getPos, pad[A] } }: Unit / emit[A] =
   tracking[A](0){body}
 
--- a/libraries/common/binstream.effekt
+++ b/libraries/common/binstream.effekt
@@ -95,29 +95,7 @@ def signedBytesLE(int: Int): Unit / emit[Byte] = signedBytesLE(int, 4)
 /// emit bytes of the given int as 4 bytes (in 2s-complement) in big-endian byte order
 def signedBytesBE(int: Int): Unit / emit[Byte] = signedBytesBE(int, 4)
 
-/// emit bits of the given int as 32 Bits in big-endian bit order
-def bitsBE(int: Int): Unit / emit[Bit] = bitsBE(int, 32)
 
-/// collect bits in big-endian bit order into an Int
-def collectBitsBE{ body: => Unit / emit[Bit] }: Int = {
-  var res = 0
-  try body() with emit[Bit] { b =>
-    res = b match {
-      case B0() => res * 2
-      case B1() => res * 2 + 1
-    }
-    resume(())
-  }
-  res
-}
-
-/// bitwise not on integers
-def bitwiseNot(n: Int): Int = {
-  // TODO currently implemented in Effekt, should use extern ?
-  collectBitsBE{
-    try bitsBE(n) with emit[Bit]{ b => resume(do emit(not(b))) }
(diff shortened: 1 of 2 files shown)


libraries/common/binstream.effekt
```diff
@@ -0,0 +1,472 @@
...
+/// A with explicit little-endian order
+record LE[A](raw: A)
+/// A with explicit width in current unit
+/// (bits for bitstreams, bytes for bytestreams)
+record OfWidth[A](raw: A, width: Int)
+/// explicitly signed A
+record Signed[A](raw: A)
+
+/// Bits
+type Bit { B0(); B1() }
```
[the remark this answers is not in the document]
jiribenes: cc @phischu: what's better on LLVM here?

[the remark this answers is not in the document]
phischu: On LLVM we desugar the builtin type `Bool` to a type exactly (up to alpha) like this.

marzipankaiser:
Another use case is https://github.com/effekt-community/effekt-rejit/ .

libraries/common/effekt.effekt
```diff
@@ -649,6 +665,12 @@ extern pure def bitwiseXor(x: Int, y: Int): Int =
   llvm "%z = xor %Int ${x}, ${y} ret %Int %z"
   vm "effekt::bitwiseXor(Int, Int)"
 
+extern pure def bitwiseNot(x: Int): Int =
+  js "(~${x})"
+  chez "(lognot ${x})"
+  llvm "%z = xor %Int ${x}, -1 ret %Int %z"
+  vm "effekt::bitwiseNot(Int)"
```
jiribenes, line 672: Is this actually implemented in the VM? 🤔

marzipankaiser: I don't think so, actually. I'm not sure how we handled this in other cases (I don't think all of them are implemented, are they?)

marzipankaiser:
Broken on LLVM, by Valgrind and missing generic stuff (for tests)

libraries/common/effekt.effekt
jiribenes: ... they are, I think? https://github.com/effekt-lang/effekt/blob/f175258a6800df6a760e7323ea9333a769e369d1/effekt/shared/src/main/scala/effekt/core/vm/Builtin.scala#L121-L152

marzipankaiser: Thx, it is now implemented.

libraries/common/binstream.effekt
```diff
@@ -0,0 +1,349 @@
+import option
```
jiribenes, line 1: Naming 🚲 🏠 
1. Should this be a top-level module like `binstream` or a nested one like `stream/binary` / `stream/bin`...?
2. Is it a _binary_ stream or a _bit_ stream?

libraries/common/binstream.effekt
```diff
@@ -0,0 +1,349 @@
...
+import stringbuffer
+import stream
+import char
+import test
```
jiribenes, line 5: Suspicious import 👀

libraries/common/binstream.effekt
```diff
@@ -0,0 +1,349 @@
...
+def bytes(n: Int): Unit / emit[Byte] = bytesBE(n)
+
+/// emit bytes of the given int as width bytes (in 2s-complement) in little-endian byte order
+def signedBytesLE(int: Int, width: Int): Unit / emit[Byte] = {
+  if (int < 0) {
+    bytesLE(bitwiseNot(neg(int)) + 1, width)
+  } else {
+    bytesLE(int, width)
+  }
+}
```
jiribenes, line 82: style-wise, we usually omit the outer braces if they're not needed
```suggestion
def signedBytesLE(int: Int, width: Int): Unit / emit[Byte] =
  if (int < 0) {
    bytesLE(bitwiseNot(neg(int)) + 1, width)
  } else {
    bytesLE(int, width)
  }
```

(though I don't care that much personally)

libraries/common/binstream.effekt
```diff
@@ -0,0 +1,349 @@
...
+    do emit((bitwiseAnd(n, pos * 255) / pos).toByte)
+    pos = pos / 256
+  }
+}
+
+/// emit bytes of the given int as a 4 bytes in big-endian byte order
+def bytesBE(n: Int): Unit / emit[Byte] = bytesBE(n, 4)
+
+/// emit bytes of the given int as a 4 bytes in big-endian byte order
+def bytes(n: Int): Unit / emit[Byte] = bytesBE(n)
```
jiribenes, line 73: Do we ever use this? If not, I'd consider removing it entirely and force the user to choose LE/BE manually.

libraries/common/binstream.effekt
```diff
@@ -0,0 +1,349 @@
...
+    bitwiseAnd(v, mask) match {
+      case 0 => do emit(B0())  
+      case _ => do emit(B1())
+    }
+    mask = mask / 2
+  }
+}
+
+/// emit bits of the given Byte as 8 Bits in big-endian bit order
+def bits(byte: Byte): Unit / emit[Bit] = bitsBE(byte)
```
jiribenes, line 184: same nitpick as in `bytes`: do we ever use this? wouldn't it be better to force the user to choose?

libraries/common/binstream.effekt
```diff
@@ -0,0 +1,349 @@
...
+    }
+    resume(())
+  }
+  with splice[OfWidth[BE[Int]]] { i =>
+    bitsBE(i.raw.raw, i.width)
+    resume(())
+  }
+  with splice[OfWidth[BE[Signed[Int]]]] { i =>
+    collectList[Bit]{
+      if(i.raw.raw.raw < 0){
```
jiribenes, line 341: _My kingdom for `newtype`s_? :)

examples/stdlib/binstream.effekt
```diff
@@ -0,0 +1,55 @@
...
+  suite("Binstream", false){
+    test("literal hex 10"){ assert(x"10${()}", 16) }
+    test("literal hex ff"){ assert(x"ff${()}", 255) }
+    test("literal char a"){ assert(x"${'a'}", x"61${()}") }
+    test("literal string ba"){ assert(x"${"ba"}", x"62${()}" * 256 + x"61${()}") }
+    test("int back-and-forth (17)"){ assert(x"${17}", 17)}
+    test("int back-and-forth (17), explicit BE"){ assert(x"${17.BE}", 17) }
+    test("int back-and-forth (17), explicit LE"){ assert(x"${17.LE}", 17 * 256 * 256 * 256) }
+    test("byte 00101010"){
+      with on[MissingValue].default{ assert(true, false) }
```
jiribenes, line 15: Since we're changing `test` anyway, could we add a `on[MissingValue].assertSome`?

jiribenes (approved):
LGTM, mostly just nitpicks / questions 👀 

:shipit:

libraries/common/binstream.effekt
marzipankaiser:
```suggestion
```

Yes.

libraries/common/binstream.effekt
marzipankaiser: Yes. It is somewhat nice use-wise, but this really should be something like a newtype - or us adding formatting specifiers to splices in some other form.

examples/stdlib/binstream.effekt
marzipankaiser: Yes, good idea. TODO.

libraries/common/binstream.effekt
marzipankaiser: Not really used (could be exchanged easily), so I removed it.

libraries/common/binstream.effekt
marzipankaiser: Not really used (could be exchanged easily), so I removed it.

dvdvgt:
Bump @b-studios, @marzipankaiser

libraries/common/stream.effekt
```diff
@@ -504,6 +504,61 @@ def tee[A]{ cons1: { => Unit / emit[A] } => Unit }{ cons2: { => Unit / emit[A] }
...
+def tracking[A]{ body: => Unit / { emit[A], getPos } }: Unit / emit[A] =
+  tracking[A](0){body}
+
+/// Return just the nth element of the given stream.
+/// Raises MissingValue if not enough elements are emitted
+def nth[A](n: Int){ body: => Unit / emit[A] }: A / Exception[MissingValue] = {
+  var m = n
+  try {
+    body()
+    val r: A = do raise[MissingValue](MissingValue(), "code in first did not emit any values")
```
jiribenes:
```suggestion
    val r: A = do raise[MissingValue](MissingValue(), "code in nth did not emit any values")
```

b-studios:
ping @b-studios

libraries/common/bytestream.effekt
```diff
@@ -58,10 +58,10 @@ def bytesLE(int: Int): Unit / emit[Byte] = bytesLE(int, 4)
...
 def bytesBE(n: Int, width: Int): Unit / emit[Byte] = {
-  var pos = pow(256, width - 1)
-  repeat(width){
-    do emit((bitwiseAnd(n, pos * 255) / pos).toByte)
-    pos = pos / 256
+  var shift = bitwiseShl(width - 1, 3)
+  repeat(width) {
+    do emit(bitwiseAnd(bitwiseShr(n, shift), 255).toByte)
+    shift = shift - 8
   }
```
jiribenes, line 65: fwiw, this might not be equivalent on JS: bit operations are restricted to 32bit integers, numeric are not.

Image

examples/stdlib/bytestream.effekt
```diff
@@ -12,36 +12,33 @@ def main() = {
     test("int back-and-forth (17), explicit BE"){ assert(x"${17.BE}", 17) }
     test("int back-and-forth (17), explicit LE"){ assert(x"${17.LE}", 17 * 256 * 256 * 256) }
     test("byte 00101010"){
-      with on[MissingValue].assertNotThrown
+      with onFail { do assert(false, "Unexpected Failure") }
```
marvinborner, line 15: We should really have some `assertNoFail` instead, no?

jiribenes: => https://github.com/effekt-lang/effekt/pull/1320 :)

PR #925, opened 2025-04-05
Fast exponentiation, streamingly 🌊

jiribenes:
Tiny, obviously correct & somewhat fast exponentiation using streams:
Screenshot 2025-04-05 at 16 04 07

Sketch of correctness:
```
n^k = n^0b10011
    = (1 * n^0b10000)
    * (0 * n^0b01000)
    * (0 * n^0b00100)
    * (1 * n^0b00010)
    * (1 * n^0b00001)
```

Also drags a few other changes along:
- `product` and `iterate` in `stream`
- temporary addition of `eachLSB` and `eachMSB` -- these would probably be better off using the toolchain from #923, but I don't want to forget about the example 

I'm adding it because:
1. it's a fun, small example,
2. there are some missed opt opportunities here, I think.


examples/stdlib/stream/fastexp.effekt
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  testBits(0)
+  testBits(1)
+  testBits(2)
+  testBits(3)
+  testBits(4)
+  testBits(123456789)
+
+  println("")
```
jiribenes, line 62: This part is just my sanity check for `eachLSB` and `eachMSB`. I'm happy to (re)move it after rebasing on top of #923 :)

examples/stdlib/stream/fastexp.effekt
```diff
@@ -0,0 +1,95 @@
...
+      if (exp == 0) {
+        acc
+      } else if (mod(exp, 2) == 0) {
+        go(n * n, exp / 2, acc)
+      } else {
+        go(n * n, exp / 2, acc * n)
+      }
+    }
+    go(n, exp, 1)
+  }
```
jiribenes, line 58: Taken verbatim from #923, I really ought to rebase :)

By the way, as an exercise in algorithms, notice that the `pow` here and the `fastexp` above are completely isomorphic, always doing the exact same steps :)

jiribenes:
TODO(Self):
add `digits` as an example & add related stream functions:
```scala
def takeWhile[A] { cond: A => Bool } { stream: => Unit / emit[A] }: Unit / emit[A] =
  try stream() with emit[A] {
    case value and cond(value) => do emit(value); resume(())
    case _                     => ()
  }

def last[A] { stream: => Unit / emit[A] }: A / Exception[MissingValue] = {
  var latest = None()
  for[A] {stream} { value =>
    latest = Some(value)
  }
  latest.value
}

def digits(n: Int) = last[Indexed[Int]] {
  with index[Int]
  with takeWhile { k => k <= n }
  iterate(1) { k => k * 10 }
}.index + 1

def main() = {
  with on[MissingValue].report
  println(digits(12345)) // ~> 5
}
```

libraries/common/stream.effekt
```diff
@@ -173,10 +173,23 @@ def replicate[A](number: Int) { action: () => A }: Unit / emit[A] =
...
 
+/// Creates an infinite iterated stream given by an `initial` seed and a `step` function:
+/// iterate(a){f} ~> a, f(a), f(f(a)), f(f(f(a))), ...
+def iterate[A](initial: A) { step: A => A }: Unit / emit[A] = {
+  var current = initial
+  while (true) {
+    do emit(current)
+    current = step(current)
+  }
+}
```
b-studios, line 184: Nice... This looks a lot like `unfold`ing a coalgebra  :)

PR #928, opened 2025-04-07
Add edit button for documentation on website

dvdvgt:
Sets the `origin` parameter in the font matter needed for https://github.com/effekt-lang/effekt-website/pull/86


b-studios:
Please use defaults in `_config.yml`

Issue #889, opened 2025-03-17
Passing same `--no-...` option twice breaks with bad error message


marzipankaiser:
When running effekt indirectly via a script, it can happen that options like `--no-optimize` get duplicated, e.g.:
```sh
effekt --no-optimize --no-optimize <some effect file>
```

When this happens, Effekt currently errors out with:
```output
Command-line error: Bad arguments for option 'optimize': ' ' - wrong arguments format
 Common Options
  -c, --compile                  Compile the Effekt program to the backend
                                 specific representation
      --no-compile
...
```

IMHO this should just work, or at least provide a better error message.

mlutze:
This seems to be a problem in the Scallop library. It seems that toggle options don't like to see other toggle options: https://github.com/scallop/scallop/blob/0f290269a69e22068629294c864ad2bd9a2c3ff1/src/main/scala/org.rogach.scallop/CliOptions.scala#L263-L278

Issue #534, opened 2024-07-30
Use CLI subcommands


jiribenes:
We have accumulated a lot of CLI flags, but many of them are just exclusive _commands_.
https://github.com/effekt-lang/effekt/blob/7de9eff01bf7ab87059a564dea4e7a103b7145e7/effekt/jvm/src/main/scala/effekt/EffektConfig.scala#L8-L165

Instead, we might want to move to subcommands, so something like:
```sh
effekt run <files>
effekt compile --backend=<backend> <files>
effekt lsp
... # etc
```
This would also allow us to then add other commands like `effekt fmt` for formatting and `effekt doc` for documentation generation in the future.

jiribenes:
If someone tackles this issue, it would be also nice to tackle https://github.com/effekt-lang/effekt/issues/540 as a drive-by :)

dvdvgt:
What about validating the options instead of using sub commands?
https://github.com/scallop/scallop/wiki/Arguments-validation#mutually-exclusive

I am afraid that while server, repl, build, compile are all mutually exclusive, they all share (more or less) the same options. and by using sub commands, these options would not be directly visible to the outside without a nice way to query them.
https://github.com/scallop/scallop/wiki/Arguments-validation#mutually-exclusive

jiribenes:
After we do this:
- we might need to update the VSCode extension,
- and it would also be nice to be able to generate a shell completion file (which gets applied on installation).

PR #930, opened 2025-04-08
Generate documentation

marvinborner:
- annotate doc comments in Tree
- add flags `show-documentation` and `write-documentation` for generating JSONs of documented nodes
  - these JSONs can be later used to generate API documentation for the website


marvinborner:
This is ready for review now.

Toplevel terms and selected 2nd level terms (such as operations or constructors in data/interface) now have a `Doc` attached. If documentation comments occur at invalid positions such as individual statements, the parser will crash (not a bug!)

We discussed my style of parsing doc comments, and I'm unsure what we've settled on. We may want to use the technique of @timsueberkrueb in #957 or the Scala 3 syntax. I think the current version looks fine though.

You can generate a JSON representation of files using `effekt.sh --show-documentation <file> .. <file>` or `--write-documentation` respectively. This JSON representation includes the top-level definitions as well as its relevant* fields. Importantly this includes definitions without documentation comments, since we don't use them everywhere yet.

*we may want to change/discuss the specifics in the future. For now this includes almost all non-statement/term fields of definitions.

Further work such as the frontend on which I worked in parallel will be merged in subsequent PRs (and in a different repository, most probably)

Also, should I add tests for documentation generation? (and where?)

marzipankaiser:
>  If documentation comments occur at invalid positions such as individual statements, the parser will crash (not a bug!)

Maybe adding a neg-Test for this would be useful (esp. since this also tests that doc-comments are at least detected in this case)?

effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
```diff
@@ -542,7 +542,7 @@ class RecursiveDescentTests extends munit.FunSuite {
       parseDefinition(
         """def foo = f
           |""".stripMargin),
-      DefDef(IdDef("foo"), None, Var(IdRef(Nil, "f"))))
+      DefDef(IdDef("foo"), None, Var(IdRef(Nil, "f")), None))
```
marzipankaiser, line 546: Maybe adding a similar test here, with an actual doc comment, would be an option?

effekt/jvm/src/main/scala/effekt/EffektConfig.scala
```diff
@@ -148,6 +148,24 @@ class EffektConfig(args: Seq[String]) extends REPLConfig(args.takeWhile(_ != "--
...
+  )
+
+  val writeDocumentation: ScallopOption[Boolean] = toggle(
+    "write-documentation",
+    descrYes = "Write all documented statements to a JSON in the output directory",
+    default = Some(false),
+    noshort = true,
+    prefix = "no-",
+    group = debugging
+  )
```
b-studios, line 167: What do we need the showDocumentation option for?

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -466,53 +484,62 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
+        case `interface` => externInterface(doc)
+        case `resource`  => externResource(doc)
+        case `include`   => externInclude(doc)
         // extern """..."""
-        case s: Str      => externString()
+        case s: Str      => externString(doc)
         case Ident(_) | `pure` =>
           // extern IDENT def ...
-          if (peek(2, `def`)) externFun()
+          if (peek(1, `def`)) externFun(doc)
```
b-studios, line 538: Why do we need to consume the extern here upfront? This deviates from the usual style that nonterminals consume their tokens.

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -580,6 +606,26 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
+    p(maybeDocumentation())
+
+  def maybeDocumentation(): Doc =
+    nonterminal:
+      peek.kind match
+        case DocComment(msg) =>
+          consume(peek.kind)
+          val next = maybeDocumentation()
+          next.map { d => msg ++ "\\n" ++ d }
+            .orElse(Some(msg))
```
b-studios, line 619: Again, I am afraid that computing this every time (even if we do not generate docs) might be slow. Also this should really be a tight loop using a string buffer (like `infix`).

effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -100,6 +100,8 @@ case object NoSource extends Tree
 // only used by the lexer
 case class Comment() extends Tree
 
+type Doc = Option[String]
```
b-studios, line 103: In spirit of speeding this up, this could be Option[Span].

effekt/shared/src/main/scala/effekt/util/DocumentationGenerator.scala
```diff
@@ -0,0 +1,286 @@
...
+// TODO: include spans
+trait DocumentationGenerator {
+  type Documentation = HashMap[String, DocValue]
+  type EffektTree[T <: Tree] = kiama.relation.Tree[AnyRef & Product, T]
+
+  // A recursive structure that resembles JSON
+  sealed trait DocValue
+  case class DocString(value: String) extends DocValue
+  case class DocArray(value: List[DocValue]) extends DocValue
+  case class DocObject(value: Documentation) extends DocValue
```
b-studios, line 16: Any reason not to use an `enum` here?

effekt/shared/src/main/scala/effekt/util/DocumentationGenerator.scala
```diff
@@ -0,0 +1,286 @@
+package effekt.util
+
+import effekt.source.*
+import kiama.util.Source
+import scala.collection.immutable.HashMap
+
+// TODO: include spans
+trait DocumentationGenerator {
+  type Documentation = HashMap[String, DocValue]
+  type EffektTree[T <: Tree] = kiama.relation.Tree[AnyRef & Product, T]
```
b-studios, line 10: We really wanted to get rid of this dependency :(

effekt/shared/src/main/scala/effekt/util/DocumentationGenerator.scala
```diff
@@ -0,0 +1,286 @@
...
+  case class DocString(value: String) extends DocValue
+  case class DocArray(value: List[DocValue]) extends DocValue
+  case class DocObject(value: Documentation) extends DocValue
+
+  def empty = DocObject(HashMap.empty)
+  def str(s: String) = DocString(s)
+  def arr(docs: List[DocValue]) = DocArray(docs)
+  def obj(doc: Documentation) = DocObject(doc)
+
+  def generate(doc: Doc): DocString = str(s"${doc.getOrElse("").replace("\"", "\\\"")}")
```
b-studios, line 23: Any reason to wrap this in `s"..."`?

b-studios (commented):
Besides the comments, this looks good, thanks!

I wonder, though, whether `documentedNonterminal` in parser with a `doc()` effect, similar to `spans` might simplify things (don't do it now).

I know that the "peek modulo whitespace and doc comments" is really annoying, but I would like to keep the style that nonterminals consume their tokens. Do you think this is possible for `extern def`s?

effekt/shared/src/main/scala/effekt/util/DocumentationGenerator.scala
marvinborner: Yep, already removed it. I just forgot to remove the type definition

effekt/shared/src/main/scala/effekt/util/DocumentationGenerator.scala
marvinborner: nope!

effekt/shared/src/main/scala/effekt/util/DocumentationGenerator.scala
marvinborner: nope!

marvinborner:
I think the manual merge succeeded, maybe @timsueberkrueb wants to look over it though

I can't reproduce the CI error, somehow the formatting is different on actions

timsueberkrueb:
> I can't reproduce the CI error, somehow the formatting is different on actions

Yes, there seems to be a non-determinism in the generated identifiers between execution environments. When they get longer/shorter pretty-printing changes.

jiribenes:
Do we have to use String equality there? What are we testing for in the (currently) failing test?

effekt/jvm/src/main/scala/effekt/EffektConfig.scala
marvinborner: Not sure, it could be used for piping the JSON directly to `jq` or some other program. Mostly it's there because of the similarities to the `showIR` and `writeIR` flags. Should I remove it?

b-studios: We already talked about subcommands at some point

jiribenes: (https://github.com/effekt-lang/effekt/issues/534 is the CLI subcommands issue, but the CLI parsing library we use doesn't really support it + it's buggy https://github.com/effekt-lang/effekt/issues/889 + we use the config in a very unstructured and ad-hoc way right now)

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
marvinborner: I thought it makes more sense this way, other definitions above (e.g. `namespaceDef`) are similar in that we first consume the documentation, then consume some identifier and then continue based on the next symbols. Otherwise we need to add a separate `maybeDocumentation()` call to all `extern...` parsers.

I see the resulting deviation in `extern...` though, so should I change it?

effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -544,7 +546,13 @@ class Lexer(source: Source) {
     }
     // exclude //
     val comment = slice(start + 2, current)
-    TokenKind.Comment(comment)
+    // TokenKind.Comment(comment)
+    if comment.startsWith("/") then
+      // exclude ///
+      val docComment = slice(start + 3, current)
+      TokenKind.DocComment(docComment)
```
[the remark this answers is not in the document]
marvinborner: Yes it's a single comment line. I don't really see a problem with slicing here, to be honest. We sliced all doc comments once before this change, I don't think slicing twice changes anything.

effekt/jvm/src/main/scala/effekt/Driver.scala
```diff
@@ -126,6 +127,25 @@ trait Driver extends kiama.util.Compiler[EffektConfig, EffektError] { outer =>
       }
   }
 
+  def generateDocumentation(ast: ModuleDecl, source: Source)(implicit C: Context): String =
```
b-studios, line 130:
```suggestion
  def generateDocumentation(ast: ModuleDecl, source: Source)(using C: Context): String =
```

effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
```diff
@@ -837,7 +848,7 @@ class RecursiveDescentTests extends munit.FunSuite {
 
       assertEquals(
         parseDefinition(source.content),
-        DefDef(IdDef("foo", Span(source, pos(0), pos(1))), None, Var(IdRef(Nil, "f", Span(source, pos(2), pos(3))))))
+        DefDef(IdDef("foo", Span(source, pos(0), pos(1))), None, Var(IdRef(Nil, "f", Span(source, pos(2), pos(3)))), None, Span(source, 0, 11)))
```
timsueberkrueb, line 851: Consider using the DSL instead (`pos`) instead.

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -380,30 +388,34 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
-          case AnyPattern(id) ~ Nil =>
+        maybeDocumentation() ~ (`val` ~> matchPattern()) ~ manyWhile(`and` ~> matchGuard(), `and`) <~ `=` match {
+          case doc ~ AnyPattern(id) ~ Nil =>
             val binding = stmt()
-            val valDef = ValDef(id, None, binding).withRangeOf(startMarker, binding)
+            val valDef = ValDef(id, None, binding, doc, span()).withRangeOf(startMarker, binding)
             DefStmt(valDef, { semi(); stmts() })
-          case p ~ guards =>
+          case doc ~ p ~ guards =>
+            // TODO: use doc
```
timsueberkrueb, line 418: TODO

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -380,30 +388,34 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
     nonterminal:
+      val doc = maybeDocumentation()
       val startMarker = nonterminal { new {} }
       def simpleLhs() = backtrack {
         `val` ~> idDef() ~ maybeValueTypeAnnotation() <~ `=`
       } map {
         case id ~ tpe =>
           val binding = stmt()
-          val valDef = ValDef(id, tpe, binding).withRangeOf(startMarker, binding)
+          val valDef = ValDef(id, tpe, binding, doc, span()).withRangeOf(startMarker, binding)
```
timsueberkrueb, line 408: I'm not sure about the spans here, notice the explicit `withRangeOf(startMarker, binding)`. Generally `span()` is always the span of the beginning of the `nonterminal` until the current position. Whenever this is not the span you want, you need to manually create a `Span`.

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -468,53 +486,63 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
 
   // effect <NAME>[...](...): ...
   def operationDef(): Def =
     nonterminal:
-      `effect` ~> operation() match {
-        case op @ Operation(id, tps, vps, bps, ret) =>
-          InterfaceDef(IdDef(id.name, id.span) withPositionOf op, tps, List(Operation(id, Many.empty(tps.span.synthesized), vps, bps, ret) withPositionOf op))
+      val doc = maybeDocumentation()
+      `effect` ~> operation(doc) match {
+        // TODO: which doc/span should be used when and where?
```
timsueberkrueb, line 498: TODO

effekt/shared/src/main/scala/effekt/util/DocumentationGenerator.scala
```diff
@@ -0,0 +1,332 @@
...
+  }
+  import DocValue.*
+
+  def empty = DocObject(HashMap.empty)
+  def num(n: Int) = DocNumber(n)
+  def str(s: String) = DocString(s)
+  def arr(docs: List[DocValue]) = DocArray(docs)
+  def obj(doc: Documentation) = DocObject(doc)
+
+  def generate(doc: Doc): DocValue = str(doc.getOrElse("").replace("\"", "\\\""))
```
timsueberkrueb, line 26: Is this still needed?

timsueberkrueb:
Great work :+1: 

There's one more TODO above for spans in a desugaring in the parser. I think the only way to guarantee they are correct is to create a test case. Let me know if I can help with that!
Otherwise it looks good to me, though I cannot guarantee that the spans not covered by the tests are correct.
I would write a test case for the TODO and then merge this.

timsueberkrueb (approved):

marvinborner:
I've just added a test and checked the implementation again. I think it is correct as is, although it's a bit unclear since an operation is syntactic sugar for an interface and therefore shares some of its fields and spans.

Other than that, let's merge! (?)

jiribenes:
@marvinborner: sorry, I merged the parser labels, it might need one last rebase :sweat_smile:

marvinborner:
Weirdly enough, git really struggled with this one, had to fix many conflicts manually. Hope it went well!

PR #931, opened 2025-04-09
JSWeb: report all collected errors upon aborting

dvdvgt:
Currently, when encountering an error (e.g. through calling `Context.abort`) in the JSWeb backend, all encountered errors are discarded and a JS exception is thrown. All information is lost there, which leads to a singular (unhelpful) error message `Cannot compile interactive.effekt`. It would be more helpful for the user to see the cause.

https://github.com/effekt-lang/effekt/blob/8abbff5c34d71f8d3396c9db4b75ad6fdb16a4f5/effekt/js/src/main/scala/effekt/LanguageServer.scala#L101-L112

In line 103 this exception is thrown. Notice that the following line attempt to catch a `FatalPhaseError`, however `mainOutputPath` is a string and thus not lazily evaluated. Furthermore, `Phase.apply` already catches `FatalPhaseError`s and converts them to `None`. Hence, the exception handler is never executed.

Instead, we report all collected error messages.

Before:
image

After:
image

(the increased resolution is not part of this PR)

### DIFF
--- a/effekt/js/src/main/scala/effekt/LanguageServer.scala
+++ b/effekt/js/src/main/scala/effekt/LanguageServer.scala
@@ -98,18 +98,12 @@ class LanguageServer extends Intelligence {
     file(path).lastModified
 
   @JSExport
-  def compileFile(path: String): String = {
-    val mainOutputPath = compileCached(VirtualFileSource(path)).getOrElseAborting {
-      throw js.JavaScriptException(s"Cannot compile ${path}")
+  def compileFile(path: String): String =
+    compileCached(VirtualFileSource(path)).getOrElseAborting {
+      // report all collected error messages
+      val formattedErrors = context.messaging.formatMessages(context.messaging.get)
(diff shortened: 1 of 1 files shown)


dvdvgt:
Note this causes some implementation details to leak, which is perhaps confusing to the user:

image

Though, I think the advantages of having a concrete error message outweigh this shortcoming.

b-studios:
It is "just" the positions that are confusing, right? I think this is fine

jiribenes:
Note/rant:

> in the JSWeb backend

We should figure out a better way to call whatever-we-do-on-the-Effekt-website to distinguish it from `--backend=js-web`, since we're calling both "JSWeb" way too often. 😅

jiribenes (approved):
LGTM, definitely better than what was there before.
The only possible improvement would be to deduplicate the error messages, but I think it's fine to push that to future work :)

PR #937, opened 2025-04-16
Remove debug print in Normalizer

jiribenes:
:)


jiribenes:
In profiling, removing this line saves about 5/25 seconds :)

PR #938, opened 2025-04-16
Simplify core.Renamer to produce smaller symbols

jiribenes:
General idea: instead of building `name_123_109238_190238_109283019283_1092830192838019283` chains, what if we use the automagic `fresh.next` to get just a darn fresh number?
This will limit the identifiers to stuff like `b_k2_123_10928310283`, which is--at least to me--a pretty big improvement.

We use the full capabilities of a `Symbol`: it has a `name` (which is really a `String`) and a fresh `id` that gets assigned lazily. Instead of changing the name to be `${oldName}_${oldId}` and then putting a fresh `id` on top of that (to get `${oldName}_${oldId}_${newId}`), we just make a new `Symbol` with the same name, which forces a new, _fresher_, `id`.

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/core/RenamerTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/core/RenamerTests.scala
@@ -5,15 +5,16 @@ package effekt.core
  */
 class RenamerTests extends CoreTests {
 
-  def assertRenamedTo(input: String,
-                      renamed: String,
-                      clue: => Any = "Not renamed to given value",
-                      names: Names = Names(defaultNames))(using munit.Location) = {
+  /**
+   * Check that the renamed input preserves alpha-equivalence using [[assertAlphaEquivalent]]
+   */
+  def assertRenamingPreservesAlpha(input: String,
+                                   clue: => Any = "Not renamed to given value",
+                                   names: Names = Names(defaultNames))(using munit.Location) = {
     val pInput = parse(input, "input", names)
-    val pExpected = parse(renamed, "expected", names)
     val renamer = new Renamer(names, "renamed")
     val obtained = renamer(pInput)
-    assertAlphaEquivalent(obtained, pExpected, clue)
+    assertAlphaEquivalent(obtained, pInput, clue)
   }
 
   test("No bound local variables"){
@@ -24,11 +25,11 @@ class RenamerTests extends CoreTests {
         |  return (bar: (Int) => Int @ {})(baz:Int)
         |}
         |""".stripMargin
-    assertRenamedTo(code, code)
+    assertRenamingPreservesAlpha(code)
   }
 
   test("val binding"){
-    val input =
+    val code =
       """module main
         |
         |def foo = { () =>
@@ -36,19 +37,11 @@ class RenamerTests extends CoreTests {
         |  return x:Int
         |}
         |""".stripMargin
-    val expected =
-      """module main
-        |
-        |def foo = { () =>
-        |  val renamed1 = (foo:(Int)=>Int@{})(4);
-        |  return renamed1:Int
-        |}
-        |""".stripMargin
-    assertRenamedTo(input, expected)
+    assertRenamingPreservesAlpha(code)
   }
 
   test("var binding"){
-    val input =
+    val code =
       """module main
         |
         |def foo = { () =>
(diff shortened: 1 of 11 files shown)


effekt/shared/src/main/scala/effekt/symbols/Symbol.scala
```diff
@@ -38,7 +38,7 @@ trait Symbol {
     case _             => false
   }
 
-  def show: String = name.toString + id
+  def show: String = s"${name}_${id}"
```
jiribenes, line 41: This is somewhat necessary, some identifiers from pattern matching have `name := "v_r_11029323"` already, so if we were to just append the `id := 42`, we'd get a much longer number like `v_r_1102932342`.
Instead, we now get `v_r_11029323_42`.
The cost is that now instead of having `Char342` and `List45`, we have `Char_342` and `List_45`...

Alternative designs:
1. choose a different symbol to append like `@` (since it cannot appear in an identifier anyways) and then sanitise for backends
2. change how all of those `v_r_...` get generated.

jiribenes: I'll try to do 2., since I think it's viable.

jiribenes:
I don't quite understand the test failures, the outputs seem identical to me both in console and in IntelliJ...

effekt/shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -211,8 +211,8 @@ case class CallTarget(symbols: List[Set[BlockSymbol]]) extends BlockSymbol { val
  * Introduced by Transformer
  */
 case class Wildcard() extends ValueSymbol { val name = Name.local("_") }
-case class TmpValue(hint: String = "tmp") extends ValueSymbol { val name = Name.local("v_" + hint + "_" + Symbol.fresh.next()) }
-case class TmpBlock(hint: String = "tmp") extends BlockSymbol { val name = Name.local("b_" + hint + "_" + Symbol.fresh.next()) }
+case class TmpValue(hint: String = "tmp") extends ValueSymbol { val name = Name.local("v_" + hint) }
+case class TmpBlock(hint: String = "tmp") extends BlockSymbol { val name = Name.local("b_" + hint) }
```
jiribenes, line 215: To explain the change, when `show`ing a `Symbol`, we _already_ lazily give a fresh `id` consisting of `Symbol.fresh.next()` to an identifier, so there's no need for this.

effekt/jvm/src/test/scala/effekt/core/RenamerTests.scala
```diff
@@ -9,9 +11,9 @@ class RenamerTests extends CoreTests {
                       names: Names = Names(defaultNames))(using munit.Location) = {
     val pInput = parse(input, "input", names)
     val pExpected = parse(renamed, "expected", names)
-    val renamer = new Renamer(names, "renamed") // use "renamed" as prefix so we can refer to it
+    val renamer = new Renamer(names, "renamed")
     val obtained = renamer(pInput)
-    shouldBeEqual(obtained, pExpected, clue)
+    assertAlphaEquivalent(obtained, pExpected, clue)
   }
```
marzipankaiser, line 17: Given that we check for alpha-equivalence anyway, we could drop one argument and not have expected, couldn't we? 🤔

(i.e. just have the input in each case)

effekt/jvm/src/test/scala/effekt/core/TestRenamer.scala
```diff
@@ -0,0 +1,120 @@
+package effekt.core
+
+import effekt.{ core, symbols }
+import effekt.context.Context
+
+/**
+ * Freshens bound names in a given term for tests.
+ * Please use this _only_ for tests.
```
marzipankaiser, line 8: Eventually, we could think about optimizing this renamer for tests slightly. E.g., choose indices that make diffs more readable (e.g. based on levels like the original one).
But fine for now :)

effekt/shared/src/main/scala/effekt/core/Renamer.scala
```diff
@@ -4,7 +4,8 @@ import effekt.{ core, symbols }
 import effekt.context.Context
 
 /**
- * Freshens bound names in a given term
+ * Freshens bound names in a given Core term.
+ * Do not use for tests!
```
marzipankaiser, line 8:
```suggestion
 * Freshens bound names in a given Core term.
 * Do not use for tests! See [[effekt.core.TestRenamer]].
```
Maybe make this more actionable.

effekt/jvm/src/test/scala/effekt/core/TestRenamer.scala
marzipankaiser:
```suggestion
 * Freshens bound names in a given term for tests.
 * Please use this _only_ for tests. Otherwise, prefer [[effekt.core.Renamer]].
```
Maybe make this more actionable.

effekt/shared/src/main/scala/effekt/symbols/Symbol.scala
jiribenes: I think having this distinction explicitly (always counting on `${name}_${id}`) is helpful, so I'll keep it :)

marzipankaiser: I would keep the `_`. I don't think it adds too much clutter, and it *guarantees* that we do not clash with some existing name (e.g. `Int12` and `Int1` renamed to `Int12` by this or sth), even if those are unlikely.

effekt/jvm/src/test/scala/effekt/core/RenamerTests.scala
jiribenes: Applied, thanks!

PR #940, opened 2025-04-17
Make 'core.Renamer' quicker by simplifying scope handling

jiribenes:
I don't know what the nested list is for, but we don't seem to use it anywhere. Therefore I'm removing the nesting and using mutable hashmaps: it's on a hot path!

Renamer then goes from 2.4-3.4s to 0.6-0.4s :) [measured a few times via profiling]
(for reference, before #938, it was somewhere between 3 and 4 seconds)

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/core/TestRenamer.scala
+++ b/effekt/jvm/src/test/scala/effekt/core/TestRenamer.scala
@@ -110,11 +110,3 @@ class TestRenamer(names: Names = Names(Map.empty), prefix: String = "") extends 
     rewrite(s)
   }
 }
-
-object TestRenamer {
-  def rename(b: Block): Block = Renamer().rewrite(b)
-  def rename(b: BlockLit): (BlockLit, Map[Id, Id]) =
-    val renamer = Renamer()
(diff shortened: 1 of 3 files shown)


effekt/jvm/src/test/scala/effekt/core/TestRenamer.scala
```diff
@@ -109,12 +109,4 @@ class TestRenamer(names: Names = Names(Map.empty), prefix: String = "") extends
...
-}
-
-object TestRenamer {
-  def rename(b: Block): Block = Renamer().rewrite(b)
-  def rename(b: BlockLit): (BlockLit, Map[Id, Id]) =
-    val renamer = Renamer()
-    val res = renamer.rewrite(b)
-    (res, renamer.renamed)
-}
+}
```
jiribenes, line 112: Drive-by: this is wrong (should refer to `TestRenamer`, not to `Renamer`), _and_ it's not used => yeet.

marzipankaiser (approved):
LGTM

jiribenes:
Thanks!
I'll only merge this in the morning, because I'm interested in seeing the diff in Effekt plots.

PR #941, opened 2025-04-17
Renamer: Add infrastructure for Checking Uniqueness of Generated Names

marzipankaiser:
After #938, the `RenamerTests` do not check that the names are actually unique.
This adds assertions checking that every `Id` only occurs in one definition-like construct.

(note that https://github.com/effekt-lang/effekt/pull/941/commits/4cf23710d2c4be2202a73d888e19d4df1ddf1645 is intentionally wrong to show it failing on shadowing :) )

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/core/RenamerTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/core/RenamerTests.scala
@@ -63,7 +63,7 @@ class RenamerTests extends CoreTests {
         }
       }
     }
-    check.query(pInput)
+    check.query(obtained)
   }
 
   test("No bound local variables"){
--- a/effekt/jvm/src/test/scala/effekt/core/RenamerTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/core/RenamerTests.scala
@@ -31,7 +31,6 @@ class RenamerTests extends CoreTests {
       assert(!seen.contains(id), clue)
       seen.add(id)
     }
-    given Unit = ()
     object check extends Tree.Query[Unit, Unit] {
       override def empty = ()
       override def combine = (_,_) => ()
@@ -63,7 +62,7 @@ class RenamerTests extends CoreTests {
         }
       }
     }
-    check.query(obtained)
+    check.query(obtained)(using ())
   }
 
   test("No bound local variables"){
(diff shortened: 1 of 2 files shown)


effekt/jvm/src/test/scala/effekt/core/RenamerTests.scala
```diff
@@ -17,6 +19,52 @@ class RenamerTests extends CoreTests {
...
+
+    val seen = mutable.HashSet.empty[Id]
+    def isFresh(id: Id): Unit = {
+      assert(!seen.contains(id), clue)
+      seen.add(id)
+    }
+    object check extends Tree.Query[Unit, Unit] {
+      override def empty = ()
+      override def combine = (_,_) => ()
+      override def visit[T](t: T)(visitor: Unit ?=> T => Unit)(using Unit): Unit = {
```
jiribenes, line 37: :)

jiribenes (approved):
Looks OK, saw it fail (as in, detect a program breaking this invariant), :shipit:

effekt/jvm/src/test/scala/effekt/core/RenamerTests.scala
```diff
@@ -17,6 +19,52 @@ class RenamerTests extends CoreTests {
...
+  def assertRenamingMakesDefsUnique(input: String,
+                                    clue: => Any = "Duplicate definition",
+                                    names: Names = Names(defaultNames))(using munit.Location) = {
+    val pInput = parse(input, "input", names)
+    val renamer = new Renamer(names, "renamed")
+    val obtained = renamer(pInput)
+
+    val seen = mutable.HashSet.empty[Id]
+    def isFresh(id: Id): Unit = {
+      assert(!seen.contains(id), clue)
```
jiribenes, line 31: Something like the following might help with the error message, I guess?
```suggestion
                                    clue: Id => Any = id => s"Duplicate definition of ${id}",
                                    names: Names = Names(defaultNames))(using munit.Location) = {
    val pInput = parse(input, "input", names)
    val renamer = new Renamer(names, "renamed")
    val obtained = renamer(pInput)

    val seen = mutable.HashSet.empty[Id]
    def isFresh(id: Id): Unit = {
      assert(!seen.contains(id), clue(id))
```

PR #943, opened 2025-04-17
Experiment: inline phase operators

jiribenes:
It's hard to say from inside the profiler, but it feels like this negates all of the costs of the `Phase` abstraction.

I'd need somebody to actually measure this well (ideally each commit separately* vs `master`), the values I get aren't really distinguishable from noise.

---

*The very first commit seems pretty good, not sure about the later ones.


b-studios:
I am surprised this makes a difference at all. I would have expected each individual phase to be much longer than calling a phase or composing it

jiribenes:
I think this might be fun to pursue, but perhaps another time. Feel free to resurrect the PR if you want to play around with this.

PR #945, opened 2025-04-17
Add tailrec annotations wherever possible

jiribenes:
IntelliJ has a button for adding `@tailrec` wherever needed.
So I clicked it. Let's see what happens now!

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/util/PathUtils.scala
+++ b/effekt/jvm/src/main/scala/effekt/util/PathUtils.scala
@@ -3,6 +3,7 @@ package util
 
 import java.io.{ File => JFile }
 import java.net.URI
+import scala.annotation.tailrec
 
 object paths extends PathUtils {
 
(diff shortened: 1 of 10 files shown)


jiribenes:
Turns out that the tailrec heuristic of IntelliJ doesn't agree with the tailrec heuristic of the Scala 3 compiler. _What fun!_

effekt/jvm/src/main/scala/effekt/util/PathUtils.scala
```diff
@@ -44,7 +44,7 @@ object paths extends PathUtils {
   implicit def file(uri: URI): File = file(uri.getPath)
   implicit def file(filepath: String) =
     if (filepath.startsWith("file:"))
-      file(new URI(filepath).getPath)
+      file(new URI(filepath))
     else
       file(new JFile(filepath))
```
jiribenes, line 49: So originally IntelliJ inferred a `tailrec` here, but that seems wrong? I'd assume we want to use the overload above, as opposed to recursing forever?

jiribenes:
I don't think anybody has anything against this PR, so I'm merging :)

jiribenes:
This change seems to have both made the compiler faster and use less memory 🥳

PR #946, opened 2025-04-18
LLVM: Emit most declared definitions as private

mattisboeckle:
I was experimenting with making emitted function definitions `private` by default in llvm.

This seemed to only reduce the code size of our *.opt.ll files, which might not even be wanted, because now all the unused/inlined function definitions are gone.

Benchmarking this showed no improvement *except* for "examples/benchmarks/are_we_fast_yet/queens.effekt" which unexpectedly got 30% faster. Not sure what is going on


mattisboeckle:
Ran all the benchmarks out of curiosity and found another outlier "effect_handlers_bench/parsing_dollars" is also 30% faster

PR #947, opened 2025-04-18
Kindless on the Surface, but kind(ed) in the Core, symbolically: Better, recoverable error messages for wrong kinds

### LINKED ISSUE #667: Confusing parser error messages on wrong kind

jiribenes:
Resolves #667 (see there for context)

Also see #949 which constituted the second part of this.

Types in surface.Tree are now kindless, not distinguished by Value/BlockType (we keep these just for docs).
This means that the parser for types is now exceedingly liberal, yet unkind(ed).

The Namer then resolves surface.Tree.Type to symbol.Value/BlockType.
The only changes there are explicit `resolveValue/BlockType` operators.

---

TODO: 
- [x] review/bikeshed new error messages
- [x] neg tests for new error messages
- [x] fix TODO comments
- [x] clean-up


effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1103,75 +1099,197 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
+  def blockType(): BlockType = guardedType(ignoreEffectSets = false, _.asBlockType) {
+    case MyType.TypeBox(tpe, captureSet) => s"Expected block type, but got boxed type ${box}. Did you mean to write just ${tpe} *without* 'at ${captureSet}'?"
+    case tpe                             => s"Expected block type, but got value type ${tpe}."
+  }
+
+  def valueType(ignoreEffectSets: Boolean): ValueType = guardedType(ignoreEffectSets, _.asValueType) {
+    case fun: MyType.TypeFun => s"Expected value type, but got function type ${fun}. Did you mean to box it, i.e., ${fun} at {}?"
+    case tpe => s"Expected value type, but got block type ${tpe}."
+  }
+  def valueType(): ValueType = valueType(ignoreEffectSets = false)
```
jiribenes, line 1292: Now it's very visible that this is the job of the Namer phase :^)

jiribenes:
Error message comparison (modulo pretty-printing, too lazy for that now):
```scala
record Fun(f: (Int, Int) => Bool)
// [BEFORE]              ^^
// Expected ) but got =>
//
// [NOW]
// Expected value type, but got function type (Int, Int) => Bool.
// Did you mean to box it, i.e., (Int, Int) => Bool at {}?

def foo { f: (Int, Int) => Bool at {} } = <>
// [BEFORE]                ^^^^
// Expected an interface type.
//
// [NOW]
// Expected block type, but got boxed type (Int, Int) => Bool at {}.
// Did you mean to write just (Int, Int) => Bool *without* 'at {}'?

def bar(f: (Int, Int) => Bool) = <>
// [BEFORE]           ^^
// Expected ) but got =>
//
// [NOW]
// Expected value type, but got function type (Int, Int) => Bool.
// Did you mean to box it, i.e., (Int, Int) => Bool at {}?
```

jiribenes:
Three failing tests:
1. missing pretty-printing
2. & 3. loss of position somewhere [?]

jiribenes:
Quick benchmarks:
- On `casestudies/prettyprinter`, the new parser is 2% slower.
- On `benchmarks/other/unify`, the new parser is 1% slower.
- On my `algj` file, the new parser is like 0.5% slower.

jiribenes:
Screenshot 2025-04-19 at 13 36 28
Now the previous code reports all of the bad types _at the same time_, at great cost of super-crappy encoding.
This too will go away if we move this thing to Namer.

examples/neg/parsing/blockargs.check
```diff
@@ -1,3 +1,4 @@
-[error] examples/neg/parsing/blockargs.effekt:3:20: Effects not allowed here. Maybe you mean to use a function type `() => T / E`?
+[error] examples/neg/parsing/blockargs.effekt:3:18: Expected block type, but got effectful type TypeEff(TypeRef(IdRef(List(),R),List()),List(TypeRef(IdRef(List(),Exc),List()))).
+Did you mean to write a function type () => TypeEff(TypeRef(IdRef(List(),R),List()),List(TypeRef(IdRef(List(),Exc),List())))?
 def foo[R] { f : R / Exc }: R = <>
-                   ^
+                 ^^^^^^^^
```
jiribenes, line 4: Yes, this needs a pretty-printer. No, I'm not going to write one yet. :)

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1100,78 +1096,251 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
-        maybeTypeParams() ~ maybeValueTypes() ~ (maybeBlockTypeParams() <~ `=>`) ~ valueType2(boxedAllowed) ~ maybeEffects() match {
-          case tparams ~ vparams ~ bparams ~ t ~ effs => FunctionType(tparams, vparams, bparams, t, effs)
-        }
+      if (peek(`/`)) {
+        TypeEff(boxed, maybeEffects())
+      } else {
+        boxed
       }
-      def parenthesized = backtrack { parens { blockType() } }
+  }
```
jiribenes, line 1286: Technically both of the `at {...}` parsing and `/ ...` parsing should be while loops / recursive functions to catch the most errors possible like `foo at {...} at {...} at {...}`, but I'm somewhat tired of parsing shenanigans, so we can keep it for a future issue. :)

jiribenes:
cc @timsueberkrueb: the LSP test "inlayHints should show the io effect" (should be "the io _capture_") seems brittle: fails on CI, but the tests succeed locally for me _and_ it also works in VSCode. [Here's the transcript from CI.](https://github.com/effekt-lang/effekt/actions/runs/14549539220/job/40819067619#step:4:876)
I'll rerun the CI to see if that was just some space ray bit flip ;)

examples/neg/parsing/blockargs.check
jiribenes: Turns out that we don't really need a pretty-printer. We just parsed something, have the source position... how about we yeet back the given source at the user? :)

effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -697,8 +701,19 @@ object Resolvable {
...
+  // XXX this should maybe be a method on `Source`?
+  def slice(range: kiama.util.Range): Option[String] = for {
+    start <- range.from.optOffset
+    end <- range.to.optOffset
+  } yield range.from.source.content.substring(start, end)
+
+  C.positions.getRange(positioned).flatMap(slice)
+}
+extension [T](positioned: T) def sourceOf(using C: Context): String =
+  positioned.sourceOfOpt.getOrElse { s"${positioned}" }
```
jiribenes, line 716: How do we legalise this? I really want this functionality: where does it belong?

effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -666,29 +672,27 @@ object Named {
 }
 export Named.symbol
 
+// TODO(jiribenes, 2024-04-21): Which of these do we need?
+// Can we actually move this to namer?
+
 // MOVE TO NAMER
 object Resolvable {
```
jiribenes, line 679: Can we actually move this to Namer?

effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -525,68 +529,73 @@ export MatchPattern.*
...
+def describe(t: Type)(using C: Context): String = t match {
+  case _: TypeRef => s"a type reference ${t.sourceOf}"
+  case _: BoxedType => s"a boxed type ${t.sourceOf}"
+  case _: FunctionType => s"a second-class function type ${t.sourceOf}"
+  case _: Effectful => s"a type-and-effect annotation ${t.sourceOf}"
+
+  // THESE TWO SHOULD NEVER BE USER-VISIBLE!
+  case ValueTypeTree(tpe) => s"a value type tree ${tpe}"
+  case BlockTypeTree(eff) => s"a block type tree ${eff}"
 }
```
jiribenes, line 588: Where does this go? I think that a `private def` in `Namer` would make the most sense?

effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -525,68 +529,73 @@ export MatchPattern.*
...
-  /**
-   * Trees that represent inferred or synthesized types (not present in the source)
-   */
-  case BlockTypeTree(eff: symbols.BlockType)
-  case FunctionType(tparams: List[Id], vparams: List[ValueType], bparams: List[(Option[IdDef], BlockType)], result: ValueType, effects: Effects)
-  case BlockTypeRef(id: IdRef, args: List[ValueType]) extends BlockType, Reference
+/**
+ * Type-and-effect annotations
+ */
+case class Effectful(tpe: ValueType, eff: Effects) extends Type
```
jiribenes, line 571: Should this be renamed to reflect that it is a `Type`?

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1100,78 +1096,113 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
-   * just to see that it either is no blocktype or it is not followed by an `at`
-   * and just "looked" like a block type...
+   * Type ::= Id ('[' Type ',' ... ']')?                                           refType
    *
-   * The parameter [[boxedAllowed]] controls whether on the right a dangling `at`
-   * can occur. This way we prevent parsing `() => S at {} at {}` and force users
-   * to manually parenthesize.
+   *       | '(' Type ',' ... ')'                                                  atomicType
+   *       | '(' Type ')'
+   *       | Id ('[' Type ',' ... ']')?
```
jiribenes, line 1106: Remove stale comment
```suggestion
```

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1100,78 +1099,109 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
-   *
-   * TODO improve errors
-   *   i.e. fail("Expected either a function type (e.g., (A) => B / {E} or => B) or an interface type (e.g., State[T]).")
-   */
-  def blockType(): BlockType = blockType2(true)
-  private def blockType2(boxedAllowed: Boolean): BlockType =
+    // Try to parse each function type variant, fall back to basic type if none match
     nonterminal:
+      functionTypeSimple orElse functionTypeComplex getOrElse atomicType()
+  }
```
jiribenes, line 1169: Here we could also get rid of the `backtrack`s, I think?

effekt/shared/src/main/scala/effekt/source/Tree.scala
jiribenes: Moved to Namer.

effekt/shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -296,6 +298,7 @@ enum BlockTypeConstructor extends BlockTypeSymbol {
 
   case Interface(name: Name, tparams: List[TypeParam], var operations: List[Operation] = Nil)
   case ExternInterface(name: Name, tparams: List[TypeParam])
+  case ErrorBlockType(name: Name = NoName, tparams: List[TypeParam] = Nil)
```
jiribenes, line 301: Screenshot 2025-04-22 at 21 55 49
Using `NoName` here shows `<anon>` when hovering and I can't see where to fix that except for making a copy of `NoName` called `ErrorName` or something...

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1100,78 +1099,109 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
-  def valueType(): ValueType = valueType2(true)
-
-  /**
-   * Uses backtracking!
+  /*
+   * Type grammar by precedence:
+   *
+   * Type ::= Id ('[' Type ',' ... ']')?                                           refType
+   *
+   *       | '(' Type ',' ... ')'                                                  atomicType
```
jiribenes, line 1107: Note to self: I wonder if representing tuples explicitly would help here?

effekt/shared/src/main/scala/effekt/source/Tree.scala
jiribenes: Note to self: Kiama's Positions has a substring method!

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -571,9 +571,13 @@ object Namer extends Phase[Parsed, NameResolved] {
       case _ => Context.abort(s"Can only assign to mutable variables.")
     }
 
-    case tpe: source.ValueType    => resolve(tpe)
     case tpe: source.FunctionType => resolve(tpe)
```
b-studios, line 574: Is this even reachable?

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -571,9 +571,13 @@ object Namer extends Phase[Parsed, NameResolved] {
...
 
+    // NOTE(jiribenes, 2025-04-21):
+    //   How do we get around this case? How do we know which one is expected? Do we even need to do this?
+    //   ... or should we just choose based on the syntax?
+    case tpe: source.Type         => Context.abort(s"Unexpected kindless type: $tpe.")
```
b-studios, line 579: Is this error user facing? Kindless type is not something we have at the moment as a concept.

jiribenes: I wasn't able to trigger it in any way...

b-studios: We can somewhat easily ("just implementation effort") get rid of all of this by defining the `resolve` functions manually for all kinds of trees (`Stmt`, `Term`, `Def`, `Implementation`, etc.).

In general, I think this is a good thing to do anyways to simplify the implementation. However, this will massively clash with other ongoing refactorings, so I don't do this now.

b-studios: Then this case does not show up by construction, since `Type` is not a `Term` and we use `resolveValueType` elsewhere.

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -675,50 +688,106 @@ object Namer extends Phase[Parsed, NameResolved] {
...
+          Context.info(pretty"A function type cannot have multiple effect sets, did you mean to use `/ ${prettyEffects}` instead of `/ ${funEffects} / ${effects}`?")
+        case source.Effectful(source.BoxedType(tpe @ source.FunctionType(tparams, vparams, bparams, result, funEffects), capt), effects) =>
+          Context.info(pretty"Did you want to write a boxed type with effects, `${tpe.sourceOf} / ${effects.sourceOf} at ${capt.sourceOf}`?")
+        case source.Effectful(innerTpe, effs) =>
+          // NOTE: We could use `isParam` to write a more precise message, but what exactly would it be?
+          Context.info(pretty"Did you mean to use a function type () => ${innerTpe.sourceOf} / ${effs.sourceOf}?")
+        case _ => ()
+
+      // Dummy interface type in order to aggregate more errors (see #947)
+      InterfaceType(ErrorBlockType(), Nil)
```
b-studios, line 772: I am not entirely happy with the constructor here. What is the tradeoff of introducing it as a BlockType (and ValueType respectively)?

effekt/shared/src/main/scala/effekt/symbols/symbols.scala
b-studios: Moving this into `BlockType` might help since then you can define a custom pretty printer for it.

effekt/shared/src/main/scala/effekt/Namer.scala
jiribenes: In at least one case, we need to make it an interface type (which is a case class only!), hence the block constructor contained within it. We can reuse it for the block type case too.
Then the value constructor is defined for symmetry for value types.

b-studios: Ah right... we use it for effects!

This is super annoying. What would you do if we were to also implement error resilient type checking?

jiribenes: My intuition is that if we were to represent things in untyped rows like `Symbol := (SymbolKind, Name, ...)`, then I'd have a `SymbolKind.Error` that can appear anywhere.
But I can't think of a good encoding of a "every symbol can be bogus" without devolving into `isNull`-style checks everywhere?

jiribenes: One potential solution is to replace every bad expression with a hole that acts like a unification variable in Typer ala Agda.

jiribenes: Yes, exactly: if we could, I'd replace the bad type with a brand new unification variable for a value type / block type / effect.

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
jiribenes: We can, see #959.

b-studios:
Ping @b-studios

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -675,50 +688,106 @@ object Namer extends Phase[Parsed, NameResolved] {
...
     }
     case source.ValueTypeTree(tpe) =>
       tpe
     // TODO reconsider reusing the same set for terms and types...
     case source.BoxedType(tpe, capt) =>
-      BoxedType(resolve(tpe), resolve(capt))
+      BoxedType(resolveBlockType(tpe), resolve(capt))
+    case other =>
+      Context.error(pretty"Expected value type, but got ${describeType(other)}.")
+      other match
```
b-studios, line 722: Everywhere else we use {} for matches.

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -675,50 +688,106 @@ object Namer extends Phase[Parsed, NameResolved] {
...
+          if isParam then Context.info(pretty"Did you mean to use braces in order to receive a block type `${funTpe.sourceOf}`?")
+          Context.info(pretty"Did you mean to use a first-class, boxed function type `${funTpe.sourceOf} at {}`?")
+        case source.Effectful(source.FunctionType(tparams, vparams, bparams, result, funEffects), effects) =>
+          val combinedEffects = funEffects.effs.toSet ++ effects.effs.toSet
+          val prettyEffects = combinedEffects.toList.map { _.sourceOf } match
+            case List(eff) => eff
+            case Nil => "{}"
+            case many => many.mkString("{", ", ", "}")
+          // TODO(jiribenes, 2025-04-22): `Effects` seem to have bad position information. Why?!
+          Context.info(pretty"A function type cannot have multiple effect sets, did you mean to use `/ ${prettyEffects}` instead of `/ ${funEffects} / ${effects}`?")
```
b-studios, line 733: This is duplicated. Does it make sense to share?

examples/neg/parsing/blockargs.check
```diff
@@ -1,3 +1,6 @@
-[error] examples/neg/parsing/blockargs.effekt:3:20: Effects not allowed here. Maybe you mean to use a function type `() => T / E`?
+[error] examples/neg/parsing/blockargs.effekt:3:18: Expected block type, but got a type-and-effect annotation R / Exc.
 def foo[R] { f : R / Exc }: R = <>
-                   ^
+                 ^^^^^^^
+[info] examples/neg/parsing/blockargs.effekt:3:18: Did you mean to use a function type () => R / Exc?
```
b-studios, line 4: Actually, in this case I always use the short hand notation `=> R / Exc` which we could also mention (additionally).

b-studios (approved):
I like this a lot and we can merge it.

Minor nitpicks: 
- no pattern matching without braces ;)
- what do we do with the comments / introduced TODOs?

Larger things (but I don't know how to address them):

- the parser for types is not as simple and elegant as I would like it to be
- the namer for types is somewhat complicated and will grow over time (every time we notice "this message could be better"). 

We don't need to address these things now, but I am not very happy about both.

b-studios:
We also need to figure out in which order to merge this and @timsueberkrueb 's #957

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -675,50 +688,106 @@ object Namer extends Phase[Parsed, NameResolved] {
...
    */
-  def resolve(tpe: source.ValueType)(using Context): ValueType = resolvingType(tpe) {
-    case source.ValueTypeRef(id, args) => Context.resolveType(id) match {
-      case constructor: TypeConstructor => ValueTypeApp(constructor, args.map(resolve))
+  def resolveValueType(tpe: source.ValueType, isParam: Boolean = false)(using Context): ValueType = resolvingType(tpe) {
+    case source.TypeRef(id, args) => Context.resolveType(id) match {
+      case constructor: TypeConstructor => ValueTypeApp(constructor, args.map(resolveValueType))
       case id: TypeVar =>
         if (args.nonEmpty) {
           Context.abort(pretty"Type variables cannot be applied, but receieved ${args.size} arguments.")
```
b-studios, line 696:
```suggestion
          Context.abort(pretty"Type variables cannot be applied, but received ${args.size} arguments.")
```

examples/neg/parsing/blockargs.check
jiribenes: I considered that, but the error message got a bit too bloated and a bit harder to explain. If we always use just `() => ...` instead of `=> ...`, it's, in a way, more obvious as to what's going on?

effekt/shared/src/main/scala/effekt/Namer.scala
jiribenes: I've noticed this, but I didn't want to do a drive-by on unrelated code :^)

Screenshot 2025-05-06 at 13 53 57

I'll fix it :)

examples/neg/parsing/blockargs.check
b-studios: Yeah... I don't like `=> ...` in a sense since it breaks the principle of definitionsite=usesite. We allowed `R / Exc` in former times but that was ambiguous to parse. Maybe we can allow it again now???

PR #949, opened 2025-04-19
Common Type experiment: part 1.5: bit-rotted surface.Tree rewrite

jiribenes:
part 2 of #947, see that PR first, this one is much more experimental than part 1 because I don't really understand Namer 🙃


jiribenes:
Original branch was `jiribenes/experiment/common-type-tree`, now it's `jiribenes/experiment/common-type-part2`. I've learned more since then in #947, so I'm changing this PR to a completely different branch. Please don't GC the old branch in case I need to relearn something from an experiment past.

PR #954, opened 2025-04-22
Cache substitutions

b-studios:
Substitutions are used multiple times but computed over and over again. This PR caches them until any modification happens.

On my machine, this brings down typer from `1150ms` to `720ms` with 

```
effekt --time text examples/casestudies/anf.effekt.md
```

This shaves of 25% of the total compile time, which is reduced from 2.5s to 2s.


jiribenes (approved):
LGTM :shipit:

PR #955, opened 2025-04-22
Add a fast, but ugly, pretty printer

b-studios:
Quick benchmarks show that it is roughly 10x faster than the pretty one (8ms instead of 80ms on anf.effekt.md).

By default the ugly printer is used, but LSP and `--debug` will show the nice output.

The implementation is purposefully in combinator style to look similar to the pretty version, since the two need to be kept in sync.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/generator/js/FastPrinter.scala
+++ b/effekt/shared/src/main/scala/effekt/generator/js/FastPrinter.scala
@@ -11,14 +11,31 @@ import scala.language.implicitConversions
 
 class FastPrinter {
 
-  var buffer = new StringBuffer()
-
   def pretty(stmt: List[Stmt]): Document =
     buffer = new StringBuffer()
-    stmt foreach { s => toDoc(s); emit("\n") }
+    stmt foreach { s => toDoc(s); line() }
     Document(buffer.toString, Nil)
 
+  // The `emit` effect
+  private var buffer = new StringBuffer()
+
   inline def emit(str: String): Unit = buffer.append(str)
+
+  // the `line` / `indent` effect
+  private var level = 0
+
+  private inline def indented[T](p: => T): T =
+    val before = level
+    level = level + 1
+    val res = p
+    level = before
+    res
+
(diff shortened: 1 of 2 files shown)


b-studios:
I am not 100% convinced I always want to use the ugly one. When performance doing performance debugging, it might be annoying to use the --debug flag and recompile (since the benchmark scripts do not use it).

b-studios:
For comparison, here is the result of nice

```javascript
function charAt_0(str_1, index_1, Exception_0, ks_1, k_1) {
  if ((index_1 < (0))) {
    return Exception_0.raise_0(new OutOfBounds_0(), ((((((((((("Index out of bounds: ") + (('' + index_1))))) + (" in string: '")))) + (str_1)))) + ("'")), ks_1, undefined);
  } else if ((index_1 >= (str_1.length))) {
    return Exception_0.raise_0(new OutOfBounds_0(), ((((((((((("Index out of bounds: ") + (('' + index_1))))) + (" in string: '")))) + (str_1)))) + ("'")), ks_1, undefined);
  } else {
    return () => k_1(str_1.codePointAt(index_1), ks_1);
  }
}
```

and ugly
```javascript
function charAt_0(str_1, index_1, Exception_0, ks_1, k_1) {
if ((index_1 < (0))) {
return Exception_0.raise_0(new OutOfBounds_0(), ((((((((((("Index out of bounds: ") + (('' + index_1))))) + (" in string: '")))) + (str_1)))) + ("'")), ks_1, undefined);
} else if ((index_1 >= (str_1.length))) {
return Exception_0.raise_0(new OutOfBounds_0(), ((((((((((("Index out of bounds: ") + (('' + index_1))))) + (" in string: '")))) + (str_1)))) + ("'")), ks_1, undefined);
} else {
return () => k_1(str_1.codePointAt(index_1), ks_1);
}
}
```

b-studios:
Not sure this is worth it, since on a 6k LoC output the pretty printer does only take 100ms. We can revisit this when we notice this is an actual bottleneck.

PR #957, opened 2025-04-23
Add initial spans to source tree 

timsueberkrueb:
This PR starts the effort on adding spans to the source syntax tree. It is just the first of several steps of this transition.

Current state:

This PR adds the span field to

* [x] `IdDef`
* [x] `IdRef`
* [x] `ModuleDecl`
* [x] `FunDef`
* [x] `Effectful`

It (partially) replaces `List` by Many and `Option` by `Maybe` for:

* [x] `InterfaceDef`
* [x] `DataDef`
* [x] `RecordDef`
* [x] `ExternType`
* [x] `ExternDef`
* [x] `Constructor`
* [x] `Operation`
* [x] `ValueTypeRef`
* [x] `FunctionType`

It tests (`RecursiveDescentTests`) the spans for

* [x] `FunDef`
* [x] `IdDef`
* [x] `IdRef`
* [x] `RecordDef`
* [x] `Effectful`

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Repl.scala
+++ b/effekt/jvm/src/main/scala/effekt/Repl.scala
@@ -312,15 +312,17 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
     def make(source: Source, expr: Term): ModuleDecl = {
 
       val body = Return(expr)
-      val fakeSpan = Span(source, 0, 0, isFake = true)
-      val fullSpan = Span(source, 0, source.content.length, isFake = true)
+      val fakeSpan = Span(source, 0, 0, origin = Origin.Synthesized)
+      val fullSpan = Span(source, 0, source.content.length, origin = Origin.Synthesized)
       ModuleDecl("interactive", includes,
-        definitions :+ FunDef(IdDef("main"), Many.empty(fakeSpan), Many.empty(fakeSpan), Many.empty(fakeSpan), Maybe.None(fakeSpan),
+        definitions :+ FunDef(IdDef("main", fakeSpan), Many.empty(fakeSpan), Many.empty(fakeSpan), Many.empty(fakeSpan), Maybe.None(fakeSpan),
           body, fakeSpan), fullSpan)
     }
 
-    def makeEval(source: Source, expr: Term): ModuleDecl =
-      make(source, Call(IdTarget(IdRef(List(), "println")), Nil, List(expr), Nil))
+    def makeEval(source: Source, expr: Term): ModuleDecl = {
+      val fakeSpan = Span(source, 0, 0, origin = Origin.Synthesized)
+      make(source, Call(IdTarget(IdRef(List(), "println", fakeSpan)), Nil, List(expr), Nil))
+    }
   }
   lazy val emptyModule = ReplModule(Nil, Nil)
 }
--- a/effekt/jvm/src/main/scala/effekt/Repl.scala
+++ b/effekt/jvm/src/main/scala/effekt/Repl.scala
@@ -108,7 +108,7 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
       }
       output.emitln("")
 
-      runFrontend(StringSource(""), module.make(UnitLit()), config) { cu =>
+      runFrontend(StringSource(""), module.make(StringSource(""), UnitLit()), config) { cu =>
         module.definitions.foreach {
           case u: Def =>
             outputCode(DeclPrinter(context.symbolOf(u)), config)
@@ -122,7 +122,7 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
     def typecheck(source: Source, config: EffektConfig): Unit =
       parse(source) match {
         case Success(e: Term, _) =>
-          runFrontend(source, module.make(e), config) { mod =>
+          runFrontend(source, module.make(source, e), config) { mod =>
             // TODO this is a bit ad-hoc
             val mainSym = mod.exports.terms("main").head
             val mainTpe = context.functionTypeOf(mainSym)
@@ -184,7 +184,7 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
    */
   def process(source: Source, tree: Tree, config: EffektConfig): Unit = tree match {
     case e: Term =>
-      runCompiler(source, module.makeEval(e), config)
+      runCompiler(source, module.makeEval(source, e), config)
 
     case i: Include =>
       val extendedIncludes = module + i
@@ -217,7 +217,7 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
 
     case d: Def =>
       val extendedDefs = module + d
-      val decl = extendedDefs.make(UnitLit())
+      val decl = extendedDefs.make(source, UnitLit())
 
       runFrontend(source, decl, config) { cu =>
         module = extendedDefs
@@ -309,17 +309,18 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
     /**
      * Create a module declaration using the given expression as body of main
      */
-    def make(expr: Term): ModuleDecl = {
+    def make(source: Source, expr: Term): ModuleDecl = {
 
       val body = Return(expr)
-
+      val fakeSpan = Span(source, 0, 0, isFake = true)
+      val fullSpan = Span(source, 0, source.content.length, isFake = true)
       ModuleDecl("interactive", includes,
-        definitions :+ FunDef(IdDef("main"), ???, ???, ???, ???,
-          body, ???), ???)
+        definitions :+ FunDef(IdDef("main"), Many.empty(fakeSpan), Many.empty(fakeSpan), Many.empty(fakeSpan), Maybe.None(fakeSpan),
+          body, fakeSpan), fullSpan)
     }
 
-    def makeEval(expr: Term): ModuleDecl =
-      make(Call(IdTarget(IdRef(List(), "println")), Nil, List(expr), Nil))
+    def makeEval(source: Source, expr: Term): ModuleDecl =
+      make(source, Call(IdTarget(IdRef(List(), "println")), Nil, List(expr), Nil))
   }
   lazy val emptyModule = ReplModule(Nil, Nil)
 }
--- a/effekt/jvm/src/main/scala/effekt/Repl.scala
+++ b/effekt/jvm/src/main/scala/effekt/Repl.scala
@@ -315,7 +315,7 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
 
       ModuleDecl("interactive", includes,
         definitions :+ FunDef(IdDef("main"), Nil, Nil, Nil, None,
-          body))
+          body), ???)
     }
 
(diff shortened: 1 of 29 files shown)


effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -3,6 +3,12 @@ package source
 
 import effekt.context.Context
 import effekt.symbols.Symbol
+import kiama.util.{Position, Source, StringSource}
+
+import scala.{None as OptionNone, Some as OptionSome}
+
+import kiama.util.Position
+import scala.{Some => OptionSome , None => OptionNone }
```
jiribenes, line 11: This seems very duplicated?
Also, why are the `OptionSome` and `OptionNone` needed?

effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -209,6 +250,73 @@ enum Param extends Definition {
 }
 export Param.*
 
+/**
+ * A `List` (usually of `Tree`s) with a `span` that spans all elements of the list
+ */
+case class Many[T](list: List[T], span: Span) {
+  def unspan: List[T] = list
```
jiribenes, line 257: I don't quite understand why the explicit `unspan` as a separate method as opposed to `list`.
I would have imagined directly
```suggestion
case class Many[T](unspan: List[T], span: Span) {
```

effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -209,6 +250,73 @@ enum Param extends Definition {
...
+  def map[B](f: T => B): Many[B] =
+    Many(list.map(f),span)
+
+  def unzip [A1, A2](implicit asPair: T => (A1, A2)): (Many[A1], Many[A2]) = {
+    val (list1 : List[A1], list2 : List[A2]) = list.unzip(asPair)
+    (Many(list1, span), Many(list2, span))
+  }
```
jiribenes, line 264: Nitpick: formatting
```suggestion
  def map[B](f: T => B): Many[B] =
    Many(list.map(f), span)

  def unzip[A1, A2](implicit asPair: T => (A1, A2)): (Many[A1], Many[A2]) = {
    val (list1: List[A1], list2: List[A2]) = list.unzip(asPair)
    (Many(list1, span), Many(list2, span))
  }
```

effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -209,6 +250,73 @@ enum Param extends Definition {
...
+}
+
+/**
+ * An `Option` with a `span`
+ *
+ * If the `Option` is `Some`, the `span` is the span of the value.
+ * If the `Option` is `None`, the `span` is an empty span pointing to where the optional value would be expected.
+ */
+case class Maybe[T](option: Option[T], span: Span){
+  def unspan: Option[T] = option
```
jiribenes, line 288: Similar comment as in Many: why both the `.option` and the `.unspan` at the same time?

jiribenes (commented):
quick general notes, does not constitute a proper review

effekt/shared/src/main/scala/effekt/source/Tree.scala
timsueberkrueb: There is a clash with `Maybe.Some`, `Maybe.None` otherwise. Maybe there's a better way to resolve that?

timsueberkrueb:
Anything else you would like to happen in this PR @jiribenes @b-studios?

effekt/jvm/src/main/scala/effekt/Repl.scala
```diff
@@ -309,17 +309,20 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
...
 
       val body = Return(expr)
-
+      val fakeSpan = Span(source, 0, 0, origin = Origin.Synthesized)
+      val fullSpan = Span(source, 0, source.content.length, origin = Origin.Synthesized)
       ModuleDecl("interactive", includes,
-        definitions :+ FunDef(IdDef("main"), Nil, Nil, Nil, None,
-          body))
+        definitions :+ FunDef(IdDef("main", fakeSpan), Many.empty(fakeSpan), Many.empty(fakeSpan), Many.empty(fakeSpan), Maybe.None(fakeSpan),
+          body, fakeSpan), fullSpan)
```
b-studios, line 319: Is there a reason the body of the function is fakeSpan and not fullSpan?

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -466,20 +468,20 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
-      EffectDef(`effect` ~> idDef(), maybeTypeParams(), `=` ~> effects())
+      EffectDef(`effect` ~> idDef(), maybeTypeParams().unspan, `=` ~> effects())
 
   // effect <NAME>[...](...): ...
   def operationDef(): Def =
     nonterminal:
       `effect` ~> operation() match {
         case op @ Operation(id, tps, vps, bps, ret) =>
-          InterfaceDef(IdDef(id.name) withPositionOf op, tps, List(Operation(id, Nil, vps, bps, ret) withPositionOf op))
+          InterfaceDef(IdDef(id.name, id.span) withPositionOf op, tps, List(Operation(id, Many.empty(tps.span.asSynthesized), vps, bps, ret) withPositionOf op))
```
b-studios, line 478: Minor nitpick: we often use the pattern `asXYZ` for unsafe casts that panic, which I thought would be happening here. Maybe just call this method `synthesized`?

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -614,9 +621,9 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
     nonterminal:
       if peek(`:`) then Some(blockTypeAnnotation()) else None
 
-  def maybeReturnAnnotation(): Option[Effectful] =
+  def maybeReturnAnnotation(): Maybe[Effectful] =
     nonterminal:
-      when(`:`) { Some(effectful()) } { None }
+      when(`:`) { Some(effectful()) } { None }.spanned
```
b-studios, line 626: Not sure whether this or

```
when(`:`) { Maybe.Some(effectful(), span()) } { Maybe.None(span()) }
```
is better

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -920,21 +926,23 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
     nonterminal:
       some(valueType, `[`, `,`, `]`)
   def valueArgs(): List[Term] =
     nonterminal:
-      many(expr, `(`, `,`, `)`)
+      many(expr, `(`, `,`, `)`).unspan
   def blockArgs(): List[Term] =
     nonterminal:
-      someWhile(blockArg(), `{`)
+      someWhile(blockArg(), `{`).unspan
```
b-studios, line 945: A lot of `unspan`s will go away once we add spans to more constructors, right?

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1351,31 +1360,34 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
   /**
    * Repeats [[p]], separated by [[sep]] enclosed by [[before]] and [[after]]
    */
-  inline def some[T](p: () => T, before: TokenKind, sep: TokenKind, after: TokenKind): List[T] =
-    consume(before)
-    val res = some(p, sep)
-    consume(after)
-    res
+  inline def some[T](p: () => T, before: TokenKind, sep: TokenKind, after: TokenKind): Many[T] =
+    nonterminal:
```
b-studios, line 1364: That's somewhat surprising to me, but makes sense.

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1395,25 +1407,27 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
 
   inline def braces[T](p: => T): T =
     consume(`{`)
-    val res = p
+    val res =  p
```
b-studios, line 1410: This might be accidental
```suggestion
    val res = p
```

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1441,7 +1455,40 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
   // Positions
 
-  inline def nonterminal[T](inline p: => T): T = {
+  // the "span" effect used to compute the positions of nodes
+  inline def span(): Span =
+    span(_start.value)
+
+  // creates a Span with a given start and the end position of the previous token
+  inline def span(start : Int ): Span =
+    val end = previous.end + 1 // since positions by lexer are inclusive, but kiama is exclusive
```
b-studios, line 1464: I noticed a change in lexer, does this still hold?

https://github.com/effekt-lang/effekt/pull/957/files#diff-b9d57417cf04b760fc643822a7f0cc8fb0827916fa671790b7f35a10f5e4774aL316-R317

effekt/shared/src/main/scala/effekt/source/ExplicitCapabilities.scala
```diff
@@ -29,11 +30,11 @@ object ExplicitCapabilities extends Phase[Typechecked, Typechecked], Rewrite {
     Some(input.copy(tree = rewritten))
 
   override def defn(using Context) = {
-    case f @ FunDef(id, tps, vps, bps, ret, body) =>
+    case f @ FunDef(id, tps, vps, bps, ret, body, span) =>
       val capabilities = Context.annotation(Annotations.BoundCapabilities, f)
       val capParams = capabilities.map(definitionFor)
-      f.copy(bparams = bps ++ capParams, body = rewrite(body))
-    case extDef @ ExternDef(capture, id, tparams, vparams, bparams, ret, bodies) =>
+      f.copy(bparams = (bps.unspan ++ capParams).spanned(bps.span), body = rewrite(body))
```
b-studios, line 36: Interesting!

b-studios (approved):
Modulo the tiny nitpicks. 

Great job @timsueberkrueb and @JakubSchwenkbeck !

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
timsueberkrueb: Removed `spanned`.

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
timsueberkrueb: Renamed.

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
timsueberkrueb: Yes for the lexer we only changed the position of EOF token. Previously, the EOF token had the position `source.content.length + 1`. This led to some spans extend outside the actual file contents. Now it has position `source.content.length = current`, which seems correct to me. The line in the parser just changes from `[start, end]` intervals to `[start, end)` intervals.

PR #963, opened 2025-04-28
Recursive check: Compare modules by path

marzipankaiser:
As noted in https://github.com/effekt-lang/kiama/pull/11, we compare by structural equality on the `ModuleDecl` currently.
This changes it to the path (which should be enough, faster, and more robust w.r.t. #957 and follow-up PRs).

---
Note: Based on #957 .

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Namer.scala
+++ b/effekt/shared/src/main/scala/effekt/Namer.scala
@@ -49,8 +49,8 @@ object Namer extends Phase[Parsed, NameResolved] {
    * Produces a cyclic import error when this is already the case
    */
   private def recursiveProtect[R](mod: ModuleDecl)(body: => R)(using Context): R = {
-    if (currentlyNaming.value.exists { m => m.path == mod.path }) {
-      val cycle = mod :: currentlyNaming.value.takeWhile{ m => m.path != mod.path }.reverse
+    if (currentlyNaming.value.exists { m => m.span.source == mod.span.source }) {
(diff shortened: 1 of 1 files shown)


marzipankaiser:
This fundamentally does not work as the `Module` for, e.g., the stdlib test for queue and the one for queue itself are *both* named `queue`. The `path` is not enough 😞

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -49,8 +49,8 @@ object Namer extends Phase[Parsed, NameResolved] {
    * Produces a cyclic import error when this is already the case
    */
   private def recursiveProtect[R](mod: ModuleDecl)(body: => R)(using Context): R = {
-    if (currentlyNaming.value.contains(mod)) {
-      val cycle = mod :: currentlyNaming.value.takeWhile(_ != mod).reverse
+    if (currentlyNaming.value.exists { m => m.path == mod.path }) {
+      val cycle = mod :: currentlyNaming.value.takeWhile{ m => m.path != mod.path }.reverse
```
marzipankaiser, line 53:
```suggestion
    if (currentlyNaming.value.exists { m => m.span.source == mod.span.source }) {
      val cycle = mod :: currentlyNaming.value.takeWhile{ m => m.span.source != mod.span.source }.reverse
```

or even just store the source (and not the whole `ModuleDecl`)

marzipankaiser:
(rebased onto #957 to make testing possible)

PR #964, opened 2025-04-28
LLVM: Add LTO

jiribenes:


jiribenes:
After the first commit, the perf for duality_of_compilation, effect_handlers_bench, and other stays the same.
(sometimes 1-2% perf wins, but hyperfine notes that changes under ~4% are not significant enough)

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -323,18 +327,52 @@ object LLVMRunner extends Runner[String] {
...
+        "-fwhole-program-vtables",   // Enable whole program devirtualization
+        "-fvisibility=hidden",       // Hide symbols by default
+      )
+      // else GCC LTO flags
+//        gccArgs ++= Seq(
+//          "-flto",                // Enable LTO
+//          "-fno-fat-lto-objects", // Don't create non-LTO object code (smaller intermediate files)
+//          "-fwhole-program",      // Consider the whole program for optimization
+//          "-fuse-linker-plugin"   // Use the GCC linker plugin for LTO
+//        )
```
jiribenes, line 353: This part is WIP, I haven't tested it on GCC at all...

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -323,18 +327,52 @@ object LLVMRunner extends Runner[String] {
...
-    if (C.config.debug()) gccArgs ++= Seq("-g", "-Wall", "-Wextra", "-Werror")
-    if (C.config.valgrind()) gccArgs ++= Seq("-O0", "-g")
-    else if (C.config.debug()) gccArgs ++= Seq("-fsanitize=address,undefined", "-fstack-protector-all")
-
-    exec(gccArgs: _*)
+    if (useLTO) {
+      // Convert to bitcode with aggressive optimizations
+      exec(opt, llPath, "-O3", "-o", bcPath)
+
+      var gccArgs = Seq(gcc, gccMainFile, bcPath, "-o", executableFile) ++ linkedLibraries
```
jiribenes, line 337: This likely only works on Clang, but it seems simpler than optimising, and then compiling to an object file...
Since we mandate LLVM anyway, I'd be willing to drop GCC support... ;)

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -323,18 +327,52 @@ object LLVMRunner extends Runner[String] {
...
+//        gccArgs ++= Seq(
+//          "-flto",                // Enable LTO
+//          "-fno-fat-lto-objects", // Don't create non-LTO object code (smaller intermediate files)
+//          "-fwhole-program",      // Consider the whole program for optimization
+//          "-fuse-linker-plugin"   // Use the GCC linker plugin for LTO
+//        )
+
+      gccArgs ++= Seq(
+        "-O3",                  // Maximum optimization
+        "-march=native",        // Optimize for current CPU
```
jiribenes, line 357: This should probably be under a flag?

mattisboeckle:
Can someone confirm what we discussed in the Meeting?

What I rememeber is:
- Throw away gcc support
- Add flag for native compilation
- Optimize/Tinker with the LTO flags for performance

Is anything wrong / missing?

b-studios:
I thought we generate the ll opt file on write all irs

jiribenes:
What I understood is, in multiple steps:
1. force using `clang`, don't generate `.opt.ll` file or `.o` file, go directly to `.bc` (like in this PR)
2. `--write-all-ir` should also dump the `.opt.ll` file (by calling `llvm-dis` or something like that)
3. tinker with LTO/perf flags (needs benchmarking ofc)
4. add a `--native` option for `-march=native` when compiling [?]

b-studios:
To be mergable, I think this PR needs at least 1 + 2

mattisboeckle:
# Benchmarking

awfy: are_we_fast_yet Benchmarks
other: duality_of_compilation, effect_handlers_bench, input_output
extra: other/unify.effekt, other/nbe.effekt

## Start from commited version
```
clangArgs ++= Seq(
  "-flto=full",                // Explicitly use full LTO
//"-fuse-ld=lld",              // Use LLVM's linker when available
//"-Wl,--lto-O3",              // Optimization level for LTO
  "-fwhole-program-vtables",   // Enable whole program devirtualization
  "-fvisibility=hidden",       // Hide symbols by default
)

clangArgs ++= Seq(
  "-O3",                  // Maximum optimization
  "-march=native",        // Optimize for current CPU
  "-fomit-frame-pointer", // Remove frame pointers when possible
  "-ffunction-sections",  // Place each function in its own section
  "-fdata-sections",      // Place each data item in its own section
)
```

### --march=native turned off
awfy: No real difference, performance increases & decreases only within std deviation
other: No difference

### -Wl,--lto-O3 and -fuse-ld=lld turned on
awfy: No difference
other: Barely different, except `match_options` which was ~30% faster without these flags

### -fvisibility=hidden turned off
awfy: No difference
other: No difference

## Start from noLTO
```
clangArgs ++= Seq(
    "-O2",
)
```

### Compare -O2 and -O3
awfy: No difference
other: No difference
extra: No difference

### Compare -O2 and -O3 + -flto=full
awfy: `queens` ~25% faster, otherwise no difference
other: mostly nothing, `word_count_ascii` ~6% faster, `nqueens` 5%, rest usually 1-2%
extra: No difference

## Result
"-O3 -flto=full" seem to perform the best

b-studios (approved):

PR #966, opened 2025-04-28
Add bytestream-based 'random' module with PRNG and /dev/urandom handlers & fix float ops on Chez

jiribenes:
Straightforward streaming randomness based on bytes, see the `examples` namespace.
The goal is to unblock #893.

The generator is specialised for 32-bit numbers, always returning only the lowest byte (which might be bad / inefficient, but oh well).
Here's more about the algorithm https://en.wikipedia.org/wiki/Lehmer_random_number_generator

Goals:
- [x] working, idiomatic PRNG
- [x] works across all backends (doesn't rely on a bit-width of `Int` through weird shifting behaviour...)
- [x] easy(-ish) to use

Non-goals:
- [ ] good performance
- [ ] cryptographic safety


libraries/common/random.effekt
```diff
@@ -0,0 +1,97 @@
...
+}
+
+def randomInt32(): Int / {read[Byte], stop} = {
+  var result = 0
+  repeat(4) {
+    val b = do read[Byte]()
+    result = result.bitwiseShl(8).bitwiseOr(b.toInt)
+  }
+  result
+}
```
jiribenes, line 55: My hope is that the optimiser will be able to unroll this into:
```scala
    val b1 = bitwiseAnd(do read[Byte]().toInt, 0xFF)
    val b2 = bitwiseAnd(do read[Byte]().toInt, 0xFF)
    val b3 = bitwiseAnd(do read[Byte]().toInt, 0xFF)
    val b4 = bitwiseAnd(do read[Byte]().toInt, 0xFF)
    
    bitwiseOr(
      bitwiseOr(
        bitwiseOr(
          bitwiseShl(b1, 24),
          bitwiseShl(b2, 16)
        ),
        bitwiseShl(b3, 8)
      ),
      b4
    )
```

libraries/common/random.effekt
```diff
@@ -0,0 +1,97 @@
...
+/// A think wrapper over `minstd` and `stream::source`
+def minstd(seed: Int) { randomnessReader: () => Unit / read[Byte] }: Unit =
+  source[Byte] { minstd(seed) } {randomnessReader}
+
+// randomness functions
+
+def randomBool(): Bool / {read[Byte], stop} = {
+  val b = do read[Byte]()
+  b.toInt.mod(2) == 1
+}
```
jiribenes, line 46: `stop` is really darn annoying here -- I know for sure that the stream is infinite.
But if I want to reuse the combinators in `stream`, I have to reuse `read`, which means seeing `stop` everywhere...

libraries/common/random.effekt
jiribenes: It does not... :(
```js
        function loop_1(i_1, ks_16, k_16) {
          if ((i_1 < (4))) {
            const s_4 = next_0.value;
            const tmp_3 = s_4;
            return tmp_3(ks_16, (v_r_1, ks_17) => {
              let b_0 = undefined;
              switch (v_r_1.__tag) {
                case 0: 
                  return SHIFT(p_0, (k_14, ks_18, k_15) =>
                    () => k_15(new None_0(), ks_18), ks_17, undefined);
                case 1:  const v_0 = v_r_1.value_0;b_0 = v_0;break;
              }
              const s_5 = result_0.value;
              result_0.set((((s_5 << (8))) | (b_0)));
              return loop_1((i_1 + (1)), ks_17, k_16);
            });
          } else {
            return () => k_16($effekt.unit, ks_16);
          }
        }
```

libraries/common/random.effekt
```diff
@@ -0,0 +1,172 @@
...
+
+  def unixRandom(): Unit = {
+    with boundary;
+    with on[IOError].report;
+    with devurandom;
+
+    repeat(2) {
+      println(randomInt32())
+    }
+  }
```
jiribenes, line 171: This is not actually called for testing purposes (since how would we figure out if it works?), but serves as a quick demo.

libraries/common/effekt.effekt
```diff
@@ -442,14 +442,14 @@ extern pure def toInt(d: Double): Int =
...
-  chez "${d}"
+  chez "(fixnum->flonum ${d})"
   llvm "%z = sitofp i64 ${d} to double ret double %z"
   vm "effekt::toDouble(Int)"
 
 
 extern pure def round(d: Double): Int =
   js "Math.round(${d})"
-  chez "(round ${d})"
+  chez "(flonum->fixnum (round ${d}))"
```
jiribenes, line 452: 👀 `round` is `flonum => flonum` in Scheme.

libraries/common/random.effekt
```diff
@@ -0,0 +1,177 @@
...
+
+  try randomnessReader() with random {
+    resume(next())
+  }
+}
+
+/// CSPRNG from `/dev/urandom`, handling a reader of random bytes.
+/// Only works on Unix-like OSes!
+def devurandom { randomnessReader: () => Unit / random }: Unit / {Exception[IOError], stop} = {
```
marzipankaiser, line 71: Is it intentional that this can raise `stop`? I would hope that `/dev/urandom` does not end 🤔

marzipankaiser: (or we'd handle this somehow, e.g. reopening)

libraries/common/random.effekt
marzipankaiser: If we'd want, we could add a test that just asserts that we get *some* values (not an error).

Note that on Mac/M1, this leads to segfaults for me on LLVM 🤔 (but probably that's the known problem)

marzipankaiser (approved):
LGTM.

Generated randomness also seems to be OK (at least for my superfluous testing).

libraries/common/effekt.effekt
marzipankaiser: That's the case in many languages AFAIK.

libraries/common/random.effekt
jiribenes: I get `stop` from the `read[Byte]()`: I could also just throw a new IOError?

jiribenes: Done, thanks

PR #970, opened 2025-05-01
Fix generic equals on Chez (#429)

### LINKED ISSUE #429: Structural equality on JavaScript backend doesn't work

jiribenes:
Resolves #429.

Experiment made using the Copilot Agent Code button, because I'm curious if it can do basic stuff like this :)
Seems like the perfect use case: Claude 3.7 -- unlike me -- probably knows enough Chez to fix `equals`...
Screenshot 2025-05-01 at 14 07 59

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
@@ -60,9 +60,6 @@ abstract class ChezSchemeTests extends EffektTests {
 
     examplesDir / "pos" / "io", // async io is only implemented for monadic JS
 
-
-    examplesDir / "pos" / "issue429.effekt",
-
     // Generic comparison
     examplesDir / "pos" / "genericcompare.effekt",
     examplesDir / "pos" / "issue733.effekt",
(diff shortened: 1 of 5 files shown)


b-studios:
Are all commits by copilot or are you manually fixing things?

jiribenes:
First it made the changes, I told it to commit them.
Then the CI failed, so I copied over the message and told it to fix the parens and commit.
Then I told it to make the test suite better and commit the changes after verifying that the checkfile is correct.
(Fun fact: the Effekt extension works in the online VSCode thing, even installs Effekt correctly)

jiribenes:
Now the JS and Chez backends seem to coincide on all of the test cases in the file, so I'm happy with the PR.

jiribenes (commented):
I'll merge this without a review since it very clearly fixes a long-standing bug by making the behaviour identical to JS, and thus improves the state-of-the-art :)

PR #973, opened 2025-05-05
Fix #971 by updating usages when inlining

### LINKED ISSUE #971: Optimizer breaks block arguments to multiple-resume bidirectional effect
When ran with optimizations, the following program ([Playground](https://effekt-lang.org/playground.html?playground=XY8xDoUgEER7TjGlVr834d%2FiH%2BAjGCnERDHGkL27y7I20sxmZ95kCdMUxgzneShwq78G2C9%2BKWbQIGoMm5iTf%2Fufhj05WAPk7UJhhWS7nkfCGfOs2eYBW9iPJaCgOKqFumZMmPpI9B3sejGoXbX8Y%2BKNld56ojB%2B1Q9pk1Y2jm4%3D&repl=bWFpbigp)):
```
effect bdeff{ body: => Unit }: Unit

def hnd{ body: => Unit / bdeff }: Unit =
  try {
    body()
  } with bdeff {
      resume { {b} => 
        b()
      }
    resume { {b} => () }
  }

def main() = {
  hnd{
    do bdeff{
      ()
    }
  }
}
```
crashes with a reference error on some `b_blockBinding_*`

The generated core looks something like:
```
interface bdeff_3421 {}

def main_3424() = reset { {p_3668} => 
  shift(p_3668) { {k_3688} => 
    val __3697: Unit_401 = resume(k_3688) {
      b_blockBinding_3682()
    };
    resume(k_3688) {
      return ()
    }
  }
}
```
which is clearly not well-scoped.

From my initial investigations, this seems to be caused by the Normalizer.

marzipankaiser:
Fixes #971 by:

Before inlining, if the block variable for the function is used more than once, change usages for all free variables
that are used once to Many. This should be a conservative approximation, then.


marzipankaiser:
Add multiply on Usage and use that.

PR #975, opened 2025-05-06
Make the language server accept config nested under the "effekt" key

timsueberkrueb:
Server-side support for https://github.com/effekt-lang/effekt-vscode/pull/71.


timsueberkrueb:
With https://github.com/effekt-lang/effekt/pull/975 and https://github.com/effekt-lang/effekt-vscode/pull/71, no restart is required for displaying the IR. It is sufficient to change the VSCode settings and save the file.

marzipankaiser (approved):
LGTM.

PR #977, opened 2025-05-07
Workaround to fix flaky LSP test

timsueberkrueb:
The generated identifiers in the IR seem non-deterministic between execution environments: one can observe different names locally vs on the CI.

This led to test failures in entirely unrelated PRs (https://github.com/effekt-lang/effekt/pull/930, https://github.com/effekt-lang/effekt/pull/923).

As this is blocking the other PRs, I propose the hotfix contained in this PR. I'm also open to better suggestions on how to write this particular test case or on how to fix the non-determinism.


jiribenes:
> I'm also open to better suggestions on how to write this particular test case or on how to fix the non-determinism.

A non-show-tree version of Core should be somewhat stable since we pretty-print it?

timsueberkrueb:
> Alternatively, we could also check modulo whitespace

This is what I implemented now.

effekt/jvm/src/test/scala/effekt/LSPTests.scala
```diff
@@ -48,8 +47,21 @@ class LSPTests extends FunSuite {
   def withClientAndServer(testBlock: (MockLanguageClient, Server) => Unit): Unit = {
     withClientAndServer(true)(testBlock)
   }
+  
+  /** Normalize the output of the IR by replacing the generated identifiers and stripping all whitespace
+   */
+  def normalizeIRString(ir: String): String = {
+    ir.replaceAll("_\\d+", "_whatever")
+      .replaceAll("\\s+", "")
```
marzipankaiser, line 55:
```suggestion
      .replaceAll("\\s+", " ")
```

timsueberkrueb: Done

PR #978, opened 2025-05-07
Fix normalizing IR strings in LSP tests

timsueberkrueb:
Replacing all occurrences of whitespace by a single space is not a valid approach since expected and actual output can differ in whitespace (e.g. newlines).


marzipankaiser (approved):
Makes the resulting failures less readable, but seems to be the only (quickly developable) solution

PR #981, opened 2025-05-08
Add a simple `readln` function to the standard library

phischu:
For advanced console input and output, in the future, we will bind to `tty` in `libuv`.

How to do it for the Web?
How to do it for the VM?

### DIFF
--- a/libraries/common/effekt.effekt
+++ b/libraries/common/effekt.effekt
@@ -47,10 +47,12 @@ extern def println(value: String): Unit =
   """
   vm "effekt::println(String)"
 
-extern def readln(): String =
+extern async def readln(): String =
+  js "$effekt.capture(callback => $effekt.readln(callback))"
   llvm """
     %result = call %Pos @c_io_readln()
-    ret %Pos %result
+    tail call void @resume_Pos(%Stack %stack, %Pos %result)
+    ret void
   """
 
 def println(value: Int): Unit = println(value.show)
--- a/libraries/common/effekt.effekt
+++ b/libraries/common/effekt.effekt
@@ -47,6 +47,12 @@ extern def println(value: String): Unit =
   """
   vm "effekt::println(String)"
 
+extern def readln(): String =
+  llvm """
+    %result = call %Pos @c_io_readln()
+    ret %Pos %result
+  """
+
 def println(value: Int): Unit = println(value.show)
 def println(value: Unit): Unit = println(value.show)
 def println(value: Double): Unit = println(value.show)
--- a/libraries/common/effekt.effekt
+++ b/libraries/common/effekt.effekt
@@ -42,7 +42,7 @@ extern def println(value: String): Unit =
   js "$effekt.println(${value})"
   chez "(println_impl ${value})"
   llvm """
-    call void @c_io_println_String(%Pos ${value})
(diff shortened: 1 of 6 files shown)


libraries/common/effekt.effekt
```diff
@@ -42,11 +42,19 @@ extern def println(value: String): Unit =
...
   chez "(println_impl ${value})"
   llvm """
-    call void @c_io_println_String(%Pos ${value})
+    call void @c_io_println(%Pos ${value})
     ret %Pos zeroinitializer ; Unit
   """
   vm "effekt::println(String)"
 
+extern async def readln(): String =
+  js "$effekt.capture(callback => $effekt.readln(callback))"
```
marzipankaiser, line 51: Maybe use `jsNode`, since this doesn't work on the web?
(You could then also have a separate `jsWeb` implementation)

Alternatively, we can check in which case we are in `$effekt.readln` and act accordingly.

marzipankaiser:
For `println` in VM we use:
https://github.com/effekt-lang/effekt/blob/bc6adfa227d43986b52a4b313d9bd7ea78767cfd/effekt/shared/src/main/scala/effekt/core/vm/Builtin.scala#L9-L27
So adding a `Runtime.in` stream there would make sense IMHO.

libraries/common/effekt.effekt
marzipankaiser: i.e.
```suggestion
  jsNode "$effekt.capture(callback => $effekt.readln(callback))"
```

jiribenes:
The biggest potential issues here are flushing & the interaction between `readln` and `println`.
The following program should _first_ print the prompt, then read the user's input, and then finally print the output.
```scala
def main() = {
  println("What's your name?") // prompt
  val name = readln()
  println("Hello, " ++ name ++ "!")
}
```

(I'm saying this since all my previous experiments with `readln` on the LLVM backend failed here, first reading, only then printing [?!])

libraries/common/effekt.effekt
jiribenes: In addition to that, we could also put `$effekt.readln` into an `extern jsNode """ .... """` block directly in `effekt.effekt` (above this defn)

marzipankaiser:
Running via sbt, the order thing fails because we run the program like this:
https://github.com/effekt-lang/effekt/blob/bc6adfa227d43986b52a4b313d9bd7ea78767cfd/effekt/jvm/src/main/scala/effekt/Runner.scala#L102-L112
and emit into the stream provided by output, not directly to stdout. This is needed for testing. See also #370 (which fixed this partially, to at least not wait until exiting to forward).

marzipankaiser:
@phischu You can try if #982 fixes it, then we can use this for now.

phischu:
I am wondering:

Do Emojis work?
Does it work on Windows?
Do Emojis work on Windows?

b-studios:
> Do Emojis work?
This you could try yourselves

> Does it work on Windows?
> Do Emojis work on Windows?

Maybe @mattisboeckle can help with these questions?

mattisboeckle:
I'll try on my win machine

marzipankaiser:
@phischu even with #982, this does not print/read in the correct order *on llvm* (only).

PR #982, opened 2025-05-09
Hotfix output forwarding when not capturing output

marzipankaiser:
Hotfix to order input/output correctly when running programs via Effekt.
This checks if the kiama output was changed from its default via a `isInstanceOf`, which we should get rid of at some point, if we decide to use this.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Runner.scala
+++ b/effekt/jvm/src/main/scala/effekt/Runner.scala
@@ -107,7 +107,7 @@ trait Runner[Executable] {
         process.run(new ProcessLogger {
 
           override def out(s: => String): Unit = {
-            C.config.output().emitln(s)
+            out.emitln(s)
           }
 
           override def err(s: => String): Unit = System.err.println(s)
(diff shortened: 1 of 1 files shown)


marzipankaiser:
Program to test (mostly stolen from #981):
```
extern jsNode """
$effekt.readln = function readln$impl(callback) {


   const readline = require('node:readline').createInterface({
     input: process.stdin,
     output: process.stdout,
   });
   readline.question('', (answer) => {
     readline.close();
     callback(answer);
   });
 }
"""

extern async def readln(): String =
  js "$effekt.capture(callback => $effekt.readln(callback))"

def main() = {
  println("What's your name?") // prompt
  val name = readln()
  println("Hello, " ++ name ++ "!")
}
```

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -99,17 +99,23 @@ trait Runner[Executable] {
...
+      case _: kiama.util.OutputEmitter =>
+        process.!< // we are not capturing the output, so forward directly
+      case out =>
+        process.run(new ProcessLogger {
 
-      override def out(s: => String): Unit = {
-        C.config.output().emitln(s)
-      }
+          override def out(s: => String): Unit = {
+            C.config.output().emitln(s)
```
marzipankaiser, line 110:
```suggestion
            out.emitln(s)
```

PR #983, opened 2025-05-12
Publish holes info via custom LSP notification

timsueberkrueb:
Porting the [`feature/llm` branch](https://github.com/effekt-lang/effekt/compare/master...feature/llm) to the current state of the Effekt compiler.

Adds a custom LSP notification `$/effekt/publishHoles` that publishes a list of holes with type and binder information.

Currently, it doesn't support user-annotated names on holes but `Namer` generates a fresh identifier for each hole based on the surrounding top-level declaration, disambiguated by a simple counter. This name should be unique wrt. to the current module and is somewhat stable to changes. It breaks when the user changes the order of holes within definitions of the same name or within the same definition. It also breaks when the user changes the name of the surrounding definition.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Server.scala
+++ b/effekt/jvm/src/main/scala/effekt/Server.scala
@@ -203,8 +203,7 @@ class Server(config: EffektConfig, compileOnChange: Boolean=false) extends Langu
     }
     try {
       publishHoles(source, config)
-    }
-    catch {
+    } catch {
       case e => client.logMessage(new MessageParams(MessageType.Error, e.toString + ":" + e.getMessage))
     }
   }
--- a/effekt/jvm/src/main/scala/effekt/Server.scala
+++ b/effekt/jvm/src/main/scala/effekt/Server.scala
@@ -175,6 +175,16 @@ class Server(config: EffektConfig, compileOnChange: Boolean=false) extends Langu
     }
   }
 
+  /**
+   * Publish holes in the given source file
+   */
+  def publishHoles(source: Source, config: EffektConfig)(implicit C: Context): Unit = {
+    if (!workspaceService.settingBool("showHoles")) return
+    val holes = getHoles(source)
+    if (holes.isEmpty) return
+    client.publishHoles(EffektPublishHolesParams(source.name, holes.map(EffektHoleInfo.fromHoleInfo)))
+  }
+
   // Driver methods
   //
   //
@@ -189,6 +199,12 @@ class Server(config: EffektConfig, compileOnChange: Boolean=false) extends Langu
     try {
       publishIR(source, config)
     } catch {
+      case e => client.logMessage(new MessageParams(MessageType.Error, e.toString + ":" + e.getMessage))
+    }
+    try {
+      publishHoles(source, config)
+    }
+    catch {
       case e => client.logMessage(new MessageParams(MessageType.Error, e.toString + ":" + e.getMessage))
     }
   }
@@ -464,12 +480,12 @@ class Server(config: EffektConfig, compileOnChange: Boolean=false) extends Langu
     contentTpe <- C.inferredTypeOption(hole.stmts)
     if holeTpe == contentTpe
     res <- hole match {
-      case Hole(source.Return(exp), span) => for {
+      case Hole(id, source.Return(exp), span) => for {
         text <- positions.textOf(exp)
       } yield EffektCodeAction("Close hole", span, text)
 
       // <{ s1 ; s2; ... }>
-      case Hole(stmts, span) => for {
+      case Hole(id, stmts, span) => for {
         text <- positions.textOf(stmts)
       } yield EffektCodeAction("Close hole", span, s"locally { ${text} }")
     }
@@ -574,6 +590,9 @@ trait EffektLanguageClient extends LanguageClient {
    */
   @JsonNotification("$/effekt/publishIR")
   def publishIR(params: EffektPublishIRParams): Unit
+
+  @JsonNotification("$/effekt/publishHoles")
+  def publishHoles(params: EffektPublishHolesParams): Unit
 }
 
 /**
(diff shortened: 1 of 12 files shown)


effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -44,6 +44,11 @@ object Namer extends Phase[Parsed, NameResolved] {
 
   /** Shadow stack of modules currently named, for detection of cyclic imports */
   private val currentlyNaming: DynamicVariable[List[ModuleDecl]] = DynamicVariable(List())
+  /** Current parent top-level definition (if any) */
+  private val currentDefinition: DynamicVariable[Option[Def]] = DynamicVariable(None)
+  /** Counter to disambiguate hole identifiers */
+  private var holeCount: mutable.HashMap[String, Int] = mutable.HashMap[String, Int]()
```
timsueberkrueb, line 50: `DynamicVariable` instead of the `var`?

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -877,6 +902,19 @@ object Namer extends Phase[Parsed, NameResolved] {
...
       case many => many.mkString("{", ", ", "}")
     }
+
+  /**
+   * Generate a fresh hole id that is unique in the current module.
+   * Ideally, this id should be somewhat stable wrt. edits to the source code.
+   * For this, we use the name of the current top-level definition, disambiguated by a counter that increments in
+   * depth-first order as Namer traverses the syntax tree.
+   */
+  private def freshHoleId: String = {
```
timsueberkrueb, line 912: Good enough for a start? Better suggestions?

jiribenes: The biggest problem I can see is that this is not stable wrt operations we plan to do: taking a hole/goal and splitting it into many smaller holes/goals.
```scala
def addOne(list) = <> // new ~> 1
val someOtherGoal = <> // 2
// ~~>
def addOne(list) = list match {
  case Nil() => <> // new ~> 1
  case Cons(_,_) => <> // new ~> 2
}
val someOtherGoal = <> // was 2 ~> 3
```
So it's impossible to cache some answers/filters/... for `someOtherGoal` as the ID of its hole went from 2 to 3.

effekt/shared/src/main/scala/effekt/Intelligence.scala
```diff
@@ -268,3 +314,30 @@ trait Intelligence {
...
   }
 }
+
+object Intelligence {
+  case class HoleInfo(id: String,
+                      span: Span,
+                      innerType: Option[String],
+                      expectedType: Option[String],
+                      importedTerms: Seq[TermBinding], importedTypes: Seq[TypeBinding],
+                      terms: Seq[TermBinding], types: Seq[TypeBinding])
```
b-studios, line 324:
```suggestion
  case class HoleInfo(
    id: String,
    span: Span,
    innerType: Option[String],
    expectedType: Option[String],
    importedTerms: Seq[TermBinding], 
    importedTypes: Seq[TypeBinding],
    terms: Seq[TermBinding], 
    types: Seq[TypeBinding]
  )
```

effekt/shared/src/main/scala/effekt/Intelligence.scala
```diff
@@ -268,3 +314,30 @@ trait Intelligence {
...
+
+  case class TermBinding(qualifier: Seq[String], name: String, `type`: Option[String])
+
+  case class TypeBinding(qualifier: Seq[String], name: String, definition: String)
+
+  case class BindingInfo(
```
b-studios, line 330: Indentation like above

effekt/shared/src/main/scala/effekt/Namer.scala
b-studios: Good point!

So there are two things that could happen to the document that invalidate the cache

1. user edits the source file (maybe adds new holes, maybe removes some, maybe moves code)
2. we perform some hole filling (removing one hole but replacing it by potentially multiple ones)

I still would go with this and then refine again; let's have it merged ASAP so that @timsueberkrueb can continue.

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -877,6 +902,19 @@ object Namer extends Phase[Parsed, NameResolved] {
...
+    val prefix = currentDefinition.value.map(_.id.name).getOrElse("hole")
```
b-studios, line 913: I like the idea of using the function name as the hole name! But why only toplevel definitions?

b-studios (approved):
Great thanks!

effekt/shared/src/main/scala/effekt/Namer.scala
marzipankaiser: This is probably a tradeoff as to how complex the computation of numbers is and under which operations we want to be stable. 🤔 

We could of course track all definitions (not just top-level ones) and have it be stable here, but this would break once it's not a local def.

We could in principle use the "path" in the tree after that (so body of the first val in ...), which would be stable under "replacing a hole with stuff", but also more effort to implement (and would probably need some "compression" to not have very long names).

effekt/shared/src/main/scala/effekt/Namer.scala
marzipankaiser: It could also be a `val` (and be cleared instead of re-initialized).
Do we ever actually re-bind it (other than in `resolve(Module)`)?

It could also be a local `val` in a custom context object that gets passed implicitly. Something like:
```scala
case class NamerContext(holeCount: mutable.HahsMap[String, Int])
//...
def ... (using Context, NamerContext): ...
```

effekt/shared/src/main/scala/effekt/Namer.scala
jiribenes: For the record, I didn't write this as a gotcha, but I just wanted to document that there was a trade-off so that there is _some_ actual record of the potential problems this might introduce in the future.
I strongly believe we forget such things all of the time, leading to problems in the future, because nobody bothers to write them down, hence the comment... :)

effekt/jvm/src/main/scala/effekt/Server.scala
```diff
@@ -585,3 +604,39 @@ trait EffektLanguageClient extends LanguageClient {
...
+ *
+ * @param uri The URI of the source file
+ * @param holes The holes in the source file
+ */
+case class EffektPublishHolesParams(uri: String, holes: List[EffektHoleInfo])
+
+/**
+ * Information about a typed hole
+ *
+ * The difference to Intelligence.HoleInfo is that it uses the appropriate LSP type for the range
```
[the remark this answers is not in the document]
timsueberkrueb: The problem is that `Intelligence` is shared between JVM and JS so it cannot use the LSP types.

effekt/shared/src/main/scala/effekt/Namer.scala
timsueberkrueb: Fair enough, it might be better to collect the stack of function scopes, e.g. for

```scala
def foo(): Unit = {
  def bar() = <>

  bar()
}
```

the hole could be called `foo_bar`

timsueberkrueb: Done.

effekt/shared/src/main/scala/effekt/Namer.scala
timsueberkrueb: Changed to `val` and `clear`

PR #986, opened 2025-05-13
Add `padLeft`/`padRight` on `String`s

marzipankaiser:


libraries/common/string.effekt
```diff
@@ -64,7 +64,17 @@ def repeat(str: String, n: Int): String = {
     if (n == 0) result
     else go(n - 1, result ++ str)
   }
-  go(n, "")
+  if(n > 0) { go(n, "") } else { "" }
```
jiribenes, line 67: Couldn't we move the check into the recursive call?
Then it would be just `if (n <= 0) result` there and `go(n, "")` here.
I think it's isomorphic to your change, it just avoids the extra comparison

marzipankaiser: Ah, yes, I was lazy when changing it 😅

marzipankaiser: Works :)

jiribenes (approved):
LGTM, though it would be nice to use a stringbuilder in the future :)

PR #987, opened 2025-05-13
Report wrong logical 'and', 'or'

jiribenes:
Is there a way to make the code not horrible?

Additionally, one common mistake people make is writing:
```effekt
case Foo(x) if x > 5
```
instead of `and`.

But specialising for that is even worse, it's usually in:
```scala
manyWhile(`and` ~> matchGuard(), `and`)
```

So we'd need something like
```scala
`and` 🙁 `if` ~> softFail("We use 'if' around here")
```

as in:
"try to parse this terminal, then try to recover with this other terminal, reporting a soft fail"

Then even the code included here would not be that horrible...

### DIFF
--- a/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
+++ b/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
@@ -825,8 +825,8 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
         case other => other
       }
 
-  def orExpr(): Term = infix(andExpr, `||`)
-  def andExpr(): Term = infix(eqExpr, `&&`)
+  def orExpr(): Term = infixConfused(andExpr, `||`, `or`)
+  def andExpr(): Term = infixConfused(eqExpr, `&&`, `and`)
   def eqExpr(): Term = infix(relExpr, `===`, `!==`)
   def relExpr(): Term = infix(addExpr, `<=`, `>=`, `<`, `>`)
   def addExpr(): Term = infix(mulExpr, `++`, `+`, `-`)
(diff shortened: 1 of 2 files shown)


jiribenes:
Fun, `and` already works, but only sometimes :)
And we clearly use it in the stdlib!
https://github.com/effekt-lang/effekt/blob/db620536ae4d4f09fde36a860bc1f6da758df583/libraries/common/string.effekt#L149

b-studios:
what?? 😲

b-studios:
> `and` 🙁 `if` ~> softFail("We use 'if' around here")

Emojis as parser combinators, I like!

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -842,6 +842,24 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
+          val after = position
+
+          if (tok.kind == confused) softFail(s"`$confused` is not a logical operator, did you mean to use `$op`?", before, after)
+
+          Token(tok.start, tok.end, op) // hack
+        }
+        val right = nonTerminal()
+        left = binaryOp(left, opToken, right).withRangeOf(left, right)
+      }
+      left
```
b-studios, line 861: I'm not very happy that we need to duplicate the code here :(

The parser (and other parts of the compiler) are increasingly filled with details that are "only" relevant for error reporting.

jiribenes: I'm also unhappy about the code duplication, hence the idea for the ☹️ combinator.

jiribenes: On the other hand, parsing is, by itself, somewhat trivial, the hard part _is, and always has been_ recovery and error reporting. The current parser design is pretty good for mere parsing, but, in a way, much worse at any attempts at recovery or error reporting (we're missing a lot of context, labels, etc., _and_ we fail on the first error!)

jiribenes: It's not that important for this exact mistake (`and` instead of `&&`) as I doubt it would deter anybody who can already code, but error recovery / usable errors are very important in the Effekt-specific bits (effects, caps, regions, captures).

Issue #915, opened 2025-03-28
Document imports that are in prelude by default


marvinborner:
Right now, the only documentation for the default imports in prelude are in `Runner.scala`. We should document these on the website, too. Maybe even automate them even, since we change them often.

PR #988, opened 2025-05-13
Add 'char' into prelude

jiribenes:
Follow-up from #664

This is mostly a refactoring: we had multiple places where we converted characters to digits... so now we have `char` directly in the prelude, I've fixed the TODOs, moved externs around, etc.


effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -695,8 +695,8 @@ class VMTests extends munit.FunSuite {
       dynamicDispatches = 783,
       patternMatches = 9464,
       branches = 14696,
-      pushedFrames = 7598,
-      poppedFrames = 7574,
+      pushedFrames = 7620,
+      poppedFrames = 7596,
```
jiribenes, line 699: VM tests now show that a few more frames were pushed&popped (I guess because of `string::toInt`-related changes). cc @b-studios

marvinborner:
By adding libraries to the prelude, should we also remove all explicit imports (`import char`) from `.effekt` samples? Otherwise users may copy redundant imports from there (related to #915)

jiribenes:
Good idea, I've removed them everywhere except:
- other prelude modules that require explicit imports
- tests in `stdlib/char/`

jiribenes:
There's a fun optimisation opportunity:
When I write `isDigit` so that it does **not** reify via `Result`:
```scala
def isDigit(char: Char): Bool = { with on[WrongFormat].default { () => false }; val _ = digitValue(char); true }
```

the core seems a bit simpler:
```scala
def main_3618() = {
  val v_r_3924: Bool_400 = reset { {p_4204} => 
    if (infixGte_3142(49, 48)) {
      val __4259: Int_409 = if (infixLte_3136(49, 57)) {
        return infixSub_106(toInt_3128(49), toInt_3128(48))
      } else {
        shift(p_4204) { {k_4278} => 
          return false
        }
      };
      return true
    } else {
      shift(p_4204) { {k_4279} => 
        return false
      }
    }
  };
  if (v_r_3924) {
    let tmp_4276 = println_1("true")
    return tmp_4276
  } else {
    let tmp_4277 = println_1("false")
    return tmp_4277
  }
}
```

---

But the committed version does reify:
```scala
def isDigit(char: Char): Bool = result[Int, WrongFormat] { digitValue(char) }.isSuccess
```

and then the Core is a bit more difficult since we're actually constructing the `Result`, even though we're just gonna tear it down:
```scala
def main_3616() = {
  val v_r_4244: Result_1961[Int_409, WrongFormat_560] = reset { {p_4261} => 
    if (infixGte_3142(49, 48)) {
      val v_r_4307: Int_409 = if (infixLte_3136(49, 57)) {
        return infixSub_106(toInt_3128(49), toInt_3128(48))
      } else {
        shift(p_4261) { {k_4361} => 
          return make Result_1961[Int_409, WrongFormat_560] Error_1987(make WrongFormat_560 WrongFormat_619(), infixConcat_36(infixConcat_36("Not a valid digit: '", toString_3126(49)), "' in base 10"))
        }
      };
      return make Result_1961[Int_409, WrongFormat_560] Success_1992(v_r_4307)
    } else {
      shift(p_4261) { {k_4362} => 
        return make Result_1961[Int_409, WrongFormat_560] Error_1987(make WrongFormat_560 WrongFormat_619(), infixConcat_36(infixConcat_36("Not a valid digit: '", toString_3126(49)), "' in base 10"))
      }
    }
  };
  v_r_4244 match {
    case Success_1992 { (a_4281: Int_409) => 
      let tmp_4359 = println_1("true")
      return tmp_4359
    }
  } else {
    let tmp_4360 = println_1("false")
    return tmp_4360}
}
```

jiribenes:
Follow-up: should we then use the hand-optimised version for the implementation (we use `isDigit` pretty often in real code, either via the scanner or via the `.toInt` method)?

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
jiribenes: The newest change rolls this back a little:
Screenshot 2025-05-13 at 20 34 28Screenshot 2025-05-13 at 20 34 48

libraries/common/char.effekt
```diff
@@ -15,56 +18,51 @@ def isWhitespace(c: Char): Bool = c match {
...
+  val digit = char match {
+    case char and char >= '0' && char <= '9' => char.toInt - '0'.toInt
+    case char and char >= 'A' && char <= 'Z' => (char.toInt - 'A'.toInt) + 10
+    case char and char >= 'a' && char <= 'z' => (char.toInt - 'a'.toInt) + 10
+    case _ => wrongFormat("Not a valid digit: '" ++ char.toString ++ "'")
   }
+  if (digit >= base) {
+    wrongFormat("Digit '" ++ digit.show ++ "' is too big for base " ++ base.show)
+  }
+  digit
```
marzipankaiser, line 52: Nice :)

libraries/common/scanner.effekt
```diff
@@ -49,6 +49,15 @@ def readSome[A, B] { convert: A => Option[B] }: B / { Scan[A], stop } =
     do stop()
   }
 
+/// like `readSome`, but for functions `A => B / Exception[WrongFormat]` such as `char::digitValue`
+def tryRead[A, B] { convert: A => B / Exception[WrongFormat] }: B / { Scan[A], stop } = {
```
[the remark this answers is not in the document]
marzipankaiser: Using `E` would be nice but can't be inferred, so I think this is fine for now.

marzipankaiser (approved):
Nice, thx. :)

PR #990, opened 2025-05-14
Fix position information for effect sets

### LINKED ISSUE #976: Fix position information for effect sets
After #947

```scala
def qux[A, B] { f: A => B / raise / log }: B = <>
```
returns
```
Expected block type, but got a type-and-effect annotation A => B / raise / log.
A function type cannot have multiple effect sets, did you mean to use `/ {raise, log}` instead of `/ Effects(List(TypeRef(IdRef(List(),raise),List()))) / Effects(List(TypeRef(IdRef(List(),log),List())))`?
```

because of wrong position information (search for the "A function type cannot have multiple effect sets" message appearing twice in `Namer.scala`)

timsueberkrueb:
Fixes #976.

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
@@ -548,22 +548,26 @@ class RecursiveDescentTests extends munit.FunSuite {
         Effectful(
           TypeRef(
             IdRef(Nil, "Int", Span(source, pos(0), pos(1))),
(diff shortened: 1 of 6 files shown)


jiribenes:
When you're done, could you please add the `def qux[A, B] { f: A => B / raise / log }: B = <>` from #976 as a neg-test in [`neg/namer/types`](https://github.com/effekt-lang/effekt/tree/master/examples/neg/namer/types)? 🙏

jiribenes (approved):
LGTM, provided that the neg test fails as expected and every other tests works :)

PR #994, opened 2025-05-14
Refactor namer to mutually recursive `resolve` functions

b-studios:
Replaces a generic (magic) traversal with explicit mutually recursive functions.

By being more explicit, I discovered two cases that might have been buggy before (see below in the comments).

To keep this PR clean, I purposefully do not change anything else. I hope this makes merging others a bit easier (sorry in advance!)

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Namer.scala
+++ b/effekt/shared/src/main/scala/effekt/Namer.scala
@@ -98,7 +98,7 @@ object Namer extends Phase[Parsed, NameResolved] {
         mod
     }
 
-    Context.timed(phaseName, src.name) { resolveGeneric(decl) }
+    Context.timed(phaseName, src.name) { resolve(decl) }
 
     // We only want to import each dependency once.
     val allIncludes = (processedPreludes ++ includes).distinct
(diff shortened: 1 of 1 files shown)


effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -403,24 +400,40 @@ object Namer extends Phase[Parsed, NameResolved] {
...
+    case source.ExternResource(id, tpe)       => ()
     case source.ExternInclude(ff, path, _, _) => ()
+  }
+
+  def resolve(t: source.Term)(using Context): Unit = Context.focusing(t) {
+
+    case source.Literal(value, tpe) => ()
+
+    case source.Hole(stmts, span) =>
+      Context.scoped { resolve(stmts) }
```
b-studios, line 416: Actually, this `scoped` was missing in the generic traversal.

b-studios:
I had to resolve a conflict with `master`. @timsueberkrueb could you review my resolution, since it conflicted with your changes?

timsueberkrueb:
> @timsueberkrueb could you review my resolution, since it conflicted with your changes?

Looks correct to me!

PR #999, opened 2025-05-16
Add DIY Binder example

marzipankaiser:
I quickly sketched this after talking to @jiribenes about how we could potentially add custom implementations of something like `while` with binders. (He said I should add it as an example.):

```effekt
effect lookupX(): Int

def whil{ cond: [R]{ => R / lookupX }{ => R } => R }{ body: => Unit / lookupX }: Unit = {
  cond{ body(); whil{cond}{body} }{ () }
}

def main() = {
  var x = Some(12)
  // desugared cond
  def cond[R]{thn: => R / lookupX}{els: => R} = {
    if(x is Some(y)){ 
      try { thn() } with lookupX{ () => resume(y) }
    } else { els() }
  }
  whil{ cond }{
    println(do lookupX())
    x = Some(do lookupX() - 1)
    if (do lookupX() == 0) {
      x = None()
    }
  }
  x = Some(12)
  whil{ cond }{
    println(do lookupX())
    x = Some(do lookupX() - 1)
    if (do lookupX() == 0) {
      x = None()
    }
  }
}
```

Surprisingly (to me), this generates reasonable JS:
```javascript
function main_2(ks_8, k_7) {
  const x_2 = ks_8.arena.fresh(new Some_0(12));
  function whil_worker_6(ks_9, k_5) {
    whil_worker_7: while (true) {
      const s_3 = x_2.value;
      switch (s_3.__tag) {
        case 1: 
          const y_3 = s_3.value_0;
          const tmp_5 = $effekt.println(('' + y_3));
          x_2.set(new Some_0((y_3 - (1))));
          if (y_3 === (0)) {
            x_2.set(new None_0());
            /* prepare call */
            continue whil_worker_7;
          } else {
            /* prepare call */
            continue whil_worker_7;
          }
          break;
        default:  return () => k_5($effekt.unit, ks_9);
      }
    }
  }
  return whil_worker_6(ks_8, (_1, ks_10) => {
    function whil_worker_8(ks_11, k_6) {
      whil_worker_9: while (true) {
        const s_4 = x_2.value;
        switch (s_4.__tag) {
          case 1: 
            const y_4 = s_4.value_0;
            const tmp_6 = $effekt.println(('' + y_4));
            x_2.set(new Some_0((y_4 - (1))));
            if (y_4 === (0)) {
              x_2.set(new None_0());
              /* prepare call */
              continue whil_worker_9;
            } else {
              /* prepare call */
              continue whil_worker_9;
            }
            break;
          default:  return () => k_6($effekt.unit, ks_11);
        }
      }
    }
    return whil_worker_8(ks_10, (tmp_7, ks_12) => () => k_7(tmp_7, ks_12));
  });
}
```


effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -819,6 +819,21 @@ class VMTests extends munit.FunSuite {
...
+      pushedFrames = 1,
+      poppedFrames = 1,
+      allocations = 30,
+      closures = 0,
+      variableReads = 28,
+      variableWrites = 29,
+      resets = 0,
+      shifts = 0,
+      resumes = 0
+    )),
```
jiribenes, line 836: Beautiful. No need to put a `optimizer-wishlist` label here, this is truly very nice already :)

PR #1002, opened 2025-05-19
Fix serialization of hole info

timsueberkrueb:
LSP4j uses the Gson Java library to (de)serialize JSON as part of the language server protocol. As a Java library, Gson doesn't special-case the Scala types Option and List (or any other Scala types for that matter), leading to undesirable output for these based on the default runtime-reflection based (de)serialization logic.

This PR fixes it by registering custom type adapters for scala `List` and `Option` with Gson.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/GsonUtil.scala
+++ b/effekt/jvm/src/main/scala/effekt/GsonUtil.scala
@@ -43,20 +43,18 @@ class ScalaListTypeAdapter extends JsonSerializer[scala.collection.immutable.Lis
   }
 }
 
-implicit class GsonBuilderScalaOps(val builder: GsonBuilder) extends AnyVal {
-  /**
-   * Make the Scala Option[_] and List[_] types (de)serialize correctly with Gson.
-   *
-   * As a Java library, Gson does not special case any Scala types. This leads to unexpected (de)serialization by default.
-   * By default, Some(x) serializes to {"value":x} and None serializes to {}.
-   * List serializes to nested JSON objects.
-   *
-   * This method adds custom (de)serializers such that
-   * - Some(x) serializes to x and None serializes to null
-   * - List serializes to a flat JSON array.
-   */
-  def withScalaSupport: GsonBuilder =
-    builder
-      .registerTypeHierarchyAdapter(classOf[Option[_]], new ScalaOptionTypeAdapter)
-      .registerTypeHierarchyAdapter(classOf[List[_]],   new ScalaListTypeAdapter)
-}
+/**
+ * Make the Scala Option[_] and List[_] types (de)serialize correctly with Gson.
+ *
+ * As a Java library, Gson does not special case any Scala types. This leads to unexpected (de)serialization by default.
+ * By default, Some(x) serializes to {"value":x} and None serializes to {}.
+ * List serializes to nested JSON objects.
+ *
+ * This method adds custom (de)serializers such that
+ * - Some(x) serializes to x and None serializes to null
+ * - List serializes to a flat JSON array.
+ */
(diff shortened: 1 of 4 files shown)


timsueberkrueb:
Manually tested against the extension.

effekt/jvm/src/main/scala/effekt/GsonUtil.scala
```diff
@@ -0,0 +1,62 @@
...
+   *
+   * This method adds custom (de)serializers such that
+   * - Some(x) serializes to x and None serializes to null
+   * - List serializes to a flat JSON array.
+   */
+  def withScalaSupport: GsonBuilder =
+    builder
+      .registerTypeHierarchyAdapter(classOf[Option[_]], new ScalaOptionTypeAdapter)
+      .registerTypeHierarchyAdapter(classOf[List[_]],   new ScalaListTypeAdapter)
+}
```
marzipankaiser, line 62: `implicit class`es are a Scala-2-ism, and I think [extension methods](https://docs.scala-lang.org/scala3/book/ca-extension-methods.html) are more Scala 3-ish (and also better IMHO):
```suggestion
  /**
   * Make the Scala Option[_] and List[_] types (de)serialize correctly with Gson.
   *
   * As a Java library, Gson does not special case any Scala types. This leads to unexpected (de)serialization by default.
   * By default, Some(x) serializes to {"value":x} and None serializes to {}.
   * List serializes to nested JSON objects.
   *
   * This method adds custom (de)serializers such that
   * - Some(x) serializes to x and None serializes to null
   * - List serializes to a flat JSON array.
   */
  extension(builder: GsonBuilder) def withScalaSupport: GsonBuilder =
    builder
      .registerTypeHierarchyAdapter(classOf[Option[_]], new ScalaOptionTypeAdapter)
      .registerTypeHierarchyAdapter(classOf[List[_]],   new ScalaListTypeAdapter)
```

effekt/jvm/src/main/scala/effekt/Server.scala
```diff
@@ -180,7 +180,7 @@ class Server(config: EffektConfig, compileOnChange: Boolean=false) extends Langu
    */
   def publishHoles(source: Source, config: EffektConfig)(implicit C: Context): Unit = {
     if (!workspaceService.settingBool("showHoles")) return
-    val holes = getHoles(source)
+    val holes = getHoles(source).toArray
```
marzipankaiser, line 183: Is this still needed, since you hooked into gson? I would prefer not to use `scala.Array` if we can help it.

effekt/jvm/src/main/scala/effekt/Server.scala
```diff
@@ -610,7 +596,7 @@ case class EffektPublishIRParams(filename: String,
  * @param uri The URI of the source file
  * @param holes The holes in the source file
  */
-case class EffektPublishHolesParams(uri: String, holes: List[EffektHoleInfo])
+case class EffektPublishHolesParams(uri: String, holes: Array[EffektHoleInfo])
```
marzipankaiser, line 599: See above, does this still need to be an `Array`?

marzipankaiser (changes requested):
I would really prefer extension methods over implicit classes.
Also, unless this is still needed, I wouldn't use `scala.Array` directly (as it is a thin-ish wrapper around java arrays and thus a bit unwieldy to use)

effekt/jvm/src/main/scala/effekt/GsonUtil.scala
timsueberkrueb: Thanks, makes sense!

marzipankaiser (approved):
LGTM. Please make sure that it still works with the extension (just as a reminder).

PR #1003, opened 2025-05-19
Stdlib: Add `OneOrBoth` and `zipLongest` on lists and streams

marzipankaiser:


libraries/common/effekt.effekt
```diff
@@ -715,6 +715,13 @@ record Tuple4[A, B, C, D](first: A, second: B, third: C, fourth: D)
 record Tuple5[A, B, C, D, E](first: A, second: B, third: C, fourth: D, fifth: E)
 record Tuple6[A, B, C, D, E, F](first: A, second: B, third: C, fourth: D, fifth: E, sixth: F)
 
+// These
+// =====
+type These[A, B] {
+  This(a: A)
+  That(b: B)
+  These(a: A, b: B)
+}
```
jiribenes, line 724: Bikeshed: [Unison](https://share.unison-lang.org/@unison/base/code/releases/4.0.0/latest/types/data/OneOrBoth) calls this type `OneOrBoth` with `This | That | Both` which is, I think, a bit cuter.

libraries/common/list.effekt
```diff
@@ -601,6 +601,40 @@ def zip[A, B](left: List[A], right: List[B]): List[(A, B)] = {
...
+///
+/// > zip(Nil[Int](), [1, 2, 3])
+/// [That(1), That(2), That(3)]
+///
+/// > zip([1, 2, 3], [42])
+/// [These(1, 42), This(2), This(3)]
+/// ```
+///
+/// O(N)
+def zipLongest[A, B](left: List[A], right: List[B]): List[These[A, B]] = {
```
jiribenes, line 623: Bikeshed: [Unison](https://share.unison-lang.org/@unison/base/code/releases/4.0.0/latest/terms/data/List/align) calls this `align` which is a bit nicer, perhaps?
(ditto for streams)

jiribenes (commented):
The code is, I think, perfectly fine, but:

🚲, I really like Unison's names for these operations :)

libraries/common/list.effekt
```diff
@@ -601,6 +601,40 @@ def zip[A, B](left: List[A], right: List[B]): List[(A, B)] = {
...
+/// [These(1, 100), These(2, 200), These(3, 300)]
+///
+/// > zip([1, 2, 3], Nil[Int]())
+/// [This(1), This(2), This(3)]
+///
+/// > zip(Nil[Int](), [1, 2, 3])
+/// [That(1), That(2), That(3)]
+///
+/// > zip([1, 2, 3], [42])
+/// [These(1, 42), This(2), This(3)]
```
jiribenes, line 619: These should be `zipLongest`, right? :)

marzipankaiser: Yes, thx.

libraries/common/effekt.effekt
marzipankaiser: I named it like I know it from haskell - but I really do not care here, if there is a consensus to rename this, I can.

examples/stdlib/list/zip.effekt
```diff
@@ -10,10 +10,25 @@ def printListOfPairs(list: List[Tuple2[Int, Int]]): Unit = {
   println(strings)
 }
 
+// used to circumvent missing universal `show` & `println` on some backends
+def printListOfPairs(list: List[These[Int, Int]]): Unit = {
```
jiribenes, line 14:
```suggestion
def printListOfThese(list: List[These[Int, Int]]): Unit = {
```
?

libraries/common/effekt.effekt
marzipankaiser: Renamed.

libraries/common/list.effekt
jiribenes: We could also "just" call it `zip` and rely on the overload resolution ;)

jiribenes (approved):
I'd still prefer a different name than `zipLongest`, but I'm fine with the contents, so feel free to merge this if you want to, @marzipankaiser

Issue #433, opened 2024-03-23
Type mismatch error gets reported as an overload error 


jiribenes:
The following Effekt program:
```scala
def run(n: Int)    { action: => Unit }: Unit = action()
def run(s: String) { action: => Unit }: Unit = action()
def add(a: Int, b: Int): Int = a + b

def main() = {
  run(42) {
    val s: Int = add(1, "Hello world!");
    ()
  }
}
```
reports a type error in _overloading_ of `run`:
```
Cannot typecheck call.
There are multiple overloads, which all fail to check:

Possible overload: module0::run of type (Int){() => Unit} => Unit
  Expected Int but got String.

Possible overload: module0::run of type (String){() => Unit} => Unit
  Expected String but got Int.
  Expected Int but got String.
```

However, the actual reported error should be that the second argument is of type `String`, whereas the expected type of the second parameter is `Int`. 
Fortunately, this error still gets reported at the site of the `add` call so the user can still technically figure this out.

---

If we combine multiple weird overloads:
```scala
def run(n: Int)    { action: => Unit }: Unit = action()
def run(s: String) { action: => Unit }: Unit = action()

def main() = {
  run(42) {
    "type error" ++ 42
  }
}
```
we get a way more confusing message, telling the user that the overload of `run` failed, but the problem is that the overload of `++` failed:
```
Cannot typecheck call.
There are multiple overloads, which all fail to check:

Possible overload: module0::run of type (Int){() => Unit} => Unit
  Expected String but got Int.
  Expected Unit but got String.

Possible overload: module0::run of type (String){() => Unit} => Unit
  Expected String but got Int.  // <- new one! (from the `++` operator)
  Expected String but got Int.
  Expected Unit but got String.
```

PR #1007, opened 2025-05-20
Fix backtracking in nested overloads

### LINKED ISSUE #996: Spurious overload issue when importing `bytearray`
The following code:
```effekt
import array
import bytearray

def main() = {
  val arr: Array[Int] = array::fromList(Nil[Int]())
  arr.foreachIndex { (i, x) =>
    println(x)
  }
}
```
fails with an overload error:
```
[error] weird_overload.effekt:6:20: Cannot typecheck call.
There are multiple overloads, which all fail to check:

Possible overload: bytearray::foreachIndex of type (ByteArray){(Int, Byte){Control} => Unit} => Unit
  Expected ByteArray but got Array[Int].
  Wrong number of block arguments, given 0, but function expects 1.

Possible overload: bytearray::foreachIndex of type (ByteArray){(Int, Byte) => Unit} => Unit
  Expected ByteArray but got Array[Int].

Possible overload: list::foreachIndex of type [A](List[A]){(Int, A){Control} => Unit} => Unit
  Expected List[A] but got Array[Int].
  Wrong number of block arguments, given 0, but function expects 1.

Possible overload: list::foreachIndex of type [A](List[A]){(Int, A) => Unit} => Unit
  Expected List[A] but got Array[Int].
  Expected Byte but got Int.

Possible overload: array::foreachIndex of type [T](Array[T]){(Int, T) => Unit} => Unit
  Expected Byte but got Int.

Possible overload: array::foreachIndex of type [T](Array[T]){(Int, T){Control} => Unit} => Unit
  Wrong number of block arguments, given 0, but function expects 1.

  arr.foreachIndex { (i, x) =>
```

However, when removing the `bytearray` import, it works.

Note the part where it says:
```
Possible overload: array::foreachIndex of type [T](Array[T]){(Int, T) => Unit} => Unit
  Expected Byte but got Int.
```
Somehow, typer gets convinced that `x` needs to be a `Byte`, probably related to the `println`.

marzipankaiser:
Fixes #996.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Typer.scala
+++ b/effekt/shared/src/main/scala/effekt/Typer.scala
@@ -1179,7 +1179,7 @@ object Typer extends Phase[NameResolved, Typechecked] {
         // use the typer state after this checking pass
         Context.restoreTyperstate(st)
         // reassign symbol of fun to resolved calltarget symbol
-        Context.assignSymbol(id, sym)
+        Context.annotateSymbol(id, sym)
 
         return tpe
 
@@ -1607,6 +1607,14 @@ trait TyperOps extends ContextOps { self: Context =>
       }
     }
 
+  private[typer] def annotateSymbol(id: source.IdRef, sym: symbols.Symbol) =
+    annotations.update(Annotations.Symbol, id, sym)
+
+  override def symbolOption(id: source.Id): Option[Symbol] = id match {
+    case x: source.IdRef => Context.annotations.get(Annotations.Symbol, x).orElse { super.symbolOption(id) }
+    case _ => super.symbolOption(id)
+  }
+
   private[typer] def bind(s: ValueSymbol, tpe: ValueType): Unit =
     annotations.update(Annotations.ValueType, s, tpe)
 
@@ -1702,6 +1710,8 @@ trait TyperOps extends ContextOps { self: Context =>
     annotations.updateAndCommit(Annotations.BoundCapabilities) { case (t, caps) => caps }
     annotations.updateAndCommit(Annotations.CapabilityArguments) { case (t, caps) => caps }
     annotations.updateAndCommit(Annotations.CapabilityReceiver) { case (t, caps) => caps }
+
+    annotations.updateAndCommit(Annotations.Symbol) { case (t, sym) => sym }
   }
 
   //</editor-fold>
--- a/effekt/shared/src/main/scala/effekt/Typer.scala
+++ b/effekt/shared/src/main/scala/effekt/Typer.scala
@@ -1610,10 +1610,8 @@ trait TyperOps extends ContextOps { self: Context =>
   private[typer] def annotateSymbol(id: source.IdRef, sym: symbols.Symbol) =
     annotations.update(Annotations.Symbol, id, sym)
 
-  override def symbolOption(id: source.Id): Option[Symbol] = id match {
-    case x: source.IdRef => Context.annotations.get(Annotations.Symbol, x).orElse { super.symbolOption(id) }
-    case _ => super.symbolOption(id)
-  }
+  override def symbolOption(id: source.Id): Option[Symbol] =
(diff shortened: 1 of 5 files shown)


b-studios:
We prefer local scopes over imports, should we also prefer explicit imports over stdlib?

marzipankaiser:
@b-studios We could, though we can also "just" disambiguate using namespaces for now

examples/pos/lists.effekt
```diff
@@ -4,7 +4,9 @@ type List[A] {
   Nil();
   Cons(head: A, tail: List[A])
 }
-
+namespace lists {
+  type List[A] = List[A]
+}
```
jiribenes, line 9: This is a good test to have, but totally baffling to read 😅

b-studios:
The usecase is to have a project (where files import each other) where you would use the same names as in the stdlib. I imagine it would be annoying if we need to manually disambiguate everywhere.

marzipankaiser:
Here, it is enough to disambiguate the type and let overload resolution deal with the rest.

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -1607,6 +1607,14 @@ trait TyperOps extends ContextOps { self: Context =>
       }
     }
 
+  private[typer] def annotateSymbol(id: source.IdRef, sym: symbols.Symbol) =
+    annotations.update(Annotations.Symbol, id, sym)
+
+  override def symbolOption(id: source.Id): Option[Symbol] = id match {
+    case x: source.IdRef => Context.annotations.get(Annotations.Symbol, x).orElse { super.symbolOption(id) }
```
b-studios, line 1614: What is the reason for the `IdRef` check?

effekt/shared/src/main/scala/effekt/context/Annotations.scala
```diff
@@ -467,7 +467,7 @@ trait TreeAnnotations { self: Context =>
       // addDefinedSymbolToSource(sym)
   }
 
-  def symbolOf(id: source.Id): Symbol = symbolOption(id) getOrElse {
+  def symbolOf(id: source.Id): Symbol = this.symbolOption(id) getOrElse {
```
b-studios, line 470: Is the `this.` here necessary? Shouldn't it refer to the same anyway?

b-studios:
If this works, this is a very clean solution. Good job! Can it somehow happen that we forget to go via the local annotations? Did the overload messages change somehow (maybe for the better)?

jiribenes:
https://github.com/effekt-lang/effekt/issues/433 could potentially, maybe, result in a better error?

effekt/shared/src/main/scala/effekt/Typer.scala
marzipankaiser: Mostly that I know that we just annotate those locally. We could drop it now (I originally had a separate annotation).

effekt/shared/src/main/scala/effekt/context/Annotations.scala
marzipankaiser: Yes, this seems to be necessary so the override works (otherwise there is an error from Scala about this being ambiguous)

marzipankaiser:
> Can it somehow happen that we forget to go via the local annotations? 

I hope that this is prevented by `override`ing the symbol-accessors on the context, so all of `.symbol`, `.definition` etc do the correct thing(TM) and first check local annotations.

effekt/shared/src/main/scala/effekt/Typer.scala
b-studios: Honestly, I don't know what's better, but I think not performing the check is conceptually simpler

marzipankaiser: Removing it.

Issue #958, opened 2025-04-23
Off-by-one position in some parser errors


jiribenes:
### Problem

The `fail` in
https://github.com/effekt-lang/effekt/blob/636136cfef13afe51258bc4d2cdcb91280106021/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala#L1096-L1101
has incorrect position information since `next` skips over a token, so it returns the error pointing to a token _after_ the bad one.

#### Example

```
parseIdent("(foo")
```
returns
```
Unexpected parse error (token index 1): Expected identifier
```
but it should return
```
Unexpected parse error (token index 0): Expected identifier
```
as the **first** zero-indexed token (index 0) was expected to be an identifier, not the **second** one (index 1)!

### Solution

We should take another look at the parser and make sure we don't do `next.kind match { ...; case _ => fail }` since it will always have bad position information.

IDK how idiomatic it is, but it can be fixed here with:
```scala
  def ident(): String =
    nonterminal:
      peek.kind match {
        case Ident(id) => { skip(); id }
        case _ => fail(s"Expected identifier")
      }
```

We should also write tests for these: currently it's a bit annoying, I don't know how to get the position out of the error (and matching on the message is hard, as it contains the whole file path...)

jiribenes:
Here's a demo you can try [in your browser today](https://effekt-lang.org/playground.html?playground=S60oSS3KU8gqVjAx4gKClNQ0hdzEzDwNTQVbBQ1NAA%3D%3D&repl=CmhlbGxvV29ybGQoKQo%3D):
<img width="272" alt="Image" src="https://github.com/user-attachments/assets/9ef0b8f5-aa20-49a6-922e-c342fea2eefd" />

The following grammar rules / functions seem to be affected:
- externString
- path
- string
- featureFlag
- + ident

jiribenes:
A potential solution might be to resurrect (and fix) the `expect` helper from b0ebb026.

jiribenes:
One more fun example for good measure ([Playground link](https://effekt-lang.org/playground.html?playground=4%2BJKSU1TMDHS0FSwVSgpKk0FAA%3D%3D&repl=CmhlbGxvV29ybGQoKQo%3D))
<img width="312" alt="Image" src="https://github.com/user-attachments/assets/aac3ba1d-2e65-4bdc-9938-ebdd921a79b5" />

b-studios:
[...]

PR #1008, opened 2025-05-20
Labels in parser

### LINKED ISSUE #960: Labels in parser
Add a `p labelledAs name` [feel free to bikeshed] combinator function that normally behaves exactly like `p`, but when `p` fails _without consuming input_, the `name` gets used for the "Expected $name, got $badToken" message when failing.
(See https://markkarpov.com/tutorial/megaparsec.html#labeling-and-hiding-things for example usage)

Example usage:
```
functionType :=
  ... <~ `=>` ~ atomicType() labelledAs "return type" ~ maybeEffects()
```

produces the following message in a failing program:
```
def show(mod: Module => ): String
                        ^ [error] Expected return type, but got ‘)’
```

instead of what I get right now (actually after #947, currently it gets stuck on `=>`):
```
def show(mod: Module => ): String
                        ^ [error] Expected identifier, but got ‘)’
```

In general, this would allow us to override "Expected $terminal, but got ')'" with "Expected $nonterminals..., but got ')'", which seems strictly better to me...

One possible implementation is to have a mutable variable with an `Option[String]` inside the `RecursiveDescent` parser with a pending label to be used by `fail` instead of its content. The only complication is that we need to correctly reset the mutable variable whenever we consume something (and we need to interact correctly in `backtrack`).

jiribenes:
Resolves #960

Most messages are now of the form "Expected $A but got $B".

Now says something like: "expected block type but got `}`" instead "expected identifier but got `}`" (wrong position is #958)
Screenshot 2025-05-20 at 15 21 33

### DIFF
--- a/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
+++ b/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
@@ -465,7 +465,7 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
 
       peek.kind match {
         case `=` => `=` ~> TypeDef(id, tps.unspan, valueType())
-        case _ => braces { DataDef(id, tps, manyWhile({ constructor() <~ semi() }, !peek(`}`))) }
+        case _ => braces { DataDef(id, tps, manyUntil({ constructor() <~ semi() }, `}`)) }
       }
 
   def recordDef(): Def =
@@ -507,7 +507,7 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
 
   def interfaceDef(): InterfaceDef =
     nonterminal:
-      InterfaceDef(`interface` ~> idDef(), maybeTypeParams(), `{` ~> manyWhile(`def` ~> operation(), `def`) <~ `}`)
+      InterfaceDef(`interface` ~> idDef(), maybeTypeParams(), `{` ~> manyUntil({ `def` ~> operation() } labelled "operation declaration", `}`) <~ `}`)
 
   def namespaceDef(): Def =
     nonterminal:
@@ -737,7 +737,7 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
         if !peek(`def`) then fail("Expected at least one operation definition to implement this interface.")
         tpe
       } map { tpe =>
-        Implementation(tpe, manyWhile(opClause(), `def`)) <~ `}`
+        Implementation(tpe, manyUntil(opClause() labelled "operation clause", `}`)) <~ `}`
       }
 
       // Interface[...] { () => ... }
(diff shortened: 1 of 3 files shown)


effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -124,6 +137,7 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
   def skip(): Unit =
     previous = tokens(position)
     position += 1;
+//    currentLabel = None // TODO(jiribenes, 2025-05-20): ???
```
jiribenes, line 140: I thought that this should be set to `None` once we consume a token, but it doesn't work (I get "expected identifier got `}`" for the example problem above)

marzipankaiser:
Turns out the original approach of this PR *does* (mostly) work, it just got hidden by #958.
So I rebased onto master, re-added the setting to None on skip, and changed the dynamic binding to a co-dynamic binding, and it seems to work:
Screenshot 2025-05-30 at 15 41 15

marzipankaiser:
Leaving it like this for now - except will later fix up the failing neg tests (and maybe add some more)

marzipankaiser:
Fixed up the interaction with `consume`, too, and slightly changed the parser to improve error messages.
I'd suggest to merge this once its through CI and reviewed; we can add more labels incrementally when we notice error messages that could be better.

jiribenes:
Great work, thanks a lot for following up on my crappy WIP implementation! :)

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -197,7 +210,7 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
           semi()
           if returnPosition then Return(e)
           else ExprStmt(e, stmts())
-      }
+      }) labelled "statements"
```
jiribenes, line 213: I think it's fine to merge it like this, but maybe we should also figure out a prefix version of `labelled`, this looks a bit weird for my taste.
(That's on me, I made the combinator like this :sweat_smile:)

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1172,8 +1188,10 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
   private def functionType(): Type = {
     // Complex function type: [T]*(Int, String)*{Exc} => Int / {Effect}
     def functionTypeComplex: Maybe[Type] = backtrack {
-      maybeTypeParams() ~ maybeValueTypes() ~ (maybeBlockTypeParams() <~ `=>`) ~ atomicType() ~ maybeEffects() match {
-        case tparams ~ vparams ~ bparams ~ t ~ effs => FunctionType(tparams, vparams, bparams, t, effs)
+      maybeTypeParams() ~ maybeValueTypes() ~ (maybeBlockTypeParams() <~ `=>`)
+    } map { case tparams ~ vparams ~ bparams =>
+      (atomicType() labelled "return type") ~ maybeEffects() match {
+        case  t ~ effs => FunctionType(tparams, vparams, bparams, t, effs)
       }
```
jiribenes, line 1195: Perfect! cuts in action ✂️

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
marzipankaiser: I agree, I mostly couldn't come up with a good name for it. 😅

PR #1009, opened 2025-05-21
Fix: Make case studies runnable on the website

dvdvgt:
Currently, the case studies are not runnable on the website or resort to the trick of having an empty REPL where you have to input `main()`. In this PR, I try to properly handle `effekt:repl` in literate Effekt files such that it has the same (observable) behaviour as on the website. Since expressions can not be top-level, I wrap them in a function `def replN() = <REPL-CONTENT>` and later append a main function where all repl cells are called:

```
def repl1() = 1 + 2
def repl2() = println("hello")

def main() = {
  repl1()
  repl2()
  ()
}
```

If there's already a user-defined `main` function, then an error occurs. Though, I argue that the user should not define a `main` function in a literate Effekt file, as this kind of defeats its purpose.

Furthermore, I also tackle the issue of importing non-stdlib modules by just merging all files in the dependency chain into the same file and adding appropriate redirects for the old links.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/util/Source.scala
+++ b/effekt/shared/src/main/scala/effekt/util/Source.scala
@@ -17,21 +17,46 @@ case class MarkdownSource(source: Source) extends Source {
 
   lazy val content = {
     var inCode = false
-    var inOtherFence = false
+    var inRepl = false
+    var repl = ""
+    var replCounter = 0
     // we map non-code lines to empty lines to preserve line numbers
-    val lines = source.content.linesIterator.map {
-      case line if fenceLine.matches(line) && !inOtherFence =>
-        inCode = !inCode
-        ""
-      case line if otherFenceLine.matches(line) =>
-        inOtherFence = !inOtherFence
-        ""
-      case line if inCode =>
-        line
-      case _ =>
-        ""
+    val lines = source.content.linesIterator.map { line =>
+      line match {
+        case fenceLine(line, ots, _) =>
+          val opts = if (ots != null) { ots.tail.split(":").toList } else { Nil }
+          if (opts.contains("repl")) {
+            inRepl = true
+            ""
+          } else if (inRepl) {
+            inRepl = false
+            val res = s"def repl$replCounter() = $repl"
+            replCounter = replCounter + 1
+            res
+          } else {
+            inCode = !inCode
+            ""
+          }
+        case line if inRepl =>
+          repl = line
+          ""
+        case line if inCode => line
+        case _ => ""
+      }
     }
-    lines.mkString("\n")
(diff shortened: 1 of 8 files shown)


jiribenes:
> Though, I argue that the user should not define a main function in a literate Effekt file, as this kind of defeats its purpose.

FWIW, I do that a lot for teaching/exercises so that I can write a more thorough description in Markdown for the task at hand without the students needing to jump between two files.
Since it's ran locally, it still has a `def main` (at the very least because of the `|> Run` button)

jiribenes:
I do, in fact, like individual `effekt:repl` more, but then I'd want the `|> Run` button to work on those somehow.

jiribenes:
One more thought is that if we had some `@main` or `@export` annotations (which we were thinking of already if I recall correctly, cc @b-studios), then we wouldn't need one main -- each REPL could be its own `@main` and then the call would be like `effekt runMain repl1 my/effekt/file/here.effekt`?

dvdvgt:
Hm yes, I indeed do not know how to still have separate run buttons or even run them separately (locally).

dvdvgt:
This should be mostly done, except for the `smc` (Monte Carlo) case study. It doesn't seem to work correctly on the web. I propose deferring fixing it to another PR so that we at least get the other case studies running.

dvdvgt:
Ready to merge. Changes:

- You may now either have one or more `effekt:repl` code fences or one manually defined `main` function
- A repl cell is translated such that it is wrapped in a function:
  ```
  def replN = {
    stmt
  }
  ```
  corresponds to
  ````
  ```effekt:repl
  stmt
  ```
  ````
  where `def replN = {` is inserted for the opening fence and `}` for the closing fence to preserve positional information.
- You may now have other languages in a literate effekt file other than Effekt, e.g. ```` ```scala ```` gives you syntax highlighting but is ignored by the effekt compiler
- Similarly, `effekt:ignore` code fences are also ignored by the compiler
- Previously, `[[()]] = #f` in chez. Now, `[[()]] = #<void>`, reasone being that `inspect(inspect(())) = inspect(())` should hold but with `[[()]] = #f` it is not since `inspect` returns `#<void>`.

libraries/chez/common/effekt_primitives.ss
```diff
@@ -2,6 +2,7 @@
   (cond
     [(number? obj) (show-number obj)]
     [(string? obj) obj]
+    [(null? obj) "()"]
```
jiribenes, line 5: Is this really needed? 👀 
```suggestion
```

jiribenes (approved):
LGTM :shipit: (see comments before)
Worst-case, you'll fix the bugs ;)

PR #1010, opened 2025-05-21
Implement hole hover for website playground

timsueberkrueb:
Fixes https://github.com/effekt-lang/effekt-website/issues/3.
The info shown still can be improved, of course, but when we do that, it will happen for both the JVM language server and the website because `getHoleHover` is now implemented on `Intelligence`.

Tested locally for the website playground:

Screenshot from 2025-05-21 16-14-04


jiribenes (approved):
LGTM :shipit:

Issue #1001, opened 2025-05-18
Eternally applicable code actions


jiribenes:
The code action can be applied infinitely, which is a bit strange:

https://github.com/user-attachments/assets/cc81e83b-f3a0-4c27-8d7b-63c40cef3a46

What I would expect is something like what happens when reifying an inlay hint:
once I apply the code action, it isn't offered anymore.
(Ideally forcing a save / recompilation...)

jiribenes:
Also I've just noticed that the title of the code action is a bit weird: it says "Update return type with inferred effects", but it adds a type-and-effect annotation?

timsueberkrueb:
Okay, the real bug here is that we don't force recompilation on `didChange`. You can't have both: up-to-date code actions/error messages *and* only recompile on save.

The proper fix is to have recoverable errors throughout the phases and recompile on `didChange`.

One possible workaround is to turn the *edit* code action into a *command* code action which forces the client to let the server know when the action is applied. We could then set a `dirty` flag in the server and recompile on the next `didChange`.

jiribenes:
I was thinking of having a "was used" flag on code actions on our side (and not show those until a save once they are used)

jiribenes:
Of course, for this specific code action, we can also just replace it by an inlay hint which does have the desired behaviour, AFAICT? ( #869 )

timsueberkrueb:
InlayHints only support `textEdits`, so the command workaround is not even possible for inlay hints:

```
/**
	 * Optional text edits that are performed when accepting this inlay hint.
	 *
	 * *Note* that edits are expected to change the document so that the inlay
	 * hint (or its nearest variant) is now part of the document and the inlay
	 * hint itself is now obsolete.
	 *
	 * Depending on the client capability `inlayHint.resolveSupport` clients
	 * might resolve this property late using the resolve request.
	 */
	textEdits?: TextEdit;
```

You could perhaps do a hack on the client-side but I rather not do that. Compiling on `didChange` seems to be the only viable solution to me.

PR #1011, opened 2025-05-21
Add inlay hints for showing omitted function return types

### LINKED ISSUE #869: Inlay hints: function type/effects
Related to #524 
Extracted from #828
Blocked on #833

Here's some code that fits into `override def getInlayHints` in `effekt/jvm/src/main/scala/effekt/Server.scala` and that mostly worked locally:
```scala
   // TODO: Also do this for un-annotated `val`s and `var`s.
    val unannotated = getTreesInRange(range)(using context).map { trees =>
      trees.collect {
        // Functions without an annotated type:
        case fun: FunDef if fun.ret.isEmpty => for {
          // TODO(question): `context.inferredTypeAndEffectOption(fun)` didn't work here. Why?
          sym <- context.symbolOption(fun.id)
          tpe <- context.functionTypeOption(sym)
          pos <- positions.getFinish(fun.bparams) // TODO(bug): This position doesn't always work. Ideally, I'd use the position of `fun.ret`, but that doesn't exist!
        res = InlayHint(InlayHintKind.Type, pos, pp": ${tpe.result} / ${tpe.effects}", resolveToLabel = true, paddingLeft = true, effektKind = "missing-type-annotation")
        } yield res
      }.flatten
    }.getOrElse(Vector())

    val allHints = captures ++ unannotated
    if allHints.isEmpty then None else Some(allHints)
```

timsueberkrueb:
Fixes #869. The attached text edits also suffer from #1001.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Server.scala
+++ b/effekt/jvm/src/main/scala/effekt/Server.scala
@@ -407,14 +407,14 @@ class Server(config: EffektConfig, compileOnChange: Boolean=false) extends Langu
               pos = fun.ret.span.range.from
             } yield {
               val prettyType = pp": ${tpe.result} / ${tpe.effects}"
-              val inlayHint = new InlayHint(convertPosition(pos), messages.Either.forLeft(pp": ${tpe.result} / ${tpe.effects}"))
+              val inlayHint = new InlayHint(convertPosition(pos), messages.Either.forLeft(prettyType))
               inlayHint.setKind(InlayHintKind.Type)
               val markup = new MarkupContent()
               markup.setValue(s"return type${prettyType}")
               markup.setKind("markdown")
               inlayHint.setTooltip(markup)
               inlayHint.setPaddingLeft(true)
-              inlayHint.setData("omitted-type-annotation")
+              inlayHint.setData("return-type-annotation")
               val textEdit = new TextEdit(convertRange(fun.ret.span.range), prettyType)
               inlayHint.setTextEdits(Collections.seqToJavaList(List(textEdit)))
               inlayHint
--- a/effekt/jvm/src/main/scala/effekt/Server.scala
+++ b/effekt/jvm/src/main/scala/effekt/Server.scala
@@ -8,7 +8,7 @@ import effekt.source.{Span, Tree}
 import effekt.symbols.Binder.{ValBinder, VarBinder}
 import effekt.symbols.BlockTypeConstructor.{ExternInterface, Interface}
 import effekt.symbols.TypeConstructor.{DataType, ExternType}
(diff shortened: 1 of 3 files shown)


timsueberkrueb:
If you like, I can also add them for the other def kinds in this PR.

effekt/jvm/src/main/scala/effekt/Server.scala
```diff
@@ -397,6 +397,32 @@ class Server(config: EffektConfig, compileOnChange: Boolean=false) extends Langu
...
+            } yield {
+              val prettyType = pp": ${tpe.result} / ${tpe.effects}"
+              val inlayHint = new InlayHint(convertPosition(pos), messages.Either.forLeft(pp": ${tpe.result} / ${tpe.effects}"))
+              inlayHint.setKind(InlayHintKind.Type)
+              val markup = new MarkupContent()
+              markup.setValue(s"return type${prettyType}")
+              markup.setKind("markdown")
+              inlayHint.setTooltip(markup)
+              inlayHint.setPaddingLeft(true)
+              inlayHint.setData("omitted-type-annotation")
```
jiribenes, line 417: Philosophically, I think it would be wise to name this as to what it adds, not what it resolves. :)
(I'm willing to compromise about this, but let's argue first if you feel strongly about `omitted-`)
```suggestion
              inlayHint.setData("return-type-annotation")
```
(don't forget the tests too)

effekt/jvm/src/main/scala/effekt/Server.scala
```diff
@@ -397,6 +397,32 @@ class Server(config: EffektConfig, compileOnChange: Boolean=false) extends Langu
...
+        val unannotated = getTreesInRange(range)(using context).map { trees =>
+          trees.collect {
+            // Functions without an annotated type:
+            case fun: FunDef if fun.ret.isEmpty => for {
+              sym <- context.symbolOption(fun.id)
+              tpe <- context.functionTypeOption(sym)
+              pos = fun.ret.span.range.from
+            } yield {
+              val prettyType = pp": ${tpe.result} / ${tpe.effects}"
+              val inlayHint = new InlayHint(convertPosition(pos), messages.Either.forLeft(pp": ${tpe.result} / ${tpe.effects}"))
```
jiribenes, line 410: ?
```suggestion
              val prettyType = pp": ${tpe.result} / ${tpe.effects}"
              val inlayHint = new InlayHint(convertPosition(pos), messages.Either.forLeft(prettyType))
```

jiribenes (commented):
Mostly LGTM, could you also add one tests which shows an _effect_ being returned in the annotation (besides the empty effect set)?

jiribenes (approved):
LGTM :shipit:

PR #1013, opened 2025-05-21
Use a ArrayBuffer in Lexer, pre-reserving space for tokens

jiribenes:
I tried benchmarking this change and the lexer part went from ~200 ms total on JS on case studies and benchmarks to ~160 ms total.

Could somebody double-check that this actually helps?

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
@@ -245,10 +245,10 @@ class RecursiveDescentTests extends munit.FunSuite {
 
   test("Peeking") {
     implicit def toToken(t: TokenKind): Token = Token(0, 0, t)
-    def peek(tokens: IndexedSeq[Token], offset: Int): Token =
+    def peek(tokens: Array[Token], offset: Int): Token =
       new RecursiveDescent(new Positions, tokens, StringSource("", "test")).peek(offset)
 
-    val tokens = Vector[Token](`(`, Space, Newline, `)`, Space, `=>`, EOF)
+    val tokens = Vector[Token](`(`, Space, Newline, `)`, Space, `=>`, EOF).toArray
     assertEquals(peek(tokens, 0).kind, `(`)
     assertEquals(peek(tokens, 1).kind, `)`)
     assertEquals(peek(tokens, 2).kind, `=>`)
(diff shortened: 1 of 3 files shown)


effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -169,7 +169,7 @@ class Lexer(source: Source) {
    */
   var current: Int = 0
   /** The sequence of tokens the source strings contains. Returned as the result by [[Lexer.run()]] */
-  val tokens: mutable.ListBuffer[Token] = mutable.ListBuffer.empty
+  val tokens: mutable.ArrayBuffer[Token] = new mutable.ArrayBuffer(initialSize = 256)
```
jiribenes, line 172: I think this is the main culprit, actually: we know we'll have enough tokens to warrant reserving a bit of space. :)

jiribenes:
One more thing worth investigating is then the `tokens.lastOption` used -- does it have the correct time complexity for ArrayBuffers?
When I go to definition in IntelliJ, it shows that it might `while true do iter.next()` which is a bit wasteful...

jiribenes:
We could also try `Vector.newBuilder` + `.result` 🤔

jiribenes:
On my machine, this makes the lexer a tad faster in most cases, but the change is often very, very small:
Screenshot 2025-05-22 at 17 47 12

effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
```diff
@@ -245,10 +245,10 @@ class RecursiveDescentTests extends munit.FunSuite {
 
   test("Peeking") {
     implicit def toToken(t: TokenKind): Token = Token(0, 0, t)
-    def peek(tokens: Seq[Token], offset: Int): Token =
+    def peek(tokens: Array[Token], offset: Int): Token =
       new RecursiveDescent(new Positions, tokens, StringSource("", "test")).peek(offset)
 
-    val tokens = List[Token](`(`, Space, Newline, `)`, Space, `=>`, EOF)
+    val tokens = Vector[Token](`(`, Space, Newline, `)`, Space, `=>`, EOF).toArray
```
jiribenes, line 251:
```suggestion
    val tokens = Array[Token](`(`, Space, Newline, `)`, Space, `=>`, EOF)
```

dvdvgt (approved):
Even if the benchmarks only show a small performance gain, I think pre-allocating the `tokens` collection is conceptually the right thing to do and should always be faster (even if only marginally).

jiribenes:
Deprecated by #1066.

PR #1015, opened 2025-05-22
Fix off-by-one-token bugs

### LINKED ISSUE #958: Off-by-one position in some parser errors

timsueberkrueb:
Fixes #958.


examples/neg/parsing/extern-path.check
```diff
@@ -0,0 +1,3 @@
+[error] examples/neg/parsing/path.effekt:1:16: Expected path as string literal
```
jiribenes, line 1: Wrong file name makes the tests fail :(
```suggestion
[error] examples/neg/parsing/extern-path.effekt:1:16: Expected path as string literal
```

jiribenes (commented):
What do we do about `literal` then? 👀
It should use `expect` (it clearly follows the same pattern already), but it needs two skips in one place for the unit literal case...

Could we at least put a TODO comment over it?

timsueberkrueb:
Added a TODO because I think this is independent of fixing this bug. Ideally, the unit literal would be a single token.

jiribenes (approved):

PR #1017, opened 2025-05-22
Change explicit box capture syntax

### LINKED ISSUE #965: Change explicit box syntax from `box <captures> <expr>` to `box <expr> at <captures>`
`box { (x, y) => compareByteString(x, y) } at {io, global}` as opposed to `box {io, global} { (x, y) => compareByteString(x, y) }` which we have right now.

To be perfectly honest, I don't think anybody is using this explicitly, it's just annoying when the parser decides "oh, I'm parsing captures!" and then whines about unexpected _stuff_.

The biggest downside is that for the implicit version without captures, the \*_inlay hint_\* is at the wrong place:
```
box *{io, global}* { (x, y) => compareByteString(x, y) }
```
But we could change it to be postfix...
```
box { (x, y) => compareByteString(x, y) } *at {io, global}* 
```

### Implementation

Here's the relevant parser code:
https://github.com/effekt-lang/effekt/blob/master/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala#L670-L676

The new parser would be something like:
```
 def boxExpr(): Term =
    nonterminal:
      val expr = if (peek(`{`)) functionArg()
        else if (peek(`new`)) newExpr()
        else callExpr()
      val captures = backtrack { `at` ~> captureSet() }
      Box(captures, expr)
```
but I think we might even be able to simplify the `expr` (as we no longer have to assume that `{` starts a capture...)
(potentially relevant: #772)

timsueberkrueb:
Closes #965.

Inlay hint:

Screenshot from 2025-05-22 17-07-16

### DIFF
--- a/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
+++ b/effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
@@ -678,7 +678,6 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
     nonterminal:
       consume(`box`)
       val expr = if (peek(`{`)) functionArg()
-      else if (peek(`new`)) newExpr()
       else callExpr()
       val captures = backtrack {
         `at` ~> captureSet()


effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1347,8 +1350,13 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
   inline def backtrack[T](inline p: => T): Maybe[T] =
     val before = position
-    try { Maybe.Some(p, span(before)) } catch {
-      case Fail(_, _) => position = before; Maybe.None(span(before))
+    val beforePrevious = previous
+    try { Maybe.Some(p, span(tokens(before).end)) } catch {
+      case Fail(_, _) => {
+        position = before
+        previous = beforePrevious
+        Maybe.None(Span(source, previous.end + 1, previous.end + 1, Synthesized))
```
timsueberkrueb, line 1358: The span here was nonsensical before.

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -674,20 +674,23 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
+  def boxExpr(): Term = {
     nonterminal:
-      val captures = `box` ~> backtrack(captureSet())
+      consume(`box`)
       val expr = if (peek(`{`)) functionArg()
-        else if (peek(`new`)) newExpr()
-        else callExpr()
-      Box(captures.unspan, expr)
-
+      else if (peek(`new`)) newExpr()
```
jiribenes, line 681: Do we need this case?
I think it was there before because of some shift-reduce conflict, but every `newExpr` is a `callExpr`? [?]

timsueberkrueb:
GitHub actions is having issues (also reflected on the [status page](https://www.githubstatus.com/).

PR #1021, opened 2025-05-24
Speed-up and cache conversion from offset to Kiama's position

jiribenes:
This makes the parser about 25-50% [!!!] quicker: (before vs after the first commit)
Screenshot 2025-05-24 at 23 12 19

Although we'll replace Kiama positions in the future, I think this still makes sense as it's a tiny fix that noticeably improves the performance...
Related to #744.

Everything works in the editor as of the second commit :)


b-studios:
Wow, that's great!

jiribenes:
Using binary search to convert an offset into a position https://github.com/effekt-lang/kiama/pull/12 makes this even a bit faster:
Screenshot 2025-05-25 at 14 55 07

jiribenes:
The measurements are sadly super unstable when I use `--time json` — sometimes some modules jump up temporarily...

PR #1023, opened 2025-05-26
Subtract block parameter captures in extern defs

### LINKED ISSUE #448: Extern definitions using blocks/functions
Using a block in an extern definition like this:
```
interface Bar {
  def bar(): Int
}
extern def foo(){ b: Bar } = 
  js "${b.bar()}"
  default {
    b.bar()
  }
```
currently fails with an error message about missing capabilities. 

(Future work from #427)

marzipankaiser:
This fixes #448 by subtracting the block param captures, so the following becomes OK:
```
extern def ifJS{ thn: => R }{ els: => R}: R =
  js { thn() }
  default { els() }
```

Note that the implementor of the extern is in charge of properly calling the block parameters with the correct calling convention.


marzipankaiser:
Should be subtracted, not annotated.

PR #1024, opened 2025-05-26
Implement Monomorphization optimization

mattisboeckle:
Aims to implement monomorphization for effekt.

- [x] Simple polymorphism
- [x] Functions with multiple polymorphic types
- [x] Different orders of types
- [x] Polymorphic data types
- [x] Polymorphic effects
- [x] Local polymorphic functions

The following programs have been used to (manually) test the implementation and should give an overview of what already works.

Simple polymorphic functions
```scala
def a[A](in: A) = { in }
def b[B](in: B): B = { a(in) }
def c[C](in: C): C = { b(in) }
def d[D](in: D): D = { c(in) }

def main() = {
  println(a(1))
  println(d("32"))
}
```
Multiple type arguments + switching the order of types
```scala
def f[C, D](in: C, in2: D) = in
def g[A, B](in: A, in2: B) = { f[B, A](in2, in) }
def main() = {
  println(g(3, "3"))
  println(g('c', false))
}
```

Polymorphic data types
```scala
type Maybe[A] {
  Nothing()
  Just(x: A)
}

def f[T](a: T): Maybe[T] = Just(a)

def main() = {
  val a = Just[Int](5)
  val b = Nothing[Char]()
  f(5)
  ()
}
```

Polymorphic effects
```scala
effect yield[A](x: A): Unit

def f() = {
  do yield[Int](123)
  do yield[String]("test")
}

def main() = {
  try {
    f()
  } with yield[Int] {
    x => println(x)
      resume(())
      resume(())
  } with yield[String] {
    x => println(x)
    resume(())
  }
  ()
}
```

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/LLVMTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/LLVMTests.scala
@@ -66,6 +66,9 @@ class LLVMTests extends EffektTests {
     examplesDir / "pos" / "type_parameters_blocks.effekt",
     examplesDir / "benchmarks" / "other" / "variadic_combinators.effekt",
     examplesDir / "benchmarks" / "other" / "unify.effekt",
+
+    // Normalizer timeout
+    examplesDir / "benchmarks" / "nofib" / "boyer.effekt"
   )
 
   override lazy val withoutOptimizations: Set[File] = Set(
--- a/effekt/jvm/src/test/scala/effekt/LLVMTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/LLVMTests.scala
@@ -58,6 +58,10 @@ class LLVMTests extends EffektTests {
     // re-entrant resumption
     examplesDir / "pos" / "issue1203.effekt",
     examplesDir / "pos" / "issue1203b.effekt",
+    
+    // Normalizer crash include again when fixed
+    examplesDir / "pos" / "raytracer.effekt",
+    examplesDir / "benchmarks" / "nofib" / "constraints.effekt",
   )
 
   override lazy val withoutOptimizations: Set[File] = Set(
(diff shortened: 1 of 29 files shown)


mattisboeckle:
This works for most programs now. There are still some open problems:
- [ ] Polymorphic recursion detection is more of a guessing game right now
- [ ] What to do with extern types
- [ ] A normalizer (?) bug, that removes `println` in ImpureApp, causing some tests to fail (because the output is empty)

examples/core/mono/boxed_type.ir
```diff
@@ -0,0 +1,28 @@
...
+}
+
+def main(f: () => Int at {io}) = {
+  val g = (id: ['A]('A) => 'A @ {})[() => Int at {io}](f: () => Int at {io});
+  (id: ['A]('A) => 'A @ {})[Int](42)
+}
+
+/// MONO_COLLECT_CONSTRAINTS
+
+<BOXED> <: id
```
b-studios: @mattisboeckle could you remind me what our design considerations here were?

b-studios: Is this just printed as "Boxed"? What if two different boxed types flow into the same type parameter?

b-studios:
@mattisboeckle Except the text file busy flaky test, the tests are now passing. Do you have something else in mind that we need to implement before merging? Could you review the changes?

b-studios:
Ready to merge :) Great job @mattisboeckle

examples/core/mono/boxed_type.ir
mattisboeckle: Then we should get two seperate flows of "Boxed" into the type parameter. We do not analyze the BlockType inside the Box.

Nevermind, just saw the next commit. That's way better

mattisboeckle:
LGTM. Thanks for cleaning this up!

PR #1025, opened 2025-05-26
Show documentation on hover

### LINKED ISSUE #1006: Show documentation on hover
> [!WARNING]
> Blocked on #930

On hover, show the attached documentation to that definition:

1. add a `documentation` to `SymbolInfo`:

https://github.com/effekt-lang/effekt/blob/5ad455b8b9ac2f84e68d7f1481ee9e4a280eb811/effekt/shared/src/main/scala/effekt/Intelligence.scala#L17-L40

2. then populate it here:

https://github.com/effekt-lang/effekt/blob/5ad455b8b9ac2f84e68d7f1481ee9e4a280eb811/effekt/shared/src/main/scala/effekt/Intelligence.scala#L194

marvinborner:
Closes #1006, blocked by #930.

### DIFF
--- a/examples/benchmarks/other/nbe.effekt
+++ b/examples/benchmarks/other/nbe.effekt
@@ -1,5 +1,6 @@
 /// Demo of an effectful Normalization by Evaluation (NbE) implementation for the pure, untyped lambda calculus.
 /// Uses the Call-by-Value reduction order, same as host.
+module examples/benchmarks/nbe
 
 import io
 import map


marvinborner:
Essentially this works now, except that the `getDefinitionOf` function not always links to the parent tree of the definition, but to its `IdDef` itself (e.g. for operations or constructors). Not sure if there's an easy fix for that, the way we lookup symbols is very mysterious.

Anyway, here's a demo: 

https://github.com/user-attachments/assets/1b549948-addb-46a2-aaed-888154c463f2

Btw, why can't I hover on `emit` in the signature? I don't think this ever worked and I don't understand why.

jiribenes:
It seems like this needs rebasing on #930, @marvinborner. 👀

timsueberkrueb:
Yes, otherwise it's a bit hard to review :) Please ping me once it's rebased!

marvinborner:
@jiribenes @timsueberkrueb rebased!

marvinborner:
Another mysterious CI error in `acme`: https://github.com/effekt-lang/effekt/actions/runs/15469750869/job/43550747148 can not reproduce!

timsueberkrueb (commented):
Great! Can you add a test case to `LSPTests.scala`?

effekt/shared/src/main/scala/effekt/Intelligence.scala
```diff
@@ -18,27 +18,36 @@ trait Intelligence {
...
+          |$sig$doc
           |""".stripMargin
     }
   }
 
+  def showDocumentation(documentation: Option[Doc]): String =
+    documentation.flatten
+      .map('\n' +: _)
+      .getOrElse("")
+      .replace("\\n", "\n")
```
timsueberkrueb, line 48: Why is this needed?

effekt/shared/src/main/scala/effekt/Intelligence.scala
```diff
@@ -110,6 +119,16 @@ trait Intelligence {
     case u => C.definitionTreeOption(u)
   }
 
+  def getDocumentationOf(s: Symbol)(using C: Context): Option[Doc] =
+    getDefinitionOf(s).asInstanceOf[Option[Any]] match {
+      case Some(p: Product) =>
+        p.productElementNames.zip(p.productIterator)
+          .collectFirst {
+            case ("doc", Some(s: String)) => Some(s)
```
timsueberkrueb: I wonder if there's a way to get rid of the magic string literal here somehow :thinking:?

timsueberkrueb (approved):
I'm happy with it if you add a test case so we notice if it breaks!

effekt/shared/src/main/scala/effekt/Intelligence.scala
jiribenes: Is .flatten.map f the same as .flatMap f?

effekt/shared/src/main/scala/effekt/Intelligence.scala
```diff
@@ -110,6 +119,16 @@ trait Intelligence {
     case u => C.definitionTreeOption(u)
   }
 
+  def getDocumentationOf(s: Symbol)(using C: Context): Option[Doc] =
```
marzipankaiser: Given that
https://github.com/effekt-lang/effekt/blob/6e8b2b515ebc45304b2d38ab161f9be0f7cd2e8d/effekt/shared/src/main/scala/effekt/source/Tree.scala#L109

Should we really return an `Option[Doc] = Option[Option[String]]`?
Is there a semantic difference between `None` and `Some(None)` and, if so, would some other way of communicating this difference be more appropriate/understandable?

effekt/shared/src/main/scala/effekt/Intelligence.scala
marzipankaiser: It seems we have:
https://github.com/effekt-lang/effekt/blob/6e8b2b515ebc45304b2d38ab161f9be0f7cd2e8d/effekt/shared/src/main/scala/effekt/source/Tree.scala#L109

But then, why do we pass an `Option[Doc]` (see other comment)?

Also, maybe instead of `.map(...).getOrElse(...)`, a `match` would be clearer?

marvinborner: Sorry for the late response.

> Why is this needed?

We only want to add a newline delimiter if the symbol actually has a doc comment, therefore we map on it and then .getOrElse. We need to unescape the newlines in order to show them in VS code as literal newlines (we build the string with escaped newlines in `RecursiveDescent`)

> Is `.flatten.map f` the same as `.flatMap f`

No, as it turns out. Also, I have changed the types from `Option[Doc]` to `Doc`, so this is not required anymore (see other comment)

> Also, maybe instead of .map(...).getOrElse(...), a match would be clearer?

It's of course a stylistic choice, but I think it is clear enough as is.

effekt/shared/src/main/scala/effekt/Intelligence.scala
marvinborner: You're right, I removed the additional Option. I initially used it to mirror the type of `signature` and `description` (`Option[String]`) and to be more flexible if we were to change the type of `Doc` in the future, but it turns out to make things a bit more unreadable.

effekt/shared/src/main/scala/effekt/Intelligence.scala
marvinborner: Not sure, I am not aware of it. I also do not really have a problem with this implementation, I think it is precisely the solution which our way of annotating `Doc`s requires us to use.

marvinborner:
The test case did indeed reveal a problem in the parsing of doc comments, where the initial doc comment was always ignored because of empty/implicit `ModuleDecl`s always parsing a doc comment. The solution is a bit ugly again and may (?) violate an invariant again, let me know if you find a better solution for only parsing doc comments in explicit `ModuleDecl`s. 

Other than that, this should be mergeable now.

effekt/shared/src/main/scala/effekt/Intelligence.scala
marzipankaiser: Could the `Tree`s that (potentially) have doc comments inherit from some `Documented` trait like:
```scala
trait Documented {
  def doc: Doc
}
// ...
enum Def extends Definition with Documented {...}
// ...
case class Constructor(..., doc: Doc, ...) extends ... with Documented
```
?
Then this could be a
```scala
case Some(d: Documented) => d.doc
```

Then we will probably get better compiler support.
Or am I misunderstanding something?

PR #1028, opened 2025-05-28
Expand stream module

phischu:


libraries/common/stream.effekt
```diff
@@ -615,6 +653,24 @@ def collectString[R] { stream: () => R / emit[Char] }: (R, String) = {
   (result, bytes.toString)
 }
 
+def readLine[R](){ body: => R / read[Char] }: R / read[Char] = {
```
jiribenes, line 656: syntax nitpick
```suggestion
def readLine[R] { body: => R / read[Char] }: R / read[Char] = {
```

libraries/common/stream.effekt
```diff
@@ -422,6 +438,10 @@ def each(string: String): Unit / emit[Char] =
 def collectString { stream: () => Unit / emit[Char] }: String =
   returning::collectString[Unit]{stream}.second
 
+def readLine(){ body: => Unit / read[Char] }: Unit / read[Char] =
```
jiribenes, line 441: syntax nitpick
```suggestion
def readLine { body: => Unit / read[Char] }: Unit / read[Char] =
```

jiribenes (commented):
👀

libraries/common/stream.effekt
```diff
@@ -30,6 +30,16 @@ effect stop(): Nothing
...
 
+/// Applies the given function on each emitted element and reemits the result.
+///
+/// e.g. map[Int,String]{ x => x.show } { range(0, 5) }
+def map[A, B] { function: A => B } { stream: () => Unit / emit[A] }: Unit / emit[B] =
+  returning::map[A, B, Unit]{function}{stream}
+
+/// Runs the given action whenever the given reader reads an element.
+def tap[A] { action: () => A / stop } { reader: () => Unit / read[A] }: Unit =
+  returning::tap[A, Unit]{action}{reader}
```
jiribenes, line 41: Is there a nice example use-case for this function? I cannot think of a way to use it for something meaningful right now...
If so, could we add it to the description? 🙏

jiribenes (approved):
LGTM, seems to work now, so I'll merge it.

PR #1029, opened 2025-05-28
Improve the context info sent for holes

timsueberkrueb:
This is the first step of improving the information the language server sends about holes and their context. In particular, we now preserve the scope that terms originate from, allowing us to display local variables first.
It does not (yet) order the bindings based on the order they were defined in because they are internally tracked in `HashMap`s which unfortunately also means that we apparently cannot show shadowed  terms at the moment?

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Intelligence.scala
+++ b/effekt/shared/src/main/scala/effekt/Intelligence.scala
@@ -140,43 +140,49 @@ trait Intelligence {
     innerType = hole.innerType.map { t => pp"${t}" }
     expectedType = hole.expectedType.map { t => pp"${t}" }
   } yield {
-    val BindingInfo(importedTerms, importedTypes, terms, types) = allBindings(scope)
-    HoleInfo(hole.name.name, hole.decl.span, innerType, expectedType, importedTerms.toList.distinct, importedTypes.toList.distinct, terms.toList.distinct, types.toList.distinct)
+    val scopeInfo = allBindings(scope)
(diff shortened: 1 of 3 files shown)


effekt/shared/src/main/scala/effekt/Intelligence.scala
```diff
@@ -140,43 +140,48 @@ trait Intelligence {
...
-  def allBindings(scope: Scope)(using C: Context): BindingInfo =
+  def allBindings(scope: Scope)(using C: Context): ScopeInfo =
     scope match {
       case Scope.Global(imports, bindings) =>
-        val (te1, ty1) = allBindings(imports)
-        val (te2, ty2) = allBindings(bindings)
-        BindingInfo(te1, ty1, te2, ty2)
+        val bindingsOut = allBindings(BindingOrigin.Imported, imports)
+          ++ allBindings(BindingOrigin.Defined, bindings)
+        ScopeInfo(None, ScopeKind.Global, bindingsOut.toList, None)
```
jiribenes, line 158: I'm a bit reluctant about this change -- is the repeated `++` on iterables here OK, especially perf-wise? 👀

timsueberkrueb: How would this change the performance characteristics? The same operations have been applied previously in `BindingInfo`, no?

b-studios (approved):

timsueberkrueb:
Can we only send hole info if panel is shown?

Issue #1026, opened 2025-05-26
Benchmarks run out of heap space in optimized JS


marvinborner:
The benchmark test suite of Effekt Plots fails since a few days.

The culprits are two JS test cases that run out of heap space: `resume_nontail 10000` and `sum_range 10000000`. Both OOMs are reproducible with a maximum heap size of `2GiB` but not with `4GiB` (the default on MacOS). Since the tests are rather minimal and require maximally a few MiB on LLVM, this is in fact a problem of the JS backend.

To continue benchmarking on Effekt Plots, a **temporary** solution would be to increase the default heap size or, better, allow passing custom flags to the JS (+ LLVM) runner from the Effekt CLI, potentially via environment variable. 

We have yet to determine the cause of the sudden change on plots, since no relevant code was changed and we cannot reproduce a working state on older versions.

PR #1031, opened 2025-06-01
Add support for custom flags based on environment

marvinborner:
Permits a temporary fix for #1026.

We can now do `NODE_FLAGS="--max-old-space-size=8192"` and it will automatically add the flags to the `node` runner flags (in shebang). I also added `VALGRIND_FLAGS`, `CHEZ_FLAGS`, and `CFLAGS` (for historic reasons)

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Runner.scala
+++ b/effekt/jvm/src/main/scala/effekt/Runner.scala
@@ -294,7 +294,7 @@ object LLVMRunner extends Runner[String] {
 
   def standardLibraryPath(root: File): File = root / "libraries" / "common"
 
-  val flags = Properties.envOrElse("LLVM_FLAGS", "").split(" ").toSeq
+  val flags = Properties.envOrElse("CFLAGS", "").split(" ").toSeq
 
   override def includes(path: File): List[File] = List(path / ".." / "llvm")
 


effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -186,7 +197,7 @@ object JSNodeRunner extends Runner[String] {
 
     os match {
       case OS.POSIX =>
-        val shebang = "#!/usr/bin/env node"
+        val shebang = s"#!/usr/bin/env -S node ${flags.mkString(" ")}"
```
marvinborner, line 200: Not sure how reliable `env -S` is for other platforms, but I do not see another elegant way of adding the flags to the node runner. See https://unix.stackexchange.com/questions/399690/multiple-arguments-in-shebang

marvinborner: I just remembered, this is very related to one of my first contributions #329, which was deemed inadequate! According to #857, this should be safe by now, though?

marvinborner:
No idea what's happening with CI, can not reproduce errors locally. CI also doesn't seem to have anything relevant in the environment. Hmmm.

PR #1032, opened 2025-06-02
Add flag to show prelude

marvinborner:
Solves #915 and allows some highlighting on `effekt-api`.

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Main.scala
+++ b/effekt/jvm/src/main/scala/effekt/Main.scala
@@ -21,7 +21,9 @@ object Main {
         return
     }
 
-    if (config.server()) {
+    if (config.showPrelude()) {
+      println(config.prelude().mkString(","))
+    } else if (config.server()) {
       val serverConfig = ServerConfig(
         debug = config.debug(),
         debugPort = config.debugPort()
--- a/effekt/jvm/src/main/scala/effekt/Repl.scala
+++ b/effekt/jvm/src/main/scala/effekt/Repl.scala
@@ -41,11 +41,6 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
    * Called by the Driver if no arguments are provided to the Effekt binary
    */
   def run(config: EffektConfig): Unit = {
-    if (config.showPrelude()) {
-      println(config.prelude().mkString(","))
-      return
-    }
-
     val out = config.output()
     out.emitln(banner)
     usingCommandHistory(config) {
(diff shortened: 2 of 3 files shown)


effekt/jvm/src/main/scala/effekt/Repl.scala
```diff
@@ -41,6 +41,11 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
    * Called by the Driver if no arguments are provided to the Effekt binary
    */
   def run(config: EffektConfig): Unit = {
+    if (config.showPrelude()) {
+      println(config.prelude().mkString(","))
+      return
+    }
```
marvinborner: I don't like this, where should we put this instead?

effekt/jvm/src/main/scala/effekt/Main.scala
```diff
@@ -21,7 +21,9 @@ object Main {
         return
     }
 
-    if (config.server()) {
+    if (config.showPrelude()) {
+      println(config.prelude().mkString(","))
+    } else if (config.server()) {
```
jiribenes, line 26: Does this interact correctly with the `--prelude` flag?

Also I've noticed (again) that we have two different prelude definitions:
1. a default one in the runner https://github.com/effekt-lang/effekt/blob/3cacf5950c337fa2c5537c0f211cb4ed3e53809f/effekt/jvm/src/main/scala/effekt/Runner.scala#L46
2. in the ScalaJS backend config https://github.com/effekt-lang/effekt/blob/3cacf5950c337fa2c5537c0f211cb4ed3e53809f/effekt/js/src/main/scala/effekt/EffektConfig.scala#L20
3. in the ScalaJVM backend https://github.com/effekt-lang/effekt/blob/3cacf5950c337fa2c5537c0f211cb4ed3e53809f/effekt/jvm/src/main/scala/effekt/EffektConfig.scala#L245

Is there a way to unify these a bit more? WDYT?

marvinborner: > Does this interact correctly with the --prelude flag?

Yes, `effekt --prelude=effekt,option --show-prelude` will return `effekt,option`

> Is there a way to unify these a bit more? WDYT?

Sure, we could probably define the default prelude list in a shared file, not sure which file this would be though.

jiribenes: OK, I'll merge this, but we ought to make a followup issue, I think :)

jiribenes (approved):

PR #1035, opened 2025-06-04
Deliver holes for literate Effekt files

timsueberkrueb:
Previously, we only published holes for `.effekt` but not for `effekt.md`. The reason for this is that for `*.md` files we construct a new `MarkdownSource` in `compileSource` and we attach the annotations to that, but we pass the old `StringSource` back to `afterCompilation`, meaning we don't find the annotations.
This PR fixes this issue by passing the correct source to `afterCompilation`.


effekt/jvm/src/main/scala/effekt/Driver.scala
```diff
@@ -35,64 +35,80 @@ trait Driver extends kiama.util.Compiler[EffektConfig, EffektError] { outer =>
...
    * In LSP mode: invoked for each file opened in an editor
    */
-  override def compileSource(source: Source, config: EffektConfig): Unit = try {
-    val src = if (source.name.endsWith(".md")) { MarkdownSource(source) } else { source }
+  override def compileSource(source: Source, config: EffektConfig): Unit = {
+    val src = if (source.name.endsWith(".md")) {
+      MarkdownSource(source)
+    } else {
+      source
+    }
```
jiribenes: This confused me at first since I couldn't tell what changed -- only the `try` got moved, right?

jiribenes (commented):
I honestly cannot tell what changed here -- the `try` shifted, but there's a lot of changes that don't seem related to what happened at all? Is there something I've missed in the new body of the `try`?

effekt/jvm/src/main/scala/effekt/Driver.scala
timsueberkrueb: It is hard to see indeed; indenting the code triggered a reformat by intelliJ. If you want I can try to use another tool to reproduce the change with smaller diff.
What changed is: `src` is being initialized before the `try` block, the statements in finally use `src` rather than `source`.

jiribenes (approved):
I _think_ this is OK, but it's pretty hard to see if the try-catch-finally logic is 100% sound.
Feel free to merge if you're happy with this, @timsueberkrueb

Issue #430, opened 2024-03-22
Misleading error "recursive functions need annotated return types" on return type mismatch


jiribenes:
Here's the whole error: notice that the real problem is that the actual return type doesn't match the expected return type --- this has nothing to do with recursive / forward-declared functions!
<img width="666" alt="Screenshot 2024-03-22 at 16 53 41" src="https://github.com/effekt-lang/effekt/assets/11269173/a4599aa3-6151-4e33-9bbe-a0887310d4da">

Interestingly, it goes away if we try just functions with numbers. 🤔 

Repro:
```
def foo(): Unit = {
  val x = Nil[Int]()
  head(x)
}
```

---

Originally reported by @mlutze

jiribenes:
Fascinatingly, the error doesn't go away if I annotate the return type as Int:
```
def foo(): Int = {
  val x = Nil[Int]()
  head(x)
}
```

So this is a bug altogether -- is it perhaps a "unhandled effect" error? 🤔

jiribenes:
As @marvinborner noted, this issue is fixed by handling the `Exception` effect.
The following code doesn't report the error anymore:
```scala
def foo(): Int = {
  val x = Nil[Int]()
  try {
    head(x)
  } with Exception[EmptyList] {
    def raise(exception: EmptyList, msg: String) = {
      println(msg); 0
    }
  }
}
```

jiribenes:
The issue is also fixed by annotating the `Exception` effect:
```scala
def foo(): Int / Exception[EmptyList] = {
  //             ^^^^^^^^^^^^^^^^^^^^
  val x = Nil[Int]()
  head(x)
}
```

b-studios:
Strangely it does report the correct error in the following case:

```
def myHead[A](l: List[A]) = head(l)

def foo(): Unit = {
  val x = Nil[Int]()
  myHead(x)
}
```

jiribenes:
[...]

PR #1038, opened 2025-06-04
Preserve orders of symbols in scope

timsueberkrueb:
For the holes panel, based on our own experiments and feedback from @jiribenes, I propose to display the bound names in the order they were defined in.
I.e., if you have the program

```scala
def foo(x: Int, y: Int, z: Int) = {
  val boo = "whatever"

  def bar(a: Int, b: Int) = <>

  val quack = "whatever"
}
```

with this PR, the context displays as follows:

image

In particular, within a scope, the variables are ordered as they were defined, rather than displayed in an arbitrary order (as they are now).

This is implemented by sorting symbols by the span of their declarations. To realize this, we had to add many more `decl` back-references to symbols.

Alternatives considered relied on the insertion order in `Namer`, which we cannot rely on as some symbols are added in `resolve`, some in `preresolve`.

### DIFF
--- a/effekt/js/src/main/scala/effekt/context/VirtualModuleDB.scala
+++ b/effekt/js/src/main/scala/effekt/context/VirtualModuleDB.scala
@@ -1,8 +1,7 @@
 package effekt
 package context
 
-import effekt.util.
-Resources
+import effekt.util.Resources
 import effekt.util.paths._
 
 import kiama.util.Source
--- a/effekt/js/src/main/scala/effekt/context/VirtualModuleDB.scala
+++ b/effekt/js/src/main/scala/effekt/context/VirtualModuleDB.scala
@@ -1,7 +1,8 @@
 package effekt
 package context
 
-import effekt.util.Resources
+import effekt.util.
+Resources
 import effekt.util.paths._
 
 import kiama.util.Source
--- a/effekt/shared/src/main/scala/effekt/Intelligence.scala
+++ b/effekt/shared/src/main/scala/effekt/Intelligence.scala
@@ -197,15 +197,10 @@ trait Intelligence {
       })
     }
 
-    sorted.flatMap((name, sym) => sym match {
-      case sym: TypeSymbol =>
-        // TODO this is extremely hacky, printing is not defined for all types at the moment
-        try { Some(TypeBinding(path, name, origin, DeclPrinter(sym))) } catch { case e => None }
-      case sym: TermSymbol =>
-        sym match {
-          case sym: ValueSymbol => Some(TermBinding(path, name, origin, C.valueTypeOption(sym).map(t => pp"${t}")))
-          case sym: BlockSymbol => Some(TermBinding(path, name, origin, C.blockTypeOption(sym).map(t => pp"${t}")))
-        }
+    sorted.map((name, sym) => sym match {
+      case sym: TypeSymbol => TypeBinding(path, name, origin, DeclPrinter(sym))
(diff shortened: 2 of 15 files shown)


b-studios:
I like this behavior! How to implement it is a different question. 

We are again in this state where we have two seemingly orthogonal set of constraints.

1. batch compilation should be fast (and the compiler should be simple)
2. editor services _might_ require some info, so we need to maintain it

Just some random observations:

- right now we only ever need this info whenever there is a hole (we could however also use it later at a cursor position).
- in other places we spam the annotations DB with additional info

Right now it looks like you could use a custom "OrderedMap" as a drop-in replacement (like https://scastie.scala-lang.org/eHJNORkYRGKoBnrCIIAtqQ). 

However, I don't understand why you keep a list of Strings and not a list of symbols.

timsueberkrueb:
Good idea with the `OrderedMap`. There are two more constraints.

1. As we have overloading, we want to track the order of terms/types with the same name. This means we need `OrderedMap[String, List[TermSymbol]` rather than `OrderedMap[String, Set[TermSymbol]`
2. We probably don't want separate maps for types and terms, but rather a single `OrderedMap`.

timsueberkrueb:
This is further complicated by the fact that `Namer` has two phases: `resolve` and `preresolve`. Some things like namespaces and definitions are inserted in `preresolve`, others in `resolve`. This means we cannot rely on the insertion order.

timsueberkrueb:
We could use the spans of definitions, though this is complicated by the symbols not carrying spans.

b-studios:
A symbol should link to its definition, which has the span, no?

jiribenes:
RE last commit: does this now fix https://github.com/effekt-lang/effekt/issues/430? 🤔

timsueberkrueb:
Looks like it might? At least the reproducing examples in the issue seem to show better errors now.

effekt/shared/src/main/scala/effekt/Intelligence.scala
```diff
@@ -182,23 +182,34 @@ trait Intelligence {
...
-    }
-    val nestedBindings = bindings.namespaces.flatMap {
-      case (name, namespace) => allBindings(origin, namespace, path :+ name)
-    }
+      case sym: TermSymbol =>
+        sym match {
+          case sym: ValueSymbol => Some(TermBinding(path, name, origin, C.valueTypeOption(sym).map(t => pp"${t}")))
+          case sym: BlockSymbol => Some(TermBinding(path, name, origin, C.blockTypeOption(sym).map(t => pp"${t}")))
+        }
+    }).toList
```
marzipankaiser: IIUC, the only reason to use a `flatMap` instead of `collect` here is the hacky catchall?
If so, how much work would it be to get rid of that? (When and where does this actually get triggered?)

timsueberkrueb: I couldn't find an example where this actually fails, so I removed the catch-all. If we find an example that fails, we can fix it properly.

effekt/js/src/main/scala/effekt/context/VirtualModuleDB.scala
```diff
@@ -1,7 +1,8 @@
 package effekt
 package context
 
-import effekt.util.Resources
+import effekt.util.
+Resources
```
marzipankaiser:
```suggestion
import effekt.util.Resources
```
Was this by accident?

PR #1042, opened 2025-06-06
Fix core type in `try ... with ...` body

### LINKED ISSUE #1040: Typed hole changes semantics of program, even when not evaluated
I was doing the `hurl` exercise from https://effekt-lang.org/docs/exercises/hurl, and noticed that typed holes don't seem to behave like noop placeholders as they should.

Here is a version that computes factorial correctly.
```
// 1. define the effect operation and any useful definitions here
effect hurl(n: Int): Unit

// 2. put your code for factorial here
//    The type of factorial should be `Unit` and it should use the effect operation you defined
def factorial(n: Int): Unit / hurl = {
  do hurl(
    if (n == 0) {
      0
    } else if (n == 1) {
      1
    } else {
      n * (
        try { factorial(n - 1); 0 } // <-- Replace '0' with '<>' and a different value is produced
        with hurl { v => v }
      )
    }
  )
}


// 3. put the try { ... } catch as x { ... } here
//    (something like `try { factorial(10); "" } with ...` in Effekt)
//    You should return the string "factorial(10) = ..." here
//
// Don't change the return type or effect set please!
def result(): String / {} = {
  try {
    factorial(10);
    "No return?" // <-- Replace '"No return?"' with '<>' and nothing gets printed
  } with hurl { n =>
    "factorial(10) = " ++ show(n)
  }
}

def main() = {
  println(result())
}
```

In the file I put two comments with "<-- Replace". If you replace either place with `<>` you get unexpected behavior, even though those places in code should never be evaluated (as far as I know).

I ran the code using `npx @effekt-lang/effekt hurl.effekt`

marzipankaiser:
Fixes #1040 by transforming the inferred type from source (instead of inferring the type of the transformed expression).


jiribenes (approved):
Seems to strictly improve state of the art, we can discuss the nuances at the meeting, but I'll already merge this.

PR #1044, opened 2025-06-10
Forward reference computation section

dvdvgt:
Resolves https://github.com/effekt-lang/effekt-website/issues/116


dvdvgt:
I also made sure the plural of computation (computations) is used consistently

PR #1045, opened 2025-06-10
Update scala version to 3.3.6

dvdvgt:
The current version of the Scala compiler (3.3.1) is no longer supported by Metals LSP. We should move to the latest LTS version.


dvdvgt:
https://github.com/effekt-lang/kiama/pull/13 needs to be merged first.

jiribenes (approved):
LGTM, I think

Issue #845, opened 2025-02-24
Universal collection syntax


dvdvgt:
While writing tests for the newly added array function in the stdlib (#839), I have found myself using the `array::fromList` function quite often. We also have briefly discussed this in one of the previous Effekt meetings; what if we had a shared syntax for creating collections?

## Idea

We re-use the `emit` effect:

```
array#[1, 2, 3]
```

is converted to

```
collectArray {
  do emit(1)
  do emit(2)
  do emit(3)
}
```

Note that `collectArray` and other functions for collecting (ha!) collections already in the [stream implementation](https://github.com/effekt-lang/effekt/blob/d57b530c328437c62b6c4a4b1e0a0651785aab3c/libraries/common/stream.effekt#L169-L214)
such this would merely be a change in the parser similar to #743.

Of course, this would also work for `bytearray`:

```
bytes#[1.toByte, 2.toByte, 3.toByte]
```

is converted to

```
collectBytes {
  do emit(1.toByte)
  do emit(2.toByte)
  do emit(3.toByte)
}
```

## Caveats

- Possible performance & overhead penalty.
- The syntax does not work for maps, sets and other collections that need a comparison function. I can see this changing if and once Effekt has type classes.

## Benefits

- Nice and unique usage of effects!
- Better ergonomics

jiribenes:
A few weeks ago, we realised that if we allow dot notation on _computations_, we could also have `[1, 2, 3].array` (where `[1, 2, 3] would be a computation by default) :^)

jiribenes:
@marzipankaiser talked about UFCS on computations a while back in the context of extension methods:
```scala
interface State[T] {
  def get(): T
  def set(x: T): Unit
}

def increment() { st: State[T] }: Unit = {
  val current = st.get()
  st.set(current + 1)
}

def st: State[T] = ...

increment() {st}
```
~>
```
st.increment()
```

@phischu talked about UFCS support for streams (which is very similar here). `[1, 2, 3].map { x => x + 2 }.collectList...`

jiribenes:
Just to add to this, we technically have _stream_ literals already: `[1, 2, 3].each` :)
If we had UFCS on computations, you could get a `set` with `[1, 2, 3].each.collectSet`.

We could invert it, have `[1, 2, 3]` as a stream literal, and have `[1, 2, 3].collectList` as a list literal, `[1, 2, 3].collectSet` as a set literal, `[1, 2, 3].collectArray` as a array literal, etc.
[...]

PR #1046, opened 2025-06-10
Feature: UFCS on blocks

dvdvgt:
This PR adds UFCS for blocks such that you can write `{ x => x + 42 }.foo { y => x * 2 }`, for example.

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/RecursiveDescentTests.scala
@@ -311,6 +311,14 @@ class RecursiveDescentTests extends munit.FunSuite {
     intercept[Throwable] { parseExpr("[,1]") }
   }
 
+  test("ufcs") {
+    parseExpr("{ l }.foo { x => x + 1}.bar { x => x * 2}")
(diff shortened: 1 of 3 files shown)


dvdvgt:
Current caveat: When expecting a statement, a UFCS call with blocks is not allowed. For example

```
val x = { f1 }.foo { f2 }
              ^
expected ; but found .
```

fails since it expects a `BlockStmt`. This only works when explicitly returning:

```
val x = return { f1 }.foo { f2 }
```

jiribenes:
Related discussion: https://github.com/effekt-lang/effekt/issues/845

jiribenes:
What about, say, 
```effekt
interface State[T] {
  def get(): T
  def set(x: T): Unit
}

def increment() { st: State[T] }: Unit = {
  val current = st.get()
  st.set(current + 1)
}

def st: State[T] = ...

increment() {st}
```

Can we now write `st.increment()` or does it have to be `{st}.increment()`? Or is it not supported at all?

jiribenes:
Superseded by #1250 which also implemented similar changes in the parser, then gave up on block literals due to inherent limitations (see there).

PR #1050, opened 2025-06-12
Only accept config under "effekt" key

timsueberkrueb:
Unifies the server side with https://github.com/effekt-lang/effekt-vscode/pull/91. I think we should wait with merging this at least until a new release of the VSCode extension has been made. We could also have a longer transition period, since there is urgency to stop accepting configuration not nested under an `"effekt"` key.


timsueberkrueb:
The VSCode extension version `v0.4.3` or later sends the new format.

jiribenes (approved):

timsueberkrueb:
@marzipankaiser suggested to show a warning prompting to update the extension if that's easy to do.

PR #1052, opened 2025-06-12
Concurrency primitives

phischu:


libraries/llvm/io.c
```diff
@@ -295,154 +295,99 @@ void c_yield(Stack stack) {
...
+            f->payload.stack = stack;
+            erasePositive(channel);
             break;
-        case RESOLVED:
-            value = p->payload.value;
-            sharePositive(value);
+        }
+        case SENDED: {
+            struct Pos value = f->payload.value;
+            f->state = EMPTY;
```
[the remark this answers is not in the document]
phischu: Nono, the value is moved!

libraries/llvm/io.c
```diff
@@ -295,154 +295,99 @@ void c_yield(Stack stack) {
...
+        }
+        case WAITED: {
+            erasePositive(channel);
+            eraseStack(stack);
+            // TODO more graceful panic
+            fprintf(stderr, "ERROR: Channel already waited\n");
+            exit(1);
+            break;
+        }
+    }
```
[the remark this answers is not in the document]
phischu: This is C and this is an assertion violation.

phischu:
Okay, I am happy for now, we should merge it and later improve based on concrete usage.

libraries/common/io/channel.effekt
```diff
@@ -0,0 +1,49 @@
...
+
+type Sender[A, B] = Ref[Node[A, B]]
+type Receiver[A, B] = Ref[Node[A, B]]
+
+namespace channel {
+  def allocate[A, B](): (Sender[A, B], Receiver[A, B]) = {
+    val node = Node(signal::allocate(), ref(None()))
+    (ref(node), ref(node))
+  }
+}
```
b-studios, line 16: What about moving it out of the namespace and calling it `channel`?

libraries/common/io/promise.effekt
```diff
@@ -0,0 +1,49 @@
...
+
+// Promises
+// --------
+
+type State[T] {
+  Resolved(value: T)
+  Pending(signals: List[Signal[T, Unit]])
+}
+
+type Promise[T] = Ref[State[T]]
```
b-studios, line 14: Is it possible to wrap JS promises in our promises?

b-studios: Could you add a doc comment mentioning examples of how to use promises?

phischu: > Is it possible to wrap JS promises in our promises?

Yes, but it would not be idiomatic Effekt. The following is how you use a function that returns a promise in JavaScript.

```
extern def fetch(url: String) at async: String =
  js "$effekt.capture(callback => fetch(${url}).then(r => r.text().then(t => callback(t))))"

def main() = {
  println(fetch("http://example.com"))
}
```

> Could you add a doc comment mentioning examples of how to use promises?

There are examples in the examples folder.

libraries/common/io/channel.effekt
```diff
@@ -0,0 +1,47 @@
...
+
+record Node[A, B](this: Signal[A, B], rest: Ref[Option[Node[A, B]]])
+
+type Sender[A, B] = Ref[Node[A, B]]
+type Receiver[A, B] = Ref[Node[A, B]]
+
+def channel[A, B](): (Sender[A, B], Receiver[A, B]) = {
+  val node = Node(signal(), ref(None()))
+  (ref(node), ref(node))
+}
```
jiribenes, line 16: @phischu Feel free to: edit this comment & add the invariant, then click the button to commit this suggestion :)
```suggestion
/// Creates a new channel consisting of a linked `Sender` and a `Receiver`.
/// Invariant: TODO @phischu
///
/// Example:
/// ```effekt
///  val (sender, receiver) = channel()
///  val closed = ref(false)
///
///  spawn(box {
///    each(0, 42) { i =>
///      sender.send(i)
///    }
///    closed.set(true)
///  })
///
///  var s = 0
///  while (not(closed.get())) {
///    s = s + receiver.receive(())
///  }
///  return s
/// ```
def channel[A, B](): (Sender[A, B], Receiver[A, B]) = {
  val node = Node(signal(), ref(None()))
  (ref(node), ref(node))
}
```

libraries/common/io/promise.effekt
```diff
@@ -0,0 +1,49 @@
...
+
+namespace promise {
+  def make[T](): Promise[T] =
+    ref(Pending(Nil()))
+}
+
+def promise[T](task: Task[T]): Promise[T] = {
```
jiribenes: cc @phischu feel free to modify, adding invariants if there are any.
```suggestion
/// Creates an async `Promise[T]` from a `Task[T]` (a callback `() => T at {io, global, async}`).
/// Use `await` to get the value of type `T` out of it.
///
/// Example:
/// ```effekt
///  val p1 = promise(box {
///    println("Start p1");
///    sleep(250)
///    println("Stop p1");
///    1
///  })
/// ```
def promise[T](task: Task[T]): Promise[T] = {
```

jiribenes: Is the `Task[T]` alias actually helpful here? IME it only confuses students. We only use this alias in two places, would it make sense to expand it?

phischu: I remove it, thanks.

PR #1055, opened 2025-06-17
Natural language holes

timsueberkrueb:
Introduces holes that can contain arbitrary natural language text. The syntax is as follows:

```scala
def foo(x: Int): Int = <"
    Holes can splice in Effekt expressions:
    ${ x }
    or statements:
    ${ do bar(); println("hey") }
">
```

The "close hole" code action now only shows if there is a hole with a single splice (this could be improved in the future, but I consider it outside the scope of this PR).

Similarly, the extension currently expects to receive a single (optional) *inner type* for each hole. This PR only provides this if there is a single splice.

This PR preserves the existing behavior of `<>` and `<{ ... }>`.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Lexer.scala
+++ b/effekt/shared/src/main/scala/effekt/Lexer.scala
@@ -493,7 +493,7 @@ class Lexer(source: Source) {
   }
 
   /** Matches a hole containing natural language text.
-   * Holes may contain unquotes, i.e., "${ do foo }" that include arbitrary expressions.
+   * Holes may contain unquotes, e.g., "${ do foo }" that include arbitrary expressions.
    * Unlike regular string literals, holes cannot contain escape sequences.
    * They do emit [[TokenKind.Str]] tokens with the `multiline` flag set to `true`.
    */
@@ -518,7 +518,7 @@ class Lexer(source: Source) {
           return TokenKind.HoleStr(stringContent.mkString)
         }
         case None =>
-          return err("Unterminated string.", start, start + offset)
+          return err("Unterminated hole.", start, start + offset)
         case Some(c) =>
           consume()
           stringContent.addOne(c)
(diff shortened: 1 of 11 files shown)


effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1128,18 +1128,35 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
+        }
         case k => fail("hole", k)
       }
     }
 
+  def holeTemplate(): Template[Stmt] =
+    nonterminal:
+      val first = holeString()
+      val (s, strs) = manyWhile((`${` ~> stmts() <~ `}`, holeString()), `${`).unzip
+      Template(first :: strs, s)
```
dvdvgt, line 1157: Can you use `labelled "hole"` here for better error messages?

effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -489,6 +492,42 @@ class Lexer(source: Source) {
     TokenKind.Str(stringContent.mkString, delim.isMultiline)
   }
 
+  /** Matches a hole containing natural language text.
+   * Holes may contain unquotes, e.g., "${ do foo }" that include arbitrary expressions.
+   * Unlike regular string literals, holes cannot contain escape sequences.
+   * They do emit [[TokenKind.Str]] tokens with the `multiline` flag set to `true`.
+   */
+  def matchHole(continued: Boolean = false): TokenKind = {
```
b-studios, line 500: I would have hoped that we could reuse the existing template infrastructure. But eventually we probably want to customize it anyways.

b-studios (approved):
Looks good. Let's see how this feels in practice before we bike-shed syntax any further.

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
timsueberkrueb: Looked into this, I don't think this will ever be displayed due to how the parser is guarded with `isHole`. We can add it later if there is an example where it would be shown.

effekt/shared/src/main/scala/effekt/Lexer.scala
timsueberkrueb: I considered this, but it made the template string implementation quite complex by adding new cases. Also there are some differences such as holes not needing escape characters. Therefore I think this is the better trade-off for now.

PR #1058, opened 2025-06-23
Add tcp servers and clients to standard library

phischu:
- [x] Fix warnings
- [x] Fix bug with `server_client` and more than 2 clients
- [x] Do todos
- [x] High-level streaming

The web does not support TCP and for Node this API is too low level. If we wanted to create node servers eventually, we would wrap a more high-level API.


libraries/llvm/io.c
```diff
@@ -178,6 +179,272 @@ void c_fs_mkdir(String path, Stack stack) {
...
+    buf->base = read_closure->data;
+    buf->len = read_closure->size;
+}
+
+void c_tcp_read(Int handle, struct Pos buffer, Int offset, Int size, Stack stack) {
+    uv_stream_t* stream = (uv_stream_t*)handle;
+
+    char* buffer_ptr = (char*)(c_bytearray_data(buffer) + offset);
+    erasePositive(buffer);
+    // TODO panic if this was the last reference
```
phischu, line 266: This is wrong here

phischu:
Perhaps the other leak is the same as this one: https://github.com/libuv/libuv/issues/559

marvinborner:
Looks indeed similar, I believe the 128 byte object to be some internal UV structure.

phischu:
The space leak was a stack that does not get freed when using `eraseStack`, but does when running it to completion. We should investigate this separately, but I do note that `@topLevelEraser` does not erase the stack's memory.

libraries/common/io/network.effekt
```diff
@@ -1,84 +1,102 @@
...
-  """
 
-  extern type JSServer // = net.Server
-  extern type JSSocket // = net.Socket
-  extern io def server(): JSServer =
-    jsNode "net.createServer()"
-  extern io def listen(server: JSServer, port: Int, host: String, listener: JSSocket => Unit at {io, async, global}): Unit =
-    jsNode "listen(${server}, ${port}, ${host}, (socket) => $effekt.runToplevel((ks, k) => (${listener})(socket, ks, k)))"
+/// A TCP handle. Should not be inspected.
+type Connection = Int
```
jiribenes, line 10:
```suggestion
/// A TCP handle. Should not be inspected.
extern type Connection 
  // = llvm "Int"
```
Could we make it an extern type so that it cannot be inspected?

(also applies to `Listener` below)

phischu: Yes, I don't like it (because then it is a `Pos`) but I guess I'll have to do it, because the representation in `js` is different.

jiribenes: Approaching the problem from the other side: does it have to be a `Pos`? We could let the user choose, no?

Couldn't we fix this "properly" by some `= llvm "i64"` annotation there instead? 🎣

libraries/common/io/network.effekt
```diff
@@ -1,84 +1,121 @@
...
-      respond("HTTP/1.1 400 Bad Request\r\n\r\n")
+/// Stops a TCP listener and releases associated resources.
+def shutdown(listener: Listener): Unit =
+  internal::shutdown(listener)
+
+
+def streaming[R](connection: Connection) { stream: () => R / {read[ByteArray], emit[ByteArray]} }: R / Exception[IOError] =
+  try {
+    val r = stream()
+    close(connection)
```
phischu: Should we close the connection? Perhaps we should not and have functions `server` and `client` that both use this `streaming` function and open and close the connection...

b-studios: At a first first glance it might be surprising that streaming closes the connection here.

mattisboeckle:
LGTM

PR #1060, opened 2025-06-25
Add shebang support

### LINKED ISSUE #1057: Shebang support
I think it's fun to have scripts written in Effekt that are directly invokable.
To that end, we might want to strip the first few lines starting with `#!` from source files...

Here's an example:
```scala
#!/usr/bin/env effekt

def main() = println("hello")
```  

Here's a different example using `effekt-nix`:
```scala
#!/usr/bin/env nix
#!nix run github:jiribenes/effekt-nix

def main() = println("hello")
```

marvinborner:
Closes #1057 by skipping shebangs in the header of a file.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Lexer.scala
+++ b/effekt/shared/src/main/scala/effekt/Lexer.scala
@@ -219,6 +219,19 @@ class Lexer(source: Source) {
       case _ => ()
     }
 
+  def consumeUntilNewline(): Unit = {
+    var reachedNewline = false
+    while (!reachedNewline) {
+      // peek to ensure the newline is not part of the comment
+      peek() match {
+        // comment is terminated when encountering a new line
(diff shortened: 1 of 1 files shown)


effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -556,6 +557,21 @@ class Lexer(source: Source) {
...
+    var reachedNewline = false
+    while (!reachedNewline) {
+      // peek to ensure the newline is not part of the command
+      peek() match {
+        // command is terminated when encountering a new line
+        case Some('\n') => reachedNewline = true
+        case None => reachedNewline = true
+        case _ => consume()
+      }
+    }
```
jiribenes: This repeats in the `matchComment` function too, could we extract it to some `consumeUntilNewline()` helper? (not sure what the naming scheme is)

PR #1062, opened 2025-06-26
Make language server send hole bodies

timsueberkrueb:


effekt/shared/src/main/scala/effekt/Intelligence.scala
```diff
@@ -414,6 +462,18 @@ object Intelligence {
...
+  }
+
+  object HoleItemKind {
+    final val NaturalLanguage = "NaturalLanguage"
+    final val Code = "Code"
+  }
+
+  case class NaturalLanguage(kind: String, text: String) extends HoleItem
+  case class Code(kind: String, text: String, `type`: Option[String]) extends HoleItem
+
```
marzipankaiser: Is there a reason to have a stringly-typed enum here in addition to the different case classes?
Wouldn't it make sense to then at least use something like
```
  case class Code(text: String, `type`: Option[String]) extends HoleItem {
      def kind: String = HoleItemKind.Code
  }
```
(or with `val` if it needs to be one).

If required for the parser, maybe it would make sense to at least document that invariant, e.g. by
```scala
case class Code(kind: String, ...) extends HoleItem {
  require(kind == HoleItemKind.Code)
}
```
?

marzipankaiser: Also, it might pay off to have a smart constructor then additionally, as in:
```
object Code {
   def apply(text: String, `type`: Option[String]): Code =
       Code(HoleItemKind.Code, text, `type`)
}
```
or sth. to prevent us from having to provide the redundant information on every manual instantiation in the tests.

timsueberkrueb: Yes, thanks for catching this!

timsueberkrueb: We do need the field, though, because otherwise the runtime reflection doesn't generate the correct JSON.

marzipankaiser (approved):
Thanks, looks good to me.

PR #1066, opened 2025-07-02
Fully resilient lexer

jiribenes:
Resolves https://github.com/effekt-lang/effekt/issues/522

The lexer is now completely resilient -- should **never** fail.
Instead, it produces an `TokenKind.Error(msg: TokenError)` token which is currently ignored by the recursive descent parser. 

The whole thing is now an iterator (`class Lexer extends Iterator[Token]`) which simplifies the design a bit.
It also allows us to go much faster, the lexer is now up to 2x faster.

There's also a drive-by in the parser -- we now report all soft fails (`softFail`) after encountering a fatal failure (`fail`). This means we can aggregate a few more parse errors when needed.

Additional drive-by: the lexer is now used to highlight code in the REPL and RecursiveDescent can finally be renamed to Parser


b-studios:
Cool!

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1105,6 +1106,18 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
+      nonterminal:
+        val startPosition = position
+        manyUntil(next(), `}$`)
+        val endPosition = position
+        if peek(`}$`) then consume(`}$`) else if peek(`}`) then consume(`}`)
+
+        val msg = s"Unexpected splice: string interpolation $${ ... } is only allowed in strings!"
+        softFail(msg, startPosition, endPosition)
+
+        UnitLit(span().synthesized) // XXX(jiribenes, 2025-07-01): HACK!
```
jiribenes: We should really add `Error` trees for this very purpose instead of misusing `UnitLit`

effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -114,593 +137,580 @@ enum TokenKind {
...
+
+      case '}' =>
+        if isAtInterpolationBoundary then
+          handleInterpolationEnd()
+          TokenKind.`}$`
+        else
+          braceTracker.decrementBraces
+          nextChar match
+            case '>' => advance2With(TokenKind.`}>`)
+            case _ => advanceWith(TokenKind.`}`)
```
jiribenes: I do actually use the `}$` token here for pure flavour, though I'm not sure it's a great idea?

b-studios: Is it somehow necessary to distinguish from `}`?

jiribenes: Not at all, though it might be a bit better for recovery, perhaps? I just thought it's cute.

dvdvgt:
Perhaps this PR is related and can be incorporated? #1013

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -89,12 +84,25 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
+        case LexerError.EmptyCharLiteral => TokenKind.Chr('?')
+        case LexerError.MultipleCodePointsInChar => TokenKind.Chr('?')
+        case LexerError.InvalidUnicodeLiteral => TokenKind.Chr('?')
+        case LexerError.InvalidIntegerFormat => TokenKind.Integer(0)
+        case LexerError.InvalidDoubleFormat => TokenKind.Float(java.lang.Double.NaN)
+        case _ => fail(err.message)
+
+      // NOTE: This relies on the fact that the `fail` above ^ throws!
+      softFail(err.message, tokenPosition, tokenPosition)
+      token.copy(kind = recoveredKind)
```
jiribenes: I'm not sure this is really worth it, but it's at least a fun demo and maybe a good starting point?

effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -115,592 +144,528 @@ enum TokenKind {
...
+
+  inline def advance(inline k: Int): Unit =
+    if k > 0 then { advance(); advance(k - 1) }
+
+  inline def advanceWith(inline k: Int, inline kind: => TokenKind): TokenKind =
+    advance(k); kind
+
+  inline def advanceWith(inline kind: => TokenKind): TokenKind = advanceWith(1, kind)
+  inline def advance2With(inline kind: => TokenKind): TokenKind = advanceWith(2, kind)
+  inline def advance3With(inline kind: => TokenKind): TokenKind = advanceWith(3, kind)
```
jiribenes: I'm not too convinced about the `advanceWith`, `advance2With`, `advance3With`. Does anybody have better names here? 🤔 
See usage below in `nextToken`.

effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -115,592 +144,528 @@ enum TokenKind {
...
+
+      // String literals, character literals, hole string literals
+      case ('"', '"') if peekAhead(2) == '"' => advance3With(lexString(MultiString))
+      case ('"',   _)                        => advanceWith(lexString(SingleString))
+      case ('\'',  _)                        => advanceWith(lexCharLiteral())
+      case ('<', '"')                        => advance2With(lexString(HoleString))
+
+      // Unicode literals
+      // TODO(jiribenes, 2025-07-01): Do we even want to keep supporting these?
+      case ('\\', 'u') => advance2With(lexUnicodeLiteral())
```
jiribenes: Do we care about the standalone `\u1234` literals at all? I'd be happy to get rid of them by now...

effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -115,592 +144,528 @@ enum TokenKind {
...
+        try java.lang.Integer.parseInt(hexStr, 16)
+        catch case _: NumberFormatException => -1
+
+      // TODO(jiribenes): revamp this syntax, count to 4 at most?
+      case c if isHexDigit(c) =>
+        val start = position.offset
+        advanceWhile { (curr, _) => isHexDigit(curr) }
+        val hexStr = source.content.substring(start, position.offset)
+        try java.lang.Integer.parseInt(hexStr, 16)
+        catch case _: NumberFormatException => -1
```
jiribenes, line 620: Does anybody use direct `\u1234` escapes or are we fine with only allowing `\u{1234}`?

effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -115,592 +144,528 @@ enum TokenKind {
...
+          advance(); advance()
+          depth -= 1
+        case _ =>
+          advance()
+
+    if depth > 0 then
+      TokenKind.Error(LexerError.UnterminatedComment)
+    else
+      val comment = getCurrentSlice.substring(2, getCurrentSlice.length - 2) // Remove /* and */
+      TokenKind.Comment(comment)
```
jiribenes: Do we want non-nestable multiline comments (like we have so far) or nestable? (I have both above)
In general, I'd be very happy to get rid of multiline comments altogether, to be honest...

effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1481,11 +1511,13 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
     val before = position
     val beforePrevious = previous
     val labelBefore = currentLabel
+    val softFailsBefore = softFails
```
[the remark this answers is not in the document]
jiribenes: Let's try that. Thanks!

effekt/shared/src/main/scala/effekt/Lexer.scala
jiribenes: I removed them, we can revert if somebody has really great arguments in favour of them...

jiribenes: It seems that nobody even used these, see https://github.com/search?q=%2F%5B%5E%27%22%5C%5C%5D%5C%5Cu%2F+path%3A*.effekt+NOT+is%3Afork+NOT+org%3Aeffekt-lang&type=code

jiribenes:
The PR grew significantly with the drive-bys and no reviewer seems to complain IRL when asked about the new Lexer,so I'll just merge the PR to avoid conflicts in the future.

PR #1067, opened 2025-07-02
Change Namespace to be immutable

timsueberkrueb:
We want to get snapshots of the current `Scope` for each hole. Currently, namespaces are mutable so this would require cloning the entire data structure.
This PR evaluates if we can switch over to an immutable data structure for tracking names without significant performance regressions.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/symbols/Scope.scala
+++ b/effekt/shared/src/main/scala/effekt/symbols/Scope.scala
@@ -56,14 +56,6 @@ case class Bindings(
       case BlockTypeConstructor.Interface(_, _, operations, _) => operations.toSet
       case _ => Set.empty
     }.groupMap(_.name.name)(op => op)
-
-  def toBindings: Bindings =
-    Bindings(
-      terms,
-      types,
-      captures,
-      namespaces.view.mapValues(_.toBindings).toMap
-    )
 }
 
 object Bindings {
@@ -310,7 +302,7 @@ object scopes {
     def define(name: String, capt: Capture)(using ErrorReporter): Unit =
       scope = scope.withBindings(scope.bindings.setCapture(name, capt))
 
-    def exports: Bindings = scope.bindings.toBindings
+    def exports: Bindings = scope.bindings
 
     def scoped[R](name: String, block: => R): R =
       val before = scope


timsueberkrueb:
This is a bit tricky:

```
==> X effekt.JavaScriptTests.examples/pos/namer/overloaded_constructor_nested.effekt (js)  0.216s munit.ComparisonFailException: /home/runner/work/effekt/effekt/effekt/jvm/src/test/scala/effekt/EffektTests.scala:120
121:        }
diff assertion failed
=> Obtained
Error:     """|[error] examples/pos/namer/overloaded_constructor_nested.effekt:7:20: Could not resolve type outer::inner::Inner
       |    def consume(c: outer::inner::Inner) = c match {
       |                   ^^^^^^^^^^^^^^^^^^^
       |""".stripMargin
-[error] examples/pos/namer/overloaded_constructor_nested.effekt:7:20: Could not resolve type outer::inner::Inner
-    def consume(c: outer::inner::Inner) = c match {
-                   ^^^^^^^^^^^^^^^^^^^
```

The way scopes are structured (inside-out), whenever a namespace in an inner scope changes, the outer scope also needs to change by updating the namespace it points to.

marzipankaiser:
For context, when are Namespaces changed? Is the mutability "just" for initializing the structure in some phase, or will there be changes later at some point? When?

EDIT: Or rather: Do I understand correctly, that, while building up the scopes, we want to remember the specific "partial" scopes at a certain position?

timsueberkrueb:
> For context, when are Namespaces changed? Is the mutability "just" for initializing the structure in some phase, or will there be changes later at some point? When?

The current `scope: Scope` tracking namespaces changes throughout `Namer` when entering/exiting scopes.

> Do I understand correctly, that, while building up the scopes, we want to remember the specific "partial" scopes at a certain position?

Yes! This is because we want the current scope at the position of any hole. Previously, the scope collected at a hole position would have been mutated after passing the hole.

marzipankaiser:
> Yes! This is because we want the current scope at the position of any hole. Previously, the scope collected at a hole position would have been mutated after passing the hole.

Is this always what we want? What about top-level scopes/namespaces? We do allow forward references for some scopes, right?

EDIT: To clarify:
```
def foo() = ...
def bar() = <" should know about baz ">
def baz() = ...
```
while
```
def outer() = {
  def foo() = ...
  def bar() = <" shouldn't know about baz ">
  def baz() = ...
  ...
}
```

i.e., in a sense the local definitions in the second case are in separate scopes that we do not want to distinguish, though?

timsueberkrueb:
> Is this always what we want? What about top-level scopes/namespaces? We do allow forward references for some scopes, right?

If I understand it correctly, this is implemented by having two phases in `Namer`: `preresolve` and `resolve`.

marzipankaiser:
Yes. I'm just wondering if we might be able to have a solution for the specific problem that does not need us to make everything immutable and re-creating the whole spine 🤔 (so one that does save dev time).

If my hopes are correct:
- we will always have just one `Global` `Scope` on the outside, which will not get new definitions (since they were already `preresolve`d) - so no copying needed
- `Named` scopes will only be below a `Global`, I don't think we allow namespaces inside functions or blocks (do we?). Those should be similar to `Global` - so no copying?
- `Local` scopes are not accessible from the outside, so we do not propagate anything to the outside?

Maybe I'm wrong, though - just wondering where the "interesting" cases are. Also, they might change if we change the language.

EDIT: We would still need to change the outer local scopes, thinking about it, and propagate them inside.

marzipankaiser:
So, in a sense, we might not want `inner.outer.namespaces(inner.name) == inner` necessarily because `inner` can only "see" part of `outer`?
Then, we could make only `Local` immutable and only update the outer scope in `scoped` (at the end)? - `Scoping` only ever changes the current scope AFAICS.

[Sorry for spamming, I just got into thinking about this and maybe it helps... - ofc just making everything immutable should also work]

marzipankaiser:
To rephrase my long text into a question: Should `outer` be:
1. The namespace that contains this namespace in `namespaces`, forming a tree from the global namespace. [This is a view "from the outside" with backpointers.]
2. The namespace that contains what's accessible here from the outside `Scope` (so the snapshotted version) [Those are the views "from the inside".]
3. Both of those in some form.

If 2, we do not need to update the `outer`s of all children on each change, since they will only see "their" version. The Scopes that allow forward and mutably recursive uses can savely be mutable (since there the accessibility does not depend on order).
Then, `namespaces` would generally contain the "last" version of a child namespace.
If 3, then we could leave `outer` mutable as-is and have an `accessibleOuter` (TODO better name) for 2.

timsueberkrueb:
Consider the test case `examples/pos/namer/overloaded_constructor_nested.effekt`:

```scala
namespace outer {
 type Outer { Constructor(n: Int) }

  namespace inner {
    type Inner { Constructor(n: Int) }

    def consume(c: outer::inner::Inner) = c match {
      case outer::inner::Constructor(x) => x + 1
    }
  }
}
```

Here, we access `outer::inner::Constructor` via the outer namespaces. In the current structure, this means that the outer namespaces need to be updated.
We *could* perhaps change the structure/lookup so this is no longer needed.

marzipankaiser:
Yes, but those are `namespace` namespaces, which allow forward references. So they should not be updated after `precheck` anyway, IIUC. So, in a sense they are "frozen" once we get to holes (in check) anyway, right?

EDIT to clarify: This relies on the namespaces allowing forward references (`namespace`, global) being the same namespaces that allow access from the outside (i.e. via outer).

This is not a "we should do it this way" - just a "maybe we can save some work here"

timsueberkrueb:
>  Then, we could make only Local immutable and only update the outer scope in scoped (at the end)? - Scoping only ever changes the current scope AFAICS.

Yes, it seems this may be possible. It would require us to create different types for Local vs Global/Namespace scopes.
I assume we would save on memory usage. Currently, I cannot measure a difference in time spent in in `Namer` compared to `master`. As in code clarity, we would save the `updateOuterScopes` function, but we would gain duplication for lookup/update functions because then we have two different namespace types.

marzipankaiser:
Yeah, just wanted to point it out after noticing.

Regarding performance: As one data point, CI ran through in 12min here, which is on the lower end of the last few runs on master AFAICT. Of course, we might want to actually measure namer. I would be surprised if it gets much slower, though, since the cases where we'd need to copy a lot are large/deeply nested local scopes (which I don't think we do much).
I think there was some cli flag to measure phase times 🤔

timsueberkrueb:
Just found the following example:

```scala
namespace A {
  def foo() = {
    val x = <>
    namespace B {
      def bar() = 42
    }
  }
}
```

The solution in this PR does the right thing here: `B::bar()` is not in scope for `x`. I think this can still be done with a hybrid mutable/immutable solution, but it is a bit tricky to reason about.

marzipankaiser (approved):

PR #1068, opened 2025-07-03
Fix cake pattern bug with more cake pattern

timsueberkrueb:
I fixed it.
Don't @ me.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/core/Transformer.scala
@@ -86,7 +86,7 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
       List(Toplevel.Val(v.symbol, transformedTpe, transformed))
 
     case v @ source.ValDef(id, _, binding, doc, span) =>
-      Context.at(d) { Context.uabort("Effectful bindings not allowed on the toplevel") }
+      Context.at(d) { Context.abort("Effectful bindings not allowed on the toplevel") }
 
     case v @ source.DefDef(id, annot, binding, doc, span) =>
       val sym = v.symbol
@@ -97,7 +97,7 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
       bindings.map(core.Binding.toToplevel) ++ List(definition)
 
     case _: source.VarDef | _: source.RegDef =>
-      Context.at(d) { Context.uabort("Mutable variable bindings not allowed on the toplevel") }
+      Context.at(d) { Context.abort("Mutable variable bindings not allowed on the toplevel") }
 
     case d @ source.InterfaceDef(id, tparamsInterface, ops, doc, span) =>
       val interface = d.symbol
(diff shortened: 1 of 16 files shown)


effekt/shared/src/main/scala/effekt/typer/Unification.scala
```diff
@@ -39,7 +39,7 @@ case class LocalScope(
  * TODO
  *   - [ ] All incoming types need to be "normalized": substituted and dealiased.
  */
-class Unification(using C: ErrorReporter) extends TypeUnifier, TypeMerger, TypeInstantiator { self =>
+trait Unification extends TypeUnifier, TypeMerger, TypeInstantiator { self: Context =>
```
b-studios: Here I would try 
```suggestion
trait Unification extends TypeUnifier, TypeMerger, TypeInstantiator { self: ErrorReporter =>
```

effekt/shared/src/main/scala/effekt/typer/Unification.scala
```diff
@@ -69,24 +69,24 @@ class Unification(using C: ErrorReporter) extends TypeUnifier, TypeMerger, TypeI
...
     substitution.substitute(tpe)
 
-  def apply(tpe: ValueType): ValueType =
+  def unification(tpe: ValueType): ValueType =
     substitution.substitute(tpe)
 
   // Lifecycle management
   // --------------------
-  def backup(): UnificationState = UnificationState(scope, constraints.clone())
+  def ubackup(): UnificationState = UnificationState(scope, constraints.clone())
```
b-studios: Maybe
```suggestion
  def backupUnification(): UnificationState = UnificationState(scope, constraints.clone())
```

effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -97,7 +97,7 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
       bindings.map(core.Binding.toToplevel) ++ List(definition)
 
     case _: source.VarDef | _: source.RegDef =>
-      Context.at(d) { Context.abort("Mutable variable bindings not allowed on the toplevel") }
+      Context.at(d) { Context.uabort("Mutable variable bindings not allowed on the toplevel") }
```
b-studios: I think this shouldn't be a unification error

effekt/shared/src/main/scala/effekt/typer/Unification.scala
timsueberkrueb: This works but requires us to write 

```scala
  protected var constraints = new Constraints()(using this)
```

and now I'm afraid this could trigger the bug again?

b-studios: How is this necessary, but wasn't before?

timsueberkrueb: If only I knew ...

b-studios: Can you point me to the position in the code where this change is necessary? I'm currently afk

timsueberkrueb: * https://github.com/effekt-lang/effekt/blob/fix-cake-pattern-bug-with-more-cake-pattern/effekt/shared/src/main/scala/effekt/typer/Unification.scala#L47
* https://github.com/effekt-lang/effekt/blob/fix-cake-pattern-bug-with-more-cake-pattern/effekt/shared/src/main/scala/effekt/typer/Unification.scala#L96

b-studios: Hmm... Passing it explicitly should already be required by your change, since you removed the given. Changing from Context to ErrorReporter should merely impose less requirements.

PR #1069, opened 2025-07-03
Add namespace name to paths

timsueberkrueb:
In conjunction with https://github.com/effekt-lang/effekt-vscode/pull/97, this addresses the issue that imported bindings appear to display twice in the holes panel. This is because we actually do have them in the namespace twice: once with the fully-qualified name (e.g. `foo::bar::baz`) and once with just their name (e.g. `baz`).
Previous to this PR, when computing `allBindings`, we just forgot adding namespace names to `path`.
After this PR:

image


marzipankaiser (approved):
LGTM

marzipankaiser (approved):

PR #1070, opened 2025-07-03
[WIP] Named arguments

marzipankaiser:
Allow naming arguments when calling a function.

Current state: Names are comments and not checked.


effekt/shared/src/main/scala/effekt/RecursiveDescent.scala
```diff
@@ -1036,12 +1036,19 @@ class RecursiveDescent(positions: Positions, tokens: Seq[Token], source: Source)
...
+      }
+    } else { of() }
+  }
   def typeArgs(): Many[ValueType] =
     nonterminal:
       some(valueType, `[`, `,`, `]`)
   def valueArgs(): List[Term] =
     nonterminal:
-      many(expr, `(`, `,`, `)`).unspan
+      many(maybeNamed(expr), `(`, `,`, `)`).unspan
```
jiribenes: I think that something like this should also work, eliminating the extra lambda?
EDIT: the suggestion rendering is very broken, sorry.
```suggestion
  inline def maybeNamed[T](inline of: () => T): T = {
    if (isIdent && peek(1).kind == `=`) {
      (ident() <~ `=`) ~ of() match {
        case _name ~ v => v
      }
    } else { of() }
  }
  def typeArgs(): Many[ValueType] =
    nonterminal:
      some(valueType, `[`, `,`, `]`)
  def valueArgs(): List[Term] =
    nonterminal:
      many({ maybeNamed(expr) }, `(`, `,`, `)`).unspan
```

marzipankaiser: Yes, that's mostly moving the lambda to the call site, though. I put it into the method because I initially had it in neither position, which leads to very weird behavior...

marzipankaiser: Just replaced it with a `valueArg` (specialized) variant.

marzipankaiser:
Hm, what should the following do:
```
def foo(a: Int) = a

def bar { fn: Int => Int } : Int = fn(b = 2)

def main() = {
  bar{foo}
}
```
🤔

If we would want to detect this, names would become part of the type. Do we want this?

jiribenes:
I'd be happy to ignore this case of blocks in general -- both named block arguments and named value arguments for blocks -- since I don't believe there's a nice solution to this 🤔 

EDIT: So in my view, this should say:
```scala
def bar { fn: Int => Int }: Int = fn(b = 2)
                                     ^^^
// error: `fn` is a block parameter, it cannot be called with named arguments
```

b-studios:
It's a bit "known versus unknown functions". Intuitively, for known ones it is ok to use named arguments, since we know them.

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -1239,8 +1244,16 @@ object Typer extends Phase[NameResolved, Typechecked] {
 
     var effs: ConcreteEffects = Pure
 
+    (vpnames.map(Some.apply).zipAll(vargs, None, source.ValueArg.Unnamed(source.UnitLit(source.Span.missing)))) foreach {
+      case (Some(expName), source.ValueArg(Some(gotName), _, _)) if expName != gotName =>
+        Context.error(s"Unexpected name in named argument: got ${gotName}, but expected ${expName}.")
```
jiribenes: Nitpick: could we also add some `if gotName in vargs.map { ...name }` condition to say something reminding the user that there _is_ an arg with this name, but the _position_ is wrong.

jiribenes (approved):

examples/neg/named_args_wrong.effekt
```diff
@@ -0,0 +1,5 @@
+def length(x: Double, y: Double): Double = sqrt(x * x + y * y)
+
+def main() = {
+  println(length(y = 3.0, x = 4.0)) // ERROR named argument
```
b-studios, line 4: Why is this wrong? Just the order is wrong, but both names exist, no?

b-studios: Namer could somehow resolve and reorder (don't know how exactly at the moment).

b-studios: This would also be half way to default arguments.

examples/pos/named_arg.effekt
```diff
@@ -0,0 +1,32 @@
...
+  println(x.foo(bar = -12))
+
+  try {
+    do foo(bar = 2); ()
+  } with Foo {
+    def foo(x) = println(x)
+  }
+
+  println(Point2D(x = 1, y = 2).x)
+}
```
b-studios, line 32: Can you mix and match named and unnamed arguments? What if I name an argument (and reorder) in-between two unnamed ones?

```
def foo(x: Int, y: Int, z: Int) = <>

foo(1, 2, y = 3)
```
?

examples/neg/named_args_wrong.effekt
marzipankaiser: Yes, the current PR just allows naming the argument, not reordering them based on the names.
The test documents this, though we could also implement reordering additionally.
Then, the question is when and how exactly to reorder based on the names.

examples/pos/named_arg.effekt
marzipankaiser: This, as of now, would tell you that the third argument is named "y" but expected "z". We do not do reordering atm.

examples/neg/named_args_wrong.effekt
marzipankaiser: `Namer` potentially can't resolve this due to overloading, so it would have to be in `Typer`.

jiribenes: I would very much prefer if we were order-sensitive, at least for the time being.
My motivation is countering boolean/literal blindness by being able to say `run(foo, steps = 42)` instead of having to write a block comment like `run(foo, /* steps = */ 42)`.

b-studios: I think default arguments that can refer to previous arguments will be very useful. This would allow us to write Koka-style copy functions like: 

```
def copy(v: Vector3D, x: Int = v.x, y: Int = v.y, z: Int = v.z): Vector3D = Vector3D(x, y, z)

v.copy(y = 0)
```

b-studios: This would be an additional argument (ha!) for not reordering.

b-studios: Btw, namer could in principle resolve it for  each overload specifically.

marzipankaiser: @b-studios Yes, in principle namer could.

I think reordering in some way could be added in a future PR, once we know how (at least I don't, esp. for programs like the one you commented below) - then, it would be a backwards-compatible change, just strictly allowing more programs.

examples/pos/named_arg.effekt
marzipankaiser: So, yes, mixing is fine, but they still need to be in order (see `increase` above)

PR #1073, opened 2025-07-07
Do not hide errors for bad type annotation in 'val' definitions

jiribenes:
Before:
image

After:
Screenshot 2025-07-07 at 13 50 16

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Parser.scala
+++ b/effekt/shared/src/main/scala/effekt/Parser.scala
@@ -485,9 +485,10 @@ class Parser(positions: Positions, tokens: Seq[Token], source: Source) {
       val startPos = pos()
       val startMarker = nonterminal { new {} }
       def simpleLhs() = backtrack {
-        `val` ~> idDef() ~ maybeValueTypeAnnotation() <~ `=`
-      } map {
-        case id ~ tpe =>
+        // Make sure there's either a `:` or `=` next, otherwise goto `matchLhs`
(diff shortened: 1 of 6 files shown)


effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -485,9 +485,10 @@ class Parser(positions: Positions, tokens: Seq[Token], source: Source) {
       val startPos = pos()
       val startMarker = nonterminal { new {} }
       def simpleLhs() = backtrack {
-        `val` ~> idDef() ~ maybeValueTypeAnnotation() <~ `=`
-      } map {
-        case id ~ tpe =>
+        // Make sure there's either a `:` or `=` next, otherwise goto `matchLhs`
+        `val` ~> idDef() <~ (if !peek(`:`) && !peek(`=`) then fail("Expected type or right-hand-side of a `val`"))
+      } map { id =>
+          val tpe = maybeValueTypeAnnotation() <~ `=`
```
jiribenes: This is the whole diff. Is there a nicer / "more idiomatic" way of writing the `if !peek ...` part? cc @dvdvgt

jiribenes: The only idiom I found feels a bit wrong:
```scala
expect("...") { case `:` | `=` => () }
```

b-studios: I think the expect version looks almost ok.

jiribenes: the `expect` doesn't work as it skips :(

PR #1074, opened 2025-07-07
Do not flatten namespaces

### LINKED ISSUE #1037: Generated documentation doesn't include namespaces
We run the parser and frontend before we generate the documentation. `BoxUnboxInference` flattens the namespaces, which is why we don't have `NamespaceDef`s in the `Tree` once we arrive at the documentation generator. Consequently, [effekt-api](https://effekt-lang.github.io/effekt-api/) also loses information about namespaces.

For documentation generation, we should either skip flattening the namespaces or skip the `BoxUnboxInference` phase completely.

marvinborner:
Flattening namespaces is annoying for documentation generation because information gets lost and somehow needs to get reconstructed (for example with scope annotations).

Now we do not flatten namespaces anymore. One case is still missing but interestingly we also get a segfault in LLVM, no idea why.

Additionally, we also deprecate namespace definitions nested in function definitions.

Closes #1037.


b-studios:
```
def helloWorld() = {
  namespace a {
    def foo(): Int = b::bar()
  }
  namespace b {
    def bar(): Int = a::foo()
  }
}
```

marvinborner:
drop local namespaces!!

effekt/shared/src/main/scala/effekt/core/Transformer.scala
```diff
@@ -201,11 +202,10 @@ object Transformer extends Phase[Typechecked, CoreTransformed] {
 
       case d: source.Def.Extern => Context.panic("Only allowed on the toplevel")
       case d: source.Def.Declaration => Context.panic("Only allowed on the toplevel")
+      case d: source.Def.NamespaceDef => Context.panic("Only allowed on the toplevel")
```
b-studios, line 205: Nice how this panic was moved around :)

b-studios:
Warning: I'll rebase to resolve the conflicts

b-studios (approved):
@marvinborner maybe you could double check whether the deleted tests are actually useless. Otherwise this looks good to me. Thanks!

effekt/jvm/src/test/scala/effekt/ParserTests.scala
```diff
@@ -928,6 +928,58 @@ class ParserTests extends munit.FunSuite {
...
+
+    val semiInsertion = parseProgram(
+      """namespace list
+        |namespace internal
+        |
+        |val x = 4
+        |val y = 5
+        |""".stripMargin)
+
+    // TODO: how to do equality on Programs?
```
marvinborner: How can we do equality on Programs (`ModuleDecl`)? Previously we did parseDefinition and `assertEqualModuloSpans`. `(a zip b).forEach(assertEqualModuloSpans)` does not work. Should I remove the equality assertions?

marzipankaiser: Aren't the tests pointless if you remove the assertions?

What does "not work" mean? Do they fail to run/compile? How?

Maybe the `assertEqualModuloSpans` needs to be extended to handle programs correctly (@timsueberkrueb might know more?).

marvinborner: Without equality assertions it is still tested whether the tests parse at all. The problem is that the programs are not equivalent anymore. Previously we parsed both as single definitions (with nested namespace definitions), now they are toplevel (either nested or in consecutive definitions).

PR #1076, opened 2025-07-07
Add Char case in Transformer

### LINKED ISSUE #1075: Char does not have a working default
The following program:
```
effect fail(): Nothing

def f(): Char / fail = do fail()

def main(): Unit =
  try {
    println(show(f()))
  } with fail {
    ()
  }
```
ran with `--backend llvm`
 fails with `Unsupported expression: Literal(a,Data(Char,List()))`.
Other data types like `Int` work fine.

mattisboeckle:
Fixes #1075


mattisboeckle:
This might not be correct. The previous case mentiones `Int` being used for `Char`, but we fall into the default when not mentioning `Char` explicitly.

PR #1078, opened 2025-07-09
Fix MatchError when printing unhandled symbols with DeclPrinter

timsueberkrueb:
Some symbols are not handled by `DeclPrinter`, previously leading to a match error in `Intelligence.scala`. This PR makes `DeclPrinter` return an `Option` so that we can handle missing cases gracefully.

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/LSPTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/LSPTests.scala
@@ -1905,6 +1905,24 @@ class LSPTests extends FunSuite {
     }
   }
 
+  test("Server does not crash on file with type variables") {
+    withClientAndServer { (client, server) =>
+      val source =
+        raw"""def foo[T]() = <>
+             |""".textDocument
+
+      val initializeParams = new InitializeParams()
+      val initializationOptions = """{"effekt": {"showHoles": true}}"""
+      initializeParams.setInitializationOptions(JsonParser.parseString(initializationOptions))
+      server.initialize(initializeParams).get()
+      val didOpenParams = new DidOpenTextDocumentParams()
+      didOpenParams.setTextDocument(source)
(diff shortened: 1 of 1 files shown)


effekt/shared/src/main/scala/effekt/symbols/DeclPrinter.scala
```diff
@@ -12,10 +12,10 @@ object DeclPrinter extends ParenPrettyPrinter {
 
   override val defaultIndent = 2
 
-  def apply(t: Symbol)(implicit context: Context): String =
-    pretty(toDoc(t, context)).layout
+  def apply(t: Symbol)(implicit context: Context): Option[String] =
+    toDoc(t, context).map { d => pretty(d).layout }
 
-  def toDoc(t: Symbol, context: Context): Doc = t match {
+  def toDoc(t: Symbol, context: Context): Option[Doc] = Some(t match {
```
jiribenes, line 18: Does this make sense when the implementation always wraps it in `Some` and all consumers just call `.get` anyways?

timsueberkrueb: Yes because there is a default case which returns `None` :)

timsueberkrueb: The existing behavior is preserved in that other places in the code keep throwing if the invariant is violated.
The behavior only changes for gathering the symbols in `Intelligence.scale`.

jiribenes: > Yes because there is a default case which returns None :)

Ah, my bad, it's an early return -- I was too confused by Effekt in which it'd return `Some(None)` in the wildcard case, sorry!

Issue #805, opened 2025-01-28
Add non-nestable OR patterns without bindings


jiribenes:
Support non-nestable OR patterns without bindings such as:
```scala
str match {
  "YES" | "NO" => ...
  _ => ...
}

car match {
  case Car("Mercedes-Benz", _) | Car("smart", _) => ...
  _ =>
}
```

only on the top-level (no nesting) and for patterns **without** bindings (those should be rejected).
This way, we get the 80% of use cases with 20% of the work of implementing actual nested OR patterns with bindings.

Then this change should be somewhat simple to add, just duplicating the branches.

jiribenes:
I'd like to bump this again since we keep encountering it both in our and students' projects; would it make sense to resurrect this issue and the corresponding PR?

PR #1081, opened 2025-07-14
Implement first draft of or-patterns

b-studios:
This is a first attempt to implement or patterns (such as requested by #805).

It also supports nested or patterns, but exhaustivity is not yet implemented.


b-studios:
Please note that this generates code (after inlining) such as

```
      if (infixEq_85("c", "a")) {
        let tmp_1303570 = println_1("hello")
        l_1303481(tmp_1303570)
      } else {
        if (infixEq_85("c", "b")) {
          let tmp_1303572 = println_1("hello")
          l_1303481(tmp_1303572)
        } else {
          let tmp_1303573 = println_1("world")
          l_1303481(tmp_1303573)
        }
      }
```
while the following would be better (for sharing code)
```
      if (infixEq_85("c", "a") || infixEq_85("c", "b")) {
        let tmp_1303570 = println_1("hello")
        l_1303481(tmp_1303570)
      } else {
        let tmp_1303573 = println_1("world")
        l_1303481(tmp_1303573)
      }
```

jiribenes:
Bikeshed: If possible, I'd like this to use `or` and also support `if (foo is "hello" or foo is "bar") { ... }` :)

Issue #770, opened 2025-01-10
Unnecessary default in a pattern match


jiribenes:
From https://github.com/effekt-lang/effekt/pull/769#issuecomment-2583063792

The following program gets compiled to a pattern match with an unnecessary `else { ... }` clause (a default clause).

```scala
def isShorterThan(x: List[Int], y: List[Int]): Bool =
  (x, y) match {
    case (_, Nil()) => false
    case (Nil(), _) => true
    case (Cons(_, xs),Cons(_, ys)) => isShorterThan(xs, ys)
  }
```

gets compiled to the following Core IR without opts:
```scala
def isShorterThan2379(x2377: List910[BoxedInt296], y2378: List910[BoxedInt296]) = {
  let v_r_24453438 = make Tuple2249[List910[BoxedInt296], List910[BoxedInt296]] Tuple2330(x2377, y2378)
  def b_k_24463439() = return false
  def b_k_24473440() = return true
  def b_k_24493441(xs2391: List910[BoxedInt296], ys2392: List910[BoxedInt296]) = {
    val v_r_24483444: Bool389 = isShorterThan2379(xs2391, ys2392);
    return v_r_24483444
  }
  val v_r_24623449: Bool389 = v_r_24453438 match {
    case Tuple2330 { (v_y_24502453: List910[BoxedInt296], v_y_24512452: List910[BoxedInt296]) => 
      v_y_24512452 match {
        case Nil1118 { () => 
          b_k_24463439()
        }
        case Cons1119 { (v_coe_36343658: BoxedInt296, v_coe_36353659: List910[BoxedInt296]) => 
          def b_tmp_36363654(v_y_24542457: Int398, v_y_24552456: List910[BoxedInt296]) = v_y_24502453 match {
            case Nil1118 { () => 
              b_k_24473440()
            }
            case Cons1119 { (v_coe_36313656: BoxedInt296, v_coe_36323657: List910[BoxedInt296]) => 
              def b_tmp_36333655(v_y_24582461: Int398, v_y_24592460: List910[BoxedInt296]) = b_k_24493441(v_y_24592460, v_y_24552456)
              b_tmp_36333655(unboxInt300(v_coe_36313656), v_coe_36323657)
            }
          }
          b_tmp_36363654(unboxInt300(v_coe_36343658), v_coe_36353659)
        }
      } else {
        v_y_24502453 match {
          case Nil1118 { () => 
            b_k_24473440()
          }
        }}
    }
  };
[...]

PR #1082, opened 2025-07-14
First stab at #770

b-studios:
If all variants are covered, we do not need to add a default. 

The PR still needs clean up. Especially the info should somehow be passed to the transformation instead of looking at the symbols.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
+++ b/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
@@ -2,6 +2,7 @@ package effekt
 package core
 
 import effekt.core.substitutions.Substitution
+import effekt.symbols.TypeConstructor
 
 import scala.collection.mutable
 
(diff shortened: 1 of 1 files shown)


b-studios:
BTW This might significantly speed up things, since cases are now used linearly more often.

b-studios:
I am quite happy with these changes since they (in general) align the pattern matching compiler closer with the exhaustivity checker and improve the quality of the generated code at the same time.

I also sneaked in an improvement where we now generate

```
      if (first_3588) {
        if (second_3587) {
          b_k0_3736()
        } else {
          if (second_3587) {
            <>
          } else {
            b_k2_3738()
          }
        }
      } else {
        if (first_3588) {
          <>
        } else {
          b_k1_3737()
        }
      }
```
instead of the dreadful
```
      if (infixEq_91(first_3602, true)) {
        if (infixEq_91(second_3601, true)) {
          b_k0_3729()
        } else {
          if (infixEq_91(second_3601, false)) {
            b_k2_3731()
          } else {
            <>
          }
        }
      } else {
        if (infixEq_91(first_3602, false)) {
          b_k1_3730()
        } else {
          <>
        }
      }
```

This could be improved further by remembering "path conditions". But I don't want to do this in this PR

(EDIT: I documented this in #1083)

jiribenes:
Both of the examples in #770 are fixed by this. 👀

jiribenes (approved):

b-studios:
Tests fail, but only due to Text File Busy

PR #1084, opened 2025-07-15
Try removing unreachable cases

### LINKED ISSUE #1083: Use information about previous pattern matches to optimize
For the following effekt code
```effekt
import args

def join(left: Bool, right: Bool): Bool = (left, right) match {
  case (true, true) => true
  case (false,     _) => false
  case (_    , false) => false
}

def main() = commandLineArgs() match {
  case Cons(input, _) => println(join(input == "a", input == "b"))
  case Nil() => ()
}
```
our pattern matching compiler currently generates
```javascript
switch (v_r_0.__tag) {
      case 1: 
        const input_0 = v_r_0.head_0;
        const tmp_3 = input_0 === "a";
        const tmp_4 = input_0 === "b";
        if (tmp_3) if (tmp_4) {
          const tmp_5 = $effekt.println("true");
          return () => k_1(tmp_5, ks_2);
        } else if (tmp_4) {
          return $effekt.hole();  // UNREACHABLE
        } else {
          const tmp_6 = $effekt.println("false");
          return () => k_1(tmp_6, ks_2);
        } else if (tmp_3) {
          return $effekt.hole(); // UNREACHABLE
        } else {
          const tmp_7 = $effekt.println("false");
          return () => k_1(tmp_7, ks_2);
        }
        break;
      case 0:  return () => k_1($effekt.unit, ks_2);
    }
```
where I marked the two holes as unreachable. We seem to lose the positive information about scrutinees (maybe look at [Sestoft's paper](https://link.springer.com/chapter/10.1007/3-540-61580-6_22) again?)

Ideally, in this case we would generate
```javascript
    if (tmp_3) {
      if (tmp_4) {
        const tmp_5 = $effekt.println("true");
        return () => k_1(tmp_5, ks_2);
      } else {
        const tmp_6 = $effekt.println("false");
        return () => k_1(tmp_6, ks_2);
      }
    } else {
      const tmp_7 = $effekt.println("false");
      return () => k_1(tmp_7, ks_2);
    }
```

b-studios:
This is a draft that attempts to fix #1083 by discovering unreachable cases and dropping them.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
+++ b/effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
@@ -104,10 +104,9 @@ object PatternMatchingCompiler {
         return core.Let(x, tpe, binding, compile(Clause(rest, target, targs, args) :: remainingClauses))
       // - We need to check a predicate
       case Clause(Condition.Predicate(pred) :: rest, target, targs, args) =>
-        return core.If(pred,
-          compile(Clause(rest, target, targs, args) :: remainingClauses),
-          compile(remainingClauses)
-        )
+        val thn = compile(Clause(rest, target, targs, args) :: remainingClauses)
+        if remainingClauses.isEmpty then return thn
+        else return core.If(pred, thn, compile(remainingClauses))
       case Clause(Condition.Patterns(patterns) :: rest, target, targs, args) =>
         patterns
     }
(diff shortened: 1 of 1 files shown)


b-studios:
I currently only do it for literals, we should do the same for tags.
Also the code is still horribly hacky.

The idea is that "there are no clauses" means it is unreachable (since we checked exhaustivity before). So we shouldn't even `compile` an empty list of clauses and instead can conclude that "empty means unreachable" and optimize locally.

b-studios:
For tags this is a bit strange, since it might result in matches like

```
e match {
  case Red() => println("")
}
```
where `e` is of a type with more than this one case. Do we even want this?

b-studios:
It turns out, this might be a non-issue. I ran this on our Effekt-lexer in Effekt casestudy (with roughly 7000 generated lines of JS code) and this does not show up once. So maybe it is not worth our time and the implementation complexity.

Enough reason for me to stop working on this now.

PR #1085, opened 2025-07-15
Spanned holes: threading positions through holes

jiribenes:
Every hole is now internally associated with a span which is printed when the hole is encountered:
```sh
$ for BACKEND in chez-callcc llvm js; do effekt hole.effekt --backend $BACKEND; done
```

```log
# Chez
Exception: hole.effekt:8:15: not implemented yet
[error] Process exited with non-zero exit code 255.

# LLVM
PANIC: hole.effekt:8:15: not implemented yet

# JS
.../out/hole.js:208
    else throw e
         ^
hole.effekt:8:15: not implemented yet
(Use `node --trace-uncaught ...` to show where the exception was thrown)

Node.js v22.16.0
```

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/vm/VM.scala
+++ b/effekt/shared/src/main/scala/effekt/core/vm/VM.scala
@@ -104,7 +104,7 @@ enum VMError extends Throwable {
   case MissingBuiltin(name: String)
   case RuntimeTypeError(msg: String)
   case NonExhaustive(missingCase: Id)
-  case Hole()
+  case Hole(span: effekt.source.Span)
   case NoMain()
 
   override def getMessage: String = this match {
@@ -113,7 +113,7 @@ enum VMError extends Throwable {
     case VMError.MissingBuiltin(name) => s"Missing builtin: ${name}"
     case VMError.RuntimeTypeError(msg) => s"Runtime type error: ${msg}"
     case VMError.NonExhaustive(missingCase) => s"Non exhaustive: ${missingCase}"
-    case VMError.Hole() => s"Reached hole"
+    case VMError.Hole(span) => s"Reached hole @ ${span.range.from.format}"
     case VMError.NoMain() => s"No main"
   }
 }
@@ -436,7 +436,7 @@ class Interpreter(instrumentation: Instrumentation, runtime: Runtime) {
           }
           State.Step(body, env, rewind(cont, stack), heap)
 
-        case Stmt.Hole() => throw VMError.Hole()
+        case Stmt.Hole(span) => throw VMError.Hole(span)
       }
     }
 
--- a/effekt/shared/src/main/scala/effekt/cps/PrettyPrinter.scala
+++ b/effekt/shared/src/main/scala/effekt/cps/PrettyPrinter.scala
@@ -150,8 +150,8 @@ object PrettyPrinter extends ParenPrettyPrinter {
     case Stmt.Resume(r, body, ks, k) =>
       "resume" <> parens(toDoc(r)) <+> toDoc(body) <+> "@" <+> toDoc(ks) <> "," <+> toDoc(k)
 
-    case Stmt.Hole() =>
-      "<>"
+    case Stmt.Hole(span) =>
+      "<>" <+> s"// @ ${span.range.from.format}"
   }
 
   def toDoc(clause: Clause): Doc = clause match {
--- a/libraries/js/effekt_builtins.js
+++ b/libraries/js/effekt_builtins.js
@@ -58,4 +58,4 @@ $effekt.unit = { __unit: true }
 
 $effekt.emptyMatch = function() { throw "empty match" }
 
-$effekt.hole = function(pos) { throw "not implemented, yet @ " + pos }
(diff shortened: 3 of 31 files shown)


effekt/shared/src/main/scala/effekt/core/vm/VM.scala
```diff
@@ -113,7 +113,7 @@ enum VMError extends Throwable {
     case VMError.MissingBuiltin(name) => s"Missing builtin: ${name}"
     case VMError.RuntimeTypeError(msg) => s"Runtime type error: ${msg}"
     case VMError.NonExhaustive(missingCase) => s"Non exhaustive: ${missingCase}"
-    case VMError.Hole() => s"Reached hole"
+    case VMError.Hole(span) => s"Reached hole @ ${span.range.from.format}"
```
jiribenes, line 116: I use `span.range.from.format`, but it's a bit of a mouthful + it has a trailing `:`.
Is there a better way to get `${file}:${line}:${column}` with the new spans? cc @timsueberkrueb

effekt/shared/src/main/scala/effekt/core/vm/VM.scala
```diff
@@ -104,7 +104,7 @@ enum VMError extends Throwable {
   case MissingBuiltin(name: String)
   case RuntimeTypeError(msg: String)
   case NonExhaustive(missingCase: Id)
-  case Hole()
+  case Hole(span: effekt.source.Span)
```
jiribenes, line 107: `Span` currently lives in `effekt.source`, but perhaps it should be somewhere else if we're threading it through.

effekt/shared/src/main/scala/effekt/cps/PrettyPrinter.scala
```diff
@@ -150,8 +150,8 @@ object PrettyPrinter extends ParenPrettyPrinter {
     case Stmt.Resume(r, body, ks, k) =>
       "resume" <> parens(toDoc(r)) <+> toDoc(body) <+> "@" <+> toDoc(ks) <> "," <+> toDoc(k)
 
-    case Stmt.Hole() =>
-      "<>"
+    case Stmt.Hole(span) =>
+      "<>" <+> s"// @ ${span.range.from.format}"
```
jiribenes, line 154: This is how I pretty-print the holes in IRs like Core or Machine in case you're interested

libraries/js/effekt_builtins.js
```diff
@@ -58,4 +58,4 @@ $effekt.unit = { __unit: true }
 
 $effekt.emptyMatch = function() { throw "empty match" }
 
-$effekt.hole = function() { throw "not implemented, yet" }
+$effekt.hole = function(pos) { throw pos + " not implemented yet" }
```
jiribenes, line 61: Question for a reviewer: how much should we try to standardise the output of these messages? 
I think it is worth it if it's just a somewhat simple change.

I already synced JS and Chez to be like `hole.effekt:8:15: not implemented yet`, but LLVM still has `"PANIC: hole.effekt:8:15: reached a hole in the program`. What should the actual message look like? What info do we want to tell the user?

b-studios: Yes, the messages should be the same on the different platforms

b-studios: Is the PANIC somewhere in the C file on the LLVM backend?

effekt/shared/src/main/scala/effekt/core/vm/VM.scala
b-studios: Same for Template if I remember correctly

libraries/llvm/panic.c
```diff
@@ -7,8 +7,8 @@
 // this should _morally_ be using `stderr`, but we don't tee it in tests
 // see PR #823 & issue #815 for context
 
-void hole() {
-    printf("PANIC: Reached a hole in the program\n");
+void hole(const char* message) {
+    printf("PANIC: %s reached a hole in the program\n", message);
```
jiribenes: cc @b-studios here

b-studios: great, then we can just drop the panic there and normalize the message.

jiribenes: I'm thinking of keeping [something like] the `PANIC` part though, Chez also prints `Exception: ...` and JS also prints extra stuff...

b-studios: Of course, we could install a catch-all handler in  main for both chez and JS that catches all exceptions (or maybe only specific ones, like not-implemented) and doesn't print the extra stuff. 

But I am fine with least resistance for now.

jiribenes: Since we're at it, what should be the message? What are we actually trying to say?
The LLVM description is operational from the point of view of the RTS: "look, I found a hole in the program", whereas the JS/Chez description is from the semantic point of view of the user/source program: "look, here's some code that's not implemented yet". Which one of these makes more sense?

jiribenes:
```suggestion
    printf("PANIC: %s not implemented yet\n", message);
```

PR #1086, opened 2025-07-16
Add back hack for DeclPrinter

timsueberkrueb:
In a previous PR, we decided to remove this workaround because we thought it was likely no longer necessary. It turns out it is necessary, because `DeclPrinter` contains a partial pattern match, leading to a `MatchError` when an unexpected symbol is encountered.
Pending a proper fix, this adds back the hack.

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/LSPTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/LSPTests.scala
@@ -1938,6 +1938,24 @@ class LSPTests extends FunSuite {
     }
   }
 
+  test("Server does not crash on file with type variables") {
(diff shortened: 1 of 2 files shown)


marzipankaiser (approved):
Could we, instead of relying on a MatchError catched by a catchall, throw a custom exception in DeclPrinter? (If we know which matches are partial?).
Alternatively, catch `MatchError`s only?

If neither works (with reasonable effort), I'm fine with this, but I'd prefer to improve upon the hack at least slightly.

marzipankaiser:
Even better would be to remove it after #1078; if this is the plan, feel free to merge as hotfix.

PR #1089, opened 2025-07-21
Continuation statements to omit braces

b-studios:
This allows us to use multiple statements in a single statement position without braces.

Current rules:

- no definitions allowed, when no braces are used(`def`) -- could potentially be relaxed
- no sequencing of expressions (`expr; expr`) outside of braces

It should be fully backwards compatible for code that parsed before.
It accepts more programs and also changes the error message for one failing test case.

Here is how it looks like (also see the added test)

```
def main() =
  with filesystem;
  val port = readInt();
  with network(port);
  println("Running!")
```

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Parser.scala
+++ b/effekt/shared/src/main/scala/effekt/Parser.scala
@@ -233,46 +233,54 @@ class Parser(positions: Positions, tokens: Seq[Token], source: Source) {
   /**
    * Statements
    */
-  def stmts(): Stmt =
+  def stmts(inBraces: Boolean): Stmt =
     nonterminal:
       (peek.kind match {
-        case `val`  => valStmt()
-        case _ if isDefinition => DefStmt(definition(), semi() ~> stmts(), span())
-        case `with` => withStmt()
-        case `var`  => DefStmt(varDef(), semi() ~> stmts(), span())
+        case `val`  => valStmt(inBraces)
+        case _ if isDefinition && inBraces => DefStmt(definition(), semi() ~> stmts(inBraces), span())
+        case _ if isDefinition => fail("Definitions are only allowed, when enclosed in braces.")
+        case `with` => withStmt(inBraces)
+        case `var`  => DefStmt(varDef(), semi() ~> stmts(inBraces), span())
         case `return` =>
-          val result = `return` ~> Return(expr() <~ maybeSemi(), span())
-          result
-        case `}` | `}$` => // Unexpected end of <STMTS> =>
+          // trailing semicolon only allowed when in braces
+          `return` ~> Return(expr() <~ (if inBraces then maybeSemi()), span())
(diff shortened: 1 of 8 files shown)


b-studios:
This should be unrelated to the changes:

```
==> X effekt.ChezSchemeCallCCTests.examples/benchmarks/other/nbe.effekt (chez-callcc)  0.898s munit.ComparisonFailException: /home/runner/work/effekt/effekt/effekt/jvm/src/test/scala/effekt/EffektTests.scala:120
121:        }
diff assertion failed
-[info] /home/runner/work/effekt/effekt/libraries/common/map.effekt:748:20: Effekt Compiler Crash: java.lang.ClassCastException: class effekt.symbols.Field cannot be cast to class effekt.symbols.ValueParam (effekt.symbols.Field and effekt.symbols.ValueParam are in unnamed module of loader sbt.internal.LayeredClassLoader @468bb9d)
-      case (Bin(_, k, v1, l1, r1), _) =>
-                   ^
-[info] /home/runner/work/effekt/effekt/libraries/common/map.effekt:748:20: java.lang.ClassCastException: class effekt.symbols.Field cannot be cast to class effekt.symbols.ValueParam (effekt.symbols.Field and effekt.symbols.ValueParam are in unnamed module of loader sbt.internal.LayeredClassLoader @468bb9d)
-	at effekt.typer.Typer$.checkPattern$$anonfun$1(Typer.scala:572)
-	at effekt.util.messages$.effekt$util$messages$ErrorReporter$$_$focusing$$anonfun$1(Messages.scala:160)
-	at effekt.util.messages$ErrorReporter.at(Messages.scala:147)
-	at effekt.util.messages$ErrorReporter.at$(Messages.scala:99)
-	at effekt.context.Context.at(Context.scala:42)
-	at effekt.util.messages$ErrorReporter.focusing(Messages.scala:160)
-	at effekt.util.messages$ErrorReporter.focusing$(Messages.scala:99)
-	at effekt.context.Context.focusing(Context.scala:42)
-	at effekt.typer.Typer$.checkPattern(Typer.scala:570)
```

Has anybody seen something like this before?

b-studios:
I added a few more examples so we can discuss this better in the meeting today.

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -233,46 +233,55 @@ class Parser(positions: Positions, tokens: Seq[Token], source: Source) {
   /**
    * Statements
    */
-  def stmts(): Stmt =
+  def stmts(inBraces: Boolean = false): Stmt =
```
jiribenes, line 236: In the future, it might be generally nice to know some context related to where we currently are.
Then we can use it not only for `inBraces` and in other places to determine the context, but also for better error messages.

b-studios:
If there are no further comments, I am going to merge it now.

Issue #878, opened 2025-03-11
RFC: Drop box inference?


jiribenes:
_Extracted from #807_

Based on an earlier idea by @b-studios, what if we:

> _never_ infer `box` in non-trivial positions

Let's say that the only patterns that would now create a box are:
```scala
given [T: CompTy], foo: T

// 1. an explicit box _anywhere_ in the program in expression position
... box foo ...

// 2. an explicit `val <name> = <computation>`
val fooBoxed = foo
```

For contrast, I think inferring `unbox` automagically is OK.

I'd love to know how many programs this would _actually_ break, because I always do one of the `1.` or `2.` to make sure that I'm not gonna go crazy later :)
(Hence the `experiment` label, I'd love to see how this helps!)

Another question is if this would help _at all_ reduce the mental overhead of trying to guess what the box-unbox inference actually does.

jiribenes:
Resolved by #1090

Issue #807, opened 2025-01-29
Unfriendly type error when using a record field as a value


jiribenes:
_Originally reported by @lenakaeufel in the final project presentation._

The following (minimised) program prints a somewhat difficult-to-understand error message:
```scala
record A(c: Int)

def repro() = {
  var result: List[Char] = []
  ['a', 'b', 'c'].foreach { x =>
    result = Cons(c, result) // here, it should be `x` instead of `c`
  }
}
```

~>

```
Possible overload: option::foreach of type [A](Option[A]){A => Unit} => Unit
  Expected Option[A] but got List[A].

Possible overload: list::foreach of type [A](List[A]){A => Unit} => Unit
  Expected Char but got A => Int at ?C.
  Expected Char but got A => Int at ?C.

Possible overload: list::foreach of type [A](List[A]){(A){Control} => Unit} => Unit
  Wrong number of block arguments, given 0, but function expects 1.
```

---

The actual problem here is that `c` here is a function `c: A => Int` which might be pretty confusing for the user who'd like to have a "`c` is not defined" error.
Could we specialise this case somehow?

jiribenes:
As of #1090, I think this message is pretty usable, though not perfect:
<img width="811" height="219" alt="Image" src="https://github.com/user-attachments/assets/cd4f105f-db4f-4ee0-ab46-a67c504aaabf" />

jiribenes:
I've decided that I'm fine with how the message looks like after #1090, so this is to be closed by #1090.

PR #1090, opened 2025-07-22
Never perform box inference

b-studios:
This is a quick experiment related to #878 to see how bad things go if we *completely* disable box inference and only keep unbox inference.


b-studios:
For funsies, I also completely removed `unbox` (and `fun`) as keywords from the language. This implies, we would need to update the tutorials on the website etc.

examples/pos/capture/defdef.effekt
```diff
@@ -4,20 +4,20 @@ def myModule { eff : Eff } = region this {
...
 
-  val f: () => Unit at {eff} = test1b;
+  val f: () => Unit at {eff} = box test1b;
 
   def test2() = {
     test1();
     var x in outer = 42; x
   };
-  val g = test2;
+  val g = box test2;
```
jiribenes, line 18: RE #878: This is exactly a place where I'd be fine with inferring `box` :)
(the RHS of a `val`)

b-studios: Yeah, I know. I wanted to see how bad it is if we don't infer boxes at all. This might drive home the point that there is a difference even more (even though it is annoying). I am actually more annoyed by `foo(box compareInt)`.

jiribenes: Interesting, I always explicitly write the `box` in `map::fromList(..., box compareInt)`

jiribenes:
My hope was that this could make Namer a bit easier in some cases, but I don't see Namer changes, so I'm not sure whether that's true 🤔
(Stuff like https://github.com/effekt-lang/effekt/issues/807 should then allow for a better error message)

b-studios:
Maybe, I didn't think about it, but just followed the type errors :)

b-studios:
We need to change the docs on the website when merging.

b-studios:
We still need to update the website, for example here

https://effekt-lang.org/tour/captures

saying

```diff
-A boxed computation may only be used after unboxing it, which is only permitted when all its captures are in scope. 
+A boxed computation may only be used after unboxing it (which happens automatically), which is only permitted when all its captures are in scope. 
```

effekt/shared/src/main/scala/effekt/typer/UnboxInference.scala
```diff
@@ -52,12 +52,18 @@ object BoxUnboxInference extends Phase[NameResolved, NameResolved] {
     case v: Var => v.definition match {
       // TODO maybe we should synthesize a call to get here already?
       case sym: (ValueSymbol | symbols.RefBinder) => v
-      case sym: BlockSymbol => Box(Maybe.None(v.span.emptyAfter), v, v.span.synthesized).inheritPosition(v)
+      case sym: BlockSymbol =>
+        C.error(pp"Computation ${sym} is used in an expression position, which requires boxing (e.g. `box ${sym}`")
```
jiribenes:
```suggestion
        C.error(pp"Computation ${sym} is used in an expression position, which requires boxing (e.g. `box ${sym}`)")
```

effekt/shared/src/main/scala/effekt/typer/UnboxInference.scala
```diff
@@ -52,12 +52,18 @@ object BoxUnboxInference extends Phase[NameResolved, NameResolved] {
...
+        v
     }
 
-    case n: New => Box(Maybe.None(n.span.emptyAfter), rewriteAsBlock(n), n.span.synthesized).inheritPosition(n)
+    case n: New =>
+      C.error(pp"Creating an instance in an expression requires boxing (e.g. `box new ${n.impl.id}[...] { ... }`")
```
jiribenes:
```suggestion
      C.error(pp"Creating an instance in an expression requires boxing (e.g. `box new ${n.impl.id}[...] { ... }`)")
```

effekt/shared/src/main/scala/effekt/typer/UnboxInference.scala
```diff
@@ -52,12 +52,18 @@ object BoxUnboxInference extends Phase[NameResolved, NameResolved] {
...
+      rewriteAsBlock(n)
 
-    case b: BlockLiteral => Box(Maybe.None(b.span.emptyAfter), rewriteAsBlock(b), b.span.synthesized).inheritPosition(b)
+    case b: BlockLiteral =>
+      C.error(pp"Function literals in expression position require boxing (e.g. `box { (${b.vparams.map(_.id).mkString(", ")}) => ... `")
```
jiribenes:
```suggestion
      C.error(pp"Function literals in expression position require boxing (e.g. `box { (${b.vparams.map(_.id).mkString(", ")}) => ... `)")
```

jiribenes (approved):

PR #1091, opened 2025-07-22
Add more (infix) operators

b-studios:
This PR adds a few more infix operators. This is mostly just lexing and parsing so doesn't come with a lot of conceptual cost for the implementation.

A few things are not clear / implemented:

- we should name things consistently (I actually would like to rename `infixAdd` to `InfixPlus`, but this will break a lot of things)
- is this slowing down the parser? (for precedence it is now deeply nested)
- is the precedence ok?
- should pipe left and right be by-name? This might open up some cool usages.
- right now the new tokens are added as one batch, we might want to clean them up by sorting them somehow
- I didn't add `a === b` and `a !== b` since the tokens `===` are already used to represent `==`, ahhh.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Parser.scala
+++ b/effekt/shared/src/main/scala/effekt/Parser.scala
@@ -1030,11 +1030,11 @@ class Parser(tokens: Seq[Token], source: Source) {
   def assignExpr(assignee: Term.Var): Term =
     peek.kind match {
       case `=` =>
-        consume (`=`)
-        Assign (assignee.id, expr(), span() )
+        consume(`=`)
+        Assign(assignee.id, expr(), span())
       case `:=` =>
-        binaryOp (assignee, next(), expr() )
-      case _ => fail ("unreachable")
+        binaryOp(assignee, next(), expr())
+      case _ => fail("unreachable")
     }
 
   /**
@@ -1083,7 +1083,9 @@ class Parser(tokens: Seq[Token], source: Source) {
 
   lazy val precedenceTable = PrecedenceTable { table =>
     given PrecedenceTable = table
+    // since by default, the associativity is left-associative, we only need to specify it for the cases where it is different
     table.declare(Associativity.None, `===`, `!==`, `<=`, `>=`, `<`, `>`)
+    // We want `+` and `-` (as well as `*` and `/`) to bind equally strong and use associativity instead to disambiguate 
     `+` =?= `-`
     `*` =?= `/`
     Set(`||`) ?< Set(`&&`)
@@ -1174,6 +1176,7 @@ class Parser(tokens: Seq[Token], source: Source) {
     case `>=` => "infixGte"
     case `:=` => "infixColonEq"
     case `+`  => "infixPlus"
+    // `+=` (`-=`, `*=`, `/=`) has the same name as `+` (`-`, `*`, `/`) due to the way it is manually desugared in [[ primExpr ]] 
     case `+=` => "infixPlus"
     case `-`  => "infixMinus"
     case `-=` => "infixMinus"
--- a/effekt/shared/src/main/scala/effekt/Precedence.scala
+++ b/effekt/shared/src/main/scala/effekt/Precedence.scala
@@ -36,8 +36,7 @@ extension (looser: Set[TokenKind]) {
  * For convenience, we break the antisymmetric property on purpose by defining `+ ?< -` as well as `- ?< +` such that for example `1 + 2 - 3` can be written without parentheses by
  * falling back to the operators' associativity for disambiguation.
  * 
- * Usage:
- * 
+ * Usage: See [[ Parser.precedenceTable ]]
  */
 class PrecedenceTable {
   var assocMap:   Map[TokenKind, Associativity] = Map.empty
--- a/effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
+++ b/effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
@@ -47,7 +47,7 @@ class Deadcode(reachable: Map[Id, Usage])
         declarations.map(rewrite),
         // drop unreachable externs
         m.externs.collect {
-          // We need to keep "show", "showBuiltin" & "" for generating show definitions (see #1123)
+          // We need to keep "show", "showBuiltin" & "infixPlusPlus" for generating show definitions (see #1123)
           case e: Extern.Def if used(e.id) || List("show", "showBuiltin", "infixPlusPlus").contains(e.id.name.name) => e
           case e: Extern.Include => e
           // We currently do not have usage information on extern types, so we have to keep all of them


effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -111,6 +111,23 @@ enum TokenKind {
...
   case `*`
 
+  case `>>`
+  case `<<`
+  case `~`
+  case `--`
+  case `~>`
+  case `<~`
+  case `<|`
+  case `|>`
```
dvdvgt: We decided to delete these.

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -1022,15 +1025,36 @@ class Parser(positions: Positions, tokens: Seq[Token], source: Source) {
...
+    case TokenKind.`<~` => "infixTildeLeft"
+    case TokenKind.`|`  => "infixPipe"
+    case TokenKind.`&`  => "infixAmp"
+
+    case `<|`  => "infixPipeLeft"
+    case `|>`  => "infixPipeRight"
+    case `..`  => "infixDotDot"
+    case `...` => "infixDotDotDot"
+    case `^^`  => "infixHatHat"
+    case `^`   => "infixHat"
```
dvdvgt: Use syntactic not semantic names since they can be overloaded anyway.

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -1168,7 +1206,20 @@ class Parser(positions: Positions, tokens: Seq[Token], source: Source) {
     case _ if isVariable     =>
       peek(1).kind match {
         case _: Str => templateString()
-        case _ => variable()
+        case _ =>
+          val lhs = variable()
+          peek.kind match {
+            case `+=` | `-=` | `*=` | `/=` =>
```
marvinborner, line 1316: what about `<<=` and `>>=`?

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -1052,18 +1076,32 @@ class Parser(positions: Positions, tokens: Seq[Token], source: Source) {
...
+
+              case _ =>
+                val member = idRef()
+                // method call
+                if (isArguments) {
+                  val (targs, vargs, bargs) = arguments()
+                  e = Term.MethodCall(e, member, targs, vargs, bargs, span())
+                } else {
+                  e = Term.MethodCall(e, member, Nil, Nil, Nil, span())
+                }
```
dvdvgt: Extract into separate PR. Needs further discussion, e.g. setting an array. Do we also want a dedicated syntax for that?

effekt/shared/src/main/scala/effekt/Parser.scala
dvdvgt: Claude suggested these names:

```
case `+`  => "infixPlus"
case `+=` => "infixPlusEq"
case `-`  => "infixMinus"
case `-=` => "infixMinusEq"
case `*`  => "infixStar"
case `*=` => "infixStarEq"
case `/`  => "infixSlash"
case `/=` => "infixSlashEq"
case `++` => "infixPlusPlus"
case `>>` => "infixGreaterGreater"
case `<<` => "infixLessLess"
case `--` => "infixMinusMinus"
case TokenKind.`~`  => "infixTilde"
case TokenKind.`~>` => "infixTildeGreater"
case TokenKind.`<~` => "infixLessTilde"
case TokenKind.`|`  => "infixBar"
case TokenKind.`&`  => "infixAmp"
case `<|`  => "infixLessBar"
case `|>`  => "infixBarGreater"
case `..`  => "infixDotDot"
case `...` => "infixDotDotDot"
case `^^`  => "infixCaretCaret"
case `^`   => "infixCaret"
```

dvdvgt: Of course, please also change the other operator names

marvinborner:
*bump*

also, what do you think about a `:=` operator? I'd like to use it to overwrite assignments e.g. for references like [here](https://effekt-lang.org/playground.html?playground=FYsxC8IwFIT3%2FIobE5AWqlMwQ0fByVUKTTGRQPpS6kOU0v%2Fel%2B3uvu9eISJRTL87h2c%2FaG%2FxCFHSCZNFb%2BDgm09gPRmlqjz7RLrOmwK%2BPiOWIm2V04140JfOCKjj1aE7o21RFk6zz%2FlvMVo3Cl7WRJxJi9a8Axu1Hw%3D%3D&repl=aGVsbG9Xb3JsZCgp)

marvinborner:
If you're already working on this, could you also support infixConcat for lists, arrays, etc.?

dvdvgt:
I implemented the required changes. I also want to note that currently `+=, -=, *=, /=` are hardcoded to be transformed to an assignment to local variables. For example

```
var x = 1
x += 1
```

is parsed as

```
var x = 1
x = x + 1
```

This is of course convenient but precludes us from overloading these operators in the future.

Furthermore, I also added the infix operator `r := x` and overloaded it for global references `Ref[T]` to have the semantics of `r.set(x)`. Additionally, `l ++ other` is now an alias for `l.append(other)`, where both `l: List[T]` and `other: List[T]`.

I refrained from overloading these new operators for any other types as this already is a pretty "bikeshed-able" PR as it is.

dvdvgt:
Also, this most definitely will break the communities builds if there any overloads for the renamed infix operators.

examples/normalizer/simple.effekt
```diff
@@ -0,0 +1,288 @@
+/*
```
jiribenes: I think this file got committed by accident :)

effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
```diff
@@ -47,8 +47,8 @@ class Deadcode(reachable: Map[Id, Usage])
         declarations.map(rewrite),
         // drop unreachable externs
         m.externs.collect {
-          // We need to keep "show", "showBuiltin" & "infixConcat" for generating show definitions (see #1123)
-          case e: Extern.Def if used(e.id) || List("show", "showBuiltin", "infixConcat").contains(e.id.name.name) => e
+          // We need to keep "show", "showBuiltin" & "" for generating show definitions (see #1123)
```
jiribenes:
```suggestion
          // We need to keep "show", "showBuiltin" & "infixPlusPlus" for generating show definitions (see #1123)
```

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -1020,29 +1021,110 @@ class Parser(tokens: Seq[Token], source: Source) {
...
-        case other => other
+
+  def assignExpr(assignee: Term.Var): Term =
+    peek.kind match {
+      case `=` =>
+        consume (`=`)
+        Assign (assignee.id, expr(), span() )
+      case `:=` =>
+        binaryOp (assignee, next(), expr() )
+      case _ => fail ("unreachable")
```
jiribenes: wobbly spacing round these parts

effekt/shared/src/main/scala/effekt/Precedence.scala
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/**
+ * Encodes the precedence of infix operators as a partial order.
+ * Additionally, each operator's associativity is kept tracked of with a default of left-associativity if not specified otherwise.
+ * For convenience, we break the antisymmetric property on purpose by defining `+ ?< -` as well as `- ?< +` such that for example `1 + 2 - 3` can be written without parentheses by
+ * falling back to the operators' associativity for disambiguation.
+ * 
+ * Usage:
+ *
```
jiribenes: Missing example [?]
but having just pointing to the relevant fn in Parser ought to be enough :)

PR #1092, opened 2025-07-23
Implement namespaced constructors and operations

b-studios:
Constructors and operations are now defined in a namespace, named after the type.

For instance:

```
type TrafficLight { Red(); Green(); Yellow() }
```
defines the constructor `TrafficLight::Red`, but also imports it as `Red`. The same holds for operations, which can now (in principle) be disambiguated like `cap.Nondet::flip()`.

As a drive-by, I simplify the way operations are looked up and drop the unused effect on `Do`, which was in preparation for `do Nondet.flip()`, which now is `do Nondet::flip()`.


b-studios:
@timsueberkrueb there is a LSP Server test failing:

```
==> X effekt.LSPTests.Server publishes term names in the correct order  0.555s munit.ComparisonFailException: LSPTests.scala:1782
1781:
1782:      assertEquals(outerBindings.length, 3)
1783:      assertEquals(outerBindings(0).name, "e1")
values are not the same
=> Obtained
5
-5
+3
```
Can I just update the number or do we want the change the behavior in the server?

b-studios:
@timsueberkrueb Could you double check w.r.t. to the "outline" and holes panel? In particular, I am afraid that operations might now show up twice (once in the namespace and once outside).

examples/pos/namespaced_constructors.effekt
```diff
@@ -0,0 +1,20 @@
...
+}
+type Bars {
+  Foo(f: Foo)
+  Boo(b: Boo)
+}
+
+def consume(b: Bars) = b match {
+  case Foo(Foo::Bar()) => 1
+  case Boo(Boo::Bar()) => 2
+}
```
b-studios, line 16: Here you can see that we now can also disambiguate equally named constructors in patterns.

timsueberkrueb:
> @timsueberkrueb Could you double check w.r.t. to the "outline" and holes panel? In particular, I am afraid that operations might now show up twice (once in the namespace and once outside).

Yes, they do indeed occur twice now.

image

b-studios:
I discussed this with @timsueberkrueb: Ideally, the semantics of

```
type Color { Red(); Green() }
```
is
```
type Color { Color::Red(); Color::Green() }
import Color::*
```
so the unprefixed constructors (and operations) should be listed as imports, not bindings.

b-studios:
I implemented it, but it isn't that simple. Assume two files with the import semantics:

```
// color.effekt
type Color { Color::Red(); Color::Green() }
import Color::*
```
and
```
// colorpicker.effekt
import color
def chooseColor(): Color = Red()
                           ^^^^^
         Cannot find a function named `Red`.
```
then the "generated" `import Color::*` will only be applied in the scope of `color.effekt`, but not in `colorpicker.effekt`.

b-studios:
So, backtracking, the semantics implemented previously (and which I plan to go with for now) is
```
type Color { Color::Red(); Color::Green() }
export Color::*
```
Since `Red` and `Color::Red` refer to the same symbol we can later choose which one to show. Maybe this is easier to solve once we reconsidered qualified names.

dvdvgt:
bump @b-studios

PR #1094, opened 2025-07-24
Visibility Modifiers

b-studios:
This PR reorganizes parsing of modifiers on declaraionts, which are now grouped in `Info`. It also adds one new modifier `private`, which is parsed, but then ignored.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/source/Tree.scala
+++ b/effekt/shared/src/main/scala/effekt/source/Tree.scala
@@ -113,6 +113,29 @@ case class Comment() extends Tree {
 type Doc = Option[String]
 
 /**
+ * The meta-data of a definition / declaration, consisting of everything up
+ * until, but not including the introducing keyword (such as `def`, or `interface`).
+ * For example:
+ *
+ *     /// documentation
+ *     private
+ *     extern
+ *     pure
+ *     def
+ */
+case class Info(
+  doc: Doc,
+  // we use Maybe[Unit] instead of Boolean to have position info for validation errors
+  isPrivate: Maybe[Unit],
+  isExtern: Maybe[Unit],
+  // TODO this should be moved from the Info to extern functions, once we changed the syntax
+  externCapture: Option[CaptureSet]
+)
+object Info {
+  def empty(span: Span) = Info(None, Maybe.None(span), Maybe.None(span), None)
+}
+
+/**
  * The origin of the span
  *
  * Possible values:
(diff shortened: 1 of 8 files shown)


b-studios:
@marvinborner I slightly refactored the parser w.r.t. to `doc`. My goal is to get rid of all usages of `documentedKind` and `maybeDocumentation`, since this will help me to generalize `Doc` to arbitrary meta data about a definition.

Could you check whether my changes so far make sense or whether you can construct an example which either:

1. does not parse
2. has different spans

b-studios:
Sidenote: I removed a couple of `nonterminal` calls since the corresponding function was always called immediately under another `nonterminal` call.

marvinborner:
I'll have to try more later, but it looks good so far.

b-studios:
There are two very annoying things:

1. `featureFlagFromCapture`, this is necessary since parsing the info already consumed the feature flag. This will automatically be better if we drop parsing captures at the front of a definition.
2. `documented` takes a flag `parseCaptures: Boolean`. This is necessary to avoid eating the name of a constructor or operation as a capture set. Again, this will be automatically better once we drop parsing captures at the front of a definition.

effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -112,6 +112,33 @@ case class Comment() extends Tree {
...
+ *     def
+ */
+case class Info(
+  doc: Doc,
+  // we use Maybe[Unit] instead of Boolean to have position info for validation errors
+  isPrivate: Maybe[Unit],
+  isExtern: Maybe[Unit],
+  // TODO this should be moved from the Info to extern functions, once we changed the syntax
+  externCapture: Option[CaptureSet]
+) {
```
b-studios, line 133: I am also very open to changing the design of this info object. In order to report some errors on the flags, they are positioned (using `Maybe[Unit]` is a hack). However, an empty `Info` object still contains a lot of position info and thus complicates the tests etc.

b-studios: We could be more precise by having a type hierarchy of `Info`. This way certain modifiers could be excluded by construction.

b-studios:
I'm fine with merging this now in order to avoid conflicts with other PRs (especially #1097). Merging it now will parse `private` and then simply ignore it. We can later decide what to do with this info :)

b-studios:
@timsueberkrueb what do you think?

b-studios:
@jiribenes is it possible to run this on the community build before merging?

timsueberkrueb:
@b-studios I'm happy with this being merged and rebasing my changes.

b-studios:
Merging since its not (yet) Friday :)

PR #1095, opened 2025-07-25
Fix char default

### LINKED ISSUE #1075: Char does not have a working default

mattisboeckle:
fixes #1075

### DIFF
--- a/examples/stdlib/char/char_default.effekt
+++ b/examples/stdlib/char/char_default.effekt
@@ -1,0 +1,10 @@
+effect fail(): Nothing
+
+def f(): Char / fail = do fail()
+
+def main(): Unit =
+  try {
+    println(show(f()))
+  } with fail {
+    ()
+  }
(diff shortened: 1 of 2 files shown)


marzipankaiser:
Fixes the issue for me, but a regression test would be nice (adding the example from the issue to the tests would be fine IMHO).

marzipankaiser (approved):

examples/stdlib/char/char_default.effekt
```diff
@@ -0,0 +1,10 @@
+effect fail(): Nothing
+
+def f(): Char / fail = do fail()
+
+def main(): Unit =
+  try {
+    println(show(f()))
+  } with fail {
+    ()
+  }
```
marzipankaiser, line 10: Thanks!

marzipankaiser (approved):

PR #1096, opened 2025-07-30
Send syntax-highlighted HTML for bindings

timsueberkrueb:
This is a simple way to get fast syntax highlighting for the holes panel.

image

`HtmlHighlight.scala` is a copy of `AnsiHighlight.scala`.

Corresponding client-side PR: https://github.com/effekt-lang/effekt-vscode/pull/123

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -2,7 +2,7 @@
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.10")
 
 // to generate a javascript file and run the compiler in the browser
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.14.0")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.16.0")
 
 // to have separate projects for jvm and js details
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "1.2.0")


effekt/jvm/src/test/scala/effekt/LSPTests.scala
```diff
@@ -1509,7 +1515,8 @@ class LSPTests extends FunSuite {
                 qualifier = List(),
                 name = "MyInt",
                 origin = BindingOrigin.Defined,
-                definition = "type MyInt = Int"
+                definition = "type MyInt = Int",
+                definitionHtml = "type MyInt = Int"
```
marzipankaiser: Is this intentionally not styled/using html?

effekt/jvm/src/test/scala/effekt/LSPTests.scala
```diff
@@ -1543,6 +1550,7 @@ class LSPTests extends FunSuite {
         |        "name": "x",
         |        "origin": "Defined",
         |        "type": "Int",
+        |        "typeHtml": "Int",
```
marzipankaiser: Same, intentionally no style?

effekt/jvm/src/test/scala/effekt/LSPTests.scala
```diff
@@ -1558,6 +1566,7 @@ class LSPTests extends FunSuite {
         |            "name": "MyInt",
         |            "origin": "Defined",
         |            "definition": "type MyInt = Int",
+        |            "definitionHtml": "type MyInt = Int",
```
marzipankaiser: Same, intentionally no style?

marzipankaiser (approved):
In the tests, some `*Html` properties do not seem to be HTML?
If this is not intentional, maybe look into this?

Otherwise LGTM.

effekt/shared/src/main/scala/effekt/util/HtmlHighlight.scala
```diff
@@ -0,0 +1,105 @@
...
+ */
+object HtmlHighlight {
+  private def htmlEscape(text: String): String = {
+    text
+      .replace("&", "&amp;")
+      .replace("<", "&lt;")
+      .replace(">", "&gt;")
+      .replace("\"", "&quot;")
+      .replace("'", "&#39;")
+  }
```
jiribenes: If were to pull in an extended stdlib module (`scala-xml`) then we could reuse [`scala.xml.Utility.escape`](https://javadoc.io/doc/org.scala-lang.modules/scala-xml_3/latest/index.html#escape-fffff2ca) to do the escaping, though I'm not sure it's worth it. 👀

jiribenes: (Actually, we could use it for the `htmlSpan` too, automagically letting it escape stuff with a `xml"""<span class="${cls}">${text}</span>"""` interpolator, but I'm not sure that one got merged to `scala-xml`?)

b-studios: For a second I thought this is about the Effekt standard library :)

PR #1098, opened 2025-08-05
Use new capture syntax in inlay hints for definitions

### LINKED ISSUE #1077: Wrong capture hint position on documented definitions
I would assume the capture would be before the `def`, _after_ the `/// ...` comments.
Image

Here's the source:
```scala
/// here is my doc comment
/// why is the capture there though?
def main() = println("hello")
```

timsueberkrueb:
Now shows capture inlay hints using the `at` syntax on (function) definitions:

image

This PR does not change the grammar, meaning it remains invalid syntax to explicitly annotate captures in these positions for now.

Resolves #1077.


b-studios (approved):
I think it looks beautiful! Thanks!

PR #1100, opened 2025-08-06
Diagnostics for main function

### LINKED ISSUE #1093: Missing diagnostics for unhandled Effekts in main
Consider the following program:

```scala
type MyError { MyError() }
def main() = {
  do raise[MyError](MyError(), "foo")
}
```

`main` has type `Nothing / { Exception[MyError] }`. When running `effekt bug.effekt`, one gets the following error:

```
[error] /home/tim/Dev/effekt-lang/playground/bug1.effekt:2:1: Main cannot have effects, but includes effects: { Exception[MyError] }
def main() = {
^
```

However, currently, there are no diagnostics provided via the language server for this. This means that the editor does not display this problem and the user only notices this when trying to execute the program.

dvdvgt:
Resolves #1093

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Repl.scala
+++ b/effekt/jvm/src/main/scala/effekt/Repl.scala
@@ -121,11 +121,11 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
     def typecheck(source: Source, config: EffektConfig): Unit =
       parse(source) match {
         case Success(e: Term, _) =>
-          runFrontend(source, module.make(source, e), config) { mod =>
-            // TODO this is a bit ad-hoc
-            val mainSym = mod.exports.terms("main").head
-            val mainTpe = context.functionTypeOf(mainSym)
-            output.emitln(pp"${mainTpe.result}")
+          // TODO this is a bit ad-hoc!!!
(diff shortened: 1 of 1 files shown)


dvdvgt:
Wrong return type:
image

Multiple main functions:
image

Unhandled effects in main:
image

dvdvgt:
Some neg tests add the wrong return type in their main function which was not detected previously

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -79,9 +79,45 @@ object Typer extends Phase[NameResolved, Typechecked] {
       // Store the backtrackable annotations into the global DB
       // This is done regardless of errors, since
       Context.commitTypeAnnotations()
+      checkMain(input.mod)
```
dvdvgt, line 82: Is this really the best place to check this?

effekt/jvm/src/test/scala/effekt/LSPTests.scala
```diff
@@ -122,6 +122,102 @@ class LSPTests extends FunSuite {
     }
   }
```
timsueberkrueb, line 124: Should we have a section

```
// Diagnostics
//
//
```

here?

timsueberkrueb (approved):
Great, thank you :)

PR #1102, opened 2025-08-06
Attach signature to BindingInfo

timsueberkrueb:


effekt/jvm/src/test/scala/effekt/LSPTests.scala
```diff
@@ -1826,81 +1826,67 @@ class LSPTests extends FunSuite {
...
           kind = "Term"
         ),
         TermBinding(
           qualifier = Nil,
           name = "theField",
           origin = "Defined",
-          `type` = Some(
-            value = "Foo1 => String"
+          signature = Some(
+            "def theField(test::N::Foo1: Foo1): String / {}"
```
timsueberkrueb: This needs fixing

PR #1103, opened 2025-08-07
Basic dictionaries

phischu:
- [ ] Documentation
- [ ] Tests/Examples
- [ ] Discuss convention for mutable data structures
- [ ] Optimize singleton branches


libraries/common/dictionary.effekt
```diff
@@ -0,0 +1,133 @@
...
+      }
+    }
+  go(trie, 0)
+}
+
+/// A record pairing a key with a value for use in streams.
+record Tagged[A](key: ByteArray, value: A)
+
+/// Collect key-value pairs from a stream of Tagged elements into a Dictionary.
+def collectTrie[A] { stream: () => Unit / emit[Tagged[A]] }: Dictionary[A] = {
```
jiribenes: To be consistent with the rest of the `collect` methods:
```suggestion
def collect[A] { stream: () => Unit / emit[Tagged[A]] }: Dictionary[A] = {
```

jiribenes:
Could you add a `unify-dictionary` example which is like `unify` (https://github.com/effekt-lang/effekt/blob/master/examples/benchmarks/other/unify.effekt), but with a Dictionary? Then we could run them to get at least one data point to see which one is quicker when (and have one program that showcases the difference)

All of the other examples are trivial, and here we could have one "for free" :)

jiribenes:
As far as methods go, in an ideal world (that is, "if you want somebody to use it" ;)), I'd like it to have `update` ("upsert"), `size`, and `remove`.
Since it's a `trie`, one thing that would make me use it (besides the not-yet-demonstrated performance) would be a `keysWithPrefix` method that would give me all of the keys / keys-and-values that start with a given prefix.

PR #1104, opened 2025-08-09
Add tparams to constructors

mattisboeckle:
The Core Parser still needs to handle tparams, but I was unsure how to add that functionality.


b-studios:
Most importantly you need to adapt type checking to substitute for the parameters.

mattisboeckle:
Do we have dedicated typechecking for core? If so in which file?

b-studios:
We don't type check, but only infer types. This is all defined here: https://github.com/effekt-lang/effekt/blob/master/effekt/shared/src/main/scala/effekt/core/Type.scala

This comment here might be relevant: 
https://github.com/effekt-lang/effekt/blob/356ccac39676753c7f6dfe31e9b1af16f9eafc66/effekt/shared/src/main/scala/effekt/core/Type.scala#L257

Maybe you could add CoreTypeTests similar to these

 https://github.com/effekt-lang/effekt/blob/master/effekt/jvm/src/test/scala/effekt/core/PatternMatchingTests.scala

where you assert for simple terms that the correct type is inferred?

Just start with a single simple test such as `Some[Int](42).tpe == Option[Int]`

PR #1105, opened 2025-08-11
Improve LLVM runner

marvinborner:
This allows for more debug symbols in the LLVM runner and makes it more robust on systems with different clang/opt installations. I also removed the requirement for llc, as we do not use it currently.


marvinborner:
hello please remove opt and use clang instead thanks

marvinborner:
blocked until @b-studios finds a workaround or macos gets fixed

b-studios:
I investigated a bit. Interestingly, the issue I am having with this PR seems to stem from running `sbt` and hence `clang` under the wrong architecture. I don't know how this happened, but my terminal was in "Rosetta"-mode

```
~ uname -m
x86_64
```

leading to the error (copied here for google and bots):
```
ld: warning: ignoring file '/opt/homebrew/lib/libuv.dylib': found architecture 'arm64', required architecture 'x86_64'
Undefined symbols for architecture x86_64:
  "_uv_default_loop", referenced from:
      _main in cc-caf16e.o
  "_uv_loop_close", referenced from:
      _main in cc-caf16e.o
  "_uv_run", referenced from:
      _main in cc-caf16e.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Fixing this on my side by starting the terminal in `arm`-mode, your PR goes through without any other changes (that is, it picks `clang-19`, but the preinstalled `clang` also works).

b-studios (approved):

PR #1109, opened 2025-08-13
Parse capture annotation syntax for (extern) function definitions

### LINKED ISSUE #1106: Parse capture annotations for definitions on RHS
Parse:
```scala
def foo() at {io, global}: Int
```
instead of
```scala
def {io, global} foo(): Int
```

(and don't forget to do the same for externs)

We already have the parser code for boxes.

dvdvgt:
Resolves #1106

```
extern? <CAPTURES> def ... : ... = ...
```

becomes

```
extern? def ... at <CAPTURES> : ... = ...
```

or to say it in terms of a regex

```
/(^( *extern) *(\w+|\{ *((\w+ *, *)| *\w+ *)* *\}) *)(?=def )(.*)(: *.*)$/gm

$2 $6 at $3$7
```


b-studios:
Thanks, this looks like a lot of work!

I haven't yet reviewed the PR, but the adjusted code examples look great to me. One thing I am wondering is: should we deprecate `pure` and use `{}` instead? It feels weird to have `pure` in the capture set position (while it looked okaish as `extern pure def`).

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -778,7 +778,7 @@ object Typer extends Phase[NameResolved, Typechecked] {
 
   def synthDef(d: Def)(using Context, Captures): Result[Unit] = Context.at(d) {
     d match {
-      case d @ source.FunDef(id, tps, vps, bps, ret, body, doc, span) =>
+      case d @ source.FunDef(id, tps, vps, bps, cpt, ret, body, doc, span) =>
         val sym = d.symbol
         // was assigned by precheck
         val functionCapture = Context.lookupCapture(sym)
```
b-studios, line 790: Since you already allow annotating captures on non-extern functions, here we might want to implement the check. That is, conceptually `functionCapture <: cpt`.

b-studios: On the other side, in precheck `functionCapture` could be (and maybe is already?) be instantiated to be `cpt` instead of a capture variable.

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -707,7 +707,7 @@ object Typer extends Phase[NameResolved, Typechecked] {
   // not really checking, only if defs are fully annotated, we add them to the typeDB
   // this is necessary for mutually recursive definitions
   def precheckDef(d: Def)(using Context): Unit = Context.focusing(d) {
-    case d @ source.FunDef(id, tps, vps, bps, ret, body, doc, span) =>
+    case d @ source.FunDef(id, tps, vps, bps, cpt, ret, body, doc, span) =>
       val fun = d.symbol
 
       // (1) make up a fresh capture unification variable and annotate on function symbol
```
b-studios: Ah, here is the location where instead of `freshCaptVar` (if there is a an annotated capture set), we need to use the symbolic representation of `cpt` (remember, `cpt` is syntax here, not its meaning).

dvdvgt:
I changed the syntax for the empty capture set for extern definitions from `at pure` to `at {}`

dvdvgt:
`AnnotateCapture` now annotates the given capture set. Otherwise, the inferred one is annotated.

effekt/shared/src/main/scala/effekt/source/ResolveExternDefs.scala
```diff
@@ -39,24 +39,24 @@ object ResolveExternDefs extends Phase[Typechecked, Typechecked] {
...
                 + s"please annotate it with a feature flag (Supported by the current backend: ${Context.compiler.supportedFeatureFlags.mkString(", ")})")
             }
 
-            val d = Def.ExternDef(capture, id, tparams, vparams, bparams, ret, List(body), doc, span)
+            val d = Def.ExternDef(id, tparams, vparams, bparams, capture, ret, List(body), doc, span)
             Context.copyAnnotations(defn, d)
             Some(d)
           case ExternBody.EffektExternBody(featureFlag, body, span) =>
-            val d = Def.FunDef(id, tparams, vparams, bparams, Some(ret).spanned(ret.span), body, doc, span)
+            val d = Def.FunDef(id, tparams, vparams, bparams, Some(capture).spanned(capture.span), Some(ret).spanned(ret.span), body, doc, span)
```
b-studios: How is this different from just `capture` and `ret`?

b-studios: Ah I see, captures on externs are always there

b-studios (approved):

b-studios:
@timsueberkrueb mentioned that, for consistency, we could also allow capture annotations on `DefDef`s, that is

```effekt
def foo at {}: Counter = new Counter {
  def get() = ...
  def inc() = ...
}
```
but I'd do this in a separate PR to get this merged ASAP

b-studios:
Please note that this is a breaking change and we will need to go through ALL of our projects and community build etc.

WDYT should we bundle it with some other PRs (such as finally changing `type` to `data`)?

timsueberkrueb:
Personally, I would rather merge it to prevent merge conflicts.

dvdvgt:
You raise a good point. If this is a breaking change anyway, I'd rather bundle it one PR and write a script for migrating (e.g. the community repos) automatically, all in one go. That shall include:

- changing `type` to `data` for data type definitions
- allow capture syntax for `DefDef`s

Am I missing anything else?

jiribenes:
> changing type to data for data type definitions

Just to be sure, you mean only for actual data definitions, right?
(so `type PersonId = Int` stays as-is)

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -707,24 +707,24 @@ object Typer extends Phase[NameResolved, Typechecked] {
...
 
-      // (1) make up a fresh capture unification variable
-      val cap = Context.freshCaptVar(CaptUnificationVar.BlockRegion(d))
+      // (1) make up a fresh capture unification variable or use the annotated captures
+      val cap = obj.caps.getOrElse(Context.freshCaptVar(CaptUnificationVar.BlockRegion(d)))
 
-      // (2) annotate capture variable and implemented blocktype
-      Context.bind(obj, Context.resolvedType(tpe).asInterfaceType, cap)
+      // (2) annotate capture variable
+      Context.bind(obj, cap)
```
dvdvgt: Is this significant? Do we need to annotate the type of the implemented interface?

dvdvgt: We overwrite it anyway in `synthDef` without ever using it again.

dvdvgt: Disregard that. Answered by failing tests :)

dvdvgt:
```
#!/bin/bash

if [ $# -eq 0 ]; then
    echo "Error: No filenames provided"
    exit 1
fi

for file in "$@"; do
    if [ ! -f "$file" ]; then
        echo "Error: not a file / not found: '$file'"
        continue
    fi
    echo "Processing file: $file"
    perl -i -pe 's/^( *extern) *(\w+|\{ *((\w+ *, *)| *\w+ *)* *\}) *((def )(.*))(: *.*)$/\1 \7 at \2\8/g' $file
done
```

Migration script for converting all `extern def`s. (Don't ask me how this REGEX works)
There shouldn't be anything else to migrate since the syntax for `FunDef`s and `DefDef`s is new and should occur nowhere as of now.

PR #1111, opened 2025-08-17
Try to fix performance regression introduced by a0033bf

b-studios:
For whatever reason, before we inlined `BODY` into the else branch

```
val x = if (...) {
  shift(p) { k => () }
} else {
  ...
};
BODY
```

but didn't anymore after a0033bf. 

This PR fixes this by always introducing joinpoints for ifs and matches. Then, when a continuation is used exactly ones in a follow-up pass, we will inline it. This way, more things become direct-style, avoiding pushes. It comes at the cost of potentially not seeing through the joinpoint (as in the case of `anf.effekt.md`) and of having more known jumps. 

However, our cost model says that known jumps are generally cheaper than pushes, so it should be worth it.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
+++ b/effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
@@ -264,6 +264,7 @@ object Normalizer { normal =>
           Stmt.Match(sc, List((id2, BlockLit(tparams2, cparams2, vparams2, bparams2,
             normalizeVal(id, tpe, body2, body)))), None)
 
+        // Introduce joinpoints that are potentially later inlined or garbage collected
         // [[ val x = if (cond) { thn } else { els }; body ]] =
         //   def k(x) = [[ body ]]
         //   if (cond) { [[ val x1 = thn; k(x1) ]] } else { [[ val x2 = els; k(x2) ]] }


b-studios:
The `examples/benchmarks/are_we_fast_yet/queens.effekt` benchmark actually requires us duplicating the continuation. The program has the following shape:

```
val _: Unit = if (v_r_6808) {
  // ...
  if (c == (n - 1)) {
    shift(p) { {k_7612} => return true } // ABORTS
  } else {
    val v: Bool = placeQueen_3868(c + 1);
    if (v) {
      shift(p) { {k_7613} => return true } // ABORTS
    } else {
      // ...
      return tmp_7573 // DOESN'T ABORT
    }
  }
} else {
  return ()  // DOESN'T ABORT
};
loop_6796(i + 1)
```
There are two usages of the continuation which do not abort, so we would need to duplicate the continuation `val _ = loop_6796(i + 1)`.

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -706,12 +706,12 @@ class VMTests extends munit.FunSuite {
...
 
     examplesDir / "casestudies" / "anf.effekt.md" -> Some(Summary(
-      staticDispatches = 6925,
+      staticDispatches = 6140,
       dynamicDispatches = 443,
       patternMatches = 5098,
-      branches = 8018,
-      pushedFrames = 4298,
-      poppedFrames = 4285,
+      branches = 8110,
```
b-studios, line 712: Here we have a few more branches

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -690,12 +690,12 @@ class VMTests extends munit.FunSuite {
...
 
     examplesDir / "casestudies" / "parser.effekt.md" -> Some(Summary(
-      staticDispatches = 12887,
+      staticDispatches = 11375,
       dynamicDispatches = 783,
       patternMatches = 9427,
-      branches = 14696,
-      pushedFrames = 7583,
-      poppedFrames = 7559,
+      branches = 14892,
```
b-studios, line 696: and here

PR #1114, opened 2025-08-25
Arity raising for record types

konradbausch:


b-studios:
Hey, thanks this looks great. 

1. do you have some insights on how this affects performance?
2. could you delete the `metals.sbt` files from git?

konradbausch:
Thanks!

It seems to affect performance positively. Our current benchmarks that use records are not running faster, since as far as I saw, they get inlined or constant-folded. But they also aren't slower. I came up with some examples that show how performance increases when the changes actually persist through compilation. The [record-passing](https://github.com/effekt-lang/effekt/blob/417cfa2d163e086aaa3c1812890ec27da996815d/examples/benchmarks/arity_raising/record_passing.effekt) example program might seem a bit constructed, so I also added a [matrix determinant](https://github.com/effekt-lang/effekt/blob/417cfa2d163e086aaa3c1812890ec27da996815d/examples/benchmarks/arity_raising/matrix_determinant.effekt) example, hoping it's a bit more realistic to an actual program.

Here are the results for doing 10 warmup runs and then 50 normal runs:

**record_passing**

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `main` | 199.1 ± 4.4 | 189.8 | 210.4 | 9.18 ± 0.33 |
| `konradbausch/arity-raising` | 21.7 ± 0.6 | 21.0 | 24.0 | 1.00 |

**matrix_determinant**

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `main` | 378.3 ± 8.8 | 362.5 | 408.3 | 14.34 ± 0.53 |
| `konradbausch/arity-raising` | 26.4 ± 0.8 | 25.2 | 28.1 | 1.00 |

It seems to only have improvements on the LLVM backend, and I'm not sure why yet. I haven't tried to find examples with negative performance effects yet and will look into that as well.

I've also removed the `metals.sbt` files.

PR #1119, opened 2025-08-27
Renormalize directly again

b-studios:
While this might be achieved by the second or third normalizer run, normalizing directly showed a bit of an performance improvement (compile time) for me on average.


b-studios:
@phischu feel free to merge once CI passes

phischu (approved):
Shouldn't hurt.

PR #1120, opened 2025-08-28
Include definition location in LSP hole bindings 

JakubSchwenkbeck:
To enable client-side functionalities in the VSCode extension like 'jump-to-definition' for the bindings, we need to also send the file uri 

See : https://github.com/effekt-lang/effekt-vscode/pull/140


timsueberkrueb:
Nice, but this only works for bindings that have been declared in the same file. If you attach the URIs to the individual bindings instead, it will also work for imports!

JakubSchwenkbeck:
the `Convert.toUri` function from Kiama cannot be accessed in `shared` so I implemented a similar function, any better ideas?

timsueberkrueb (approved):
Great work!

PR #1121, opened 2025-08-28
Add benchmarks "folkore to fact" and "no fib"

b-studios:


b-studios:
Please rename (and check) to folklore_to_fact and fix tests in VMTests

PR #1123, opened 2025-08-31
Show instances for ground types

mattisboeckle:
Implement new Core -> Core Phase which
synthesizes `show` definitions for any type based on ground show definitions

For example 
```scala
type MaybeInt {
  Nothing()
  Just(x: Int)
}
```
could generate something along the lines of
```scala
def show(value: MaybeInt): String = value match {
  case Nothing() => "Nothing"
  case Just(x: Int) => "Just(${x.show})"
}
```

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/Show.scala
+++ b/effekt/shared/src/main/scala/effekt/core/Show.scala
@@ -47,7 +47,7 @@ object Show extends Phase[CoreTransformed, CoreTransformed] {
 
   override def run(input: CoreTransformed)(using Context): Option[CoreTransformed] = input match {
     case CoreTransformed(source, tree, mod, core) => {
-      // 3. Synthesize `show` definitions, create (ValueType -> (show) Id) map for context
+      // Synthesize `show` definitions, create (ValueType -> (show) Id) map for context
 
       given ctx: ShowContext = ShowContext(collection.mutable.Map.empty, collection.mutable.Map.empty, collection.mutable.Map.empty)
       given dctx: DeclarationContext = DeclarationContext(core.declarations, core.externs)
@@ -79,7 +79,6 @@ object Show extends Phase[CoreTransformed, CoreTransformed] {
   }
 
   def transform(operation: Operation)(using Context, ShowContext, DeclarationContext): Operation = operation match {
-    // Maybe need to pass tparams here to be able to lookup while transforming body
     case Operation(name, tparams, cparams, vparams, bparams, body) => Operation(name, tparams, cparams, vparams, bparams, transform(body))
   }
 
@@ -177,7 +176,6 @@ object Show extends Phase[CoreTransformed, CoreTransformed] {
           Put(ref, annotatedCapt, value_, transform(body))
         }
       case Reset(body) => Reset(transform(body))
-      // FIXME: Does k need to be handled?
       case Shift(prompt, k, body) => Shift(transform(prompt), k, transform(body))
       case Resume(k, body) => Resume(transform(k), transform(body))
       case Hole(tpe, span) => Hole(tpe, span)
@@ -228,45 +226,20 @@ object Show extends Phase[CoreTransformed, CoreTransformed] {
 
     vt match {
       case Type.TString =>
-        // (value: String) { return value }
         val stmt = Stmt.Return(paramValueVar)
         Some(generateDef(stmt))
       case Type.TUnit =>
-        /*
-        (value: Unit) {
-          "()"
-        }
-        */
         val stmt = Stmt.Return(Expr.Literal("()", TString))
         Some(generateDef(stmt))
       case Type.TInt =>
-        /*
-        (value: Int) {
-          %z = call %Pos @c_bytearray_show_Int(%Int ${value})
-          ret %Pos %z
-        }
-        */
         Some(generateShow(Type.TInt))
       case Type.TChar =>
         Some(generateShow(Type.TChar))
       case Type.TByte =>
-        /*
-        (value: Byte) {
-          %z = call %Pos @c_bytearray_show_Byte(%Byte ${value})
-          ret %Pos %z
-        }
-        */
         Some(generateShow(Type.TByte))
       case Type.TDouble =>
-        /*
-        (value: Double) {
-          %z = call %Pos @c_bytearray_show_Double(%Double ${value})
-          ret %Pos %z
-        }
-        */
         Some(generateShow(Type.TDouble))
       case Type.TBoolean =>
-        // if (value) "true" else "false"
         val stmt = Stmt.If(paramValueVar, Stmt.Return(Expr.Literal("true", TString)), Stmt.Return(Expr.Literal("false", TString)))
         Some(generateDef(stmt))
       case ValueType.Data(name, targs) =>


mattisboeckle:
The current implementation can already generate the following core code for

```scala
type MaybeInt {
  Nothing()
  Just(x: Int, y: Int)
}
```

```
def myshow_3970(value_3971: MaybeInt_3714) = {
  def b_k0_3972() = return "Nothing"
  def b_k1_3973(x_3722: Int_389, y_3723: Int_389) = return infixConcat_39("Just(", infixConcat_39(show_15(x_3722), infixConcat_39(", ", infixConcat_39(show_15(y_3723), ")"))))
  val v_r_3974: String_394 = value_3971 match {
    case Nothing_3718 { () => 
      b_k0_3972()
    }
    case Just_3719 { (x_3722: Int_389, y_3723: Int_389) => 
      b_k1_3973(x_3722, y_3723)
    }
  };
  return v_r_3974
}
```
but when being called in main like this
```
let tmp_3960 = make MaybeInt_3714 Just_3719(42, 45)
let tmp_3963 = myshow_3970(tmp_3960)
let tmp_3964 = println_1(tmp_3963)
```
the compiler fails in Transformer with
`Unsupported expression: PureApp(BlockVar(myshow,Function(List(),List(),List(Data(MaybeInt,List())),List(),Data(String,List())),Set()),List(),List(ValueVar(tmp,Data(MaybeInt,List()))))`

b-studios:
Since it is a "user defined" function, you need to call it with App, not PureApp

effekt/shared/src/main/scala/effekt/core/Type.scala
```diff
@@ -62,6 +62,8 @@ object Type {
   val TString = ValueType.Data(builtins.StringSymbol, Nil)
   val TDouble = ValueType.Data(builtins.DoubleSymbol, Nil)
 
+  val Builtins = List(TUnit, TInt, TChar, TByte, TBoolean, TString, TDouble)
```
b-studios: Nitpick:

```suggestion
  val builtins = List(TUnit, TInt, TChar, TByte, TBoolean, TString, TDouble)
```

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -726,8 +726,8 @@ class VMTests extends munit.FunSuite {
       dynamicDispatches = 3224986,
       patternMatches = 1497945,
       branches = 304507,
-      pushedFrames = 2643313,
-      poppedFrames = 2066048,
+      pushedFrames = 2643512,
+        poppedFrames = 2066247,
```
b-studios: Great that the VMTests are back to normal :) Also great to see that the optimizer actually inlines all of it.

mattisboeckle: Sadly not everything works anymore. Sometimes I'm missing a type variable which I need to investigate

effekt/shared/src/main/scala/effekt/core/DeadCodeElimination.scala
```diff
@@ -0,0 +1,20 @@
+package effekt.core
+
+import effekt.PhaseResult.CoreTransformed
+import effekt.context.Context
+import effekt.core.optimizer.Deadcode
+import effekt.Phase
+
+object DeadCodeElimination extends Phase[CoreTransformed, CoreTransformed] {
```
b-studios: Does it make sense to just turn `Deadcode` into a Phase?

That is,

```scala
class Deadcode(reachable: Map[Id, Usage]) extends core.Tree.TrampolinedRewrite, Phase[CoreTransformed, CoreTransformed] {
    val phaseName: String = "deadcode-elimination"

    def run(input: CoreTransformed)(using Context): Option[CoreTransformed] =
        input match {
                val term = Context.ensureMainExists(mod)
                val dce = Context.timed("deadcode-elimination", source.name) {
                    Deadcode.remove(term, core)
                }
                Some(CoreTransformed(source, tree, mod, dce))
        }
        ...
}
```
This way we do not need a separate file.

effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
```diff
@@ -44,7 +44,7 @@ class Deadcode(reachable: Map[Id, Usage]) extends core.Tree.TrampolinedRewrite {
         declarations.map(rewrite),
         // drop unreachable externs
         m.externs.collect {
-          case e: Extern.Def if used(e.id) => e
+          case e: Extern.Def if used(e.id) || e.id.name.name == "show" || e.id.name.name == "showBuiltin" || e.id.name.name == "infixConcat" => e
```
b-studios: Can you add a comment that we keep those externs because they are needed for generating show instances?

effekt/shared/src/main/scala/effekt/core/DeadCodeElimination.scala
mattisboeckle: Sure I'll do that

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
mattisboeckle: My code seems to break when using `inspect` we can discuss how to handle this

effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
```diff
@@ -3,8 +3,11 @@ package core
 package optimizer
 
 import util.Trampoline
+import effekt.PhaseResult.CoreTransformed
+import effekt.context.Context
 
-class Deadcode(reachable: Map[Id, Usage]) extends core.Tree.TrampolinedRewrite {
+class DeadcodeRewrite(reachable: Map[Id, Usage])
```
b-studios: Why did you rename this?

b-studios: Couldn't this be named `Deadcode` as well?

timsueberkrueb: Fixed.

effekt/shared/src/main/scala/effekt/core/DeclarationContext.scala
```diff
@@ -71,6 +75,7 @@ class DeclarationContext(
   }
 
   def findExternDef(id: Id): Option[Extern.Def] = externDefs.get(id)
+  def findExternDef(name: String, params: List[ValueType]): Option[Extern.Def] = externDefsName.get((name, params))
```
b-studios, line 78: Note: this uses Scala equality and hashCode on `ValueType` which is not what we might need

effekt/shared/src/main/scala/effekt/core/Show.scala
```diff
@@ -0,0 +1,404 @@
...
+
+    def emit(id: Id, stmt: Stmt): Unit =
+      bindings.append(Binding.Val(id, stmt))
+  }
+
+  // This will check if we have already generated a Show instance for the given type and generate it if we didn't
+  def getShowBlockVar(vt: ValueType)(using ctx: ShowContext)(using Context, DeclarationContext): Block.BlockVar =
+
+    val showId = ctx.showNames.getOrElse(vt, {
+      generateShowInstance(vt) match {
```
b-studios, line 38: Again, this uses equality and hashCode, which might not be robust enough

effekt/shared/src/main/scala/effekt/core/Show.scala
```diff
@@ -0,0 +1,404 @@
...
+    case New(impl) => New(transform(impl))
+    case Unbox(pure) => Unbox(transform(pure))
+  }
+
+  def transform(implementation: Implementation)(using Context, ShowContext, DeclarationContext): Implementation = implementation match {
+    case Implementation(interface, operations) => Implementation(interface, operations map transform)
+  }
+
+  def transform(operation: Operation)(using Context, ShowContext, DeclarationContext): Operation = operation match {
+    // Maybe need to pass tparams here to be able to lookup while transforming body
```
b-studios: what does this comment entail?

effekt/shared/src/main/scala/effekt/core/Show.scala
```diff
@@ -0,0 +1,404 @@
...
+
+    var bindings = mutable.ListBuffer.empty[Binding]
+
+    def withBindings(block: => Stmt): Stmt =
+      val outer = bindings
+      bindings = mutable.ListBuffer.empty[Binding]
+      Binding(outer.toList, block)
+
+    def emit(id: Id, stmt: Stmt): Unit =
+      bindings.append(Binding.Val(id, stmt))
```
b-studios, line 31: I just refactored it, but do not understand why it is in this order? Why do we reify the _outer_ bindings and not those generated by `block`?

effekt/shared/src/main/scala/effekt/core/Show.scala
```diff
@@ -0,0 +1,404 @@
...
+        */
+        val stmt = Stmt.Return(Expr.Literal("()", TString))
+        Some(generateDef(stmt))
+      case Type.TInt =>
+        /*
+        (value: Int) {
+          %z = call %Pos @c_bytearray_show_Int(%Int ${value})
+          ret %Pos %z
+        }
+        */
```
b-studios: Why these comments?

examples/casestudies/parser.effekt.md
```diff
@@ -35,7 +35,7 @@ input stream and fails, if it does not match.
 def accept { p: Token => Bool } : Token / Parser = {
   val got = do next();
   if (p(got)) got
-  else do fail("Unexpected token " ++ got.show)
+  else do fail("Unexpected token " ++ lexer::show(got))
```
b-studios, line 38: This is surprising, don't we keep existing shows?

examples/pos/probabilistic.effekt
```diff
@@ -72,17 +67,6 @@ def choose3[R] { x: => R } { y: => R } { z: => R } : R / fork =
...
-def showRounded(d: Double): String = show(d.round(2))
-
-def show(w: Weighted[Bool]): String = w match {
-  case MkWeighted(weight, value) => "MkWeighted(" ++ showRounded(weight) ++ ", " ++ show(value) ++ ")"
-}
-
-def show(w: Weighted[Var]): String = w match {
-  case MkWeighted(weight, Observed(value)) => "MkWeighted(" ++ showRounded(weight) ++ ", Observed(" ++ show(value) ++ "))"
-  case MkWeighted(weight, Unobserved()) => "MkWeighted(" ++ showRounded(weight) ++ ", Unobserved())"
-}
```
b-studios, line 84: fantastic!

b-studios:
@timsueberkrueb is this CI failure a known flaky one?

effekt/shared/src/main/scala/effekt/core/Show.scala
mattisboeckle: I think this is just my general lack of understanding which tparams may be relevant for generating show instances (or some ideas of mono got mixed in).

The only relevant params for show should be directly in show, meaning this can be removed

effekt/shared/src/main/scala/effekt/core/Show.scala
mattisboeckle: They are unnecessary now. I'll do another cleanup pass and remove them

examples/casestudies/parser.effekt.md
mattisboeckle: We do, but we need to fully qualify the name here.
It would not be a problem if the show were in the same file
```
Ambiguous overload.
  There are multiple overloads, which all would type check:
  - effekt::show: [R](R) => String
  - lexer::show:  Token => String
```

mattisboeckle:
Just for reference because I'm rerunning the CI. The failing (flaky?) tests were
- Internal: `effekt.LSPTests.inlayHint shows the io capture in explicit box syntax`
- Chez: `effekt.ChezSchemeCPSTests.examples/neg/issue548c.effekt (chez-cps)`

PR #1143, opened 2025-10-02
Jump with arguments in machine

### LINKED ISSUE #1134: Jump with arguments in machine
The intermediate representation in `machine/Tree.scala` currently passes arguments with explicit substitutions. We should use a normal jump with arguments instead. This might require us to pass a substitution during transformation.

phischu:
Fixes #1134


marvinborner (approved):
LGTM, very useful for readability and removes some code and makes things clearer, great job!

jiribenes:
This made `product_early` 22% slower btw 👀

PR #1145, opened 2025-10-02
Show `extern def` capture inlay hint if it is the default capture

### LINKED ISSUE #1138: Mandatory capture annotation in extern definitions
I was confused by the definition of `println`:

```
extern def println(value: String): Unit = ...
```

There is no capture annotation and I briefly thought the default is `pure`, but it is `io`, which is good.
However, we could make the capture annotation mandatory to users would have to write:

```
extern def println(value: String) at io: Unit =
```

dvdvgt:
Fixes #1138


jiribenes (approved):
Nice, thanks!
Feel free to merge once the tests pass (LSP tests might also need to be changed)

Issue #1080, opened 2025-07-11
API for processes


jiribenes:
I'd like there to be an API that can spawn processes similarly to `popen(3)` / `subprocess.run`.

The high-level API would be something like:
```scala
import process

process::spawn("cp", ["from.txt", "to.txt"]) // : ???
```

There are two main questions:

1. How to actually implement this / what properties do we want?

Should `spawn` only return the process ID? What would be the callback when the command gets run? Is it (a-)synchronous?

NodeJS has [`child_process[.spawn]`](https://nodejs.org/api/child_process.html) for this, libuv (for the LLVM backend) has [`spawn`](https://docs.libuv.org/en/latest/guide/processes.html). It's necessary to have such an API that works on both the [Node.]JS and the LLVM backends.

2. How do we communicate with the process?

In simpler terms, how do we get the stdin/stdout/stderr of the new process?
Can/should it be streaming? How do we process failures (`exit = 1`) and return codes in general? Can we do something with signals? IPC?...

Issue #1087, opened 2025-07-16
Effekt program ran with `effekt` executes IO operations in incorrect order


timsueberkrueb:
Consider the following program:

```scala
import stringbuffer

def prompt_for_user_name(): String = {
  println("What's your name?")
  readln()
}

def main() = {
  val username = prompt_for_user_name()
  println(s"Hello ${username}")
}
```

Output:

```console
> ./out/foo
What's your name?
xxx
Hello xxx
> effekt foo.effekt
xxx
What's your name?
Hello xxx
```

b-studios:
Really bad publicity for a language that's called Effekt ;)

This is probably also related to all the jvm / sbt io piping that @marzipankaiser had to go through

jiribenes:
We talked about this being potentially bad multiple times in https://github.com/effekt-lang/effekt/pull/981 which added it:

> The biggest potential issues here are flushing & the interaction between `readln` and `println`. The following program should _first_ print the prompt, then read the user's input, and then finally print the output.
> 
> ```scala
> def main() = {
>   println("What's your name?") // prompt
>   val name = readln()
>   println("Hello, " ++ name ++ "!")
> }
> ```


And then @marzipankaiser [already reported on the PR that it does not work on LLVM](https://github.com/effekt-lang/effekt/pull/981#issuecomment-2873020650) with #982 which already fixed (or tried to fix) the jvm/sbt-related flushing — but yeah, it looks like something to do with flushing...

PR #1146, opened 2025-10-02
Minimal subprocess API

marzipankaiser:
First steps to implement #1080


marzipankaiser:
No Valgrind errors 🎉  - Now there's just a weird ordering issue left.

marzipankaiser:
The order thing looks to be related to #1087 . There, too, the `println`s are delayed until after other io operations 🤔

marzipankaiser:
Locally, merging #1198 fixes the issue for me.

marzipankaiser:
[Nvm, would be unsafe, too]

If we (I) could handle blocks better in FFI, a nicer API would (IMHO) be:
```
spawn(cmd: String, args: Array[String]){ options: {SpawnOptions} => Unit }: ...
withStdoutTo{ opts: SpawnOptions }{ onChunk: String => Unit }: Unit
```
etc.
This would also allow, e.g. emitting the output etc.

libraries/common/process.effekt
```diff
@@ -7,3 +8,101 @@ extern def exit(errorCode: Int) at io: Nothing =
...
+extern def default() at {}: SpawnOptions =
+  jsNode """{ stdio: [0,1,2], onSpawn: (p) => {} }"""
+  llvm """
+    %opts = call %Pos @c_spawn_options_default()
+    ret %Pos %opts
+  """
+extern def onStdout(opts: SpawnOptions, callback: String => Unit at {io, global}) at {}: SpawnOptions =
+  jsNode """(function() {
+    let old = ${opts};
+    old.stdio[1] = 'pipe';
```
jiribenes, line 38: This looks mutable to me.
Couldn't there be an issue when we try to reuse existing `SpawnOptions`?

Should we instead copy `opts` here?

libraries/common/process.effekt
```diff
@@ -7,3 +8,101 @@ extern def exit(errorCode: Int) at io: Nothing =
...
+    let oldSpawn = old.onSpawn;
+    old.onSpawn = function(p) {
+      oldSpawn(p);
+      p.stdout.on('data', (chunk) => {
+        $effekt.runToplevel((ks, k) => (${callback})(chunk, ks, k));
```
jiribenes, line 43: Is this `runTopLevel` or `run`? 🤔

libraries/common/process.effekt
```diff
@@ -7,3 +8,101 @@ extern def exit(errorCode: Int) at io: Nothing =
...
+extern def withStdin(opts: SpawnOptions, callback: WritableStream => Unit at {io, global}) at {}: SpawnOptions =
+  jsNode """(function () {
+    let old = ${opts};
+    old.stdio[0] = 'pipe'
+    let oldSpawn = old.onSpawn;
+    old.onSpawn = function(p) {
+      oldSpawn(p);
+      p.on('spawn', () => {
+        $effekt.runToplevel((ks, k) => (${callback})(p.stdin, ks, k));
+      });
```
jiribenes, line 77: Can the process fail to spawn at all? 🤔

libraries/common/process.effekt
```diff
@@ -7,3 +8,101 @@ extern def exit(errorCode: Int) at io: Nothing =
...
+    };
+    return old;
+  })()"""
+  llvm """
+    %opts = call %Pos @c_spawn_options_pipe_stdin(%Pos ${opts}, %Pos ${callback})
+    ret %Pos %opts
+  """
+extern def write(to: WritableStream, s: String) at io: Unit =
+  jsNode """${to}.write(${s})"""
```
jiribenes, line 86: This likely works now, but NodeJS can (and will) sometimes return `false`, and then we should be more async and wait for the `'drain'` event.
https://nodejs.org/api/stream.html#writablewritechunk-encoding-callback

libraries/common/process.effekt
```diff
@@ -7,3 +8,101 @@ extern def exit(errorCode: Int) at io: Nothing =
...
+  llvm """
+    call void @c_close_stream(%Pos ${to})
+    ret %Pos zeroinitializer
+  """
+
+extern def spawn(cmd: String, args: Array[String], options: SpawnOptions) at async: Int  =
+  jsNode """$effekt.capture(k => {
+    let p = spawn(${cmd}, ${args}, ${options});
+    ${options}.onSpawn(p);
+    p.on('close', (exitcode, sig) => { k(exitcode) });
```
jiribenes, line 102: This ignores situations in which the child process is terminated.

I _thought_ we had a string->number map for this somewhere in `io/error`, but I couldn't find it...
Anyways, here's something like what could be there (but I don't like the implementation) 
```js
  p.on('close', (exitcode, sig) => { 
    if (sig) {
      const sigMap = {
        'SIGHUP':1, 'SIGINT':2, 'SIGQUIT':3, 'SIGILL':4, 'SIGTRAP':5,
        'SIGABRT':6, 'SIGBUS':7, 'SIGFPE':8, 'SIGKILL':9, 'SIGUSR1':10,
        'SIGSEGV':11, 'SIGUSR2':12, 'SIGPIPE':13, 'SIGALRM':14, 'SIGTERM':15
      };
      k(128 + (sigMap[sig] || 15)); // '15' is default, this is obviously weird
    } else {
      k(exitcode || 0); // the `|| 0` is just defensive here
    }
  });
```

libraries/common/process.effekt
marzipankaiser: It is mutable. I'm still unsure how an API that is both safe and does not extensively copy things on the C/JS side.
Having `SpawnOptions` be a `extern interface` (and then explicitly mutated) would be more ideal but I'm not sure how to pass `extern interface`s over the FFI boundary.
I.e. something like `spawn(...){ options: {SpawnOptions} => Unit }: ...`

libraries/common/process.effekt
marzipankaiser: It is a callback called from the JS side, so from somewhere in the event loop.
AFAICT, this means `runToplevel`

libraries/common/process.effekt
marzipankaiser: It might. This is not currently handeled (but could, and probably should, be)

libraries/common/process.effekt
marzipankaiser: If we can reasonably from FFI, it would also be nice to just `do signal('SIGHUP')` or sim. instead of encoding it into an int 🤔

marzipankaiser:
Since there don't seem to be any new comments, merging so we can build on the current state.
Opening issue for the missing/to-be-improved parts: #1238

Issue #801, opened 2025-01-24
Bad codegen for negative types on LLVM


jiribenes:
_Extracted from https://github.com/effekt-lang/effekt/issues/795#issuecomment-2609496261_

Hypothesis: The LLVM backend has bad codegen for negative types.

Here are the two test cases that fail:
```
$ EFFEKT_DEBUG=1 effekt --backend=llvm --ir-write-all --debug --optimize examples/stdlib/json.effekt
opt: ./out/json.ll:4001:9: error: multiple definition of local value named 'd_24_132_3_22162'
        %d_24_132_3_22162 = insertvalue %Neg %vtable_temporary_865, %Object %closure_861, 1
        ^

$ EFFEKT_DEBUG=1 effekt --backend=llvm --ir-write-all --debug --optimize examples/stdlib/buffer.effekt
Instruction does not dominate all uses!
  %buffer_24_4387 = insertvalue %Neg %vtable_temporary_1257, ptr %closure_1239, 1
  call void @shareNegative(%Neg %buffer_24_4387)
Instruction does not dominate all uses!
  %buffer_24_4387 = insertvalue %Neg %vtable_temporary_1257, ptr %closure_1239, 1
  store %Neg %buffer_24_4387, ptr %buffer_24_4387_pointer_1255, align 8, !noalias !0
Instruction does not dominate all uses!
  %buffer_24_4387 = insertvalue %Neg %vtable_temporary_1257, ptr %closure_1239, 1
  call void @eraseNegative(%Neg %buffer_24_4387)
opt: ./out/buffer.ll: error: input module is broken!
```

### Analysis

I think it's the same problem, actually.
For negative types, we generate:
```
define tailcc void @xyz_clause_1(...) { ... }
define tailcc void @xyz_clause_2(...) { ... }
```
which always share what would be:
```
; new xyz_clause_6769, 0 parameters
        %xyz_6769_pointer_492 = getelementptr {%reference, %reference, %neg, %pos, i64}, %environment %environment_489, i64 0, i32 2
        %xyz_6769 = load %neg, ptr % xyz_6769_pointer_492, !noalias !2
```
so the same value is defined under the same name in multiple different functions.

PR #1150, opened 2025-10-02
Work around bad codegen of negative types

marvinborner:
Related: issue #801. Eventually we should do these kinds of transformations automatically, in the future we should do something smarter.


phischu:
The source of the problem is that recursive object definitions (with `new`) are not supported by the llvm backend. It is possible to work around this, and also to automate this workaround.

PR #1154, opened 2025-10-21
Fix tail resumptive analysis for vars

### LINKED ISSUE #1153: Incorrectly applies tail resumption optimization under var scope
Effekt's optimizations are currently not semantics-preserving for the following program:

<details>

```scala
effect flip(): Bool

def all{prog: => Unit / {flip}} = {
  try {
    prog()
  } with flip {
    resume(true)
    resume(false)
  }
}

interface State[T] {
  def get(): T
  def set(x: T): Unit
}

interface allocate {
  def fresh[T, R](init: T){k: {State[T]} => R}: R
}

def newRegion[T]{prog: {r: allocate} => T}: T = {
  try {
    prog{r}
  } with r: allocate {
    def fresh[T, R](init) = {
      var x = init
      def s = new State[T] {
        def get() = x
        def set(y) = {
          x = y
        }
      }
      resume{{k} => k{s}}
    }
  }
}

def main() = {
  newRegion { {r} =>
    all {
      with def x = r.fresh(0)
      if (do flip()) {
        x.set(3)
      } else {
      }
      println(x.get())
    }
  }
}
```

</details >

with optimizations:

```console
> effekt regions.effekt
3
0
```

without optimizations:

```console
> effekt regions.effekt --no-optimize
3
3
```

timsueberkrueb:
Fixes #1153.


effekt/shared/src/main/scala/effekt/core/optimizer/RemoveTailResumptions.scala
```diff
@@ -35,7 +35,9 @@ object RemoveTailResumptions {
       } && default.forall { stmt => tailResumptive(k, stmt) }
       case Stmt.Region(BlockLit(tparams, cparams, vparams, bparams, body)) => tailResumptive(k, body)
```
b-studios: Shouldn't this also be false then?

b-studios:
Nice that you apparently didn't have to change the VMTests!

b-studios (approved):
I pre-approve :) Once you fixed / commented out the Scheme tests this is ready to go.

effekt/shared/src/main/scala/effekt/core/optimizer/RemoveTailResumptions.scala
```diff
@@ -35,7 +35,9 @@ object RemoveTailResumptions {
...
       case Stmt.Alloc(id, init, region, body) => tailResumptive(k, body) && !freeInExpr(init)
-      case Stmt.Var(ref, init, capture, body) => tailResumptive(k, body) && !freeInExpr(init)
+      // Conceptually, a mutable variable definition can be seen as a handler for get and put operations.
+      // Treating this as tail-resumptive leads to a failure of semantics preservation
+      case Stmt.Var(ref, init, capture, body) => false
```
jiribenes, line 41: Not too relevant for the current PR, but would a use-mention distinction help here?
If we have variables/prompts that are merely mentioned, can we still consider the function to be tail-res?

As a very trivial example, is it safe to do something similar to the example, but not use `x` at all after declaring it? 🤔

effekt/shared/src/main/scala/effekt/core/optimizer/RemoveTailResumptions.scala
jiribenes: If/when we change this, could we add an example where this fires / doesn't fire?

effekt/shared/src/main/scala/effekt/core/optimizer/RemoveTailResumptions.scala
timsueberkrueb: Is there an example that isn't just dead code (and would be handled by dead code elimination)?

batencle (approved):
LGTM

effekt/shared/src/main/scala/effekt/core/optimizer/RemoveTailResumptions.scala
jiribenes: see #1453: use-mention distinction helps a lot :)

PR #1159, opened 2025-10-23
Soften 'functions are second-class' statement in language tour

dvdvgt:
Feel free to also add other grievances you have with the language tour. Then we can just bundle them all in one PR.

### DIFF
--- a/examples/tour/captures.effekt.md
+++ b/examples/tour/captures.effekt.md
@@ -48,7 +48,7 @@ This is called a boxed computation. A boxed computation may only be used after u
 
 ## Boxing
 
-Effekt distinguishes between _values_ (that can be returned and stored) and _computation_ (that can be passed, but not returned). Computation can (invisibly) close over arbitrary capabilities, while values make this capture explicit in their type. In the below example we use the term-level construct `box`, to explicitly convert a computation to a value.
+Effekt distinguishes between _values_ (that can be returned and stored) and _computation_ (that can be passed, but not returned). Computation can (invisibly) close over arbitrary capabilities, while values make this capture explicit in their type. In the example below we use the term-level construct `box`, to explicitly convert a computation to a value.
 
 ```
 def example() =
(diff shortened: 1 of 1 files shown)


dvdvgt:
> We might need to update the parts that mention unboxing, since there is no syntax anymore

examples/tour/captures.effekt.md
```diff
@@ -5,8 +5,10 @@ permalink: tour/captures
 
 # Captures
 
-Functions are second-class in Effekt, but for a good reason. Functions closing over capabilities must not be returned, otherwise it can not be guaranteed that these capabilities are still in scope.
-Thus, we have to keep track of these captures and ensure they are in scope upon invoking.
+[Computations](/tour/computation), such as functions and objects, are second-class _by default_ in Effekt, but for a good reason.
```
jiribenes, line 8: Does the "but for a good reason" add anything? 🤔

kyay10:
> > We might need to update the parts that mention unboxing, since there is no syntax anymore

Note that #1172 added back the explicit syntax.

examples/tour/captures.effekt.md
b-studios: I think it is fine, since it leads over to the next sentence.

examples/tour/captures.effekt.md
```diff
@@ -5,8 +5,10 @@ permalink: tour/captures
...
+Since computations can capture capabilities, they can only be used where those capabilities remain in scope.
+For statically preventing capabilities from escaping their valid scope, Effekt tracks captured capabilities on the type level and verifies they're still in scope upon usage.
+We call this mechanism _boxing_ and will discuss in greater details in the next [section](/tour/captures#boxing).
```
b-studios: This is a bit misleading IMO. I would rather say something like

```suggestion
Functions and capabilities can still be treated as first-class values by means of _boxing_, which we discuss in greater details in the next [section](/tour/captures#boxing).
```

PR #1160, opened 2025-10-24
Chez Scheme Backend using the CPS IR

martin-ilgner:
Uses the CPS IR like the javascript backend does. But, Chez Scheme supports tailcall optimization, so this should be faster.

### DIFF
--- a/libraries/chez/cps/effekt.ss
+++ b/libraries/chez/cps/effekt.ss
@@ -75,10 +75,10 @@
     (let ([store (meta-cont-store ks)])
         (make-ref store init (store-generation store))))
 
-; Box -> Value
+; Ref -> Value
 (define (get ref) (ref-value ref))
 
-; Box, Value -> void
+; Ref, Value -> void
 (define (put ref value)
     (let* ([rGen (ref-generation ref)]
            [store (ref-store ref)]
@@ -96,7 +96,7 @@
 ; MetaCont -> MetaCont
 (define (create-region ks) ks)
 
-; Value, MetaCont -> Box
+; Value, MetaCont -> Ref
 (define allocate var)
 
 ; Ref | MetaCont -> void
--- a/effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/ChezSchemeTests.scala
@@ -101,6 +101,7 @@ class ChezSchemeCPSTests extends EffektTests {
     examplesDir / "benchmarks" / "input_output" / "interleave_promises.effekt",
     examplesDir / "benchmarks" / "input_output" / "financial_format.effekt",
     examplesDir / "benchmarks" / "input_output" / "server_client.effekt",
+    examplesDir / "pos" / "issue1087.effekt",
 
     // the test is specific to JS
     examplesDir / "pos" / "multiline_extern_definition.effekt",


libraries/chez/cps/effekt.ss
```diff
@@ -0,0 +1,135 @@
...
+(define (return x ks)
+    (let* ([stacks (meta-cont-stacks ks)]
+           [current-stack-cell (car stacks)]
+           [other-stack-cells (cdr stacks)]
+           [k (stack-cell-cont current-stack-cell)]
+           [ks (make-meta-cont (stack-cell-prompt current-stack-cell) 
+                               (stack-cell-state current-stack-cell)
+                               (stack-cell-backup current-stack-cell)
+                               other-stack-cells)])
+    (k x ks)))
```
b-studios: Let's see how the benchmark results are, but in JS we specifically pass `k` and `ks` separately to avoid the lookups like these here.

libraries/chez/cps/effekt.ss
```diff
@@ -0,0 +1,135 @@
+; Cont a = (a, MetaCont -> #)
+; State = [box]
+; Backup = [(box value)]
+; StackCell = (Prompt State Backup Cont)
+; MetaCont = (Prompt State Backup [StackCell])
```
b-studios: Why do the these have backup fields? Isn't this only needed on captured continuations?

libraries/chez/cps/effekt.ss
```diff
@@ -0,0 +1,135 @@
...
+           [ks (make-meta-cont prompt '() '() (cons k-cell rest-cells))])
+        (prog prompt ks return)))
+
+; StackCell -> StackCell
+(define (create-backup cell)
+    (let* ([state (stack-cell-state cell)]
+           [construct-cell (lambda (r) (make-backup-cell r (unbox r)))]
+           [backup (map construct-cell state)])
+        (stack-cell-backup-set! cell backup)
+        cell))
```
b-studios: Ah, I see you use the same type for live and captured stack segments

libraries/chez/cps/effekt.ss
martin-ilgner: Yes, Philipp also mentioned this. Currently fixing that.

libraries/chez/cps/effekt.ss
martin-ilgner:
```js
const RETURN = (x, ks) => ks.rest.stack(x, ks.rest)
```
I think the difference comes from me using cons-cells instead of saving the rest in the `meta-cont`-record. Maybe I should change that.

I ran the benchmarks from the examples folder (some fail for chez because file I/O is not implemented) and in most of the benchmarks that work, this implementation is already faster. The cases where chez is slower maybe arise from the fact that I haven't yet implemented the state with the snapshot data structure.

b-studios: Cool, did you also compare with the other Chez backends?

martin-ilgner: Current version:
Compared to chez-callcc the average speedup is ~1.27
Compared to chez-monadic the average speedup is ~1.80

(Compared to js its ~3.10. On some benchmarks the speedup here is above 8, the three worst are 0.31, 0.77 and 0.999

jiribenes:
Hey, once you rebase, you'll likely need to add your backend tests here:
https://github.com/effekt-lang/effekt/blob/b2e4c779bb7ccda6f92a3b329bc80ffa443158af/build.sbt#L220-L225

and here:
https://github.com/effekt-lang/effekt/blob/b2e4c779bb7ccda6f92a3b329bc80ffa443158af/build.sbt#L238-L249

so that your tests run correctly in the Chez group in CI. :)

martin-ilgner:
Preliminary results for current version:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/marti/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/marti/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">


benchmark | chez-callcc | chez-cps | chez-monadic | js | llvm
-- | -- | -- | -- | -- | --
bounce 100000 | 1,795s (1,200) | 2,154s (1,000) | 2,720s (0,792) | 1,401s (1,537) | 0,414s (5,204)
list_tail   100 | 1,226s (1,010) | 1,238s (1,000) | 2,216s (0,559) | 2,318s (0,534) | 0,853s (1,451)
mandelbrot 1000 | 2,603s (1,042) | 2,713s (1,000) | 2,642s (1,027) | 2,083s (1,302) | 0,341s (7,949)
nbody   50000 | 0,293s (0,992) | 0,290s (1,000) | 0,285s (1,020) | 0,105s (2,777) | 0,041s (7,027)
permute 10 | 3,939s (1,676) | 6,602s (1,000) | 9,681s (0,682) | 5,103s (1,294) | 1,485s (4,445)
queens   20 | 0,045s (0,764) | 0,034s (1,000) | 0,132s (0,258) | 0,320s (0,107) | 0,128s (0,267)
sieve 10000000 | 0,476s (1,176) | 0,559s (1,000) | 0,657s (0,851) | 0,490s (1,142) | 0,696s (0,804)
storage   5000 | 1,356s (0,934) | 1,267s (1,000) | 1,915s (0,662) | 2,235s (0,567) | 2,180s (0,581)
towers 25 | 0,635s (1,083) | 0,688s (1,000) | 1,391s (0,494) | 2,489s (0,276) | 1,329s (0,517)
erase_unused   10000 | 0,143s (1,020) | 0,146s (1,000) | 0,145s (1,005) | 0,379s (0,385) | 1,241s (0,118)
factorial_accumulator 500000000 | 8,725s (1,005) | 8,769s (1,000) | 8,683s (1,010) | 10,119s (0,867) | 0,000s   (87691180,730)
fibonacci_recursive   42 | 3,063s (1,276) | 3,910s (1,000) | 20,773s (0,188) | 29,466s (0,133) | 2,845s (1,374)
iterate_increment 500000000 | 0,844s (0,987) | 0,833s (1,000) | 0,835s (0,997) | 0,961s (0,867) | 0,000s   (8330526,620)
lookup_tree   10000000 | 1,020s (1,477) | 1,506s (1,000) | 1,899s (0,793) | 7,323s (0,206) | 0,870s (1,732)
match_options 10000000 | 0,269s (2,180) | 0,586s (1,000) | 0,764s (0,767) | 4,025s (0,146) | 0,302s (1,939)
sum_range   5000000 | 0,866s (1,152) | 0,997s (1,000) | 1,901s (0,525) | 4,865s (0,205) | 0,479s (2,083)
countdown 100000000 | 0,194s (1,009) | 0,196s (1,000) | 0,194s (1,012) | 0,571s (0,343) | 0,000s   (979812,405)
iterator   100000000 | 0,275s (0,975) | 0,268s (1,000) | 0,276s (0,971) | 0,293s (0,916) | 0,016s (16,503)
nqueens 10 | 0,088s (0,647) | 0,057s (1,000) | 0,077s (0,742) | 0,093s (0,616) | 0,021s (2,741)
parsing_dollars   10000 | 0,194s (1,000) | 0,194s (1,000) | 0,194s (1,000) | 0,185s (1,049) | 0,081s (2,408)
product_early 100000 | 0,419s (0,997) | 0,418s (1,000) | 1,470s (0,284) | 1,769s (0,236) | 0,478s (0,873)
resume_nontail   100000 | 23,518s (0,329) | 7,736s (1,000) | 14,101s (0,549) | error | 1,867s (4,145)
tree_explore 14 | 0,124s (0,698) | 0,086s (1,000) | 0,148s (0,584) | 0,317s (0,272) | 0,061s (1,427)
triples   100 | 0,046s (0,402) | 0,018s (1,000) | 0,030s (0,609) | 0,082s (0,225) | 0,008s (2,245)
  |   |   |   |   |  
average   speedup relative to chez-cps | 1,042965488 | 1 | 0,724248553 | 0,695665919 | 4041732,733
median speedup relative to   chez-cps | 1,007019524 | 1 | 0,754463951 | 0,534236882 | 2,164395151



</body>

</html>


The numbers inside the parentheses is the speedup of this benchmark on the specific backend compared to chez-cps. So for example, js is 1.537 times faster than chez-cps on the bounce benchmark.

Speedup = chez-cps time / other-backend time

So 1 means same speed as chez-cps, <1 means slower than chez-cps and >1 means faster than chez-cps

Edit: Note that I did not compare the benchmarks using files, because file I/O is not implemented for chez

martin-ilgner:
Also, this was a one-run measurement, which is pretty stupid now that I think about it

effekt/shared/src/main/scala/effekt/generator/chez/ChezSchemeCPS.scala
```diff
@@ -47,8 +47,10 @@ class ChezSchemeCPS extends Compiler[String] {
   lazy val CPSTransformed = Optimized map {
     case (mainSymbol, mainFile, core) =>
       val cpsTransformed = effekt.cps.Transformer.transform(core)
-      val contified = cps.Contify.rewrite(cpsTransformed)
-      (mainSymbol, mainFile, core, contified)
+      // (11.11.25) Makes the code actually slower
+      // can be turned back on when improved
+      //val contified = cps.Contify.rewrite(cpsTransformed)
+      (mainSymbol, mainFile, core, cpsTransformed)
```
b-studios, line 50: It is not too surprising, since Chez already will optimize code in CPS.

This phase was mostly for JS which is really bad at CPS :)

martin-ilgner: Ok thanks, then I will just remove it

martin-ilgner:
@b-studios This is what made me think that region deallocation is not really supported:
https://github.com/effekt-lang/effekt/blob/100d19aa8631e4591abf87ee661df305857496fb/libraries/js/effekt_runtime.js#L37-L38

martin-ilgner:
I tried checking for memory leaks with this program:

```scala
effect noop(): Unit

def main() = {
    val allocationsPerRegion = 100
    println("Working: " ++ show(allocationsPerRegion))
    repeat(1000){
        repeat(1000){
            try {
                region r {
                    repeat(allocationsPerRegion){
                        var x in r = 42
                        x = x + 1
                        //trick optimizer
                        if (x == 2) {
                            do noop()
                        }
                    }
                }
            } with noop { 
                resume(())
            }
        }
    }
}
```

I checked for chez-cps by first compiling the effekt code, then `echo '(compile-file "out/memoryLeak.ss")' | scheme` to create a scheme executable and then measuring heap + stacks with `valgrind --tool=massif --stacks=yes scheme --program out/memoryLeak.so`.
But it turns out that even changing the allocationsPerRegion from 100 to 1000 does not increase memory usage, probably because `x` gets collected when leaving the innermost repeat. I just gave up for now trying to search for memory leaks and just assume there are none.

b-studios:
This is not exactly, what I meant. The `noop` introduces a fresh stack and thus also a fresh region. Could you try modifying it a bit?

```
effect noop(): Unit

def allocateHere { p: => Unit / noop } =
  try p() with noop { resume(()) }

def main() = {
  val allocationsPerRegion = 100
  println("Working: " ++ show(allocationsPerRegion))


  with allocateHere; // <- we will allocate into this region, even though we shouldn't
  repeat(1000) {
    repeat(1000) {
      region r {
        repeat(allocationsPerRegion) {
          var x in r = 42
          x = x + 1
          if (x == 2) { do noop() } // trick optimizer
        }
      }
    }
  }
  println("Done.")
}
```

Try moving the `allocateHere` slowly over the repeat calls (until it hits the `region r`). Make sure to look at the generated code to check whether there is something different besides the position of the delimiter.

I would assume that moving `allocateHere` closer to the `region` will improve memory usage.

b-studios:
Looking at the _optimized_ version of the above example in JS reveals that the `noop` effect AND the handler are removed completely. 

```
function main_0(ks_5, k_3) {
  const allocationsPerRegion_0 = 100;
  const v_r_0 = $effekt.println(("Working: " + ('' + allocationsPerRegion_0)));
  function loop_0(i_0, ks_3, k_2) {
    if ((i_0 < (1000))) {
      function loop_1(i_1, ks_0, k_1) {
        if ((i_1 < (1000))) {
          const r_0 = ks_0.arena.newRegion();
          function loop_2(i_2, ks_1, k_0) {
            loop_3: while (true) {
              if ((i_2 < allocationsPerRegion_0)) {
                const x_0 = r_0.fresh(42);
                const s_0 = x_0.value;
                x_0.set((s_0 + (1)));
                const s_1 = x_0.value;
                if (s_1 === (2)) {
                  /* prepare call */
                  const tmp_i_0 = i_2;
                  i_2 = (tmp_i_0 + (1));
                  continue loop_3;
                } else {
                  /* prepare call */
                  const tmp_i_1 = i_2;
                  i_2 = (tmp_i_1 + (1));
                  continue loop_3;
                }
              } else {
                return () => k_0($effekt.unit, ks_1);
              }
            }
          }
          return loop_2(0, ks_0, (tmp_0, ks_2) =>
            loop_1((i_1 + (1)), ks_2, k_1));
        } else {
          return () => k_1($effekt.unit, ks_0);
        }
      }
      return loop_1(0, ks_3, (_0, ks_4) => loop_0((i_0 + (1)), ks_4, k_2));
    } else {
      return () => k_2($effekt.unit, ks_3);
    }
  }
  return loop_0(0, ks_5, (_1, ks_6) => {
    const v_r_1 = $effekt.println("Done.");
    return () => k_3(v_r_1, ks_6);
  });
}
```
This seems great, however, now we allocate the region into `ks_0` which is the toplevel meta-continuation, potentially leaking memory!

Running with `--no-optimize` prevents this and should be fine for the purpose of verifying memory leaks.

<details>

Weirdly, moving `allocateHere` immediately above `region r`, the optimizer "constant folds" the example to 
```
function main_0(ks_0, k_0) {
  const v_r_0 = $effekt.println(("Working: " + ('' + (100))));
  const v_r_1 = $effekt.println("Done.");
  return () => k_0(v_r_1, ks_0);
}
```

Note, this constant folding can be prevented by summing up all results:

```
def main() = {
  val allocationsPerRegion = 100
  println("Working: " ++ show(allocationsPerRegion))

  var sum = 0

  repeat(1000) {
    repeat(1000) {
      with allocateHere; // <- we will allocate into this region, even though we shouldn't
      region r {
        repeat(allocationsPerRegion) {
          var x in r = 1
          x = x + 1
          sum = sum + x
          if (x == 1337) { do noop() } // trick optimizer
        }
      }
    }
  }
  println("Done.")
  println(sum)
}
```

</details>

martin-ilgner:
There is some LSP test failing which I don't know why. What does this test test? (nevermind, now it runs through)

b-studios:
Hey, did you try my modified examples above to check for memory leaks?

martin-ilgner:
No, I think it is out of scope for this PR/thesis, Philipp shared my opinion. I just brought it up because I thought there was maybe something that I overlooked that made this approach obviously memory-leak-safe

PR #1162, opened 2025-10-27
Add 'did you mean' help for mistaken type names

jiribenes:
Add a system to help the users when the compiler cannot find a name of a type:

#### Heuristic 1: Some types are named differently in Effekt than in other languages

Screenshot 2025-10-27 at 15 06 41

This specific one (`Boolean` instead of `Bool`) is done often by students and LLMs alike.

#### Heuristic 2: Try case-insensitive comparison

Screenshot 2025-10-27 at 15 06 26

#### Heuristic 3: Try full on edit distance!

Screenshot 2025-10-27 at 22 31 43

### DIFF
--- a/effekt/shared/src/main/scala/effekt/util/EditDistance.scala
+++ b/effekt/shared/src/main/scala/effekt/util/EditDistance.scala
@@ -22,21 +22,25 @@ def editDistance(a: String, b: String, limit: Int): Option[Int] = {
   val s1 = a.substring(prefixLen, n - suffixLen)
   val s2 = b.substring(prefixLen, m - suffixLen)
 
-  // Levenshtein
-  val N = s1.length
-  val M = s2.length
-
   // After stripping, if one is empty, distance is just the other's length
-  if (s1.isEmpty) return Some(M)
-  if (s2.isEmpty) return Some(N)
+  if (s1.isEmpty) return Some(s2.length)
+  if (s2.isEmpty) return Some(s1.length)
 
-  val d = Array.ofDim[Int](N + 1, M + 1)
-  for (i <- 0 to N) { d(i)(0) = i }
-  for (j <- 0 to M) { d(0)(j) = j }
+  val distance = levenshtein(s1, s2)
+  if (distance <= limit) Some(distance) else None
+}
+
+inline def levenshtein(s1: String, s2: String): Int = {
+  val n = s1.length
+  val m = s2.length
+
+  val d = Array.ofDim[Int](n + 1, m + 1)
+  for (i <- 0 to n) { d(i)(0) = i }
+  for (j <- 0 to m) { d(0)(j) = j }
 
   for {
-    i <- 1 to N; s1_i = s1(i - 1)
-    j <- 1 to M; s2_j = s2(j - 1)
+    i <- 1 to n; s1_i = s1(i - 1)
+    j <- 1 to m; s2_j = s2(j - 1)
   } do {
     val costDelete     = d(i - 1)(j) + 1
     val costInsert     = d(i)(j - 1) + 1
@@ -50,6 +54,5 @@ def editDistance(a: String, b: String, limit: Int): Option[Int] = {
     }
   }
 
-  val distance = d(N)(M)
-  if (distance <= limit) Some(distance) else None
+  d(n)(m)
 }


effekt/shared/src/main/scala/effekt/symbols/Scope.scala
```diff
@@ -208,8 +208,61 @@ object scopes {
...
+      case "UInt32" | "UInt64" | "UInt" | "U32" | "U64" | "u32" | "u64" =>
+        Some(pp"Effekt only supports signed integers, did you mean `Int`?")
+      case "Int8" | "UInt8" | "I8" | "U8" | "i8" | "u8" =>
+        Some(pp"Did you mean `Byte` (8 bits) or `Char` (32 bits)?")
+      case "Float" | "float" | "F64" | "F32" | "f64" | "f32" =>
+        Some(pp"Did you mean `Double`?")
+      case "Maybe" | "maybe" => Some(pp"Did you mean `Option`?")
+      case "Void" | "void" => Some(pp"Did you mean `Unit`?")
+      case _ => None
+    }
```
jiribenes, line 252: Are there any other ones that you see people struggling with in a new language? 🤔 
Are any of them too easy to figure out or flat out useless?

The main mistakes I've seen were: `Boolean`, `bool`, `i32/64`, `Integer`, `UInt`, `u32/64`, `Float` (or some variant thereof), `Maybe` and `Void`.

effekt/shared/src/main/scala/effekt/symbols/Scope.scala
```diff
@@ -208,8 +208,61 @@ object scopes {
...
+
     def lookupType(id: IdRef)(using E: ErrorReporter): TypeSymbol =
-      lookupTypeOption(id.path, id.name) getOrElse { E.abort(pp"Could not resolve type ${id}") }
+      lookupTypeOption(id.path, id.name) getOrElse {
+        // If we got here, we could not find an exact match.
+        // But let's try and find a close one!
+        def availableTypes = all(id.path, scope) {
+          _.types.keys.toList
+        }.flatten.toSet
```
jiribenes, line 260: I'm very much not a fan of this, can we do better here somehow?

effekt/shared/src/main/scala/effekt/symbols/Scope.scala
```diff
@@ -208,8 +208,61 @@ object scopes {
...
+            candidates
+              .toSeq
+              .flatMap { candidate =>
+                effekt.util.editDistance(nameNotFound, candidate, threshold).map((_, candidate))
+              }
+              .sorted
+              .headOption
+              .map { case (_, name) => pp"Did you mean $name?" }
+          }
+      } foreach { msg => E.info(msg) }
```
jiribenes, line 235: I tried very hard to structure this sanely, but it's still not what I'd prefer, is there a better way?

effekt/shared/src/main/scala/effekt/util/EditDistance.scala
```diff
@@ -0,0 +1,55 @@
...
+  val N = s1.length
+  val M = s2.length
+
+  // After stripping, if one is empty, distance is just the other's length
+  if (s1.isEmpty) return Some(M)
+  if (s2.isEmpty) return Some(N)
+
+  val d = Array.ofDim[Int](N + 1, M + 1)
+  for (i <- 0 to N) { d(i)(0) = i }
+  for (j <- 0 to M) { d(0)(j) = j }
```
jiribenes: If this performs too badly, we can always use two or three rows of the array only... (it just makes it harder to read)

effekt/shared/src/main/scala/effekt/symbols/Scope.scala
b-studios: I think it is not too bad. It is modular enough to also use it for terms. Well done!

Of course the general concern of better-error-messages-complicate-compiler still applies. At least you managed to keep the actual change in `lookupType` to just a few lines.

effekt/shared/src/main/scala/effekt/symbols/Scope.scala
b-studios: What would you expect? We somehow need to gather all types from all scopes. The only thing that looks a bit weird at first to me is the `.keys`, but that makes complete sense.

examples/neg/boolean_wrong_name.effekt
```diff
@@ -0,0 +1,3 @@
+module boolean_wrong_name
+
+def alwaysTrue(): Boolean = true
```
b-studios, line 3: Once we have it for terms, we can add a hint `True` ~> `true` for @BinderDavid ;)

And `S(S(Z()))` ~> "Did you mean `2`?"

effekt/shared/src/main/scala/effekt/symbols/Scope.scala
marzipankaiser: All the ones I could come up with from the top of my head (that aren't way obscure) are already covered by case-insensitive compare (e.g. `int` and `string`, ...)

effekt/shared/src/main/scala/effekt/symbols/Scope.scala
marzipankaiser: [prob. overengineering, but since you asked:]
Maybe there'd be a sufficiently common pattern between the two others that could be extracted into a `foo(candidates: ...)(comp: (String,String) => ?) => Option[String]` function?
Also depending on how much we care about efficiency here.

jiribenes:
Thanks for the reviews!

PR #1163, opened 2025-10-27
Add an LLVM implementation of 'random' to standard library

phischu:


jiribenes:
fwiw, we also have a `random` module with a PRNG written in Effekt (just not a great one)
(from #966)

Issue #1165, opened 2025-10-28
Unhelpful error message where box would have been inferred


jiribenes:
_Originally reported by @Plixo2_

As of #1090, we no longer infer boxes.

The following program reports a badly formatted position-less error:
```scala
def toString(): String = ""
def main() = {
  println(toString)
}
```

The error in question reported at the last character in the file is:
```
Computation toString_5890 is used in an expression position, which requires boxing (e.g. `box toString_5890`)
```

coming from unbox inference:
https://github.com/effekt-lang/effekt/blob/664897873ec95088b0e8d51a2895208143aef1e9/effekt/shared/src/main/scala/effekt/typer/UnboxInference.scala#L56

---

The error message should:
1. have better position information
2. use user-given names instead of internal ones

jiribenes:
For some expressions, the position information seems usable, though both the message and the position are still improvable:
```scala
effect exc(): Nothing
def foo() = try { 
  //             ^ error: 
  //       Computation e_5648 is used in an expression position, which requires boxing (e.g. `box e_5648`)
  e
} with e: exc { <> }
```

PR #1168, opened 2025-10-29
Rephrase UnboxInference errors using a 'expected X, got Y' scheme

jiribenes:
Follow-up from #1167, trying to tweak the phrasing of the error messages coming from UnboxInference.

Here's examples from #1165:
Screenshot 2025-10-29 at 21 25 10
Screenshot 2025-10-29 at 21 25 33

I tried to utilise the 'expected X, got Y; try Z' scheme, which isn't the best, but it is at least a consistent framework. The jargon has been cut down as we no longer need to mention "expression position".

I also added a heuristic based on a misunderstanding I've seen floating around: "as a rule, in Effekt, I can skip `()` in a function/method/... call". This does not work for simple function calls since, to us, this looks like a missing `box`.
Therefore we have a more specific message for such functions/block params/..., but we try to offer the "did you mean" only in cases where this misconception would make sense, **not** in the general case of _all_ functions/block params/... (which might be a bit limiting, but I think works well to prevent spurious messages)

### DIFF
--- a/examples/neg/must_box_block_param.check
+++ b/examples/neg/must_box_block_param.check
@@ -1,3 +1,3 @@
-[error] examples/neg/must_box_block_param.effekt:3:33: Expected a value, but block is a computation. Use `box block` to pass it as a value instead
+[error] examples/neg/must_box_block_param.effekt:3:36: Expected a value, but block is a computation. Use `box block` to pass it as a value instead
 def foo { block: (Int) => Unit } = block
                                    ^^^^^


effekt/shared/src/main/scala/effekt/typer/UnboxInference.scala
```diff
@@ -53,16 +53,16 @@ object UnboxInference extends Phase[NameResolved, NameResolved] {
...
     }
 
     case n: New =>
-      C.error(pp"Creating an instance in an expression requires boxing (e.g. `box new ${n.impl.id}[...] { ... }`)")
+      C.error(pp"Expected a value, but `new` creates a computation (an object instance). Use `box new ${n.impl.id}[...] { ... }` to pass it as a value instead")
       rewriteAsBlock(n)
 
     case b: BlockLiteral =>
-      C.error(pp"Function literals in expression position require boxing (e.g. `box { (${b.vparams.map(_.id).mkString(", ")}) => ... `)")
+      C.error(pp"Expected a value, but block (function) literals are computations. Use `box { (${b.vparams.map(_.id).mkString(", ")}) => ... }` to pass it as a value instead")
```
jiribenes, line 75: How can I write some surface code that hits this branch? 🤔

b-studios: Not sure that's possible since only `box` and blockarguments use `functionArg`:

https://github.com/effekt-lang/effekt/blob/e9d4ab999f8948c77b98f6f65f1b4129b6fa8150/effekt/shared/src/main/scala/effekt/Parser.scala#L1136C22-L1136C34

PR #1169, opened 2025-10-29
Mark unlikely calls in LLVM as cold

jiribenes:


jiribenes:
@phischu could you check if you want this / if it makes any difference for the LLVM benchmarks? if so, feel free to merge it :)

libraries/llvm/forward-declare-c.ll
```diff
@@ -6,8 +6,8 @@ declare %Pos @c_get_arg(i64)
 declare void @c_io_println(%Pos)
 declare %Pos @c_io_readln()
 
-declare void @hole(i8*)
-declare void @duplicated_prompt()
+declare void @hole(i8*) cold
+declare void @duplicated_prompt() cold
```
jiribenes, line 10: My hypothesis is that the `hole` (`panic: not implemented yet`) being `cold` doesn't matter at all, but the `duplicated_prompt` might by a tiny bit since I assuming used stacks are non-null seems good...

PR #1148, opened 2025-10-02
Feat byte hex notation

PhictionalOne:
Adds support for a Bytes in accordance to #814

PR #1170, opened 2025-10-30
Add hexadecimal literals for 'Byte'

### LINKED ISSUE #814: Add hexadecimal literals for `Byte`
Format: `'0x' <hexa> <hexa>` where `<hexa> ::= 0 | 1 | ... | a | A | b | B | ... | f | F`.
Note that it's only for bytes, i.e. values between `0x00` (0) and `0xFF` (255) that have the `Byte` type, everything else should fail.
Specifically, this notation is _not_ for integers of type `Int`, it's just for us being able to write `0xEF` directly instead of relying on `239.toByte`.

jiribenes:
Originally by @PhictionalOne, started in #1148 (was on a repo fork).

Resolves #814 by adding hexadecimal literals for bytes: `0xAA`, `0x00`, `0xFF`, `0xA0`, etc, exclusively of type `Byte`.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/vm/Builtin.scala
+++ b/effekt/shared/src/main/scala/effekt/core/vm/Builtin.scala
@@ -185,12 +185,6 @@ lazy val integers: Builtins = Map(
   },
 )
 
-lazy val bytes: Builtins = Map(
-  builtin("effekt::show(Byte)") {
-    case As.Byte(n) :: Nil => Value.String(n.toString)
-  },
-)
-
 lazy val booleans: Builtins = Map(
   builtin("effekt::not(Bool)") {
     case As.Bool(x) :: Nil => Value.Bool(!x)


libraries/common/bytearray.effekt
```diff
@@ -133,6 +133,40 @@ extern def toString(arr: ByteArray) at {}: String =
   chez "(utf8->string ${arr})"
   vm "bytearray::toString(ByteArray)"
 
+def toHexCharBytes(b: Byte): Unit / emit[Byte] = {
+  val unsigned: Int = b.toInt.bitwiseAnd(0xFF.toInt)
+
+  val high = unsigned.bitwiseShr(4).bitwiseAnd(0xFF.toInt)
+  val low = unsigned.bitwiseAnd(0x0F.toInt)
```
jiribenes: It would be nice to have bitwise operators on `Byte`s

libraries/common/bytearray.effekt
```diff
@@ -133,6 +133,40 @@ extern def toString(arr: ByteArray) at {}: String =
...
+
+/// Converts a bytearray like `bytearray.[0xAB, 0xFF]` into a bytearray `bytearray.['A', 'B', 'F', 'F']"`
+def toHexCharBytes(arr: ByteArray): ByteArray = collect {
+  for[Byte] {arr.each} { b => b.toHexCharBytes }
+}
+
+// NOTE: It seems like these cannot really be defined earlier... Oh well.
+
+def show(b: Byte): String = b.toHexCharBytes.toString
+def println(b: Byte): Unit = println(show(b))
```
jiribenes: Only tested locally, needs a proper test.

libraries/common/bytearray.effekt
```diff
@@ -133,6 +133,40 @@ extern def toString(arr: ByteArray) at {}: String =
...
+  def toHexDigit(n: Int): Byte =
+    if (n < 10) {
+      (n + 48)       // '0' = 48
+    } else {
+      (n - 10 + 65)  // 'A' = 65
+    }.toByte
+
+  do emit(high.toHexDigit)
+  do emit(low.toHexDigit)
+}
```
jiribenes: The `show` instance for `Byte`s is a bit weird since we're trying to write it in Effekt directly.
In order to do _anything_ at all, we need to either convert to characters or use bytearrays -- I've chosen to do the latter here, but as a consequence, the `println` and `show` for `Byte` are defined awkwardly in `bytearray`...

`show(b: Byte) := bytearray::collect { b.toHexCharBytes }`

libraries/common/effekt.effekt
```diff
@@ -109,15 +108,6 @@ extern def show(value: Char) at {}: String =
   """
   vm "effekt::show(Char)"
 
-extern def show(value: Byte) at {}: String =
-  js "'' + ${value}"
-  chez "(string ${value})"
-  llvm """
-    %z = call %Pos @c_bytearray_show_Byte(i8 ${value})
```
jiribenes, line 108: TODO: now we should be able to deprecate `c_bytearray_show_Byte` 😌

jiribenes: Maybe I'll revert the custom `show` instance defined in Effekt directly and provide them via FFI instead.
Then we don't have to change the `bytearray` module.

effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
```diff
@@ -131,6 +132,7 @@ class PrettyPrinter(printDetails: Boolean, printInternalIds: Boolean = true) ext
     case Literal((), _)         => "()"
     case Literal(n, Type.TInt)  => n.toString
     case Literal(n, Type.TChar) => s"'\\${n.toString}'"
+    case Literal(b: Byte, Type.TByte) => UByte.unsafeFromByte(b).toHexString
```
jiribenes, line 135: This might be a bit confusing, but we can't match on a `UByte` as it's opaque, so we have to match on a `Byte`, just to zero-cost convert it to a `Byte`, just to print the `hexString` consistently.
... I really don't know how to do this better without either wrapping it in a custom type and dealing with the overhead, or pulling a newtype library :/

libraries/common/bytearray.effekt
jiribenes: ... and equality and stuff :)

effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
phischu: It's ok. You could have `toHexString` operate directly on `Byte` and use `toUnsignedInt` everywhere.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -476,10 +478,12 @@ object Transformer {
...
       }
 
-    case core.Literal(value: Int, core.Type.TByte) =>
+    case core.Literal(value: Byte, core.Type.TByte) =>
       shift { k =>
         val unboxed = Variable(freshName("byte"), Type.Byte())
-        LiteralByte(unboxed, value, Coerce(variable, unboxed, k(variable)))
+        val b = UByte.unsafeFromByte(value)
+        // TODO: Literal bytes could now also be just byte-sized in Machine (= use 'UByte'; they are byte-sized in LLVM anyway).
+        LiteralByte(unboxed, b.toInt, Coerce(variable, unboxed, k(variable)))
```
phischu, line 470: I would prefer `toUnsignedInt` here instead of the newtype wrapping, but ok.

examples/benchmarks/input_output/dyck_one.effekt
```diff
@@ -12,13 +12,13 @@ def emitTree(n: Int): Unit / emit[Byte] =
...
+    do emit(0x28)
     repeat(n) {
-      do emit(40.toByte)
-      do emit(41.toByte)
+      do emit(0x28)
+      do emit(0x29)
     }
     emitTree(n - 1)
-    do emit(41.toByte)
+    do emit(0x29)
```
phischu, line 21: Beautiful!

libraries/llvm/bytearray.c
```diff
@@ -115,8 +117,8 @@ struct Pos c_bytearray_show_Char(const uint64_t n) {
 
 // TODO do this in Effekt
 struct Pos c_bytearray_show_Byte(const Byte n) {
-    char str[4];  // Byte values range from 0 to 255, 3 characters + null terminator
-    snprintf(str, sizeof(str), "%" PRIu8, n);
+    char str[5]; // |"0x" ++ <2 hex digits> ++ '\0'| = 5
+    snprintf(str, sizeof str, "0x%02" PRIX8, (uint8_t)n);
```
phischu, line 121: Shouldn't `n` already be of type `uint8_t`? We have `typedef uint8_t Byte;`. Or is this best practice in C?

phischu (approved):

libraries/llvm/bytearray.c
jiribenes: Could be both `int8_t` and `uint8_t` if the source of `inttypes.h` is correct. 🤔

phischu: Doesn't matter too much.

jiribenes:
One more thing that would sometimes be useful is byte character literals (without the need to convert them).
But we could also just build that into the normaliser as `when c is charLiteral, rewrite c.toByte to <lowest byte of c>`.

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
jiribenes: The biggest issue I have with `toUnsignedInt` is that you need to remember to use that whenever you see a `Byte`, which felt like a footgun (that I stumbled into like three times when making this PR in the first place).
The alternative is weird since we can't match on `UByte` as a type, so it's all quite unsatisfactory :/

PR #1172, opened 2025-10-30
Reintroduce explicit 'unbox' for didactic use

jiribenes:
Motivation: explicit unbox was removed in #1090, but it's actually pretty good for didactic reasons (no magic!).

Since essentially only the parser and lexer of `unbox` were removed, this PR reintroduces them.


effekt/shared/src/main/scala/effekt/core/Parser.scala
```diff
@@ -365,6 +366,7 @@ class CoreParsers(names: Names) extends EffektLexers {
   // ------
   lazy val block: P[Block] =
     ( blockVar
+    | `unbox` ~> expr ^^ Block.Unbox.apply
```
jiribenes, line 369: I believe that the core parser should to be able to parse `unbox`, shouldn't it?

PR #1173, opened 2025-10-30
Multiple CI jobs for PRs

jiribenes:
Trying to speed-up CI by having multiple jobs: total time goes from some 20-30 minutes down to 12!

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -47,7 +47,10 @@ jobs:
       - uses: ./.github/actions/restore-build-cache
 
       - name: Run internal tests
-        run: sbt effektJVM/testRemaining
+        run: |
+          sbt kiamaJVM/test
+          sbt effektJVM/testRemaining
+          sbt effektJS/test
 
       - name: Assemble fully optimized JS file
         run: sbt effektJS/fullOptJS
--- a/build.sbt
+++ b/build.sbt
@@ -17,7 +17,7 @@ lazy val bumpMinorVersion = taskKey[Unit]("Bumps the minor version number (used 
 lazy val testBackendJS = taskKey[Unit]("Run JavaScript backend tests")
 lazy val testBackendChez = taskKey[Unit]("Run Chez Scheme backend tests")
 lazy val testBackendLLVM = taskKey[Unit]("Run LLVM backend tests")
-lazy val testRemaining = taskKey[Unit]("Run all non-backend tests (internal tests)")
+lazy val testRemaining = taskKey[Unit]("Run all non-backend tests (internal tests) on effektJVM")
 
 lazy val noPublishSettings = Seq(
   publish := {},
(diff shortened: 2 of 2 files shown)


.github/workflows/ci-pr.yml
```diff
@@ -0,0 +1,132 @@
...
+          path: |
+            target
+            project/target
+            project/project/target
+            kiama/jvm/target
+            kiama/js/target
+            effekt/jvm/target
+            effekt/js/target
+            ~/.ivy2/cache
+            ~/.sbt
```
timsueberkrueb, line 32: Is it save to cache this? Is `sbt` smart enough to clean if needed?

.github/actions/restore-build-cache/action.yml
```diff
@@ -0,0 +1,22 @@
...
+          project/project/target
+          kiama/jvm/target
+          kiama/js/target
+          effekt/jvm/target
+          effekt/js/target
+          ~/.ivy2/cache
+          ~/.sbt
+          ~/.cache/coursier
+        key: effekt-build-${{ github.sha }}
+        fail-on-cache-miss: true
```
timsueberkrueb: This defaults to `false`, why did you decide to set it to `true`?

.github/workflows/ci-pr.yml
jiribenes: No idea tbh, the caching is only for running the the test suite here, which I would assume is safe. In my head, I'm making a snapshot of the built files and then reusing those to "just" call the test endpoints further down the line

.github/actions/restore-build-cache/action.yml
jiribenes, line 22: Great question: the `cache` here is a bit misleading, the whole point is to allow multi-machine builds by packaging up the build env after building and then restoring it just before calling `sbt testWhatever` in order to prevent recompiling

jiribenes: Therefore, if this cannot find the build env. with prebuilt class files, something went terribly wrong and we want to quit!

.github/workflows/ci-pr.yml
timsueberkrueb: Nvm, understood that it's only used within a single run and not intended to be shared between runs.

.github/workflows/ci-pr.yml
```diff
@@ -0,0 +1,132 @@
+name: CI
+
+on:
+  pull_request:
+
```
timsueberkrueb, line 5: One thing we could consider is adding a [concurrency group](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency) and cancel running jobs when a new commit is pushed:

```
concurrency:
  group: ci-pr-${{ github.workflow }}-${{ github.head_ref || github.ref }}
  cancel-in-progress: true
```

This saves CI minutes but has the disadvantage that some commits may not be tested and situations can arise where one has to wait longer for any CI feedback (in case of frequent pushes).

.github/workflows/ci-pr.yml
jiribenes: No, it's a good question if it was not immediately obvious 😅

PR #1179, opened 2025-11-04
Add InlayHint filtering on the language server

Plixo2:
Instead of filtering InlayHints on the Language Client, they should be filtered by the Language Server. 

For more details see https://github.com/effekt-lang/effekt-vscode/pull/155

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/LSPTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/LSPTests.scala
@@ -718,7 +718,6 @@ class LSPTests extends FunSuite {
       markup.setValue("captures: `{io}`")
       inlayHint.setTooltip(markup)
       inlayHint.setPaddingLeft(true)
-      inlayHint.setData("capture")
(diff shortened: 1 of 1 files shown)


kiama
Plixo2, line 1: I dont know if this submodule is the old one, or the new commit. 

[This has to be updated anyway](https://github.com/effekt-lang/kiama/pull/16).

timsueberkrueb: Just checked, the submodule has been updated to latest `main`.

Plixo2:
> CI is failing

Fixed it, CI 'should' work now. 
Tests did not work for me, since i am on Windows. Should i file a issue for this? A fix would be simple

marvinborner:
If it's 'simple' and doesn't impact other systems, please do. We appreciate any insight on Windows compatibility.

Plixo2:
Sorry for the commit and subsequent reset on my master, the workflow has to be approved again 🥲

PR #1181, opened 2025-11-05
Reach reparsability of core pretty-printer

timsueberkrueb:
Currently, it is very hard to write tests for core. There are two main reasons for this:

1. The core parser is currently very rudimentary and it is not easy to write core programs. This would be addressed by a new core frontend (#1152).
2. The core pretty-printer is currently unable to produce code that can be re-parsed with the core parser. I would like to address this with this PR.

While a new core frontend would subsume both of these problems, it is a lot more effort than adding a re-parsable mode to the current core pretty-printer. This would allow us to compile source programs to core and pretty-print them to use them as golden output in core tests.

In particular, when developing the new normalizer (#1127), I would like to track its behavior in unit tests, but they are currently very time-consuming to write due to the fact I cannot just use pretty-printed core for comparison.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -242,8 +242,8 @@ lazy val effekt: CrossProject = crossProject(JSPlatform, JVMPlatform).in(file("e
 
       val allTests = (Test / definedTestNames).value.toSet
 
-      // Explicit list of backend tests (union of all testBackend targets)
-      val backendTests = Set(
+      // Explicit list of tests run separately (union of all testBackend targets)
+      val separatedTests = Set(
         "effekt.JavaScriptTests",
         "effekt.StdlibJavaScriptTests",
         "effekt.ChezSchemeMonadicTests",
@@ -256,7 +256,7 @@ lazy val effekt: CrossProject = crossProject(JSPlatform, JVMPlatform).in(file("e
         "effekt.core.ReparseTests",
       )
 
-      val remaining = allTests -- backendTests
+      val remaining = allTests -- separatedTests
 
       if (remaining.isEmpty) {
         log.info("No remaining tests")


effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
```diff
@@ -74,50 +86,60 @@ object PrettyPrinter extends ParenPrettyPrinter {
...
   }
 
   def toDoc(p: ValueParam): Doc = toDoc(p.id) <> ":" <+> toDoc(p.tpe)
-  def toDoc(p: BlockParam): Doc = braces(toDoc(p.id))
+  def toDoc(p: BlockParam): Doc = braces(toDoc(p.id) <> ":" <+> toDoc(p.tpe))
 
   //def toDoc(n: Name): Doc = n.toString
 
-  def toDoc(s: symbols.Symbol): Doc = s.show
+  def toDoc(s: symbols.Symbol): Doc = s.name.name ++ "$" ++ s.id.toString
```
timsueberkrueb: `$` is a valid character in identifiers, so we need to find another symbol. How about `#`?

effekt/shared/src/main/scala/effekt/core/Parser.scala
```diff
@@ -238,21 +258,37 @@ class CoreParsers(names: Names) extends EffektLexers {
...
+        Extern.Def(
+          id, tparams, cparams, vparams, bparams, result, captures,
+          ExternBody.StringExternBody(ff, templ)
+        )
     })
 
+  lazy val externBodyTemplate: P[Template[Expr]] =
+    (
+      (multilineString | stringLiteral) ~
+        rep((`++` ~> expr) ~ (`++` ~> (multilineString | stringLiteral)))
```
timsueberkrueb, line 304: I chose the syntax `"string" ++ (<expr> ++ "string")*` for extern body template strings because it is easy to parse. Open to to other suggestions!

effekt/shared/src/main/scala/effekt/core/Parser.scala
```diff
@@ -30,17 +42,18 @@ class EffektLexers extends Parsers {
...
-    )
+      | failure("Expected an identifier")
+      )
 
   lazy val identRef =
     (not(anyKeyword) ~> qualifiedName
-    | failure("Expected an identifier")
-    )
+      | failure("Expected an identifier")
+      )
```
timsueberkrueb: Undo some of these spurious white space changes.

effekt/shared/src/main/scala/effekt/core/Parser.scala
```diff
@@ -472,9 +517,21 @@ class CoreParsers(names: Names) extends EffektLexers {
     `{` ~> (id <~ `:` | wildcard) ~ blockType <~ `}` ^^ { case id ~ tpe => id -> tpe }
 
   lazy val interfaceType: P[BlockType.Interface] =
-    ( id ~ maybeTypeArgs ^^ { case id ~ tpe => BlockType.Interface(id, tpe) : BlockType.Interface }
-    | failure("Expected an interface")
-    )
+    (
+      id ~ maybeTypeArgs ^^ {
+        case id ~ targs =>
+          // Special-case Resume[result, answer] to use the canonical ResumeSymbol
```
timsueberkrueb: This is also not very nice.

effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
jiribenes: TIL, I thought we didn't allow it in source exactly for this reason. Does anybody actually use it anywhere [in source]?

effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
```diff
@@ -269,7 +269,15 @@ object PrettyPrinter extends ParenPrettyPrinter {
...
-      "<>" <+> s"// @ ${span.range.from.format}"
+      val from = span.from
+      val to = span.to
+      val name = span.source.name
+      val scheme = span.source match {
+        case _: kiama.util.FileSource   => "file"
+        case _: kiama.util.StringSource => "string"
+        case _                          => "source"
+      }
+      "<>" <+> s"@ \"$scheme://$name\":$from:$to"
```
timsueberkrueb, line 280: TODO: proper escaping

effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
timsueberkrueb: No longer relevant, I've reverted these parts of the PR because it was breaking various Barendregt-related assumptions. I still think it's a good idea to improve the name handling, but it should be done separately from this PR.

effekt/jvm/src/test/scala/effekt/core/ReparseTests.scala
```diff
@@ -38,6 +38,10 @@ class ReparseTests extends CoreTests {
     File("examples/pos/bidirectional/selfrecursion.effekt"),
     // FIXME: Wrong number of type arguments
     File("examples/pos/type_omission_op.effekt"),
+    // FIXME: There is currently a limitation in TestRenamer in that it does not rename captures.
+    // This means, that in this example, captures for State[Int] and State[String] are both printed as "State",
+    // leading to a collapse of the capture set {State_1, State_2} to just {State}.
+    File("examples/pos/parametrized.effekt")
```
timsueberkrueb, line 44: @marzipankaiser this is what I just mentioned in person.

build.sbt
```diff
@@ -245,7 +252,8 @@ lazy val effekt: CrossProject = crossProject(JSPlatform, JVMPlatform).in(file("e
         "effekt.StdlibChezSchemeCallCCTests",
         "effekt.LLVMTests",
         "effekt.LLVMNoValgrindTests",
-        "effekt.StdlibLLVMTests"
+        "effekt.StdlibLLVMTests",
+        "effekt.core.ReparseTests",
```
jiribenes, line 256: Nitpick: `backendTests` probably no longer applies here 😅

effekt/jvm/src/test/scala/effekt/core/TestRenamer.scala
```diff
@@ -14,19 +14,32 @@ import effekt.context.Context
  *
  * @param C the context is used to copy annotations from old symbols to fresh symbols
  */
-class TestRenamer(names: Names = Names(Map.empty), prefix: String = "") extends core.Tree.Rewrite {
+class TestRenamer(names: Names = Names(Map.empty), prefix: String = "_") extends core.Tree.Rewrite {
```
jiribenes, line 17: Just to check, do we still need both Renamer and TestRenamer as of this PR?

effekt/shared/src/main/scala/effekt/core/Parser.scala
```diff
@@ -291,12 +359,26 @@ class CoreParsers(names: Names) extends EffektLexers {
...
+      case (name ~ from ~ to) =>
+        val source = if (name.startsWith("file://")) {
+          kiama.util.FileSource(name.stripPrefix("file://"))
+        } else if (name.startsWith("string://")) {
+          kiama.util.StringSource("", name.stripPrefix("string://"))
+        } else {
+          sys error s"Unsupported source scheme in hole source name: $name"
+        }
+        Hole(effekt.source.Span(source, from.toInt, to.toInt))
+      }
```
jiribenes, line 381: One alternative here would be already the pre-formatted position as a string, i.e. `<> @ "..."`. We always just show the start pos (see backends), it could technically be a string. The reason why I didn't do that in the first place is if we have a debugger / source maps or something, we might want actual ranges/spans.

timsueberkrueb: Alternatively, we could replace `core.Hole` with an instruction `core.Panic` that aborts the program and prints a message.

jiribenes: This would alleviate the current issue, yes, but I was hoping we could later use `core.Hole` to partially evaluate programs ala Hazel / for teaching purposes :(

(edit: although that is the far future, so short term your change would definitely work)

effekt/jvm/src/test/scala/effekt/core/TestRenamer.scala
timsueberkrueb: It may be possible to merge them, but this requires some more work, so it should be done separately.

PR #1192, opened 2025-11-07
Experiment: namespaced fields / accessors

jiribenes:
This is just an experiment, I'm interested to see if this breaks any existing code 👀 

Example:
```scala
////////////////////////
module testey

def age[A](x: A): Int = 100
record Person(name: String, age: Int)

/////////////////////////
module main

import testey

def main() = {
  val p = Person(name = "Jolene", age = 42)
  val a: Int = age(p) // can now be qualified as `Person::age`!
  // error: Ambiguous overload.
  // There are multiple overloads, which all would type check:
  //  - testey::age: [A](A) => Int
  //  - testey::age: Person => Int
  println(a)
}
```

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Namer.scala
+++ b/effekt/shared/src/main/scala/effekt/Namer.scala
@@ -585,11 +585,18 @@ object Namer extends Phase[Parsed, NameResolved] {
         val fieldId = paramTree.id.clone
         val name = Context.nameFor(fieldId)
         val fieldSym = Field(name, paramSym, constructor, paramTree)
-        if (defineAccessors) {
-          Context.define(fieldId, fieldSym)
-        } else {
-          Context.assignSymbol(fieldId, fieldSym)
+
+        // Constructor::field
+        Context.namespace(constructor.name.name) {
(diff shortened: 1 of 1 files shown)


jiribenes:
Ideally, I'd like to allow `person.age` for the field using dot access, but only `Person::age(person)` as a function.
But I have to admit that this would, in a way, go against the other rules for UFCS-like things we have. :/

jiribenes:
This actively worsens the error messages, though I'm not really sure why: 👀
```scala
type MyList {
  Bar(myHead: Int)
}

def foo(l: MyList) = myHead(l)
```

~>

```diff
=> Diff (- obtained, + expected)
-[error] examples/neg/selecting_from_datatype.effekt:8:22: Cannot find type for myHead -- forward uses and recursive functions need annotated return types.
+[error] examples/neg/selecting_from_datatype.effekt:8:22: Cannot find a function named `myHead`.
 def foo(l: MyList) = myHead(l)
-                     ^^^^^^^^^
+                     ^^^^^^
```

jiribenes:
I don't believe this is worth it for now. Oh well.

PR #1194, opened 2025-11-07
Infer operation type params based on lexically closest valid capability

### LINKED ISSUE #1189: infer type parameters for effect operations based on the locally-available effects
If an effect operation doesn't use all the interface's type parameters, then currently one is forced to type those parameters explicitly to get code compiling:
```effekt
interface Greet[T] { def sayHello(): Unit }

def helloWorld() = try {
  do sayHello() // Effects need to be fully known, but effect Greet[T]'s type parameter(s) T could not be inferred.
} with Greet[Int] {
  def sayHello() = { println("Hello!"); resume(()) }
}
```
Potentially related to #1157, since `do sayHello()` can be thus inferred to need `Greet[_]` (as in *any* `Greet` effect will do), and then it'd typecheck since `Greet[Int] <: Greet[_]`

jiribenes:
Resolves #1189.

The gist of the approach is: if you don't find "our" capability and that's because of some unknowns, try using the closest one in the scope.

This **does** fix inference of:
```scala
interface Greet[T] { def sayHello(): Unit }

def helloWorld() = try {
  do sayHello() // Greet[Int] effect
} with Greet[Int] {
  def sayHello() = { println("Hello!"); resume(()) }
}
```

But this does **not** fix inference of:
```scala
  for { [1, 2, 3].each } { x =>
    println(x) // error: ambiguous 'println'
  }
```

---

This also picks up where https://github.com/effekt-lang/effekt/pull/678 left off :)


effekt/shared/src/main/scala/effekt/typer/CapabilityScope.scala
```diff
@@ -24,7 +24,19 @@ case object GlobalCapabilityScope extends CapabilityScope {
...
+        case Nil => None
+        case _ => capabilities.collectFirst {
+          case (handlerTpe, cap) if handlerTpe.typeConstructor == tpe.typeConstructor =>
+            (tpe.args zip handlerTpe.args).foreach { case (inferred, concrete) =>
+              C.requireSubtype(inferred, concrete, ErrorContext.MergeTypes(inferred, concrete)) // TODO(jiribenes, 2025-11-07): better error message?
+            }
+            cap
+        }
+      }
+    }.getOrElse(parent.capabilityFor(tpe))
```
jiribenes: Note the following where, perhaps confusingly, if you remove the `[A]` in `val a = do read[A]()`, you get "got A but expected B" on the use site (last line):
https://github.com/effekt-lang/effekt/blob/795b14e1c2bec5de6d44742b67c19db318bf537c/libraries/common/stream.effekt#L154-L163

jiribenes: Also, it doesn't work here -- removing the `[Char]` annotations causes the good old "Effects need to be fully known, but effect Scan[A]'s type parameter(s) A could not be inferred." to trigger again.
Note that the `try` there in the snippet handles something else, but the inner `exponent` function is tricky: it does have the `Scan[A]` effect inferred, but it's "too far" for this method to discover it? (or there's some logic error I've made)

https://github.com/effekt-lang/effekt/blob/795b14e1c2bec5de6d44742b67c19db318bf537c/libraries/common/json.effekt#L94-L99

effekt/shared/src/main/scala/effekt/typer/CapabilityScope.scala
```diff
@@ -24,7 +24,19 @@ case object GlobalCapabilityScope extends CapabilityScope {
...
   def capabilityFor(tpe: symbols.InterfaceType)(using C: Context): symbols.BlockParam =
-    capabilities.getOrElse(tpe, parent.capabilityFor(tpe))
+    capabilities.get(tpe).orElse { // TODO(jiribenes, 2025-11-07): WIP, this is pretty horrible...
+      // If the requested type has unbound type parameters, try lexical resolution
+      unknowns(tpe).toList match {
+        case Nil => None
+        case _ => capabilities.collectFirst {
+          case (handlerTpe, cap) if handlerTpe.typeConstructor == tpe.typeConstructor =>
+            (tpe.args zip handlerTpe.args).foreach { case (inferred, concrete) =>
+              C.requireSubtype(inferred, concrete, ErrorContext.MergeTypes(inferred, concrete)) // TODO(jiribenes, 2025-11-07): better error message?
```
jiribenes: I'd be interested to see if we can even trigger the error here.

effekt/shared/src/main/scala/effekt/typer/CapabilityScope.scala
```diff
@@ -24,7 +24,19 @@ case object GlobalCapabilityScope extends CapabilityScope {
...
+            }
+            cap
+        }
```
jiribenes: The following program currently compiles, choosing the `Int` variant, which I'm not a fan of, I'd prefer if this reported an error, esp. if we believe that `try-with-with` and `try-(try-with)-with` are different:
```scala
interface Greet[T] { def sayHello(): Unit }

def helloWorld() = try {
  do sayHello()
} with Greet[Int] {
  def sayHello() = { println("Hello!"); resume(()) }
} with Greet[String] {
  def sayHello() = { println("Hello!"); resume(()) }
}
```

I quickly tried post-processing the results by using `collect` and then `match`ing on the result:
```scala
 capabilities.collect { ... } match {
          // Only allow this if there's exactly one solution in this scope!
          case Nil => None
          case single :: Nil => Some(single)
          case _ => None
        }
```

But this seemingly _still_ chooses the `Int` version (based on the "`Greet[String]` is not used" error) and, in addition, throws the following error:
```
Different arms of a conditional/match have incompatible types.

One arm has type
  T
while another one has type
  String

Type mismatch between String and Int.
```

I assume that in order to fix this, it would really only need to `checkSubtype`, not `requireSubtype` (so not do any unification, ideally, until we choose the correct candidate).

---

If we cannot do that because of technical reasons, perhaps the unification here is an overkill, couldn't we just check "this op has a typever and this handler does not, let's use _that_"?

jiribenes:
I tried a dummy unifier + `.collect` instead of `.collectFirst`, but it didn't really work (is a regression to the original state, essentially) :/

EDIT: a dummy unifier + `.collectFirst` also doesn't seem to work for me 😭 

EDIT 2: Yes, I see it now 🤦‍♂️ 

<details>

```patch

commit 43df2f27467a9327eba1676a3299722b25790fba
Author: Jiří Beneš <mail@jiribenes.com>
Date:   Tue Nov 11 20:16:17 2025 +0100

    Try using a dummy unifier

diff --git a/effekt/shared/src/main/scala/effekt/typer/CapabilityScope.scala b/effekt/shared/src/main/scala/effekt/typer/CapabilityScope.scala
index afd6a5f94..700260a8a 100644
--- a/effekt/shared/src/main/scala/effekt/typer/CapabilityScope.scala
+++ b/effekt/shared/src/main/scala/effekt/typer/CapabilityScope.scala
@@ -2,7 +2,8 @@ package effekt
 package typer
 
 import effekt.context.Context
-import effekt.{ source, symbols }
+import effekt.symbols.{Captures, Type, TypeVar, ValueType}
+import effekt.{source, symbols}
 import effekt.util.messages.ErrorMessageReifier
 
 /**
@@ -21,20 +22,51 @@ case object GlobalCapabilityScope extends CapabilityScope {
   def capabilityFor(tpe: symbols.InterfaceType)(using C: Context): symbols.BlockParam =
     C.abort(pretty"Effect ${tpe} is not allowed in this context.")
 }
+
+object DummyUnifier extends TypeUnifier {
+  var hasError = false
+
+  def checkSubtype(t1: effekt.symbols.ValueType, t2: effekt.symbols.ValueType)(using C: Context): Boolean = {
+    hasError = false
+    unifyValueTypes(
+      C.substitution.substitute(t1),
+      C.substitution.substitute(t2),
+      ErrorContext.MergeTypes(t1, t2))
+    val hadError = hasError
+    hasError = false
+    hadError
+  }
+
+  override def requireLowerBound(x: TypeVar.UnificationVar, tpe: ValueType, ctx: ErrorContext): Unit = ()
+  override def requireUpperBound(x: TypeVar.UnificationVar, tpe: ValueType, ctx: ErrorContext): Unit = ()
+  override def requireEqual(x: TypeVar.UnificationVar, tpe: ValueType, ctx: ErrorContext): Unit = ()
+  override def requireSubregion(lower: Captures, upper: Captures, ctx: ErrorContext): Unit = ()
+
+  override def abort(msg: String, ctx: ErrorContext): Nothing = ??? // shouldn't ever be triggered [?]
+
+  override def error(msg: String, ctx: ErrorContext): Unit = { hasError = true }
+  override def error(left: Type, right: Type, ctx: ErrorContext): Unit = { hasError = true }
+}
+
 class BindSome(binder: source.Tree, capabilities: Map[symbols.InterfaceType, symbols.BlockParam],val parent: CapabilityScope) extends CapabilityScope {
   def copy: CapabilityScope = BindSome(binder, capabilities, parent.copy)
   def capabilityFor(tpe: symbols.InterfaceType)(using C: Context): symbols.BlockParam =
-    capabilities.get(tpe).orElse { // TODO(jiribenes, 2025-11-07): WIP, this is pretty horrible...
+    capabilities.get(tpe).orElse {
       // If the requested type has unbound type parameters, try lexical resolution
       unknowns(tpe).toList match {
         case Nil => None
-        case _ => capabilities.collectFirst {
-          case (handlerTpe, cap) if handlerTpe.typeConstructor == tpe.typeConstructor =>
-            (tpe.args zip handlerTpe.args).foreach { case (inferred, concrete) =>
-              C.requireSubtype(inferred, concrete, ErrorContext.MergeTypes(inferred, concrete)) // TODO(jiribenes, 2025-11-07): better error message?
-            }
-            cap
-        }
+        case _ =>
+          def isValidFit(handlerTpe: symbols.InterfaceType): Boolean =
+            handlerTpe.typeConstructor == tpe.typeConstructor &&
+              (tpe.args zip handlerTpe.args).forall { case (inferred, concrete) => DummyUnifier.checkSubtype(inferred, concrete) }
+
+          capabilities.collect { case (handlerTpe, cap) if isValidFit(handlerTpe) => (handlerTpe, cap) }.toList match {
+            case Nil => None
+            case List((handlerTpe, cap)) =>
+              (tpe.args zip handlerTpe.args).foreach { case (inferred, concrete) => C.requireSubtype(inferred, concrete, ErrorContext.MergeTypes(inferred, concrete)) }
+              Some(cap)
+            case _ => None // TODO: Could we have a better error message here?
+          }
       }
     }.getOrElse(parent.capabilityFor(tpe))
   override def toString: String = s"BindSome(${binder.getClass.getSimpleName}, ${capabilities}, ${parent})"

```

</details>

effekt/shared/src/main/scala/effekt/typer/CapabilityScope.scala
```diff
@@ -20,11 +22,60 @@ case object GlobalCapabilityScope extends CapabilityScope {
...
+              Some(cap)
+            case Nil => None
+            case caps => None
+          }
+        }
+      }
+    }.getOrElse(parent.capabilityFor(tpe))
+
+  def relevantInScopeFor(tpe: BlockType.InterfaceType)(using C: Context): Set[BlockType.InterfaceType] =
+    capabilities.collect { case (handlerTpe, _) if handlerTpe.typeConstructor == tpe.typeConstructor && unknowns(handlerTpe).isEmpty => handlerTpe }.toSet ++ parent.relevantInScopeFor(tpe)
```
jiribenes, line 84: I don't understand why
```
&& unknowns(handlerTpe).isEmpty
```
is needed here.

Try the `examples/neg/lexical_capability_selection_ambiguous.effekt` example without these additional conditions, and you'll also see `Greet[T]` suggested, which is whack.

examples/pos/lexical_capability_selection_proxy.effekt
```diff
@@ -0,0 +1,16 @@
...
+
+def greet[T, U](p: Proxy[T]) = do sayHello[T, U]()
+
+def helloWorld() = try {
+  Proxy[Int].greet()
+} with Greet[Int, Int] {
+  def sayHello() = { println("Hello, world!"); resume(()) }
+} with Greet[String, String] {
+  def sayHello() = { println("Hello, world!"); resume(()) }
+}
```
jiribenes, line 14: I was quite surprised that this example does **not** work [!]
```
examples/neg/lexical_capability_selection_proxy.effekt:10:3:
Effects need to be fully known, but effect Greet[Int, U]'s type parameter(s) U could not be inferred.
```
But I haven't debugged it yet, so I don't know why it happens...

I thought that the `Greet[Int, Int]` should be a valid cap here, as we're looking for `Greet[Int, ?U]`.

effekt/shared/src/main/scala/effekt/typer/CapabilityScope.scala
jiribenes: Resolved using a dummy unifier.

examples/pos/lexical_capability_selection_proxy.effekt
jiribenes: Oh, this doesn't even call `BindSome.capabilityFor`. Interesting...
(I tried to essentially do `do sayHello[Int, _]`, be this is not actually equivalent, is it...)

effekt/shared/src/main/scala/effekt/typer/ConcreteEffects.scala
```diff
@@ -80,11 +80,14 @@ implicit def asConcrete(effs: Effects)(using C: Context): ConcreteEffects =
...
-  effs.effects.foreach(assertConcreteEffect)
+  effs.effects.foreach { eff => assertConcreteEffect(eff) }
 
-private[typer] def assertConcreteEffect(eff: InterfaceType)(using C: Context): Unit =
+private[typer] def assertConcreteEffect(eff: InterfaceType, hints: Set[InterfaceType] = Set.empty)(using C: Context): Unit =
   unknowns(eff) match {
     case us if us.nonEmpty =>
+      hints.foreach { hint =>
+        C.info(pretty"Did you mean to use an operation of the concrete interface ${hint}?")
+      }
```
jiribenes, line 90: This might be a bit controversial, this is what the `relevantInScopeFor` was added for.
Not sure if it's worth it, you can see the result below in `examples/neg/lexical_capability_selection_ambiguous.effekt`, where it prints:
```
examples/neg/lexical_capability_selection_ambiguous.effekt:6:3:
[info] Did you mean to use an operation of the concrete interface Greet[Int]?
[info] Did you mean to use an operation of the concrete interface Greet[String]?
[error] Effects need to be fully known, but effect Greet[T]'s type parameter(s) T could not be inferred.
  do sayHello()
  ^^^^^^^^^^^^^
```

Screenshot 2025-11-11 at 21 29 52

examples/pos/lexical_capability_selection_proxy.effekt
b-studios: That's indeed weird

effekt/shared/src/main/scala/effekt/typer/CapabilityScope.scala
jiribenes: The latter now at least has a better error message 🌸
Screenshot 2025-11-11 at 21 45 22

jiribenes: But note that the following does not trigger the nice error message:
```
def myskip[A]() = do skip[A]()
/// Read a double value, in the common format (TODO document)
def readDouble(): Double / Scan[Char] = {
  def exponent(pre: Double) = {
    try {
      if(do peek() == 'e' || do peek() == 'E') {
        myskip() // !
```

Even though it's conceptually the same code.
This is also shown below in the Proxy example.

examples/pos/lexical_capability_selection_annot.effekt
```diff
@@ -0,0 +1,12 @@
...
+interface Greet[T] { def sayHello(): Unit }
+
+// unlike in lexical_capability_selection.effekt,
+// the _annotation_ here forces `sayHello` to be of `Greet[Int]`
+def helloWorld(): Unit / Greet[Int] =
+  do sayHello()
+
+def main() = try helloWorld() with Greet[Int] {
+  def sayHello() = { println("Hello there!"); resume(()) }
+}
```
jiribenes, line 12: This example is pretty cute, the type annotation introduces a capability, so the `sayHello` operation here is resolved correctly! 🥳

examples/pos/lexical_capability_selection_proxy.effekt
jiribenes: Works now, thanks, @b-studios!

effekt/shared/src/main/scala/effekt/typer/CapabilityScope.scala
```diff
@@ -20,11 +21,67 @@ case object GlobalCapabilityScope extends CapabilityScope {
...
+        }
+
+    val compatible = capabilities.filter { case (handlerTpe, cap) => isCompatible(handlerTpe) }.toList
+    compatible match {
+      case List((handlerTpe, cap)) =>
+        (tpe.args zip handlerTpe.args).foreach { case (inferred, concrete) =>
+          // Enforce subtype constraints for the type params of the chosen handler
+          C.requireSubtype(inferred, concrete, ErrorContext.MergeTypes(inferred, concrete))
+        }
+        Some(cap)
```
jiribenes, line 77: cc @b-studios: when I try an example from #1189:
```scala
interface Greet[T] { def sayHello(x: T): Unit }

def helloWorld() = try {
  do sayHello(box {})
  // ⚠️error: Effects need to be fully known, but effect Greet[() => Unit at ?C]'s type parameter(s) ?10319 could not be inferred.
  // info: Did you mean to use an operation of the concrete interface Greet[() => Unit at {}]?
} with Greet[=> Unit at {}] {
  def sayHello(x) = { println("Hello!"); resume(()) }
}
```

and trace it, I get:
```
trying compatibility of Greet[() => Unit at {}] [GreetList(BoxedType(FunctionType(List(),List(),List(),List(),ValueTypeApp(Unit,List()),Effects(List())),CaptureSet(Set())))] with our found Greet[() => Unit at ?C] [GreetList(BoxedType(FunctionType(List(),List(),List(),List(),ValueTypeApp(Unit,List()),Effects(List())),?5576))] (cap Greet)
// and then
 Greet[() => Unit at {}] with found Greet[() => Unit at ?C]: concrete () => Unit at {} vs inferred () => Unit at ?C: compat? Y
```

So `DummyUnifier.checkSubtype` claims that they _are_ compatible, but then the captures aren't propagated correctly [?]

jiribenes: One potential explanation would be that we're doing
```
() => Unit at ?C <: () => Unit at {}
```
which is not the direction that helps us solve `?C` (it just adds a lower bound of `{}` to `?C` which, well, not very helpful...) [?].
But should it be the other way around? I'm a bit stuck here 🤔

jiribenes: Doing `inferred <+: concrete` and `concrete <+: inferred` (where τ <+: ρ means "`<:` in a positive context") does not actually help.

b-studios: Hmm, 

```
() => Unit at ?C <: () => Unit at {}
```
should add an upper bound to ?C, no?

Anyways, I think the problem could be hidden in the implementation of the dummy comparison (`requireSubregion`) which doesn't do anything when comparing capture sets, right?

jiribenes: I'm not sure the problem is there though — it succeeds there, same as the C.requireSubtype which also succeeds. But then the capture variable in question is not-yet-solved as we solve them only later [?].

jiribenes: Moving into a follow-up issue.

jiribenes:
I'll merge this; am happy to improve anything later -- for now, I think this is good enough to be merged and released tomorrow.

PR #1199, opened 2025-11-11
Try fixing the IO tour

jiribenes:
The captures changed in #1109, but were not changed in the IO Tour 👀


jiribenes:
I need[ed] this on the website _yesterday_, so I'll just merge this when the tests pass and update it on the website 👀

PR #1204, opened 2025-11-14
Delegate platform error code mapping to node's getSystemErrorName

### LINKED ISSUE #702: Reading a non-existent file on JS on Windows crashes
Possible repro (awaiting confirmation)
```scala
import io
import io/error
import io/filesystem
import exception

def peek(filename: String): Option[Char] / Files = {
  val contents = do readFile(filename)
  if (contents.length == 0) None()
  else Some(contents.unsafeCharAt(0))
}

def run(filename: String) = {
  with default[IOError, String]{ "File not found" };
  with filesystem;
  peek(filename) match {
    case None() => "Empty file"
    case Some(c) => "The first letter is: " ++ show(c)
  }
}

def main() = println(run("ThisFileDoesNotExist"))
```

works on my machine (M1).
But on Windows, it seems to crash with:

Screenshot 2024-11-20 at 11 53 42

Here's a trace of how I _think_ this might have happened:

https://github.com/effekt-lang/effekt/blob/898c86c1f090285521c950b464acc187d02e6e57/libraries/common/io/error.effekt#L455

https://github.com/effekt-lang/effekt/blob/898c86c1f090285521c950b464acc187d02e6e57/libraries/common/io/error.effekt#L356

timsueberkrueb:
I checked that this fixes #702 on Windows, where the provided reproducing example now returns

```console
$ effekt issue702.effekt 
File not found
```

I have not tested this on other platforms, yet.

### DIFF
--- a/libraries/common/io/error.effekt
+++ b/libraries/common/io/error.effekt
@@ -362,97 +362,110 @@ namespace internal {
     declare %Int @c_error_number(%Int)
   """
 
-  extern jsNode """
-    const os = require('node:os');
+extern jsNode """
+    const { getSystemErrorName } = require('node:util');
 
     /**
      * Maps the error code to a Effekt-stable (platform independent) numeric value.
      *
      * Tries to use most common errno integer values, but introduces fresh values (> 200)
      * for those without common errno values.
+     *
+     * The mapping from platfrom-specific error codes to Effekt error codes is handled by
+     * getSystemErrorName from 'node:util', which in turn uses the error code conventions from libuv.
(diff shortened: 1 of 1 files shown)


jiribenes:
As far as I'm concerned, if this works on a Windows machine for the one given example + works on CI (which does I/O stuff on Linux), then I'm happy to merge this (and worst case scenario, push a patch 0.52.1 release on Monday/Tuesday if something's horribly wrong).
Thanks!

jiribenes:
Can confirm that it also works on MacOS! 🥳 
```
$ effekt repro.effekt
File not found
```

PR #1205, opened 2025-11-14
A more complete error message when args/params do not match

jiribenes:
Before, Effekt would early exit on "expected 1 value argument but got 0", now it also says that, in addition, it did _not_ expect a block argument!
There's also some useful hints:

Screenshot 2025-11-14 at 19 12 15

Screenshot 2025-11-14 at 18 53 47

Screenshot 2025-11-14 at 19 13 04

Screenshot 2025-11-14 at 19 14 05

Screenshot 2025-11-14 at 19 21 00

---

I don't want to add new neg tests for this because I think the message could be improved even more and I don't want to stabilise it for now.

### DIFF
--- a/examples/neg/blocks_wrong_arguments.effekt
+++ b/examples/neg/blocks_wrong_arguments.effekt
@@ -4,7 +4,7 @@ def foo { f : Int => Int } =
   f(2)
 
 def bar { f : (Int, Bool) => Unit } =
-  f(2, true, 4) // ERROR Wrong number of value arguments
+  f(2, true, 4) // ERROR Wrong number of arguments to f: expected 2 value arguments, but got 3 value arguments
 
 def main() = {
   val r = foo { (x: Int, y: Bool) =>


effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -1275,6 +1268,66 @@ object Typer extends Phase[NameResolved, Typechecked] {
...
+        val diff = gotBlocks - expectedBlocks
+        Context.info(pretty"Did you mean to pass ${pluralized(diff, "block argument")} as a value? e.g. box it using `box { ... }`")
+      } else if (gotValues > expectedValues) {
+        val diff = gotValues - expectedValues
+        Context.info(pretty"Did you mean to pass ${pluralized(diff, "value argument")} as a block (computation)?")
+      }
+    }
+
+    if (!targsOk || !vargsOk || !bargsOk) {
+      Context.abort(s"Wrong number of arguments to ${name}: expected ${expected}, but got ${got}")
```
jiribenes, line 1327: WDYT about the message?


We could have some alternate messages here like:

- "${name} expected ... but got ..."
- "Argument mismatch: ${name} expected ... but got ..."

We could also avoid repeating the word _arguments_ everywhere:

"Wrong number of _arguments_ to XYZ: expected 4 value _arguments_ and 2 block _arguments_, but got 3 value _arguments_ and no block _arguments_" is a real mouthful.
Would something like "Wrong number of _arguments_ to XYZ: expected 4 values and 2 blocks, but got 3 values and no blocks" be better?

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -1275,6 +1268,66 @@ object Typer extends Phase[NameResolved, Typechecked] {
...
+  /**
+   * Asserts that number of {type, value, block} arguments is the same as
+   *          the number of {type, value, block} parameters.
+   * If not, aborts the context with a nice error message.
+   */
+  private def assertArgsParamsAlign(
+    name: String,
+    gotTypes: Int, gotValues: Int, gotBlocks: Int,
+    expectedTypes: Int, expectedValues: Int, expectedBlocks: Int
+  )(using Context): Unit = {
```
jiribenes, line 1280: The signature uses `Int`s because I couldn't find reasonable common types between the type checking of a block literal and of a function call 😭. (I could also use `List[_]` everywhere, but I'm not a fan of pretending like that's a better design)
I'd much prefer these to be some _specific_ lists of concrete types \:)

jiribenes:
Merging, happy to revise this later.

PR #1212, opened 2025-11-17
Don't call block literals by name

### LINKED ISSUE #1195: Performance regression in product_early with jump with arguments in machine
#1134  made the `product_early` benchmark run significantly slower. We must investigate.

serkm:
This fixes #1195

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Compiler.scala
+++ b/effekt/shared/src/main/scala/effekt/Compiler.scala
@@ -3,6 +3,7 @@ package effekt
 import effekt.PhaseResult.{AllTransformed, CoreTransformed}
 import effekt.context.Context
 import effekt.core.Transformer
+import effekt.core.optimizer.BindSubexpressions
 import effekt.namer.Namer
 import effekt.source.{AnnotateCaptures, ExplicitCapabilities, ModuleDecl, ResolveExternDefs}
 import effekt.symbols.Module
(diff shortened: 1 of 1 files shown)


effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -189,7 +189,18 @@ object Transformer {
...
+                  core.Let(arg.id, argExpr.tpe, argExpr, rest)
+              }
+              val bindBlocks = bargs.zip(bparams).foldRight((bindValues)) {
+                case ((argBlock, arg), rest) =>
+                  core.Def(arg.id, argBlock, rest)
+              }
+              val vnames = vparams.map(p => core.ValueVar(p.id, p.tpe))
+              val bnames = bparams.map(p => core.BlockVar(p.id, p.tpe, p.capt))
+
+              transform(substitute(Block.BlockLit(tparams, cparams, vparams, bparams, bindBlocks), targs, vnames, bnames))
```
b-studios: Weird, don't we enforce the invariant before machine.Transformer to bind all arguments to names?

b-studios:
We have this phase: https://github.com/effekt-lang/effekt/blob/master/effekt/shared/src/main/scala/effekt/core/optimizer/BindSubexpressions.scala

It is called here in the optimizer:

https://github.com/effekt-lang/effekt/blob/100d19aa8631e4591abf87ee661df305857496fb/effekt/shared/src/main/scala/effekt/core/optimizer/Optimizer.scala#L39

However, maybe we do not maintain it and inline things again...

Could you try whether running it before machine transformer also fixes the issue?

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
serkm: It's enforced in `Optimizer`, which runs before `PolymorphismBoxing`

serkm:
Yes, `BindSubexpressions` also works.

phischu (approved):
Yes, this re-establishes the invariant broken by PolymorphismBoxing. Thanks @serkm

PR #1216, opened 2025-11-19
Fix off-by-one in spans

jiribenes:
See example below that previously failed and now works.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/source/Spans.scala
+++ b/effekt/shared/src/main/scala/effekt/source/Spans.scala
@@ -10,7 +10,8 @@ object Spans {
   def findNodesContaining[T <: Tree](nodes: Vector[T], position: Position): Vector[T] = {
     nodes.filter { t =>
       val r = t.span.range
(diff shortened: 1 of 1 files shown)


effekt/jvm/src/test/scala/effekt/LSPTests.scala
```diff
@@ -659,6 +659,31 @@ class LSPTests extends FunSuite {
...
 
+  test("right edge of definition returns expected range") {
+    withClientAndServer { (client, server) =>
+      val (textDoc, positions) =
+        raw"""
+             |def foo() = <>
+             |↑             ↑
+             |def bar() = foo()
+             |               ↑
+             """.textDocumentAndPositions
```
jiribenes, line 670: When you double-click an identifier to highlight it, say the `readFile` in `val contents = do readFile(filename)`, your cursor is in the following position where the `‸` is
```scala
val contents = do readFile‸(filename)
```

but previously, we did not consider this position as it's technically beyond the `[start, end)` range where `readFile` is defined!

PR #1218, opened 2025-11-19
Experiment: use JSDoc types in JS runtime

jiribenes:
Most of the contents are scaffolding, what this really adds is:
1. types in `libraries/js/effekt_*.js` using the JSDoc syntax
2. a `// @ts-check` comment at the top of the file

This means that we get some light form of type checking when investigating the JS output in VSCode.
It's especially useful when writing FFI. :)

I have not tested this properly, I really only tried it on two examples.

To me, this is just a first step: the type annotations currently present are a combination of my own vibes and of Claude's ideas of what the types should be. There's also a lot of `@ts-ignore` comments and casts to `any` and similar things to make TS shut up, this could likely be much more improved, hence the _draft_ status of this PR.

### DIFF
--- a/libraries/js/effekt_builtins.js
+++ b/libraries/js/effekt_builtins.js
@@ -28,8 +28,8 @@ $effekt.equals = function(obj1, obj2) {
 }
 
 /**
- * @param {number} n1
- * @param {number} n2
+ * @param {*} n1
+ * @param {*} n2
  */
 function compare$prim(n1, n2) {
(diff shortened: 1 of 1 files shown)


jiribenes:
There are still a few spurious errors like the following, but it seems that it works fine now:
image

jiribenes:
But the types propagate somewhat well if there aren't any local functions in the way -- this is from `parsing_dollars`:
image

PR #1219, opened 2025-11-21
Disable subtyping, introduce coercions, and type check core

jiribenes:
See the PR comments.

### DIFF
--- a/.github/actions/setup-effekt/action.yml
+++ b/.github/actions/setup-effekt/action.yml
@@ -15,7 +15,7 @@ inputs:
     required: false
     default: '18'
   install-dependencies:
-    description: 'Whether to install system dependencies (Linux only)'
+    description: 'Whether to install system dependencies (Linux & macOS only)'
     required: false
     default: 'false'
   install-valgrind:
@@ -40,6 +40,7 @@ runs:
       uses: actions/setup-node@v4
       with:
         node-version: ${{ inputs.node-version }}
+
 
     # Only run the following on Linux-based runners
     - name: Update apt database
@@ -66,3 +67,17 @@ runs:
       if: ${{ inputs.install-dependencies == 'true' && runner.os == 'Linux' }}
       run: sudo apt-get install -y libuv1-dev
       shell: bash
+
+
+    # Only run the following on macOS-based runners
+    - name: Install LLVM ${{ inputs.llvm-version }} (macOS)
+      if: ${{ inputs.install-dependencies == 'true' && runner.os == 'macOS' }}
+      run: |
+        brew install llvm@${{ inputs.llvm-version }}
+        echo "$(brew --prefix llvm@${{ inputs.llvm-version }})/bin" >> $GITHUB_PATH
+      shell: bash
+
+    - name: Install libuv (macOS)
+      if: ${{ inputs.install-dependencies == 'true' && runner.os == 'macOS' }}
+      run: brew install libuv
+      shell: bash 
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -96,6 +96,25 @@ jobs:
       - name: Run Windows smoke test
         run: sbt "effektJVM/testOnly effekt.JavaScriptTests -- --tests=.*examples[\\/]*pos[\\/]*sideeffects.*"
 
+  macos-llvm-tests:
+    name: "macOS LLVM Smoke Tests"
+    needs: build-and-compile
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v6
+        with: 
+          submodules: 'true'
+
+      - uses: ./.github/actions/setup-effekt
+        with:
+          install-dependencies: 'true'
+
+      - name: Compile project
+        run: sbt Test/compile
+
+      - name: Run macOS LLVM smoke tests (with debug)
+        run: EFFEKT_DEBUG=1 sbt "effektJVM/testOnly effekt.LLVMTests -- --tests=.*benchmarks[\\/]*input_output[\\/].*"
+
   js-tests:
     name: "JS Backend"
     needs: build-and-compile
@@ -146,5 +165,5 @@ jobs:
 
       - uses: ./.github/actions/restore-build-cache
 
-      - name: Run LLVM backend tests
+      - name: Run LLVM backend tests (with debug & valgrind)
         run: EFFEKT_VALGRIND=1 EFFEKT_DEBUG=1 sbt effektJVM/testBackendLLVM


jiribenes:
The following program breaks down for reasons I don't understand:
```scala
effect Raise(msg: String): Unit

def main() = try {
  def block(msg: String): Unit / {} = { do Raise(msg) }

  val f: () => Unit at raise = {
    val thn = box { block("first") }
    val els = box { block("second") }

    if (true) { // <  error: 
      thn       // <      Not allowed {raise}
    } else {    // <
      els       // <
    }
  }
} with raise: Raise { msg => panic(msg) }
```

`thn` and `els` are both inferred to be `() => Unit at raise`
Screenshot 2025-11-21 at 14 23 27

but the `if-then-else` stmt claims "Not allowed {raise}".

If I _manually_ annotate the type of either `thn` or `els` with `: () => Unit at raise`, then everything works again. Strange.

jiribenes:
The following commutation also works:
```scala
// [...]
    val thn = box { block("first") }
    val els = box { block("second") }

    def res() /* at raise */ = if (true) {
      thn()
    } else {
      els()
    }

    box res
  }
// [...]
```

b-studios:
I disabled subtyping in contravariant positions, since there we cannot (easily) insert coercions. In particular, we do not want to eta-expand functions to insert coercions.

b-studios:
Current plan to insert coercions: 

1. since Typer annotated inferred all types but used subtyping in some positions, in the translation to core we now need to insert the coercions where necessary. In order to determine whether coercions are necessary, core-Transformer could pass down the expected return type of a statement / expression (hence only covariant coercions). 
2. If there is a type mismatch between the expected (Core Type?) and the type of the generated core expression that can be resolved by coercions, insert them there.

jiribenes:
btw, in case we'd want other names than `isZeroLike` and `isOneLike`, we could try borrowing names from type theory with `isUninhabited` and `isContractible` :^)

b-studios:
Quick status report: yesterday night, I tried to thread the expected type through the `core.Transformer` in order to insert coercions. I am not happy with the result since I am partially reimplementing type checking.

So I am currently tempted to go back to my original idea where Typer detects coercions and annotates them in the DB (yes... more side channeling---broken window).

Ideally, typer would perform elaboration. However, if we have unification based (HM-like) type inference, this is a bit tricky to achieve in general. How do other languages (like Koka) implement elaboration if things might become known only much later (and non-locally)?

It seems an elaborating typer is easier to implement in a bidirectional setting.

b-studios:
Note: of course we can collect coercions also by `Source` and then display them in LSP as inlay information. This way some errors might be less confusing. For example in the test case

```
def foo(): Option[Int] = if (true) Some(42)
```
we infer a coercion from `Some(42)` to Unit since the else branch is missing.

b-studios:
TODOs

- [x] Look into machine and LLVM at all.
- [x] potentially drop type annotations on `Val` and `Let` in core since typing should be unique, now that `Match` is annotated.
- [x] look into VMTests (I expect there are a lot of failing tests since we now add coercions regardless of the backend and I disabled some type-based optimizations)
- [x] implement proper type comparison on core (alpha equivalence, etc)
- [x] implement a proper type checker for core, now that the type system is simpler again.
- [ ] consider replacing `<>: tpe` by `<> match [tpe] {}` and assign `Nothing` as a type for all holes.

b-studios:
Explicitly annotating match with the motif together with limited checks of type-correctness in core surfaced a few bugs:

A lot of times, transformations like `Normalizer`, `RemoveTailResumptions`, or `DirectStyle` change the result type and hence these transformations also need to change the motif annotated on the match. While this wasn't necessary before, this might have led to problems in PolymorphismBoxing, which we just haven't discovered yet.

b-studios:
The VMTests look roughly ok with one exception: `parsing_dollars` that now suddenly has a lot of shifts and resumes that weren't there before.

```diff
   branches = 210,
-  pushedFrames = 68,
-  poppedFrames = 67,
+  pushedFrames = 1,
+  poppedFrames = 1,
   allocations = 0,
   variableWrites = 88,
-  resets = 2,
-  shifts = 68,
-  resumes = 66
+  resets = 1,
+  shifts = 1,
+  resumes = 0
```

Before we generated this code in core:

```
val v_r_7058: Nothing_383 = shift(p_7035) { {k_7106} =>
  return ()
};
v_r_7058 match {}
```
while now we generate
```
shift(p$14160: Prompt[Unit] @ {pCapt$14159}) { (){k$14237 @ resume$14236: Resume[Unit, Unit]} => 
  return ()
}
```

The former is arguably better, since it is trivially tail resumptive (empty match) in any continuation other than `k_7106`. Hence the other shift and resumes have been removed originally, but are now still there.

b-studios:
All of this is still WIP, but here is my glorious plan to fix a lot of bugs related to free variables altogether and once and for all™️  :)

- [x] get rid of `DeclarationContext` AND `ErrorReport` (transitively) in `typecheck`
- [x] Rename `Result` -> `Typing`
- [x] make `val typing: Typing[ValueType]` a field on `Stmt`
- [x] define `def tpe = typing.result`, similar for `capt`
- [x] define `def free: Free = typing.free` and use it in the phases that require free variable computation
- [x] delete `core.Variables` (defined in core/Tree.scala) and replace all usages by `stmt.free`
- [x] delete `inferType(expr: Expr)` and friends.

Ultimately, the new core typer shouldn't add too many lines of code (maybe it is even fewer lines) and shouldn't be that costly (we can still optimize later). Maybe it even speeds up things since all free variables are always already precomputed and cached.

b-studios:
I don't have good ideas on how to get rid of the declaration context though. While it seems somewhat feasible to annotate calls to `make` with the constructor call, I am very hesitant to annotate calls to `new` with the interface type.

b-studios:
Thanks @marzipankaiser (and @jiribenes) for the idea of recording the requirements on constructors and interfaces Damas-style as well. I will implement this naively and later we can think about how to represent the constraints more (space) efficiently.

effekt/shared/src/main/scala/effekt/typer/TypeComparer.scala
```diff
@@ -22,33 +22,56 @@ trait TypeUnifier {
...
-  def unify(c1: Captures, c2: Captures, ctx: ErrorContext): Unit = ctx.polarity match {
+  def unifyCaptures(c1: Captures, c2: Captures, ctx: ErrorContext): Unit = ctx.polarity match {
     case Covariant     => requireSubregion(c1, c2, ctx)
     case Contravariant => requireSubregion(c2, c1, ctx)
     case Invariant     => requireSubregion(c1, c2, ctx); requireSubregion(c2, c1, ctx)
   }
-  def unify(c1: Capture, c2: Capture, ctx: ErrorContext): Unit = unify(CaptureSet(Set(c1)), CaptureSet(Set(c2)), ctx)
+  def unifyCapture(c1: Capture, c2: Capture, ctx: ErrorContext): Unit = unifyCaptures(CaptureSet(Set(c1)), CaptureSet(Set(c2)), ctx)
+
+  def isZeroLike(tpe: ValueType): Boolean = tpe match {
```
b-studios: we can delete those

effekt/shared/src/main/scala/effekt/typer/TypeComparer.scala
```diff
@@ -22,33 +22,56 @@ trait TypeUnifier {
...
-    case (ValueTypeRef(s: UnificationVar), t: ValueType, Contravariant) => requireLowerBound(s, t, ctx)
-    case (s: ValueType, ValueTypeRef(t: UnificationVar), Contravariant) => requireUpperBound(t, s, ctx)
+    case (ValueTypeRef(s: UnificationVar), t: ValueType, Invariant, _) => requireEqual(s, t, ctx)
+    case (s: ValueType, ValueTypeRef(t: UnificationVar), Invariant, _) => requireEqual(t, s, ctx)
 
-    case (ValueTypeRef(s: UnificationVar), t: ValueType, Invariant) => requireEqual(s, t, ctx)
-    case (s: ValueType, ValueTypeRef(t: UnificationVar), Invariant) => requireEqual(t, s, ctx)
+    // coercing is the last resort...
+    case (TBottom, to, Covariant, Some(coerce)) => coerce(Coercion.FromNothing(to))
+    case (from, TUnit, Covariant, Some(coerce)) => coerce(Coercion.ToUnit(from))
```
b-studios: delete this

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -548,15 +548,15 @@ class VMTests extends munit.FunSuite {
       dynamicDispatches = 0,
       patternMatches = 0,
       branches = 210,
-      pushedFrames = 1,
-      poppedFrames = 1,
+      pushedFrames = 68,
+      poppedFrames = 67,
```
b-studios, line 552: We can merge but should investigate this.

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -40,6 +40,13 @@ import effekt.source.Implementation
  */
 case class Result[+T](tpe: T, effects: ConcreteEffects)
 
+/**
+ * Coercion as inferred by Typer
+ */
+enum Coercion {
+  case ToAny(from: ValueType)
+  case FromNothing(to: ValueType)
+}
```
b-studios, line 49: Again, let's merge but also investigate later whether we want to have coercions _at all_.

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -1603,13 +1607,32 @@ object Typer extends Phase[NameResolved, Typechecked] {
...
+      case first :: rest => rest.foldLeft[ValueType](first) { (t1, t2) =>
+        Context.mergeValueTypes(t1, t2, ErrorContext.MergeTypes(Context.unification(t1), Context.unification(t2)))
+      }
+    }
+    // check again to insert coercions
+    // TODO this duplicates errors
+    tpesAndTerms.foreach { case (tpe, term) =>
+      Context.requireSubtype(tpe, result, ErrorContext.Expected(tpe, result, term.getOrElse(Context.focus),
+        term.map(t => c => Context.coerce(t, c))))
+    }
```
b-studios, line 1630: If we do not insert coercions, we can also delete this duplicate check here, which improves errors.

effekt/shared/src/main/scala/effekt/core/PatternMatchingCompiler.scala
```diff
@@ -85,9 +85,9 @@ object PatternMatchingCompiler {
    * - a sequence of clauses that represent alternatives (disjunction)
    * - each sequence contains a list of conditions that all have to match (conjunction).
    */
-  def compile(clauses: List[Clause]): core.Stmt = {
+  def compile(clauses: List[Clause], motif: ValueType): core.Stmt = {
```
b-studios, line 88: I like "motif" but of course we can also use the english term "motive" that Conor used.

effekt/shared/src/main/scala/effekt/core/Tree.scala
```diff
@@ -822,3 +721,24 @@ object substitutions {
   def substitute(capt: Captures)(using subst: Substitution): Captures =
     Type.substitute(capt, subst.captures)
 }
+
+@targetName("preserveTypesStmt")
+inline def preserveTypes(before: Stmt)(inline f: Stmt => Stmt): Stmt = {
```
b-studios, line 726: Do we want to keep these in "production" as well? This might be costly.

effekt/shared/src/main/scala/effekt/core/Type.scala
```diff
@@ -48,12 +49,111 @@ enum BlockType extends Type {
...
+// so `type Later = Expr.Make | Implementation`
+//
+// How do we check the dual method calls and pattern matches then against declarations?
+enum Constraints {
+  case Make(ctor: Id, result: ValueType.Data, existentialTypeArgs: List[ValueType], arguments: List[ValueType])
+  case Implementation(interface: BlockType.Interface, operations: Map[Id, BlockType.Function])
+
+  // These are always never traversed (accept when checking on the module level, so we use the free structure)
+  case Empty
+  case Append(lhs: Constraints, rhs: Constraints)
```
b-studios, line 73: We are not yet done here (but can complete this in a follow up PR)! We need to actually check whether the constraints hold when encountering the declarations.

effekt/shared/src/main/scala/effekt/core/Type.scala
```diff
@@ -202,89 +279,246 @@ object Type {
...
+      //      val existentialParams = ctor.tparams
+      //      // [T, S](S, S => S at {}, S => T at {}) => Coalg[T]
+      //      val sig: BlockType.Function = BlockType.Function(universalParams ++ existentialParams, Nil, ctor.fields.map(_.tpe), Nil,
+      //        ValueType.Data(data.name, universalParams.map(ValueType.Var.apply)))
+      //      // (Int, Int => Int at {}, Int => Bool at {}) => Coalg[Bool]
+      //      val BlockType.Function(_, _, paramTypes, _, retType) = instantiate(sig, data.targs ++ targs, Nil)
+      //      // TODO factor out callLike things
+      //      shouldEqual(data, retType)
+      //      val Typing(argTypes, argCapt, argFree) = all(vargs, typecheck)
+      //      valuesShouldEqual(paramTypes, argTypes)
```
b-studios, line 325: I'm leaving this here, since we will need a very similar logic when checking constraints at the declaration site.

libraries/common/union_find.effekt
```diff
@@ -31,11 +31,11 @@ namespace internal {
     }
   }
   def setParent(u: UnionFind, child: Int, parent: Int) = {
-    with on[OutOfBounds].default { do raise(MissingValue(), "No node " ++ child.show) }
+    with default[OutOfBounds, Unit] { do raise(MissingValue(), "No node " ++ child.show) }
```
b-studios, line 34: It is a bit sad that we need the type annotations here.

b-studios:
~@phischu once CI passes (which can take a time due to reparse tests), we can merge this~

b-studios:
I retract. The LLVM tests using files are still failing on my Mac M1. I would be nice to have an ARM runner; maybe we could set up one at some point.

PR #1220, opened 2025-11-22
Safer array primitives

phischu:
- [x] Review resizable array (what is the intended semantics even)
- [x] Change names of vm primitives.
- [x] Check allocateZeros and allocateNones on JS
- [x] Measure performance cost

### DIFF
--- a/examples/stdlib/resizable_array/resizable_array.effekt
+++ b/examples/stdlib/resizable_array/resizable_array.effekt
@@ -4,6 +4,7 @@ import test
 def main() = {
   suite("ResizableArrayTests", false) {
     test("usage as stack") {
+      with on[OutOfBounds].default { assertTrue(false, "out of bounds") }
       val a = resizableArray()
       a.add(1)
       a.add(1)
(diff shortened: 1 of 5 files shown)


phischu:
For some reason the stdlib tests go into an endless loop now.

libraries/common/resizable_array.effekt
```diff
@@ -21,112 +21,77 @@ def size[T](arr: ResizableArray[T]) = arr.rawSizePtr.get
 
 /// Allocate a new, empty dynamic array with given initial capacity
 def resizableArray[T](capacity: Int): ResizableArray[T] = {
-  ResizableArray(ref(0), ref(allocate(capacity)))
+  ResizableArray(ref(0), ref(unsafeAllocate(capacity)))
 }
 
-/// Allocate a new, empty dynamic array
+/// Aallocatellocate a new, empty dynamic array
```
jiribenes: typo 👀

examples/stdlib/resizable_array/resizable_array.effekt
```diff
@@ -27,19 +26,6 @@ def main() = {
...
-        with on[OutOfBounds].default { () }
-        arr.get(index)
-        assertTrue(false, "not out of bounds")
-      }
-      with on[OutOfBounds].default { assertTrue(false, "out of bounds") }
-      val a = resizableArray()
-      a.checkOutOfBounds(-1)
-      a.add(1)
-      a.add(2)
-      a.checkOutOfBounds(2)
```
marzipankaiser, line 41: Can we somehow have a test for something like this, still? Even if just checking that it panics afterwards - for the resizable arrays, it might return the value at index 2 still (if capacity is >2) even if the length is just 2. And this might be hard to catch otherwise. 🤔

jiribenes: I'd personally be fine with the resizable arrays still using `Exception[OutOfBounds]` everywhere (and perhaps later a `fail` instead).

PR #1225, opened 2025-11-27
Nicer error messages on argument list / tuple mismatch

### LINKED ISSUE #668: Nicer error messages on argument list / tuple mismatch
There are two general situations where the users confuse argument lists with tuples.
We could potentially special-case the messages here:

### 1. Want a tuple, got an argument list

```scala
def foo { f: (Int, Int) => Bool }: Unit = <>

def main() = foo { case (x, y) => true }
//                 ~~~~~~~~~~~~~~~~~~~
// Wrong number of value arguments, given 1, but function expects 2.
// ~> Did you mean `(x, y) => true`?
```

### 2. Want two arguments, got a tuple

```
def bar { f: ((Int, Int)) => Bool }: Unit = <>

def main() = bar { (x, y) => true }
//                 ~~~~~~~~~~~~~~~
// Wrong number of value arguments, given 2, but function expects 1.
// ~> Did you mean `case (x, y) => true`
```

jiribenes:
Resolves #668

Continues on from #1205 by having a much more structured representation: instead of just numbers, we actually `align` the two argument lists, seeing what's in both, what's only expected, and what's only received.

Then once we have the information, we can (very hackily!) resolve #668 by providing slightly more specialised messages when argument lists and tuples are mismatched. Note that this message is only shown somewhat conservatively: only if the swap for a tuple could work based on the arg count!

Screenshot 2025-11-27 at 16 11 49
Screenshot 2025-11-27 at 16 12 37

Screenshot 2025-11-27 at 16 12 16
Screenshot 2025-11-27 at 16 46 54

I think that all of these could eventually have quick fixes attached to them :)

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Typer.scala
+++ b/effekt/shared/src/main/scala/effekt/Typer.scala
@@ -1279,8 +1279,6 @@ object Typer extends Phase[NameResolved, Typechecked] {
     def gotCount: Int = matched.size + extra.size
     def expectedCount: Int = matched.size + missing.size
     def delta: Int = extra.size - missing.size // > 0 => too many
-
-    def show: String = pp"Aligned(matched=${matched}, gotExtra=${extra}, expectedMissing=${missing})"
   }
 
   object Aligned {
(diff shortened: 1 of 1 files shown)


effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -1268,63 +1268,134 @@ object Typer extends Phase[NameResolved, Typechecked] {
...
+        case _ => ()
+      }
+
+      // 3. User wrote `foo((x, y))` (tuple literal) but function expects multiple arguments
+      // => Single matched arg that's a Call to a Tuple constructor, with some missing params
+      // HACK: Hardcoded "tuple constructor is IdRef to TupleN in effekt namespace"
+      (values.matched, values.extra, values.missing) match {
+        case (List((arg: source.ValueArg, _)), Nil, missing@(_ :: _)) =>
+          (arg.value, name) match {
+            case (source.Call(source.IdTarget(source.IdRef(List("effekt"), tupleName, _)), _, tupleArgs, _, _), Some(givenName))
```
b-studios, line 1352: Polymorphism boxing used to use `findPrelude` to find these more precisely.

jiribenes:
I'll merge this since I've made a follow-up for the ad-hoc detection of tuples/lambda-case.

PR #1227, opened 2025-11-28
Fix error on namespace

### LINKED ISSUE #1222: Internal Compiler Error on empty namespace
_Originally reported by @phischu_

The following code crashes the compiler:
```cpp
namespace empty {}
```

Image

The only occurrence of `reduceLeft` is here in Typer, introduced in https://github.com/effekt-lang/effekt/pull/1074 by @marvinborner:
https://github.com/effekt-lang/effekt/blob/6532fa53b57581c751d472317624e348874993ec/effekt/shared/src/main/scala/effekt/Typer.scala#L948-L951

dvdvgt:
Closes #1222

### DIFF
--- a/examples/pos/empty_namespace.effekt
+++ b/examples/pos/empty_namespace.effekt
@@ -1,0 +1,3 @@
+namespace empty {}
+
+def main() = ()


jiribenes:
Could you please add an example as a test to make sure we'll catch this regression next time?

jiribenes (approved):

PR #1230, opened 2025-11-28
Allow (multi-)pattern matching in 'with val'

### LINKED ISSUE #707: `with val` syntax sugar for more complex types
**Problem**: There's no way to unpack more complex types the using `with val _ = ...` syntax sugar.

When working with `stream`, I tried to do something like:
```scala
def main() = {
  with val (i, x) = [(1, 'a'), (2, 'b'), (3, 'c')].foreach;
  println(show(i) ++ ": " ++ show(x))
}
```
in order to unpack the tuple.  
However, that's wrong since this sugar would imply that foreach takes two arguments
as it desugars to something like:
```scala
...foreach { (i, x) => ... }
```

However the _correct_ diet, sugar-free version is the following:
```scala
def main() =
  [(1, 'a'), (2, 'b'), (3, 'c')].foreach { case (i, x) => 
    println(show(i) ++ ": " ++ show(x))
  }
```

I tried to remedy this by using `Tuple2(i, x)` explicitly, but that reports a parse error:
```scala
def main() = {
  with val Tuple2(i, x) = [(1, 'a'), (2, 'b'), (3, 'c')].foreach;
  //             ^ Expected = but got (

  println(show(i) ++ ": " ++ show(x))
}
```
Similarly if I try to use a custom type:
```scala
record Pos(i: Int, x: Char)
def main() = {
  with val Pos(i, x) = [Pos(1, 'a'), Pos(2, 'b'), Pos(3, 'c')].foreach;
  //          ^ Expected = but got (

  println(show(i) ++ ": " ++ show(x))
}
```

What _does_ work is the following:
```scala
def main() = {
  with val tup = [(1, 'a'), (2, 'b'), (3, 'c')].foreach;
  val (i, x) = tup;
  println(show(i) ++ ": " ++ show(x))
}
```

jiribenes:
Resolves #707

In this PR, we allow more general `with val`, specifically `with val <PATTERN> (, <PATTERN>)* (and <GUARD>)* = <EXPR> (else <STMT>)?; <STMTS>`.
This comes at a cost: `with val x: Int = foreach([1,2,3]);` becomes unsupported as the `: Int` is no longer possible (since it's a pattern and those cannot have type annotations right now).

This adds a lot of pos tests since it's the only way I can be sure this does what I want it to :D

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/core/ReparseTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/core/ReparseTests.scala
@@ -45,7 +45,7 @@ class ReparseTests extends CoreTests {
     // FIXME: Added in #1230, blocked on #1258 (https://github.com/effekt-lang/effekt/pull/1258)
     //        doesn't work since any program importing `json` fails due to string escapes.
     examplesDir / "pos" / "with_val_and_else.effekt",
-    examplesDir / "pos" / "with_val_and_two_args_and_else.effekt",
+    examplesDir / "pos" / "with_val_two_args_and_else.effekt",
     examplesDir / "pos" / "with_val_pair_and_else.effekt",
   )
 


examples/pos/withstatement.effekt
```diff
@@ -11,8 +11,8 @@ def bar { f: (Int, String) => Unit / {}}: Unit =
 
 def user(): Unit = {
     with printer;
-    with val x: Int = foreach([1,2,3]);
-    with val (a, b) = bar;
+    with val x = foreach([1,2,3]);
+    with val a, b = bar;
```
jiribenes, line 15: This is a breaking change ^ as type annotations are no longer supported and multiple arguments have to be written _without_ the parenthesis.

b-studios: Does it make sense to support `else` clauses for exhaustive pattern matching?

Like:

```
with val Some(v) = foo() else do raise()
...
```

jiribenes: I don't see a reason why not :)
In an ideal case, I'd like to share as much parser code as possible with destructuring ala `val (x, Some(y)) = ...` (and with lambda-case)

jiribenes: This is now done.

jiribenes:
Since the core reparser test failures are investigated in #1258, I've taken the liberty of disabling the new tests that add import (for now, anyways).

PR #1231, opened 2025-11-29
Relax 'unifyEffects', improving error messages for effect set mismatches

### LINKED ISSUE #813: Spurious type error when joining boxed types with effects
The following Effekt program (or a similar one with `match`):

```effekt
effect Go[T](): Unit

def dropTill[T2](l: (Int => Int / { Go[T2] } at {})): (Int => Int / { Go[T2] } at {}) = 
  if (false) { dropTill[T2](l) } else { l }
```

it fails with the error message:
```
{ Go[T2] } is not equal to { Go[T2] }
  l match {
  ^
```
raised from `unifyEffects`:
https://github.com/effekt-lang/effekt/blob/cfa951ebe18f830be9e08b07e865ed16080da083/effekt/shared/src/main/scala/effekt/typer/TypeComparer.scala#L83
called from the `join` between the result types in the `if`/`match`.

The two `T2` have a different id (when inspected via the debugger).

### LINKED ISSUE #866: { emit[T] } is not equal to { emit[T] }
> EDIT: [Playground link](https://effekt-lang.org/playground.html?playground=bY%2B9CsMwDIR3P8WN9pTdkD5FOoUUQitTQxJTR%2BkPxu9exenf0EmI73R3IufoyKDRc9t0%2BtoPC1k0xmI%2FeVbqRA7O34UhIUwrQ70rENX7DFngzJH60f6jmxlqJAVwfLy02uDm%2BfzRJZR4cRAZ1lgkiVzNZWzdDHKBVYUDWooxxM6K4NvEz5iCfHRZ%2BgEcflm5jDQvI2ltjKxZ5Sc%3D&repl=KCk%3D)

The following code causes an weird "{ emit[T] } is not equal to { emit[T] }" error:
```scala
def fix[T] { one: T => Unit / emit[T] } { stream: => Unit / emit[T] }: Unit = {
  try stream() with emit[T] { value =>
    fix {one} { one(value) }
    // ^ [error]: { emit[T] } is not equal to { emit[T] }
    resume(())
  }
}
```

Fixable by 
```
fix[T] {one} { one(value) }
```

### LINKED ISSUE #1215: Bad error message for requiring a type annotation for `stream::source`
Here's a program written in Effekt with a _particularly_ incomprehensible message:

```scala
def push123(): Unit / emit[Int] = [1, 2, 3].each
def printAll(): Bool / read[Int] = {
  exhaustively { println(do read()) }
  true
}

def main() = {
  // note(jirka): originally was just `source`, then the error is much worse.
  val a = returning::source {push123} {printAll}
  //                                   ^^^^^^^^ { read[A] } is not equal to { read[Int] }
  //                         ^^^^^^^ { emit[A] } is not equal to { emit[Int] }
}
```

This is not the fault of #1194, it also has the same message in 0.48.0.

I think this message is quite bad as the user never wrote an explicit `A` type param to begin with,
and the error doesn't point to the things which are wrong -- shouldn't it point to `source` here instead?

Btw, if you use it the way we usually write it (with a `with`), it triggers the "Effects need to be fully known" message:

```scala
def main() = {
  val a = {
    with source { [1, 2, 3].each };
    //            ^^^^^^^^^^^^^^ Effects need to be fully known, but effect emit[A]'s type parameter(s) A could not be inferred.
    //                           Maybe try annotating them?
    exhaustively { println(do read()) }
    true
  }
}
```

jiribenes:
Resolves #1215
Resolves #866
Resolves #813

No idea if this works in general, but it fixes multiple different bugs 🥳

Before, `unifyEffects` outright rejected programs where effect set 1 != effect set 2,
but this doesn't make a lot of sense if some of the effect sets have yet-unresolved unification variables.
This patch just relaxes the condition, hoping that we catch this later when we convert to `ConcreteEffects` (and when we do the "actual" unification*)

### #1215

Before:
```scala
def push123(): Unit / emit[Int] = [1, 2, 3].each
def printAll(): Bool / read[Int] = {
  exhaustively { println(do read()) }
  true
}

def main() = {
  val a = returning::source {push123} {printAll}
  //                                   ^^^^^^^^ { read[A] } is not equal to { read[Int] }
  //                         ^^^^^^^ { emit[A] } is not equal to { emit[Int] }
}
```

After:

```scala
def push123(): Unit / emit[Int] = [1, 2, 3].each
def printAll(): Bool / read[Int] = {
  exhaustively { println(do read()) }
  true
}

def main() = {
  val a = returning::source {push123} {printAll}
  //      ^ Cannot infer type argument A, maybe consider annotating it?       
}
```

### #866

Before:

```scala
def fix[T] { one: T => Unit / emit[T] } { stream: => Unit / emit[T] }: Unit = {
  try stream() with emit[T] { value =>
    fix {one} { one(value) }
    // ^ [error]: { emit[T] } is not equal to { emit[T] }
    resume(())
  }
}
```

After:

\<just works\>

### #813 

Before:

```scala
effect Go[T](): Unit

def dropTill[T2](l: (Int => Int / { Go[T2] } at {})): (Int => Int / { Go[T2] } at {}) = 
  if (false) { dropTill[T2](l) } else { l }
  // error: { Go[T2] } is not equal to { Go[T2] }
```

After:

\<just works\>

### DIFF
--- a/libraries/common/stream.effekt
+++ b/libraries/common/stream.effekt
@@ -44,7 +44,7 @@ def fix[T] { one: T => Unit / emit[T] } { stream: => Unit / emit[T] }: Unit =
   try {
     stream()
   } with emit[T] { value =>
-    fix[T] {one} { one(value) }
+    fix {one} { one(value) }
     resume(())
   }
 


libraries/common/stream.effekt
```diff
@@ -44,7 +44,7 @@ def fix[T] { one: T => Unit / emit[T] } { stream: => Unit / emit[T] }: Unit =
   try {
     stream()
   } with emit[T] { value =>
-    fix[T] {one} { one(value) }
+    fix {one} { one(value) }
```
jiribenes, line 47: This was the problem exactly outlined in #866 requiring an explicit type param as a workaround, now the explicit type param is no longer needed 👀

effekt/shared/src/main/scala/effekt/typer/TypeComparer.scala
```diff
@@ -80,7 +80,15 @@ trait TypeUnifier {
...
-    if (eff1.toList.toSet != eff2.toList.toSet) error(pp"${eff2} is not equal to ${eff1}", ctx)
+    val effs1 = eff1.toList.toSet
+    val effs2 = eff2.toList.toSet
+
+    val hasUnknowns = (effs1 ++ effs2).exists { eff => unknowns(eff).nonEmpty }
+
+    // If we're sure that the effect sets are different, throw an error, otherwise defer it for later.
+    if (effs1 != effs2 && !hasUnknowns) {
+      error(pp"${eff2} is not equal to ${eff1}", ctx)
+    }
```
b-studios, line 91: I am a little bit worried that we will not detect that the two are the same if we do not fail here immediately. But I don't have a concrete counter example.

libraries/common/stream.effekt
b-studios: Great!

b-studios (approved):

PR #1237, opened 2025-12-01
Make pretty-printed core more readable

timsueberkrueb:
In #1181 we made the core pretty printer a lot more verbose in order to enable re-parsing. However, the additional verbosity hinders human inspection and debugging. This PR aims to mitigate this by splitting the core pretty printer in two instances:

* `ReparsablePrettyPrinter` is fully verbose and can be re-parsed with `core.Parser`
* `HumanReadablePrettyPrinter` is less verbose (and thus more readable) but cannot be re-parsed

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
+++ b/effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
@@ -9,7 +9,7 @@ import kiama.output.PrettyPrinterTypes.Document
 import scala.language.implicitConversions
 import effekt.symbols.{ Name, Wildcard, builtins }
 
-class PrettyPrinter(printDetails: Boolean) extends ParenPrettyPrinter {
+class PrettyPrinter(printDetails: Boolean, printInternalIds: Boolean = true) extends ParenPrettyPrinter {
   override val defaultIndent = 2
 
   def format(t: ModuleDecl): Document =
@@ -124,7 +124,7 @@ class PrettyPrinter(printDetails: Boolean) extends ParenPrettyPrinter {
     // In reparsable mode, we just show the string part, which should be freshened by the TestRenamer before printing.
     // The TestRenamer does not rename the Barendregt id because that would violate the internal invariant of having
     // just a single global Barendregt namespace.
-    builtins.coreBuiltinSymbolToString(s).getOrElse(if printDetails then s.name.name else s.show)
+    builtins.coreBuiltinSymbolToString(s).getOrElse(if printInternalIds then s.show else s.name.name)
   }
 
   def toDoc(e: Expr): Doc = e match {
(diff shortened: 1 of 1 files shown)


effekt/shared/src/main/scala/effekt/core/TestRenamer.scala
```diff
@@ -31,14 +31,26 @@ class TestRenamer(names: Names = Names(Map.empty), prefix: String = "_") extends
...
+    // This assumes that user-provided names to not include `prefix`.
+    val userName = if (id.name.name.contains(prefix)) {
+      id.name.name.substring(0, id.name.name.lastIndexOf(prefix))
+    } else {
+      id.name.name
+    }
+    // For many purposes, we would like to include the user-annotated name in the fresh id,
+    // so that printed terms are more readable.
+    // However, we do not want to include it when checking for alpha-equivalence in tests.
+    val userPart = if preserveUserAnnotatedPrefix then userName else ""
```
timsueberkrueb, line 45: We can probably get rid of this by moving the `userName` logic to `idFor`. We need to be careful with this, though, as `idFor` is also used by `core.Parser`.

effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
```diff
@@ -341,3 +339,18 @@ object PrettyPrinter extends ParenPrettyPrinter {
...
+/**
+ * Instance of PrettyPrinter that produces output that can be parsed back by the core parser.
+ */
+object ReparsablePrettyPrinter extends PrettyPrinter(false) {}
+
+/**
+ * Instance of PrettyPrinter that produces less verbose, more human-readable output.
+ */
+object HumanReadablePrettyPrinter extends PrettyPrinter(true) {
+  def renameAndFormat(t: ModuleDecl): Document = {
```
timsueberkrueb: There is one critical caveat with this: this only prints the core module modulo alpha conversion, which is perhaps not what we want. Especially for debugging, we probably want to see the exact Barendregt ids.
Sadly, this means adding another flag and conditional to the pretty printer.

marvinborner:
You may want to use `--debug` instead of `--detailed-ir`, as we do for more debug information in the LLVM IR.

timsueberkrueb:
Also remove some of the captures currently printed by default. It would also be good to highlight `get` and `put` as keywords in the VSCode plugin.

timsueberkrueb:
Corresponding VSCode extension PR: https://github.com/effekt-lang/effekt-vscode/pull/163

jiribenes:
Q from the meeting: does `util.show(...)` show the ID in addition to the string?

timsueberkrueb:
I just checked: it always prints some ids:

* If `--debug` is **not** set, it prints the Barendregt ids actually used internally in the compiler
* If `--debug` is set, it runs in the reparsable mode (`printDetailed`, currently conflated with the debug mode) and prints a deterministic id starting from zero.

```
  def toDoc(s: symbols.Symbol): Doc = {
    // In human-readable mode, we show the name together with the actual Barendregt id.
    // This allows the user to connect the symbol to the internal representation when debugging.
    // In reparsable mode, we just show the string part, which should be freshened by the TestRenamer before printing.
    // The TestRenamer does not rename the Barendregt id because that would violate the internal invariant of having
    // just a single global Barendregt namespace.
    builtins.coreBuiltinSymbolToString(s).getOrElse(if printDetails then s.name.name else s.show)
  }
```

I can change this behavior to always print the internal Barendregt ids except when deterministic ids are needed for test cases.

PR #1239, opened 2025-12-02
Remove redundant transform in core.Transformer

jiribenes:
> [!CAUTION]
> although #1219 has been merged, this is kept alive as it's a deterministic reproducer of https://github.com/effekt-lang/effekt/issues/1240

Cherry-picked from #1219


jiribenes:
I think this is deprecated by #1219 getting merged.
However, it's a deterministic reproducer of #1240, so I won't close it yet.

PR #1243, opened 2025-12-03
Partial functions in standard library

phischu:


libraries/common/partial.effekt
```diff
@@ -0,0 +1,36 @@
...
+
+def optionally[R] { program: () => R / fail }: Option[R] =
+  try {
+    Some(program())
+  } with fail {
+    None()
+  }
+
+/// Run `program` forever until `fail` is thrown.
+def successfully { program: () => Unit / fail }: Unit =
```
marzipankaiser: Feel free to ignore, but since I just stumbled over this: I think the name is not really fitting, we're executing it until it *isn't* successful, right?
Maybe `untilFail` or `whileSuccessful` (or some other name) would be clearer?

jiribenes: dtto ↑, when I gave it to the students, I called it `untilFailDo` which isn't that great, but the spirit is there 😆

phischu:
@timsueberkrueb I am seeing null pointer exceptions in the core reparse tests, is this known? I added a module to the prelude, perhaps I need to add it somewhere else too?

PR #1244, opened 2025-12-03
Add more streaming combinators

jiribenes:
These are the utilities I wanted to find in 'stream' (and related modules) when I was doing AoC'24 and '25:
- `stream::countIf`
- `stream::takeWhile`
- `stream::all`
- `stream::any`
- `list::tails`

Each one has a test and a doc comment.
I tried very hard to add only a single variant of each function so that we don't have too many single-use functions.
(I also have `countTrue`, `all`, and `any` that work on boolean streams directly + some variants that work on pull streams)

### DIFF
--- a/libraries/common/list.effekt
+++ b/libraries/common/list.effekt
@@ -956,7 +956,7 @@ def each[A](list: List[A]): Unit / emit[A] =
 /// ```
 def tails[A](list: List[A]): Unit / emit[List[A]] =
   list match {
-    case Nil() => ()
+    case Nil()            => do emit(Nil())
     case Cons(head, tail) => do emit(list); tails(tail)
   }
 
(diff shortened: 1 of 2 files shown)


libraries/common/list.effekt
```diff
@@ -944,6 +944,22 @@ def each[A](list: List[A]): Unit / emit[A] =
...
+/// ```effekt
+/// > list::collect[List[Int]] { [1, 2, 3].tails }
+/// [[1, 2, 3], [2, 3], [3], []]
+///
+/// > list::collect[List[Int]] { [].tails }
+/// [[]]
+/// ```
+def tails[A](list: List[A]): Unit / emit[List[A]] =
+  list match {
+    case Nil() => ()
```
phischu: Shouldn't this emit an empty list in the nil case?

phischu (approved):

libraries/common/list.effekt
jiribenes: Sure, why not.

jiribenes:
Thanks for the quick review! :)

PR #1246, opened 2025-12-04
Rename exhaustively to loop

phischu:


jiribenes:
Here's the current users of `.continue` and `.break`:
https://github.com/search?q=NOT+owner%3Aeffekt-lang+NOT+owner%3Ase-tuebingen+NOT+is%3Afork+path%3A*.effekt+%2F%5C.continue%2F+OR+%2F%5C.break%2F&type=code

if I'm looking correctly, only one or two are used with numerical `each`, all others are in some `collection::foreach` construct

phischu:
Very good, thank you. How can we test the snippets on the website?

phischu:
Rename `Control` to `Label`

phischu:
#985 is also relevant here

PR #1248, opened 2025-12-05
Experiment/uniform representation

phischu:
Here we see what happens when we use the same representation for all values in the LLVM backend.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
@@ -199,7 +199,15 @@ object Transformer {
 
                 // Known Jump
                 case BlockInfo.Definition(freeParams, blockParams) =>
-                  Jump(Label(transform(id), blockParams ++ freeParams), values ++ blocks ++ freeParams)
+                  // TODO properly distinguish between async calls and jumps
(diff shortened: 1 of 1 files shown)


effekt/shared/src/main/scala/effekt/core/Type.scala
```diff
@@ -408,10 +408,13 @@ object Type {
         Typing(result, armCapt, armFree ++ Free.make(id, sc.tpe.asInstanceOf[ValueType.Data], tparams.map(id => ValueType.Var(id)), vparams))
       }
 
+      // TODO assert that scrutinee actually has type being matched on
       def join(typings: List[Typing[ValueType]], annotated: ValueType): Typing[ValueType] =
         fold(typings, annotated) { (tpe1, tpe2) => valueShouldEqual(tpe1, tpe2); tpe1 }
 
-      join(clauseTypings ++ default.toList.map { stmt => stmt.typing }, annotatedTpe)
+      val Typing(tpe, capt, free) = join(clauseTypings ++ default.toList.map { stmt => stmt.typing }, annotatedTpe)
+      Typing(tpe, scCapt ++ capt, scFree ++ free)
```
b-studios, line 416: Thanks, I missed this!

Issue #462, opened 2024-05-21
`x!=y` interpreted as `x! = y` due to `!` being allowed in identifiers


jiribenes:
As discovered by @marzipankaiser, right now, `x!=y` without spaces gets parsed as:

    `x!` (identifier), `=` (binary operation assign), `y` (identifier)

whereas I'd expect it to be the same as before:

    `x` (identifier), `!=` (binary operation not equal to), `y` (identifier)

This is caused by supporting `!` in identifiers introduced in https://github.com/effekt-lang/effekt/commit/e8a14778f99524601aa8bcd62541aa46529241a0.

marzipankaiser:
Note, this is still the case on current master.

Also, similarly: `n-1` lexes as `n` `-1` (not as `n` `-` `1`) and fails during parsing.

b-studios:
Just tagging @dvdvgt since this might be relevant for him.

b-studios:
To be honest, I think the latter (`n-1` not parsing) is a feature, not a bug :) I am happy with nudging people to use whitespace.

dvdvgt:
I fail to see what we can do about it to completely avoid this issue:
- Either disallow `!` occurring in identifiers,
- force the user to add a space
- or have it the other way around such that there may be a falsely lexed token identifier where it is actually a `!=`.

How else should the lexer know that `val n!=1` contains the identifier `n!` and and the token `=` but not `!=` instead without taking the context into consideration?

dvdvgt:
> Also, similarly: `n-1` lexes as `n` `-1` (not as `n` `-` `1`) and fails during parsing.

This should be fixable by allowing the lexer only to lex unsigned numbers and let parser figure out the precedence.

b-studios:
As I said, I wouldn't want to fix this. People should use whitespaces if they mean `n - 1`.

dvdvgt:
As much as I like the frequent and consistent use whitespaces to structure code, I disagree. I don't think arithmetics is the right place to force this onto the user. Parsing should be robust enough to handle something like `1-1` (instead of `1 - 1`) without aborting the whole compilation.

marzipankaiser:
> As I said, I wouldn't want to fix this. People should use whitespaces if they mean `n - 1`.

I'm somewhat fine with this (though I sometimes do prefer the version without whitespace), but then we should be consistent and also disallow `n- 1` (which is parsing fine).

Issue #1183, opened 2025-11-05
Inconsistent handling of line endings on Windows vs Linux


Plixo2:
Currently 35 tests on Windows fail. 

Git converts all line endings of text to CRLF on Windows (default git behavior). 
Using the git config `core.autocrlf=false` will fix most of the tests. 
Adding a .gitattributes with `* text eol=lf` should work for everyone, although it would hide the underlying issue.

Many errors stem from some incorrect handling in the Lexer, e.g. here:
https://github.com/effekt-lang/effekt/blob/a4c916fa3506db9899bc4387bdba14c8fee3578f/effekt/shared/src/main/scala/effekt/Lexer.scala#L627

Other errors come from string literal comparisons that fail, since the tests only expects '\n'.

As stated, there are a total of 35 tests that are failing and i am willing to share more insights and even fix them, when needed.

PR #1249, opened 2025-12-05
Improve whitespace handling in Lexer & Parser, warn on bad whitespace around binops

jiribenes:
Resolves https://github.com/effekt-lang/effekt/issues/462

The parser soft fails (fails with recovery) if it encounters a binary operator that is not surrounded by some whitespace:
```scala
// OK
val _ = 1 + 2
val _ = 1  + 2
val _ = 1 +  2
val _ = 1 +
  2

// FAIL
val _ = 1 +2
val _ = 1+ 2
val _ = 1+2
```

The implementation is a bit hacky, see the note in the comment.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Parser.scala
+++ b/effekt/shared/src/main/scala/effekt/Parser.scala
@@ -82,7 +82,10 @@ class Parser(tokens: Seq[Token], source: Source) {
 
       // Report recoverable diagnostics
       reportRecoverableDiagnostics()
-      if recoverableDiagnostics.isEmpty then res else None
+      // Don't fret unless we had a recoverable error
+      if recoverableDiagnostics.forall { d =>
+        d.kind != kiama.util.Severities.Error
+      } then res else None
     } catch {
       case Fail(msg, pos) =>
         // Don't forget soft fails!
--- a/examples/neg/parsing/bin_op_spaces_recovers.check
+++ b/examples/neg/parsing/bin_op_spaces_recovers.check
@@ -1,0 +1,22 @@
+[error] examples/neg/parsing/bin_op_spaces_recovers.effekt:4:38: Cannot typecheck call.
+There are multiple overloads, which all fail to check:
+
+Possible overload: effekt::infixAdd of type (Int, Int) => Int
+  Expected type
+    JustCauseATypeErrorPlease
+  but got type
+    Int
+
+Possible overload: effekt::infixAdd of type (Double, Double) => Double
+  Expected type
+    JustCauseATypeErrorPlease
+  but got type
+    Double
+  Expected Double but got Int.
+  Expected Double but got Int.
+
+  val _: JustCauseATypeErrorPlease = 1 +2
+                                     ^^^^
+[warning] examples/neg/parsing/bin_op_spaces_recovers.effekt:4:40: Missing whitespace around binary operator
+  val _: JustCauseATypeErrorPlease = 1 +2
+                                       ^
--- a/examples/neg/parsing/bin_op_spaces_recovers.effekt
+++ b/examples/neg/parsing/bin_op_spaces_recovers.effekt
@@ -1,0 +1,5 @@
+// this example causes a type error, showing that the parser recovers correctly
+record JustCauseATypeErrorPlease()
+def main() = {
+  val _: JustCauseATypeErrorPlease = 1 +2
+}
--- a/examples/pos/parser_warnings_recover_completely.check
+++ b/examples/pos/parser_warnings_recover_completely.check
@@ -1,0 +1,4 @@
+2
+[warning] examples/pos/parser_warnings_recover_completely.effekt:2:24: Missing whitespace around binary operator
+def main() = println(1 +1)
+                       ^
--- a/examples/pos/parser_warnings_recover_completely.effekt
+++ b/examples/pos/parser_warnings_recover_completely.effekt
@@ -1,0 +1,2 @@
+// the following code should still print 2
+def main() = println(1 +1)


jiribenes:
Follow-up: try and see if we can make Parser more aware of the whitespace tokens that it throws out now

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -141,11 +142,19 @@ class Parser(tokens: Seq[Token], source: Source) {
...
+      if position < 0 then fail("Unexpected start of file")
+
+      tokens(position).failOnErrorToken(position) match {
+        case token if isSpace(token.kind) && token.kind != Newline => go(position - 1)
+        case token if token.kind == Newline => true
+        case _ => false
+      }
+
+    go(position - 1)
+  }
```
jiribenes, line 162: Previously we did `lookbehind(1) == Newline`, but that no longer works, the `lookbehind` would need to ignore all whitespace tokens _except_ for a newline... I found that to be too weird, so I replaced it with a specialised function to just check if the last non-space token was a newline (feel free to suggest a better name)

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -1053,6 +1063,17 @@ class Parser(tokens: Seq[Token], source: Source) {
...
+  // Check that the current token is surrounded by whitespace. If not, soft fail.
+  private def checkBinaryOpWhitespace(): Unit = {
+    // position points to the operator token in the raw token array
+    val wsBefore = position     > 0             && isSpace(tokens(position - 1).kind)
+    val wsAfter  = position + 1 < tokens.length && isSpace(tokens(position + 1).kind)
+
+    if (!wsBefore || !wsAfter) {
+      softFail(s"Missing whitespace around binary operator", position, position)
+    }
+  }
```
jiribenes, line 1079: Now I'm very happy about the implementation :)

effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -303,11 +303,11 @@ class Lexer(source: Source) extends Iterator[Token] {
...
       currentChar match {
-        case ' ' | '\t' => advance()
+        case ' ' | '\t' => return // Stop here, let spaces be handled as a token
         case '\n' => return // Stop here, let newline be handled as a token
         case '\r' =>
           if nextChar == '\n' then return // Stop here for \r\n
-          else advance() // Treat standalone \r as whitespace
+          else advance() // Skip standalone \r
         case _ => return
       }
```
jiribenes: This function is now somewhat weird, it only handles `\r\n` (weirdly to beat), perhaps this is more an issue for #1183.

effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -303,11 +303,11 @@ class Lexer(source: Source) extends Iterator[Token] {
   private def skipWhitespace(): Unit =
```
dvdvgt: Should this be called `skipNewline` instead maybe?

effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -380,6 +385,8 @@ class Lexer(source: Source) extends Iterator[Token] {
     }
 
     (currentChar, nextChar) match {
+      case (' ',     _) => advanceSpaces()
+      case ('\t',    _) => advanceSpaces()
       case ('\n',    _) => advanceWith(TokenKind.Newline)
       case ('\r', '\n') => advance2With(TokenKind.Newline)
```
dvdvgt: I am a bit confused now when newlines and spaces are skipped and when they are emitted. `next` calls `skipWhitespace` but also not always. Perhaps a brief comment on whitespace handling would be nice?

effekt/shared/src/main/scala/effekt/Lexer.scala
jiribenes: I removed it outright, it's not needed anymore, I think.

effekt/shared/src/main/scala/effekt/Lexer.scala
jiribenes: I redid this part, whitespace ought to be always emitted. :)

dvdvgt: Nice, seems reasonable

jiribenes:
I have changed it to only emit a warning :)

jiribenes:
cc @marvinborner: do your feelings change now that it's merely a warning, and one that doesn't crash, allowing you to even run a program with parser warnings? :) [see tests]

effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -1053,6 +1068,16 @@ class Parser(tokens: Seq[Token], source: Source) {
...
 
+  // Check that the current token is surrounded by whitespace. If not, soft fail.
+  private def checkBinaryOpWhitespace(): Unit = {
+    val wsBefore = position     > 0             && isSpace(tokens(position - 1).kind)
+    val wsAfter  = position + 1 < tokens.length && isSpace(tokens(position + 1).kind)
+
+    if (!wsBefore || !wsAfter) {
+      warn(s"Missing whitespace around binary operator", position, position)
+    }
+  }
```
jiribenes, line 1079: We could ever-so-slightly generalise this:
```scala
  // Require that the current token is surrounded by whitespace. If not, soft fail.
  private def requireSpacesAround(): Unit = {
    val wsBefore = position     > 0             && isSpace(tokens(position - 1).kind)
    val wsAfter  = position + 1 < tokens.length && isSpace(tokens(position + 1).kind)

    if (!wsBefore || !wsAfter) {
      warn(s"Missing whitespace around '${TokenKind.explain(peek.kind)}'", position, position)
    }
  }
```

and then use it for other places where we really want spaces on both sides (like the `=` in `val foo = ...` / `def foo = ...` / ...) :)

marvinborner:
Fine by me, sure. In the future we could maybe add flags to hide warnings, including such parser warnings, if desired.

PR #1251, opened 2025-12-08
Parse all definitions even locally, distinguish top-level-only in Namer

### LINKED ISSUE #1191: Compilation error "Only allowed on the toplevel" with unhelpful error message and no LSP support
When compiling:
```effekt
def bar() = {
  type Foo {
    F(block: () => Foo at {})
  }
}
```
I get no LSP errors, but a compilation error with strange line numbers:
```
PlainTextError(Only allowed on the toplevel,Some(Range(Position(7,28,VirtualFileSource(module0.effekt)),Position(7,32,VirtualFileSource(module0.effekt)))),Error)
```
Note that a similar case of using `record` or `interface` instead of `type` has a proper error message and LSP support.

jiribenes:
Resolves #1191

From now on, Namer is to check if a given defn is allowed at top-level.
This also allows us to get a bit further in the pipeline, not everything has to fail in the parser.
The grammar, and hence the parser, get a bit simpler.

I believe this is morally correct since scoping is the task of the Namer, not the Parser.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Namer.scala
+++ b/effekt/shared/src/main/scala/effekt/Namer.scala
@@ -124,10 +124,10 @@ object Namer extends Phase[Parsed, NameResolved] {
       ()
 
     case d @ source.VarDef(id, annot, binding, doc, span) =>
-      Context.requireNotToplevel("Variable")
+      Context.requireNotToplevel("Mutable variable")
 
     case d @ source.RegDef(id, annot, region, binding, doc, span) =>
-      Context.requireNotToplevel("Variable-in-a-region")
+      Context.requireNotToplevel("Mutable variable in a region")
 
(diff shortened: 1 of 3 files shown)


examples/neg/namer/local_interface.check
```diff
@@ -0,0 +1,3 @@
+[error] examples/neg/namer/local_interface.effekt:4:8: Could not resolve type Greet
+} with Greet {
+       ^^^^^
```
jiribenes, line 3: I'm not too happy about this error for:
```
def main() = try {
  interface Greet { def sayHello(): Unit }
  do sayHello()
} with Greet {
  def sayHello() = { println("Hello!"); resume(()) }
}
```

Can we somehow also report the bad interface declaration here in a principled way?

PR #1254, opened 2025-12-11
Disallow block parameters on extern forever

marvinborner:
Previously there was an error message stating that block parameters are _currently_ not supported, but they most probably never will be (@phischu)

### DIFF
--- a/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
+++ b/effekt/shared/src/main/scala/effekt/machine/Transformer.scala
@@ -53,7 +53,7 @@ object Transformer {
 
   def transform(extern: core.Extern)(using BlocksParamsContext, ErrorReporter): Declaration = extern match {
     case core.Extern.Def(name, tps, cparams, vparams, bparams, ret, capture, body) =>
-      if bparams.nonEmpty then ErrorReporter.abort("Foreign functions currently cannot take block arguments.")
+      if bparams.nonEmpty then ErrorReporter.abort("Foreign functions cannot take block arguments.")
 
       val transformedParams = vparams.map(transform)
       noteDefinition(name, transformedParams, Nil)


b-studios:
We should also change the source and core representation. However, it would be good to do this after merging #1219

marzipankaiser:
Note that there are specific use cases for external block parameters:
- https://github.com/effekt-lang/effekt/pull/1020/files#diff-95b6afb9b993424af9c118c39adf1b71ab6cc1f4dd061358b4a19659e450a99aR575 uses this for the `isJSNode` etc functions.

That is why I fixed the captures for them in https://github.com/effekt-lang/effekt/issues/448 .
This *does work* on LLVM, because they are compiled away before reaching machine.

marzipankaiser:
Otherwise, is there a reason they won't be supported? Apart from having to pass them through?
As in every extern fn, the llvm side would be required to handle the calling convention correctly in any case and it can just be a `Neg`, no?

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -53,7 +53,7 @@ object Transformer {
 
   def transform(extern: core.Extern)(using BlocksParamsContext, ErrorReporter): Declaration = extern match {
     case core.Extern.Def(name, tps, cparams, vparams, bparams, ret, capture, body) =>
-      if bparams.nonEmpty then ErrorReporter.abort("Foreign functions currently cannot take block arguments.")
+      if bparams.nonEmpty then ErrorReporter.abort("Foreign functions cannot take block arguments.")
```
marzipankaiser, line 56: Given that this error message will only show up on LLVM (i.e. machine-based backends), maybe we'd want to specify this in the error message?

phischu:
We can also just support it, if you tell me what the semantics should be.

marzipankaiser:
@phischu I would assume that the block parameters are also passed as parameters (of type `%Neg` probably?) and the foreign llvm code then would have to make sure to use them correctly (it is ffi after all).
That is, the llvm side is required to use the correct calling convention etc, as already is the case when passed a boxed function.
We might need to allow `${box b}` in llvm, or allow `${b}` for block variables `b` for this to be usable in LLVM (I have the latter stashed away somewhere, and think it is reasonable).

phischu:
Beware, this will probably be subsumed by #1298

PR #1258, opened 2025-12-15
Fix printing of literals with escape sequences in core

timsueberkrueb:
Previously, translating the following source program to core:

```
def main() = println("\"\ta\n\ra\"")
```

and printing it, led to the following output

```
def main() = {
    let ! v_r = println: (String) => Unit @ {io}("""" a
a"""")
    return v_r: Unit
}
```

which tripped the core parser because it contains unescaped quotes immediately preceding the multiline end string literal marker.

This parser code cannot handle this because (presumably) it eagerly consumes the first three quotes at the end of the string literal, leaving an orphan `"` behind:

```scala
  // Delimiter for multiline strings
  val multi = "\"\"\""

  // multi-line strings `(?s)` sets multi-line mode.
  lazy val multilineString: P[String] = regex(s"(?s)${ multi }[\t\n\r ]*(.*?)[\t\n\r ]*${ multi }".r) ^^ {
    contents => contents.strip().stripPrefix(multi).stripSuffix(multi)
  }
```

This PR currently contains a conservative fix: we just print single-line string literals with escape sequences whenever the string contains "problematic" characters.

Other suggestions are welcome!

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/Parser.scala
+++ b/effekt/shared/src/main/scala/effekt/core/Parser.scala
@@ -170,7 +170,9 @@ class EffektLexers extends Parsers {
           case '"'  => sb.append('"');  i += 2
           case 'r'  => sb.append('\r'); i += 2
           case 't'  => sb.append('\t'); i += 2
+          case 'n'  => sb.append('\n'); i += 2
           case other =>
+            sb.append('\\')
             sb.append(other)
             i += 2
         }
--- a/effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
+++ b/effekt/shared/src/main/scala/effekt/core/PrettyPrinter.scala
@@ -333,12 +333,18 @@ object PrettyPrinter extends ParenPrettyPrinter {
       case '"'  => "\\\""
       case '\r' => "\\r"
       case '\t' => "\\t"
+      case '\n' => "\\n"
       case c    => c.toString
     }
   }
 
+  // If we encounter any of the following characters, we print a single-line string literal with escape sequences
+  // instead of a multi-line string literal.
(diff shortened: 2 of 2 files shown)


marvinborner:
This seems kind of unstable, especially because I've been missing further escape sequences anyway (at least [the ones supported by JS](https://en.wikipedia.org/wiki/Escape_character#JavaScript)). Now there are many places we'd have to modify/test (and again in some backends). Might there be an idiomatic way to implement this?

timsueberkrueb:
> This seems kind of unstable

Can you clarify? Are you worried of the source parser getting out-of-sync with the core parser wrt. the supported escape sequences or that not all escape codes needed in extern strings are expressible?

marvinborner:
> Are you worried of the source parser getting out-of-sync with the core parser wrt. the supported escape sequences

Yes, mostly this. Not a big deal (and a bit off-topic), but we hardcode these escape sequences a lot. See also

https://github.com/effekt-lang/effekt/blob/04220781176d2cacb4c3c7ce539a472c149a51a8/effekt/shared/src/main/scala/effekt/util/package.scala#L10

https://github.com/effekt-lang/effekt/blob/master/effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala#L786

https://github.com/effekt-lang/effekt/blob/04220781176d2cacb4c3c7ce539a472c149a51a8/effekt/shared/src/main/scala/effekt/generator/chez/Transformer.scala#L47

https://github.com/effekt-lang/effekt/blob/04220781176d2cacb4c3c7ce539a472c149a51a8/effekt/shared/src/main/scala/effekt/generator/js/Transformer.scala#L21

timsueberkrueb:
I see, let's do that in a separate PR though, since this is just a bug fix.

Issue #1203, opened 2025-11-13
JS: "ReferenceError: tmp_XXX is not defined"


kyay10:
I'll try to minimize the example further, but it seemingly needs a lot of the "box"ing to get it to happen. Seems to be an optimization issue.
```effekt
import option
import dequeue
import list

def basicScheduler{reg: Region}: {((=> Unit at reg) => Unit at reg) => Unit} => Unit at reg = box { {block} =>
  var queue: Dequeue[=> Unit at {reg, block}] in reg = emptyQueue()
  def run(): Unit = queue.popBack match {
    case None() => ()
    case Some((k, q)) =>
      queue = q
      k()
      run()
  }
  block(box { c => queue = queue.pushFront(c) })
  run()
}

interface Suspend[C] {
  def suspend[T](block: (T => C at {}) => C at {}):T
}

def scheduler{reg: Region}: {((=> Unit / Suspend[=> Unit at reg] at reg) => Unit at reg) => Unit} => Unit at reg = box { {block} =>
  (basicScheduler{reg}) { s =>
     def handleSuspend(c: => Unit / Suspend[=> Unit at reg] at reg) = try c() with Suspend[=> Unit at reg] {
       def suspend[T](block2) = block2(box { t => box{resume(t)} })()
     }
     block(box handleSuspend)
  }
}

interface Channel[T, C] {
  def send(value: T): => Unit / Suspend[C] at {}
  def receive: => T / Suspend[C] at {}
}

def send[T, C](value: T){channel: Channel[T, C]} = channel.send(value)()
def receive[T, C]{channel: Channel[T, C]} = channel.receive()()

namespace internal {
  type ChannelQueue[T, C] {
    Receivers(receive: T => C at {}, rest: Dequeue[T => C at {}])
    Senders(senders: Dequeue[(T, C)])
  }

  def asReceivers[T, C](q: Dequeue[T => C at {}]): ChannelQueue[T, C] = q.popBack match {
    case None() => Senders(emptyQueue())
    case Some((receive, rest)) => Receivers(receive, rest)
  }


  def channel[T]{this: Region}: ((=> Unit / Suspend[=> Unit at this] at this) => Unit at this) => (Channel[T, => Unit at this] at {}) at this = box { s =>
    def s2(block: => Unit at this) = s(box {block()})
    var queue: ChannelQueue[T, => Unit at this] in this = Senders(emptyQueue())
    box new Channel[T, => Unit at this] {
[...]

PR #1262, opened 2025-12-16
Remember mapping in the other direction to address #1203

b-studios:
This PR aims to address #1203 where usage information got lost because the input of the renamer didn't follow Barendregt and contained duplicate bindings.

As a consequence, we only remember the last freshened binding occurence and lost usage information. Now I invert the mapping to remember ALL freshened binding occurrences.

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/LLVMTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/LLVMTests.scala
@@ -54,6 +54,10 @@ class LLVMTests extends EffektTests {
 
     // Generic comparison
     examplesDir / "pos" / "issue733.effekt",
+
+    // re-entrant resumption
+    examplesDir / "pos" / "issue1203.effekt",
+    examplesDir / "pos" / "issue1203b.effekt",
   )
 
   override lazy val withoutOptimizations: Set[File] = Set(


b-studios:
The failing test looks like a spurious network error.

jiribenes:
Could we please also add a test for the other variant that doesn't depend on the `random` library as `pos/issue1203b.effekt` or something, that is the following:

```scala
interface Suspend {
  def suspend[T](block: (T => Unit at {}) => Unit at {}): T
}

def handleSuspend(block: => Unit / Suspend at {}) = try block() with Suspend {
  def suspend[T](block2) = block2(box resume)
}

def foo(flag: Bool): Unit = {
  handleSuspend(box { do suspend[=> Unit at {}](box { resume => 
      def r() = resume(box{}) // inlining this makes the bug not trigger
      if (flag) (box r)() else resume(box r)
    })()
  })
}

effect random(): Bool

def main() = {
  var next = box { true }
  next = box { false } // this seems to be necessary in order to trigger the bug [?]
  
  try { foo(do random()) } with random { resume(next()) }
}
```

jiribenes:
Need to ignore issue1203.effekt and issue1203b.effekt on LLVM ("reentrant resumption")

PR #1267, opened 2025-12-17
Fix #603

### LINKED ISSUE #603: Incorrect captures for interface operations with block parameters
In the following example, the captures are annotated as indicated in the comments:
```
interface Bar {
  def bar(): Unit
}
interface Foo {
  def foo(){b: Bar}: Unit
}

// {b}
def baz() = {
  // {b}
  def x = new Foo { 
    def foo(){b} = { b.bar() }
  }
  // {}
  def b2 = new Bar { def bar() = () }
  x.foo{b2}
}

// {}
def foof(){b: Bar} = b.bar()

// {b}
def main() = {
  baz()
}
```

Here, the capture for `x` contains `b`, which is not even in scope outside of the definition of `foo`.
(This is also inconsistent with the captures of `foof`).

timsueberkrueb:


jiribenes:
Before I forget, I'd like to note down that it would be really nice to include _all_ the repros from #603 (so that we can be sure it's resolved no matter which flavour of objects is chosen). 🙏

effekt/jvm/src/test/scala/effekt/JavaScriptTests.scala
```diff
@@ -15,9 +15,9 @@ class JavaScriptTests extends EffektTests {
   def backendName = "js"
 
   override def positives: Set[File] = Set(
-    examplesDir / "pos",
-    examplesDir / "casestudies",
-    examplesDir / "benchmarks",
+//    examplesDir / "pos",
```
b-studios: this is a mistake, no?

timsueberkrueb:
Also flaky `ReparseTest` issue in the CI again.

effekt/shared/src/main/scala/effekt/source/AnnotateCaptures.scala
```diff
@@ -75,7 +75,14 @@ object AnnotateCaptures extends Phase[Typechecked, Typechecked], Query[Unit, Cap
...
 
     case b @ source.BlockLiteral(tps, vps, bps, body, _) =>
-      query(body) -- boundCapabilities(b) -- CaptureSet(bps.map(_.symbol.capture))
+      query(body) -- CaptureSet(bps.map(_.symbol.capture))
+  }
+
+  override def query(h: OpClause)(using Context, Unit) = h match {
+    case source.OpClause(id, tps, vps, bps, ret, body, resume, span) =>
+      // bps also contains any capabilities as this phase is run after ExplicitCapabilities,
+      // so we don't need to special-case them here.
```
b-studios:
```suggestion
```

Issue #1240, opened 2025-12-02
Investigate long running "Internal Tests & Verify Binary" CI


jiribenes:
"Internal Tests & Verify Binary" sometimes takes up to 6 hours [!] before GitHub kills it.
As far as we know, this was added in #1181 

As a quick workaround, we might want to add timeout to PR CI anyways.

But a proper investigation is due here.

jiribenes:
Now timed out _twice_ in https://github.com/effekt-lang/effekt/pull/1239

timsueberkrueb:
As mentioned previously, this started happening in #1181 after adjusting the `build.sbt` file to separate the core reparse tests. It also happened to me locally, but non-deterministically. This is going to be hard to debug. I personally suspect a bug somewhere in our tooling stack.

PR #1268, opened 2025-12-17
Workaround poor reparse test performance and #1240

timsueberkrueb:
There are two issues related to the core reparse tests that currently slow down our development:

1. They take a long time to execute because the core parser is very slow. The full test suite currently takes ~13m in CI 
2. Since separating out the reparse tests from the rest of the test suite we sometimes see long-running (likely diverging) test execution (see #1240). The exact cause is still unclear, but I hypothesize the issue will disappear by reverting this change.

This PR

* runs reparse tests only on the "benchmarks" subset of the full test suite
* reverts the separation of the reparse test from the remaining internal tests

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -61,24 +61,6 @@ jobs:
       - name: Test effekt binary
         run: effekt.sh --help
 
-  # These are currently separated as they take a long time to run.
-  core-reparse-tests:
-    name: "Core Reparse Tests"
-    needs: build-and-compile
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v6
-        with:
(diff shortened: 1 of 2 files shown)


timsueberkrueb:
With this PR, runtime for the internal tests still increases from ~3.5m to 7m. I could select an even smaller subset if you like.

jiribenes:
Screenshot 2025-12-17 at 15 35 28
I just noticed this in #1170 cc @timsueberkrueb

jiribenes:
Screenshot 2025-12-17 at 15 36 51


... oh, `separateTests` has `"effekt.StdlibChezCPSTests",`
but it's stdlibchez**SCHEME**cpstests.

=> https://github.com/effekt-lang/effekt/pull/1274

PR #1269, opened 2025-12-17
Add chez-cps benchmark config

marvinborner:
see https://github.com/effekt-lang/effekt-plots/pull/12


phischu (approved):
Let's use those.

PR #1270, opened 2025-12-17
Remove indirection of partial functions from core rewrites

b-studios:


b-studios:
Should I rename rewrite for blocklits to rewriteBlocklit and similar before merging?

PR #1271, opened 2025-12-17
Add Euclidean-based `div` and `mod`

dvdvgt:
Since this is based on effekt definitions and not FFI, this has the added side-effect of also unifying the behavior across all backends.

### DIFF
--- a/libraries/common/effekt.effekt
+++ b/libraries/common/effekt.effekt
@@ -319,6 +319,12 @@ extern def infixNeq(x: Bool, y: Bool) at {}: Bool =
 
 // Math ops
 // ========
+
+// Numeric representations in the backends:
+// LLVM: int64_t (64-bit integer)
+// JS: 64-bit floating-point (64-bit IEEE 754)
+// Chez: 64-bit integer
+
 extern def infixAdd(x: Int, y: Int) at {}: Int =
   js "(${x} + ${y})"
   chez "(+ ${x} ${y})"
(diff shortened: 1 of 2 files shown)


dvdvgt:
All backends except LLVM used to use floor-based integer division prior to this PR such that `8 / -3 = -3` and not `8 / -3 = -2`. Thus, I changed the `infixDiv` implementations for JS and Chez to make it consistent with LLVM

libraries/common/effekt.effekt
```diff
@@ -343,11 +343,27 @@ extern def infixSub(x: Int, y: Int) at {}: Int =
...
+/// Computes the remainder based on truncated-division
+ extern def rem(x: Int, y: Int) at {}: Int =
+   llvm "%z = srem %Int ${x}, ${y} ret %Int %z"
+   default { x - y * (x / y) }
+
+/// Computes the result of Euclidean division
+def div(x: Int, y: Int): Int = {
+  val q = x / y
+  val r = x.rem(y)
+  q - (r.bitwiseShr(31).bitwiseAnd(y.bitwiseShr(31).bitwiseOr(1)))
```
jiribenes: Stupid question: Will this work on LLVM where `Int` is i64 as opposed to f64 (~i32 for the purposes of bin ops) like in JS?

dvdvgt: It appears to be working and yielding consistent results with the remaining backends. I'll do some further testing, though. Good point, thanks!

phischu:
Please add a fast implementation of `mod` for llvm and other platforms where this is possible:
```
extern def mod(x: Int, y: Int): Int at {} =
  llvm " ... "
  default { current code }
```

PR #1274, opened 2025-12-17
Fix wrong entry in test suite

jiribenes:
Follow-up from #1268:

Screenshot 2025-12-17 at 15 36 51

`separateTests` has `"effekt.StdlibChezCPSTests",` but it's stdlibchez**SCHEME**cpstests.


jiribenes:
@timsueberkrueb this makes Reparser tests fail :(

Issue #1257, opened 2025-12-15
Delete `effekt.sh`


phischu:
Since we now produce an `effekt` executable, which works on all platforms, there is no need to also produce `effekt.sh`.

jiribenes:
This would entail "just" deleting the `bin/effekt.sh` file, right?

https://github.com/effekt-lang/effekt/blob/master/bin/effekt.sh

marvinborner:
We should also remove it from `package.json` and some CI files.

https://github.com/search?q=repo%3Aeffekt-lang%2Feffekt%20effekt.sh&type=code

We also explain it on the website

https://github.com/search?q=repo%3Aeffekt-lang%2Feffekt-website%20effekt.sh&type=code

phischu:
In passing, we could also decrease the jvm stack size again.

jiribenes:
> In passing, we could also decrease the jvm stack size again.

I tried to do so here, no results yet: https://github.com/effekt-lang/effekt/pull/1266

PR #1278, opened 2025-12-19
Add build.mill setup

timsueberkrueb:
This PR adds an initial [mill](https://mill-build.org/) setup. It is intended for evaluating `mill` as an alternative to `sbt` for `effekt`. We run a subset of the test suite using `mill` instead of `sbt`. This PR does not remove any of the existing config and touches no deployment logic. If we decide against `mill`, we will revert this PR.

Cheat sheet for sbt users:

* `sbt clean` -> `mill clean`
* `sbt 'project effektJVM; compile'` -> `mill project.testCompile`
* `sbt 'project effektJVM; install'` -> `mill project.install`
* `sbt 'project effektJVM; testOnly'` -> `mill project.test`

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -1,3 +1,5 @@
+//| mill-version: 1.0.6
+
 // Building Effekt is managed through Mill (https://com-lihaoyi.github.io/mill/).
 // 
 // To compile the project, targeting JVM:


jiribenes:
The CI says:
> No mill version specified.
> You should provide a version via a '//| mill-version: ' comment or a '.mill-version' file.

timsueberkrueb:
Closes #1257.

PR #1280, opened 2026-01-03
Allow 'is' and 'and' as ordinary expressions

### LINKED ISSUE #481: [To be discussed] Allow match guards as boolean expressions

jiribenes:
Resolves #481

`foo is bar` and `true and false` are now parseable expressions on their own, so the example from #481 should now work:
```
while(not(cmd is Some("quit"))) {
  cmd = readCmd()
}
```

A small downside is that we don't want `foo is Some(x)` to _actually_ bind stuff, instead emitting a warning
Also, I haven't really tested this (merely ran tests recommended by Claude), but I think it shouldn't break existing code; we'll see if that's the case after the CI runs.

As a follow-up, I'd also like to add `or` as a keyword for consistency's sake (and then possibly deprecate `&&` and `||` altogether). In addition to that, this makes `is` into a proper keyword instead of a context-sensitive one.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Parser.scala
+++ b/effekt/shared/src/main/scala/effekt/Parser.scala
@@ -1043,7 +1043,7 @@ class Parser(tokens: Seq[Token], source: Source) {
 
   def isExpr(): Term = nonterminal:
     val scrutinee = relExpr()
-    when(`is`) { desugarIsExpr(scrutinee, matchPattern(), span()) } { scrutinee }
+    when(`is`) { desugarIsExpr(scrutinee, position, matchPattern(), span()) } { scrutinee }
 
   def relExpr(): Term = infix(addExpr, `<=`, `>=`, `<`, `>`)
   def addExpr(): Term = infix(mulExpr, `++`, `+`, `-`)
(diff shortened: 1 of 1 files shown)


jiribenes:
Darn, our parser tests suck. I see a way out of this, but it seems like much more work, so I'll kindly bail :)
(but do ask me if interested in following up)

PR #1282, opened 2026-01-07
Increase JVM stack size for compiler

phischu:
This fixes the stack overflows of the compiler we are seeing.


jiribenes:
This also needs followup changes in effekt-nix used by the community projects.

PR #1284, opened 2026-01-11
Hotfix crash on joining boxes with different capture sets

### LINKED ISSUE #1283: Normalizer crash when boxing functions with different captures
When boxing two functions with different captures inside a conditional like match or if leads to a crash in the normalizer.

```scala
def main(): Unit = {
  val boxed = 1 match {
    case 2 => box { () }
    case _ => box { println("message") }
  }
}
```
or alternatively
```scala
def main(): Unit = {
  val boxed2 = if (1 < 2) {
    box { () }
  } else {
    box { println("message") }
  }
}
```

Lead to the following when executing
```
[error] java.lang.AssertionError: assertion failed: Then and else branch have different types: () => Unit at {} != () => Unit at {io}
[error] 
[error] b_k0_7132()
[error] 
[error] b_k1_7134()
[error] 
[error] if (tmp_9245) {
[error]   b_k0_7132()
[error] } else {
[error]   b_k1_7134()
[error] }
```

jiribenes:
Resolves #1283 by using `Type.equals`


jiribenes:
Since the tests pass, I'll merge this so that it's included in tomorrow's release.
If I understood the issue correctly, in the very worst case, this will "just" not catch something that will later crash. Oh well. :)

PR #1286, opened 2026-01-13
More benchmarks for tasks and channels

phischu:
This benchmark program is faster than the equivalent one in other languages for small inputs when compiled with the llvm backend, but much much slower for large inputs. We suspect there to be a performance bug that makes it have the wrong time complexity. One should investigate, for example by plotting the timing for different input sizes, and comparing to other backends.


phischu:
Also does #1314 
I still believe there is a time complexity bug in our channel and thread implementation, we should investigate.

PR #1292, opened 2026-01-27
Add some utility functions

marvinborner:
Utility functions via my EPE project.

The assertion functions are taken from #923 because it did not get merged.

Maybe we can use this PR to merge some other utility functions that would not otherwise make it into the standard library (e.g. by being stuck in some other PR)?

Also, I'm not sure if we should also add functions such as a nondeterministic stream selection:

```scala
def choose[A] { stream: () => Unit / emit[A] }: A / fail =
  try {
    def body(): A = { stream(); first { stream() } }
    body()
  }
  with emit[A] { a =>
    if (random() > 0.5) a
    else resume(())
  }
```

(there were some other presenters who wanted more nondeterministic functions in stdlib, right?)

### DIFF
--- a/libraries/common/stream.effekt
+++ b/libraries/common/stream.effekt
@@ -90,13 +90,6 @@ def replicate[A](number: Int) { action: () => A }: Unit / emit[A] =
     replicate(number - 1) {action}
   }
 
-def size[A] { s: => Unit / emit[A] }: Int = {
-  var size = 0
-  try s()
-  with emit[A] { _ => size = size + 1}
-  size
-}
-
 /// Creates an infinite iterated stream given by an `initial` seed and a `step` function:
 /// iterate(a){f} ~> a, f(a), f(f(a)), f(f(f(a))), ...
 def iterate[A](initial: A) { step: A => A }: Unit / emit[A] = {
@@ -198,6 +191,10 @@ def countIf[A] { shouldCount: A => Bool } { stream: () => Unit / emit[A] }: Int 
   }
   count
 }
+
+/// Count the number of emitted values.
+def count[A] { stream: () => Unit / emit[A] }: Int =
+  countIf[A] { _ => true } {stream}
 
 def source[A] { stream: () => Unit / emit[A] } { reader: () => Unit / read[A] }: Unit =
   returning::source[A, Unit]{stream}{reader}


b-studios:
The question about nondeterministic functions is related to @phischu's work on #874

effekt/shared/src/main/scala/effekt/core/vm/Builtin.scala
```diff
@@ -19,6 +19,11 @@ def builtin(name: String)(impl: Runtime ?=> List[Value] ~> Value): (String, Buil
 type Builtins = Map[String, Builtin]
 
 lazy val printing: Builtins = Map(
+  builtin("effekt::print(String)") {
```
jiribenes, line 22: I welcome the addition of this function, but I'm not too happy about its name: I'm worried that most our users will be surprised that `print` doesn't act like in Python.
I think we used to have some name like `write` — I'd even be in favour of namespacing it like `stdout::write` 🤔

b-studios: TBH I am fine with `print` vs `println`. Users will use it, be annoyed and then figure it out. LLMs however pose more of a problem, unless they run as an agent. But even then, if they see the prelude they will likely choose `println`.

b-studios: On the other side, it would be nice to have a "low level" streaming API for input and output of the console. There `write` and `read` makes a lot of sense. So for actual TUI applications I wouldn't use `print` but `write`.

marvinborner: Wouldn't LLMs write `print` with the expectation that it behaves like in Scala? At least that is my expectation and confusion appeared multiple times when switching between the languages (I often use multiple `print`without `ln` for debugging)

phischu:
> The question about nondeterministic functions is related to @phischu's work on #874

Yes, I am working on it, please don't add it in this pull request, but this motivates me to speed up.

libraries/common/stream.effekt
```diff
@@ -90,6 +90,13 @@ def replicate[A](number: Int) { action: () => A }: Unit / emit[A] =
     replicate(number - 1) {action}
   }
 
+def size[A] { s: => Unit / emit[A] }: Int = {
+  var size = 0
+  try s()
+  with emit[A] { _ => size = size + 1}
```
phischu: Shouldn't you resume here? Also, I would prefer `stream: () => Unit / emit[A]` but okay, who writes the code, decides the names.

phischu: Hmm, this is related to `countIf`, and indeed I would prefer the name `count` (`size` sounds like it is a finite datum).

libraries/common/list.effekt
```diff
@@ -74,6 +74,16 @@ def tail[A](l: List[A]): List[A] / Exception[MissingValue] = l match {
   case Cons(a, rest) => rest
 }
 
+/// Return all elements of a given list except the last element.
+/// Throws a `MissingValue` exception if it's empty.
+///
+// O(n)
```
jiribenes: missing slash here; I really don't want this function in particular to miss the complexity :D
```suggestion
/// O(n)
```

effekt/shared/src/main/scala/effekt/core/vm/Builtin.scala
jiribenes: Personally, I'm more worried about humans precisely _because of Scala_ -- in 99% of cases, I want `println`, yet tooling suggests `print` first, which always frustrates me. 🫠

PR #1293, opened 2026-01-28
Warn when encountering implicitly interpolated strings

dvdvgt:


effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -1346,12 +1346,20 @@ class Parser(tokens: Seq[Token], source: Source) {
...
         // s"a${x}b${y}" ~> s { do literal("a"); do splice(x); do literal("b"); do splice(y); return () }
-        case id ~ SpannedTemplate(strs, args) =>
-          val target = id.getOrElse(IdRef(Nil, "s", id.span.synthesized))
+        case Maybe(id, range) ~ SpannedTemplate(strs, args) =>
+          val target = id match {
+            case Some(id) => id
+            case None =>
+              val synthSpan = range.synthesized
+              val end = position - 1
+              warn("interpolated string literals need to be explicit marked with an interpolator functions, e.g. 's\"...\"'", start, end)
```
jiribenes: grammar
```suggestion
              warn("interpolated string literals need to be explicitly marked with an interpolator function, e.g. 's\"...\"'", start, end)
```

jiribenes: though personally, I'd prefer a more succinct message here
```suggestion
              warn("String interpolation requires an explicit interpolator, e.g. 's\"...\"'", start, end)
```

PR #1294, opened 2026-02-02
Rename literal effect to write

phischu:


libraries/common/effekt.effekt
```diff
@@ -795,17 +795,18 @@ def locally[R] { p: => R } : R = p()
...
+//     do literal("foo "); do splice(42); do write(" bar "); do splice(43)
 //
 // The stream is wrapped into a handler function, which defaults to `stringbuffer::s`
 //
 //    s { ... }
 //
 // but can be specified by prefixing the string interpolation `custom"..."`.
 
-effect literal(s: String): Unit
+effect write(s: String): Unit
```
b-studios, line 803: Won't this clash with `write(b: Byte)`?

phischu: Where? There is no `write(b: Byte)`. We use `emit[Byte]` and will continue doing so.

phischu:
Okay, this is strange @timsueberkrueb I have restarted CI five times by now but the LSP tests still fail... is this known?

timsueberkrueb:
Unfortunately, this seems like yet another flaky issue. I checked this locally: all tests ran through, so I restarted the CI again.

timsueberkrueb:
I restarted the full CI and the error still persists in the CI, even though I could not reproduce it locally.

phischu:
Interesting, it seems reproducible in CI....

PR #1295, opened 2026-02-20
Delete skip from scanner

phischu:


libraries/common/scanner.effekt
```diff
@@ -2,21 +2,12 @@ module scanner
...
 
-interface Scan[A] {
-  /// Return the next character, not advancing.
-  /// That is, this does not change what any future calls on the Scanner return.
-  def peek(): A / stop
-  /// Advance the Scanner to the next character.
-  def skip(): Unit / stop
-}
 
+effect peek[A](): A / stop
```
jiribenes, line 8: I'd like to keep the existing docs here since they document the invariant pretty well.

```suggestion
/// Return the next character, not advancing.
/// That is, this does not change what any future calls on the Scanner return.
effect peek[A](): A / stop
```

libraries/common/scanner.effekt
```diff
@@ -2,21 +2,12 @@ module scanner
...
 
-/// Advance the Scanner to the next token, returning it.
-def read[A](): A / { Scan[A], stop } = {
-  val t = do peek()
-  do skip()
-  return t
-}
+effect Scan[A] = {read[A], peek[A]}
+
+def skip[A](): Unit / {read[A], stop} = { do read(); () }
```
jiribenes, line 13: dtto
```suggestion
/// Advance the pull stream to the next element, discarding the current element.
def skip[A](): Unit / {read[A], stop} = { do read(); () }
```

PR #1298, opened 2026-02-27
Unify `PureApp`, `ImpureApp`, and `AsyncApp`

phischu:


phischu:
@mattisboeckle I renamed `ImpureApp` to `ExternApp`, deleted `PureApp`, and added a field for purity. Currently there are 173 compile errors. Please fix those that are easy (renaming, etc.) and I will take it from there.

mattisboeckle:
Okay, I've done all the "easy" stuff. Renaming, fixing matches and removing some obsolete stuff. This technically compiles now, but is ***very*** incorrect in most places.
I marked locations that definitely need some thought with "FIXME EXTERNAPP" for easy searching. There will be others as well...

I did try and look into fixing some of my own code ( Show.scala ) and am no longer convinced this is a good idea. 
The translation from ImpureApp => ExternApp works pretty well because it is basically the same thing.
Going from PureApp => ExternApp barely functions as PureApp didn't require most of the data in the ExternApp type and you additionally need to deal with Ids everywhere which does break things

phischu:
Okay, I have done some of the "hard" stuff. Try to fix some of the failing tests and come up with a syntax for pretty printing and bring back the type inference tests that I have commented out.

b-studios:
May I ask why? You probably have good reasons. In the end this is a huge refactoring.

phischu:
We used to secretely have extern async calls be normal `App` and that could lead to bugs and confusion and special casing. This refactoring should delete quite a lot more lines than adding.

b-studios:
But it also makes the "dropping" pass for JS irrepresentible in core

phischu:
Perhaps we can finish this after we have introduced a type `Args` for the triple of `targs`, `vargs`, `bargs` that is so common.

phischu:
Problem is that the following program

```
def main() = {
  println(utf8ByteCount('a'))      // 1
  println(utf8ByteCount('\u0a00')) // 3
}
```

produces the following Core without optimization

```
def utf8ByteCount_6029(codepoint_6028: Char) = {
  def b_k0_8160(c_6130: Char) = {
    return 1
  }
  def b_k1_8169(c_6133: Char) = {
    return 2
  }
  def b_k2_8178(c_6136: Char) = {
    return 3
  }
  def b_k3_8187(c_6139: Char) = {
    return 4
  }
  def b_k4_8198(c_6142: Char) = {
    val v_r_8197 = {
      let !! v_r_8196 = panic_58("""Not a valid code point""")
      return v_r_8196
    };
    v_r_8197 match[Int] {}
  }
  val v_r_8166 = {
    let ! v_r_8165 = infixGte_6027(codepoint_6028, '\0')
    return v_r_8165
  };
  if (v_r_8166) {
    val v_r_8168 = {
      let ! v_r_8167 = infixLte_6021(codepoint_6028, '\127')
      return v_r_8167
    };
    if (v_r_8168) {
      b_k0_8160(codepoint_6028)
    } else {
      val v_r_8175 = {
        let ! v_r_8174 = infixGte_6027(codepoint_6028, '\128')
        return v_r_8174
      };
      if (v_r_8175) {
        val v_r_8177 = {
          let ! v_r_8176 = infixLte_6021(codepoint_6028, '\2047')
          return v_r_8176
        };
        if (v_r_8177) {
          b_k1_8169(codepoint_6028)
        } else {
          val v_r_8184 = {
            let ! v_r_8183 = infixGte_6027(codepoint_6028, '\2048')
            return v_r_8183
          };
          if (v_r_8184) {
            val v_r_8186 = {
              let ! v_r_8185 = infixLte_6021(codepoint_6028, '\65535')
              return v_r_8185
            };
            if (v_r_8186) {
              b_k2_8178(codepoint_6028)
            } else {
              val v_r_8193 = {
                let ! v_r_8192 = infixGte_6027(codepoint_6028, '\65536')
                return v_r_8192
              };
              if (v_r_8193) {
                val v_r_8195 = {
                  let ! v_r_8194 = infixLte_6021(codepoint_6028, '\1114111')
                  return v_r_8194
                };
                if (v_r_8195) {
 ...
```

It not only duplicates code but also contains unwanted `val` `return` sequences. I have no idea why.

Here is the offender:

```
def utf8ByteCount(codepoint: Char): Int = codepoint match {
  case c and c >= '\u0000'  and c <= '\u007F'   => 1
  case c and c >= '\u0080'  and c <= '\u07FF'   => 2
  case c and c >= '\u0800'  and c <= '\uFFFF'   => 3
  case c and c >= '\u10000' and c <= '\u10FFFF' => 4
  case c => panic("Not a valid code point")
}
```

effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
```diff
@@ -467,7 +466,6 @@ object Normalizer { normal =>
 
     // congruences
     // [[ let x = f(y); f(y) ]] = let x = f(y); x
-    case Expr.PureApp(f, targs, vargs) => available(Expr.PureApp(f, targs, vargs.map(normalize)))
```
jiribenes, line 470: I would have expected a case for `ExternApp(..., Pure, ...)` here, together with the case in `shareable`, otherwise we can't share computation as much...
[if I understood the role of ExternApp, that is]

EDIT: or a similar case somewhere else, anyways

PR #1130, opened 2025-09-24
Basic Bump allocator

danielflat:

PR #1299, opened 2026-02-27
Constant-time Reference Counting in Effekt

danielflat:
@phischu Here is my final PR for my thesis. Can you check it, please?


jiribenes:
Hey @danielflat, thanks for the PR.
Does it subsume #1130 and #1182? If so, could you please close those that are subsumed by this PR?

danielflat:
hi @jiribenes, yes. I closed the other ones.

effekt/jvm/src/main/scala/effekt/Runner.scala
```diff
@@ -306,17 +306,24 @@ object LLVMRunner extends Runner[String] {
...
         case os if os.contains("win") => None
         case os if os.contains("linux") => Some(file("/usr/local/lib"))
         case os => None
       }
     }
     val includes = C.config.clangIncludes.toOption.map(file).orElse {
       OS match {
-        case os if os.contains("mac")  => Some(file("/opt/homebrew/include"))
+        case os if os.contains("mac") && arch.equals("aarch64") => Some(file("/opt/homebrew/include"))
+        case os if os.contains("mac") => Some(file("/usr/local/include"))
```
phischu, line 326: Someone with a Mac should test and approve this

effekt/jvm/src/test/scala/effekt/CTRCTests.scala
```diff
@@ -0,0 +1,23 @@
...
+  def backendName = "llvm"
+  
+  override def valgrind = false // works on linux only
+  
+  override lazy val positives: Set[File] = Set(
+    examplesDir / "ctrc" / "optimized"
+  )
+  
+  override lazy val withoutOptimizations: Set[File] = Set(
+    examplesDir / "ctrc" / "nooptimized",
```
phischu: Why are the tests in a folder nooptimized? I can't find the "optimized" folder.

I personally would not add any tests at all. Others surely disagree.

phischu (commented):
Looks very good so far, impressive work.

I still need to review `llvm/Transformer.scala`.

jiribenes:
If kept separate, the CRTC tests should likely be grouped with LLVM tests (i.e. added to both `testBackendLLVM` and `separatedTests` below)
https://github.com/effekt-lang/effekt/blob/12001cf9c9c0093210b44349873be79f8db16c0d/build.sbt#L227-L251

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -569,32 +566,39 @@ object Transformer {
...
+      loadEnvironment(environmentReference, environment, Object)
+
+      // Step 02: we create the release function for this object
+      val release = ConstantGlobal(freshName("release"));
+      defineFunction(release.name, List(Parameter(objectType, "object"))) {
+        createReleaseFunction(release.name, environment, freeInBody)
+      }
+
+      // Step 03: we call the release function
+      emit(Call("_", Ccc(), VoidType(), release, List(`object`)));
```
phischu: Please avoid defining a function here that is then immediately called. This should let us get rid of the "Directly" variants too. Does this make sense?

effekt/jvm/src/main/scala/effekt/Runner.scala
jiribenes: fwiw, this works on my machine (though I use nix-darwin, so I'm not the best guinea pig)

effekt/jvm/src/test/scala/effekt/CTRCTests.scala
danielflat: Yes, there is no optimized folder because during the thesis, it was helpful to work with the pure program rather than the optimized one. These tests were done to "debug" better through Effekt.

It would be cool if we could test stronger by asserting the number of used chunks or other invariants. Then these tests would have more value

PR #1305, opened 2026-03-01
Port 'io/network' to the jsNode backend

jiribenes:
This is a very approximate implementation in that the only thing I barely tested is the `client_server` benchmark (which I can't run right now, my LLVM backend doesn't work).

It does work with my LSP-in-Effekt demo locally :)

### DIFF
--- a/examples/benchmarks/input_output/networked_numbers.effekt
+++ b/examples/benchmarks/input_output/networked_numbers.effekt
@@ -26,22 +26,22 @@ def readBE(): Int / { read[Byte], fail } = must {
 }
 
 /// Write a 4-byte big-endian int
-def writeInt(conn: Connection, n: Int): Unit / Exception[IOError] = {
-  val buf = bytearray::collect { emitBE(n) }
-  val _ = write(conn, buf, 0, 4)
+def writeInt(connection: Connection, n: Int): Unit / Exception[IOError] = {
+  val buffer = bytearray::collect { emitBE(n) }
+  val _ = write(connection, buffer, 0, 4)
 }
 
 /// Read exactly 4 bytes and decode into a big-endian int
-def readInt(conn: Connection): Int / Exception[IOError] = {
-  val buf = bytearray::allocate(4)
+def readInt(connection: Connection): Int / Exception[IOError] = {
+  val buffer = bytearray::allocate(4)
   var offset = 0
   while (offset < 4) {
-    val r = read(conn, buf, offset, 4 - offset)
+    val r = read(connection, buffer, offset, 4 - offset)
     if (r <= 0) { panic("unexpected EOF reading int") }
     offset = offset + r
   }
   try {
-    feed(buf) { readBE()}
+    feed(buffer) { readBE()}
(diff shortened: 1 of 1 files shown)


libraries/common/io/network.effekt
```diff
@@ -22,8 +22,9 @@ def connect(host: String, port: Int): Connection / Exception[IOError] =
   internal::checkResult(internal::connect(host, port))
 
 /// Closes a TCP connection and releases associated resources.
-def close(handle: Connection): Unit =
-  internal::close(handle)
+def close(handle: Connection): Unit / Exception[IOError] = {
+  internal::checkResult(internal::close(handle)); ()
+}
```
phischu: I don't think this can fail. Indeed, the callback is always called with `0` as far as I can see.

libraries/common/io/network.effekt
```diff
@@ -39,9 +40,9 @@ def listen(listener: Listener, handler: Connection => Unit at {io, async, global
 }
 
 /// Stops a TCP listener and releases associated resources.
-def shutdown(listener: Listener): Unit =
-  internal::shutdown(listener)
-
+def shutdown(listener: Listener): Unit / Exception[IOError] = {
+  internal::checkResult(internal::shutdown(listener)); ()
+}
```
phischu: Same here, as far as I can see this never fails.

libraries/common/io/network.effekt
jiribenes: Thanks, fixed; returns a `Unit` now.

libraries/common/io/network.effekt
jiribenes: dtto, thanks

jiribenes:
I added one more simple benchmark for `io/network` + tried to ensure that int->byte conversion is the same across all backends. PTAL @phischu 🙏

examples/benchmarks/input_output/networked_numbers.effekt
```diff
@@ -0,0 +1,86 @@
...
+/// Read a 4-byte big-endian integer, failing if any byte missing or invalid
+def readBE(): Int / { read[Byte], fail } = must {
+  do read().toInt.bitwiseShl(24).bitwiseOr(
+    do read().toInt.bitwiseShl(16).bitwiseOr(
+      do read().toInt.bitwiseShl(8).bitwiseOr(
+        do read().toInt
+      )
+    )
+  )
+}
```
jiribenes, line 26: These are also likely in #923 which we really ought to merge. :)

phischu (approved):
Yes, very good job! The fact that connections are integer indices into a map is not ideal, we could change that later. Since this adds a new benchmark, when running it a good input size is `100_000` @marvinborner .

marvinborner:
Note that if you want benchmarks to be tracked in effekt-plots, you need to add them to the config files, e.g. [config_llvm.txt](https://github.com/effekt-lang/effekt/blob/main/examples/benchmarks/config_llvm.txt). Then they should appear in effekt-plots automagically.

PR #1306, opened 2026-03-02
Add primitive (in)equality on 'Byte's

jiribenes:
Only tested on JS.

Alternative: derive via conversion from/to `Int` and their equality...


jiribenes:
As a drive-by, I added `bytearrray::build`. Hopefully this side-steps the cryptic flakiness of the LSPTests.

PR #1310, opened 2026-03-04
Add getter for environment variables to stdlib

### LINKED ISSUE #543: Add a getter for environment variables to stdlib
Add a function `getEnv : (key: String) => Option[String]` which can read from environment variables.
Feel free to use a better name :)

A possible design is some subset of:
- an opaque type called `Environment`
- `def getOption(env: Environment, key: String): Option[String]`
- `def get(env: Environment, key: String): String / Exception[???]`
- `def getOrElse(env: Environment, key: String) { otherwise : => String }: String`

Backend availability:
- libuv (LLVM): [`uv_os_getenv`](https://docs.libuv.org/en/v1.x/misc.html#c.uv_os_getenv)
- nodejs: `process.env.get`
- chez: [`getenv`](https://practical-scheme.net/wiliki/schemexref.cgi?getenv)

Use cases:
- secrets
- environment variables like `TERM`, `COLORTERM`, `TERM_PROGRAM`, and [`NOCOLOR`](http://no-color.org) are useful for detecting (the lack of) color support in terminal emulators (relevant for #542)

---

<details>
  <summary>A few imprecise notes on how to implement this for the LLVM backend</summary>

> **Warning:**
> I have not thought this through, I got this after talking to Claude for a while, but I haven't tested it at all.

In `io.effekt`, add something like:
```scala
def getenv(name: String): Option[String] / AsyncIO =
  do performIO(box { k =>
    callback::getenv(name,
      box { value => k(Some(value)) },
      box { error =>
        if (error == 2) { // ENOENT (No such file or directory)
          k(None())
        } else {
          panic("Error getting environment variable: " ++ error.toString)
        }
      })
  })

namespace callback {
  extern io def getenv(name: String,
      onSuccess: (String) => Unit at {io, global},
      onFailure: (Int) => Unit at {io, global}): Unit =
    js "getenv(${name}, ${onSuccess}, ${onFailure})"
    llvm """
      ; stack allocate
      %onSuccess_ptr = alloca %Neg
      store %Neg %onSuccess, %Neg* %onSuccess_ptr

      %onFailure_ptr = alloca %Neg
      store %Neg %onFailure, %Neg* %onFailure_ptr

      call void @c_getenv(%Pos %name, ptr %onSuccess_ptr, ptr %onFailure_ptr) #0
      ret %Pos zeroinitializer
    """
}
```

In `io.c`, add something like:
```c
void c_getenv_handler(uv_fs_t* req) {
    // Extract the result from the uv_fs_t structure
    int result = req->result;

    // Load the callbacks
    Callbacks* callbacks = (Callbacks*)req->data;
    struct Neg success = callbacks->on_success;
    struct Neg failure = callbacks->on_failure;

    // Free request structure
    uv_fs_req_cleanup(req);
    free(req);
    free(callbacks);

    if (result >= 0) {
        // The environment variable exists
        char* buffer = (char*)req->ptr;
        struct Pos value = c_buffer_from_string(buffer);
        free(buffer);
        run_Pos(success, value);
    } else {
        // An error occurred or the environment variable doesn't exist
        run_i64(failure, uv_error_to_errno(result));
    }
}

void c_getenv(struct Pos name, struct Neg* success, struct Neg* failure) {
    uv_fs_t* req = (uv_fs_t*)malloc(sizeof(uv_fs_t));

    // Convert the Effekt String to a 0-terminated string
    char* name_str = c_buffer_as_null_terminated_string(name);
    erasePositive((struct Pos) name);

    // Allocate Callbacks on the heap
    Callbacks* callbacks = (Callbacks*)malloc(sizeof(Callbacks));
    callbacks->on_success = *success;
    callbacks->on_failure = *failure;

    // Store the callbacks in the req's data field
    req->data = callbacks;

    // Get the default loop
    uv_loop_t* loop = uv_default_loop();

    // First, get the size of the environment variable by waiting for UV_ENOBUFS
    size_t size = 0;
    int result = uv_os_getenv(name_str, NULL, &size);

    if (result == UV_ENOENT) {
        // Environment variable doesn't exist
        free(name_str);
        free(req);
        free(callbacks);
        run_i64(*failure, 2); // ENOENT
        return;
    }

    if (result != UV_ENOBUFS) {
        // Unexpected error
        free(name_str);
        free(req);
        free(callbacks);
        run_i64(*failure, uv_error_to_errno(result));
        return;
    }

    // Allocate buffer for the environment variable value
    char* buffer = (char*)malloc(size);
    result = uv_os_getenv(name_str, buffer, &size);

    if (result == 0) {
        // Success
        req->ptr = buffer;
        req->result = result;
        c_getenv_handler(req);
    } else {
        // Error
        free(buffer);
        free(req);
        free(callbacks);
        run_i64(*failure, uv_error_to_errno(result));
    }

    free(name_str);
}
```

</details>

jiribenes:
Resolves #543 since nobody else wanted to do it :)

The design is a bit weird: I don't know how to return something like `undefined` from the LLVM backend, so since we need two calls to decide the size of the resulting buffer anyways, I opted to reuse the `size` for indicating failure when it's `-1`.
This _should_ also work on Windows, but I haven't tested it there.

### DIFF
--- a/libraries/llvm/io.c
+++ b/libraries/llvm/io.c
@@ -826,15 +826,14 @@ void c_spawn(struct Pos cmd, struct Pos args, struct Pos options, Stack stack) {
     }
 }
 
-// Returns the required buffer size (including null terminator) if the
-// variable exists, or -1 if it does not.
+// Returns the required buffer size if the variable exists, or -1 if it does not.
 // This avoids allocating when the variable is not set.
 Int c_getenv_size(struct Pos name) {
     char* name_str = c_bytearray_into_nullterminated_string(name);
     erasePositive(name);
 
     char dummy;
-    size_t size = 1;
+    size_t size = 1; // Note: size = 1 is required, otherwise libuv always returns EINVAL.
     int result = uv_os_getenv(name_str, &dummy, &size);
     free(name_str);
 


libraries/llvm/io.c
```diff
@@ -826,4 +826,42 @@ void c_spawn(struct Pos cmd, struct Pos args, struct Pos options, Stack stack) {
...
 
+// Returns the required buffer size (including null terminator) if the
+// variable exists, or -1 if it does not.
+// This avoids allocating when the variable is not set.
+Int c_getenv_size(struct Pos name) {
+    char* name_str = c_bytearray_into_nullterminated_string(name);
+    erasePositive(name);
+
+    char dummy;
+    size_t size = 1;
```
phischu:
```suggestion
    size_t size = 0;
```

libraries/llvm/io.c
```diff
@@ -826,4 +826,42 @@ void c_spawn(struct Pos cmd, struct Pos args, struct Pos options, Stack stack) {
...
+    int result = uv_os_getenv(name_str, &dummy, &size);
+    free(name_str);
+
+    if (result == UV_ENOENT) return -1;
+    if (result == 0 || result == UV_ENOBUFS) return (Int)(size + 1);
```
phischu:
```suggestion
    if (result == 0 || result == UV_ENOBUFS) return (Int)(size);
```

libraries/llvm/io.c
jiribenes: This one does not work, the call to the `uv` function _needs_ `size ≥ 1`. (I've tried before with `NULL` and `size = 0`)
https://docs.libuv.org/en/v1.x/misc.html#c.uv_os_getenv

phischu: Good thanks!

jiribenes: Here's the tests, the last one suggests that they always return `EINVAL` when the `size=0`
https://github.com/libuv/libuv/blob/50ed2fd7bdd42830ff7327773f62540899b0e9a3/test/test-env-vars.c#L49-L59

libraries/llvm/io.c
jiribenes: Yeah, you're right, I overcorrected here, thanks!

PR #1311, opened 2026-03-05
Add 'bytearray::sort' as a counting, in-place sort

jiribenes:
I had this on my computer for some time and it makes no sense not to upstream it.

### DIFF
--- a/libraries/common/bytearray.effekt
+++ b/libraries/common/bytearray.effekt
@@ -130,15 +130,17 @@ def resize(source: ByteArray, size: Int): ByteArray = {
   go(0)
 }
 
-/// In-place counting sort for byte arrays
+/// Sorts a given bytearray `arr` in-place using a counting sort,
+/// allocating a temporary array of 256 integers for said counting.
 ///
-/// Allocates a temporary array of 256 integers
+/// O(N)
 def sort(arr: ByteArray): Unit = {
   val length = arr.size
   if (length <= 1) {
     ()
   } else {
     // Count occurrences of each byte value
+    // (maps 0..255 to its count in the byte array)
     val counts = array::allocateZeros(256)
     each(0, length) { i =>
       val byte = arr.get(i).toInt


libraries/common/bytearray.effekt
```diff
@@ -3,6 +3,7 @@ module bytearray
 import effekt
 import stream
 import control
+import array
```
jiribenes, line 6: The only potential caveat: since we need to alloc a 256-element array, we need to `import array` here :/

examples/stdlib/bytearray/bytearray_sort.effekt
```diff
@@ -0,0 +1,76 @@
+import bytearray
+
+// mocking the test framework here
+def test(name: String) { test: () => Unit / stop }: Unit = {
```
jiribenes, line 4: These were originally tests using the `test` module, but I didn't want to import it here in the test (and wanted to try out an even smaller testing framework).

libraries/common/bytearray.effekt
```diff
@@ -129,6 +130,32 @@ def resize(source: ByteArray, size: Int): ByteArray = {
   go(0)
 }
 
+/// In-place counting sort for byte arrays
+///
+/// Allocates a temporary array of 256 integers
```
jiribenes: The docs could be a bit better here and perhaps it will make CI work again. 
```suggestion
/// Sorts a given bytearray `arr` in-place using a counting sort,
/// allocating a temporary array of 256 integers for said counting.
///
/// O(N)
```

jiribenes:
This PR deterministically fails in CI on JS, even when changing the doc comment [!], cc @timsueberkrueb 
Screenshot 2026-03-05 at 22 30 58

EDIT: solved by a rebase.

jiribenes:
This is somewhat niche and, at the same time, doesn't break any existing code, so I'm just merging this without a review :)

Issue #1224, opened 2025-11-25
Unbox requires a boxed type, but got B.


phischu:
The following program:

```
type Status[B] {
  Done(value: Int)
  More(state: B)
}

def main(): Unit = {
  Done(123) match {
    case Done(value) => value
    case More(task) => task()
  }
  ()
}
```

yields the error message `Unbox requires a boxed type, but got B.`.

Desired behavior is either to accept the program and infer `B` to be instantiated with `() => Int at {}`, or reject the program with an error message like "could not infer type argument for `B` in `Done(123)`".

jiribenes:
One reason why we don't instantiate is here in a TODO comment:
https://github.com/effekt-lang/effekt/blob/4946a1cd95911ed759f9042266fbe7efbdb767ed/effekt/shared/src/main/scala/effekt/Typer.scala#L556-L566

EDIT: It's interesting that we don't propagate the expected type here; in this example it's always `None`, even if you do `val res: Int = ...`. We _could_ still pass a `expectedResult: Option[ValueType]` since we know it from the aforementioned annotation...

jiribenes:
Best I can do without that is:
```scala
// ...
    case More(task) => task()
                       ^^^^
// [info] Type parameter(s) B of type B could not be inferred.
// Consider providing explicit type annotations where this type originates.

// [error] Unbox requires a boxed type, but got B.
```

where the only reasonable workaround I have is
```scala
// ...
    case More(task) =>
      def t: () => Int = task
      t()
```

jiribenes:
As of #1315, the error message becomes:
```
Expected task to be a function, but got a value of type B instead, which cannot be called as a function.
```

which seems a bit better, though the user never wrote `B` anywhere...

jiribenes:
With #1308, we might be able to put a hint annotation into the `Some`, i.e.:
```scala
  Done(123) match {
    case Done(value) => value
    case More(task: () => Int at {}) => task()
  }
```
which is one more solution (besides the two mentioned above and implemented in #1317)

phischu:
I believe there is an oversight in typer. https://github.com/effekt-lang/effekt/pull/1317#issuecomment-4016217294

jiribenes:
I don't think it's an oversight, we just don't unify block types

PR #1317, opened 2026-03-06
Improve type inference of function calls by propagating expectations

jiribenes:
Improves the error message of #1224 by propagating more information for boxed functions whenever possible, specifically when we know the expected type of the call, but were previously too shy to try and apply it:

```scala
type Status[B] {
  Done(value: Int)
  More(state: B)
}

def ex1(): Unit = {
  val res: Int = Done(123) match {
  //     ~~~~~ 
  // this is enough to help resolve the issue
    case Done(value) => value
    case More(task) => task()
  }
  println(res)
}

def ex2() =
  println(Done(123) match {
  // the println is enough to force it to be a number
    case Done(value) => value
    case More(task) => task()
  })

def original_issue1224(): Unit = {
  // still fails with "Unbox requires a boxed type, but got B." on the `task()` call
  Done(123) match {
    case Done(value) => value
    case More(task) => task()
  }
  ()
}

def another_fails(): Unit = {
  val result: Int = Done(123) match {
    case Done(v) => v
    // f() checked with expected = None
    case More(f) => f() match { // still fails with "Unbox requires a boxed type, but got B."
      case n => n
    }
  }
  result
}
```


jiribenes:
@phischu if you still remember the original problem that inspired the issue, does this fix solve it?

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -579,14 +587,29 @@ object Typer extends Phase[NameResolved, Typechecked] {
...
-                Context.abort(pretty"Expected the right-hand side of a `def` binding to be a block, but got a value of type $vtpe.\nMaybe try `val` if you're defining a value.")
-              case _ =>
-                Context.abort(pretty"Unbox requires a boxed type, but got $vtpe.")
+            // `expected` wasn't enough to resolve the type (e.g. unconstrained type variable ?B).
+            // Now we *lazily* construct a partial function type to avoid polluting
+            // the unification state in the normal case where the first branch succeeds.
+            expectedCallHint match {
+              case Some((arity, retTpe)) =>
+                val freshVps = List.fill(arity) {
+                  ValueTypeRef(Context.freshTypeVar(TypeParam(Name.local("_")), u))
```
jiribenes, line 597: I'm still a bit worried about this `_` leaking (again), I'm trying to fake "just enough" block unification to get by here...

jiribenes:
Rebased on top of 117f72092b7ac36fd3bdf22c0f7ef9a0621ce3f3 (from merging #1315).

effekt/shared/src/main/scala/effekt/Typer.scala
jiribenes: Yep, that seems to have broken #1315 :(
Screenshot 2026-03-06 at 17 18 18

jiribenes: In a way, the unexpected error message _is_ better, but I don't want to have ad-hoc "anything fits here" type variables without them having some reasonable semantics a la #276.

phischu:
No, this does not fix #1224.  The following type checks:

```
def main(): Unit = {
  Done[() => Int at {}](123) match {
    case Done(value) => value
    case More(task) => task()
  }
  ()
}
```

But in the following:

```
def main(): Unit = {
  Done[String](123) match {
    case Done(value) => value
    case More(task) => task()
  }
  ()
}
```

I get:

> Expected task to be a function, but got a value of type B instead, which cannot be called as a function.

What I would want is

> Expected task to be a function, but got a value of type _String_ instead, which cannot be called as a function.

So we are forgetting to do index unification in matches or something.

jiribenes:
Yeah, I know it doesn't fix #1224, I was just trying to improve the inference when you annotate at least something. But good point, I can take a look at the trace again and see if I can come up with some other trick — the issue here is that we never unify block types... (so we would derive the `B ~ String` fact too late, only after the err)

phischu:
No need to fix #1224, I misunderstood. I am the wrong person to review this, but I'll just approve anyway.

phischu (approved):
Nothing controversial

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -578,21 +586,38 @@ object Typer extends Phase[NameResolved, Typechecked] {
...
+                Context.annotationOption(Annotations.UnboxParent, u) match {
+                  case Some(source.DefDef(id, captures, annot, block, doc, span)) =>
+                    // Since this `unbox` was synthesized by the compiler from `def foo = E`,
+                    // it's possible that the user simply doesn't know that they should have used the `val` keyword to specify a value
+                    // instead of using `def`; see [issue #130](https://github.com/effekt-lang/effekt/issues/130) for more details
+                    Context.abort(pretty"Expected the right-hand side of a `def` binding to be a block, but got a value of type $vtpe.\nMaybe try `val` if you're defining a value.")
+                  case Some(source.IdTarget(id)) =>
+                    // Since this `unbox` was synthesized by the compiler when trying to call a value as if it was a function (e.g., `myArray(0)`),
+                    // we provide a more helpful error message; see [issue #737](https://github.com/effekt-lang/effekt/issues/737)
+                    Context.abort(pretty"Expected $id to be a function, but got a value of type $vtpe instead, which cannot be called as a function.")
```
jiribenes: I wonder what this will do. :)
```suggestion
                    Context.abort(pretty"Expected $id to be a function, but got a value of type ${Context.unification(vtpe)} instead, which cannot be called as a function.")
```

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -578,21 +586,38 @@ object Typer extends Phase[NameResolved, Typechecked] {
...
+                  case Some(source.DefDef(id, captures, annot, block, doc, span)) =>
+                    // Since this `unbox` was synthesized by the compiler from `def foo = E`,
+                    // it's possible that the user simply doesn't know that they should have used the `val` keyword to specify a value
+                    // instead of using `def`; see [issue #130](https://github.com/effekt-lang/effekt/issues/130) for more details
+                    Context.abort(pretty"Expected the right-hand side of a `def` binding to be a block, but got a value of type $vtpe.\nMaybe try `val` if you're defining a value.")
+                  case Some(source.IdTarget(id)) =>
+                    // Since this `unbox` was synthesized by the compiler when trying to call a value as if it was a function (e.g., `myArray(0)`),
+                    // we provide a more helpful error message; see [issue #737](https://github.com/effekt-lang/effekt/issues/737)
+                    Context.abort(pretty"Expected $id to be a function, but got a value of type ${Context.unification(vtpe)} instead, which cannot be called as a function.")
```
jiribenes:
```suggestion
                    Context.abort(pretty"Expected $id to be a function, but got a value of type $vtpe instead, which cannot be called as a function.")
```
Yeah, it didn't do anything :(

PR #1321, opened 2026-03-09
Delete redundant jvm option

dvdvgt:
The option `Xms` specifies the initial amount of reserved memory the JVM is starting with. The option `Xmx` specifies the maximum amount of memory the JVM is allowed to use. `Xmx` is given with `Xmx4g` and `Xmx8g` and are thereby contradictory.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,4 +1,3 @@
 -Xms1g
--Xmx4g
 -Xss32m
 -Xmx8g


marvinborner:
Is there any reason for not having an even higher memory limit? I regularly run into Java Heap OOMs when compiling with `--optimize` (though this might also be a separate bug)

dvdvgt:
> Is there any reason for not having an even higher memory limit? I regularly run into Java Heap OOMs when compiling with `--optimize` (though this might also be a separate bug)

Memory is expensive these days ;)

More seriously; this should probably be a separate issue.

PR #1323, opened 2026-03-16
CPS: Tag literals with their core type

marzipankaiser:
This tags literals in the CPS IR with their core type, since in Scala.js matching against Byte also matches Doubles in the respective range (thus, converting literal `-126.0` to the byte `129`).

Still to be tested.

Thanks @jiribenes for helping to identify the root cause of this.


marzipankaiser:
Btw, do we actually run the `WebTests`? 🤔

PR #1326, opened 2026-03-18
Experiment: flatten overloading hierarchy between fns and ops

jiribenes:


jiribenes:
This is not a great solution of the underlying problem. I'd much rather consider desugaring ops into `def op_($vargs...) {self: $Interface} {$bargs...} = self.op_($vargs...) {$bargs...}` and then use overloading.

PR #1327, opened 2026-03-18
Experiment: objects instead of stack frames

serkm:
This mimics cps for benchmarking purposes.
Stack frames are allocated as closures, which are called when returning.
Stacks are still allocated as before, and used by `var`s.

cc @phischu 

benchmarks_normalized


phischu:
Very interesting, thank you. Easier than I thought... but also a lot slower than I thought.

PR #1329, opened 2026-03-20
Fix backtracked mutable state on Chez CPS

jiribenes:
TLDR Backtracking over an unconditional variable write results in bad output on Chez CPS.

---

I got nerdsniped at @martin-ilgner's presentation today looking at the implementation of backtrackable state via snapshottable stores in Chez.

Compare the JS version
https://github.com/effekt-lang/effekt/blob/da69e03c4c5bb248a6cb8fb40e4355b83e5ce8ce/libraries/js/effekt_runtime.js#L64-L69
with the Chez CPS version
https://github.com/effekt-lang/effekt/blob/72795e70899370ea22b5e92e9bc0615ac73032a0/libraries/chez/cps/effekt.ss#L57-L63

The latter does **not** actually restore the generation!

Same for
https://github.com/effekt-lang/effekt/blob/72795e70899370ea22b5e92e9bc0615ac73032a0/libraries/chez/cps/effekt.ss#L40-L51
and
https://github.com/effekt-lang/effekt/blob/da69e03c4c5bb248a6cb8fb40e4355b83e5ce8ce/libraries/js/effekt_runtime.js#L58-L61

Putting together the example wasn't easy. Here's my strategy:
1. Create a store and a ref with value 0
2. Take a snapshot S
3. Set ref to 1
4. Restore S ~> ref back to 0
5. Set ref to 2 (no diff created because `gen` didn't increment)
6.  Restore S again ~> should get 0, but 2 appears [!]

This was directly reproducible in Chez, but I spent a bit more time trying to encode it in an Effekt source program (see the added test):
```
bash-5.3$ effekt --backend=chez-cps --no-optimize --ir-write-all $FILE.effekt
44
bash-5.3$ effekt --backend=js --no-optimize --ir-write-all $FILE.effekt
43
bash-5.3$ effekt --backend=llvm --no-optimize --ir-write-all $FILE.effekt
43
bash-5.3$ effekt --backend=chez-callcc --no-optimize --ir-write-all $FILE.effekt
43
```

The fix is simple, just do the same thing as JS does: always restore the generation correctly (see the attached patch).

### DIFF
--- a/libraries/chez/cps/effekt.ss
+++ b/libraries/chez/cps/effekt.ss
@@ -47,6 +47,7 @@
                    [r (diff-ref realDiff)]
                    [oldValue (ref-value r)])
                 (ref-value-set! r (diff-value realDiff))
+                (ref-generation-set! r (diff-generation realDiff))
                 (set-box! n (make-diff r oldValue (ref-generation r) currentDiff))
                 (applyDiffs currentDiff (cdr diffs)))]))
 
@@ -60,7 +61,8 @@
            [store (snap-store snap)]
            [storeRoot (store-root store)])
         (reroot snapRoot storeRoot)
-        (store-root-set! store snapRoot)))
+        (store-root-set!       store snapRoot)
+        (store-generation-set! store (+ (snap-generation snap) 1))))
 
 ; Cont a = a, MetaCont -> #
 ; Prompt = Int


examples/pos/multishot-forgotten-generation.effekt
```diff
@@ -0,0 +1,18 @@
...
+      val b = do flip()
+      x = x + 1            // unconditional write triggers the bug
+      x + (if (b) 10 else 20)
+    } with flip {
+      val a = resume(true)   // resume 1: x goes 0 ~> 1, returns 11
+      val b = resume(false)  // resume 2: x badly restored to 0, x goes 0 ~> 1, returns 21
+      val c = resume(true)   // resume 3: x stays 1, goes 1 ~> 2, returns 12 <== BUG!
+      a + b + c
+    }
+  println(result)
```
jiribenes, line 17: I think that two resumes should be enough, but I didn't manage to write a program where two resumes actually are enough...
Let this be an open challenge for all delimcc enthusiasts :^)

jiribenes:
Since this is a bug and the CI witnesses that it's fixed now, I'll just merge this once the CI finishes its run.

PR #1330, opened 2026-03-21
Port Chez CPS runtime 'diffs' to JS runtime

jiribenes:
Port the tail recursive rerooting from Chez CPS runtime (#1160) to JS runtime.

The original port was slightly slower (geomean 0.1% slower), I optimised it by hand by reusing the small `path` array by using a global `_rerootPath` that just gets overwritten every time (so allocations for the path are rare).
After the optimisation, some benchmarks like triples are ~1-2% faster, now geomean is 0%.

One bonus that hasn't shown up in the benchmarks is that this is no longer recursive so it shouldn't blow the JS stack anymore.

Further potential optimisations are:
1. fast path for diff length 1 (a check for `cur.root.value === Mem` in `reroot`)
2. inlining the snapshots directly into the frame 
3. if the backend emitted `RESUME_ONCE` for one-shots, we could have a special version of reroot that doesn't write reverse diffs (no need to set `cur.value    = ...`)

### DIFF
--- a/libraries/js/effekt_runtime.js
+++ b/libraries/js/effekt_runtime.js
@@ -6,6 +6,9 @@
 //  restore: O(|write operations since capture|)
 const Mem = null
 
+// reusable buffer for rerooting
+const _rerootPath = []
+
 function Arena() {
   const s = {
     root: { value: Mem },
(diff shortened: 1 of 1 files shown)


jiribenes:
My plan is to merge this and see if it does anything on effekt-plots. If the benchmarks get worse, we can always revert this commit/PR before the release tomorrow.

PR #1331, opened 2026-03-21
Use builtin symbols for prompt identifiers at runtime

jiribenes:
on Chez, use `(gensym)` for prompts instead of an int counter
on JS, use `Symbol()` for prompts instead of an int counter

This is likely [and hopefully] not noticeably faster, but I find it more semantically correct than mutable int counters, not to mention that it's closer to what LLVM does.

The top-level prompt has a pretty name `"toplevel"` so that we can identify it if needed [?]

I didn't try this locally, the test suite should be enough here 😇


jiribenes:
Rebased on top of https://github.com/effekt-lang/effekt/commit/cf6e906531bc8befd970896f6d3e415aca11c392.

Issue #1309, opened 2026-03-03
collection::collectInto / collection::collect(reserve: Int)


jiribenes:
When using streams, we often do stuff like:
```scala
newCollection::collect {
  for { oldCollection.each } { x =>
    do emit(x * x)
  }
}
```

or `string::collect { readMany { c => c.isAlphabetic } }`
or `bytearray::collect { do emit(b1); ...; do emit(b4) }`

but we do know exactly or approximately how long the resulting collection should be, so we should be able to pre-reserve space for string/bytestring/array. Currently, we can only use `::collect` which always allocates a 1-size collection, then doubles and copies it, then doubles and copies it, etc...
This is fine in the general case, but sometimes we know how big the collection will be...

Alternatively, for mutable collections, we might be able to provide the [byte]array to collect into [and at some position], overwriting some elements?

jiribenes:
as a note, some concept of "sized" push-streams could also be helpful here
[...]

PR #1350, opened 2026-03-27
Collect with initial capacity

phischu:
We now have collecting functions with initial capacity as a performance hint. Addresses #1309.

### DIFF
--- a/libraries/common/array.effekt
+++ b/libraries/common/array.effekt
@@ -679,16 +679,25 @@ def feed[T, R](array: Array[T]) { reader: () => R / read[T] }: R = {
 
 /// Collects elements emitted by a push stream producer `stream` into a new array,
 /// discarding the result of the producer.
+/// The provided capacity is a hint for the expected number of elements.
+/// It must be strictly larger than zero.
+def collect[A](capacity: Int) { stream: () => Unit / emit[A] }: Array[A] =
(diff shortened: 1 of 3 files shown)


libraries/common/array.effekt
```diff
@@ -677,18 +677,24 @@ def feed[T, R](array: Array[T]) { reader: () => R / read[T] }: R = {
...
+/// discarding the result of the producer. The provided capacity is a hint for
+/// the expected number of elements.
+def collect[A](capacity: Int) { stream: () => Unit / emit[A] }: Array[A] =
+  returning::collect[A, Unit](capacity){stream}.second
+
 /// Collects elements emitted by a push stream producer `stream` into a new array,
 /// discarding the result of the producer.
 def collect[A] { stream: () => Unit / emit[A] }: Array[A] =
-  returning::collect[A, Unit]{stream}.second
+  collect[A](1){stream}
```
jiribenes: IMO the default capacity should be `4` for arrays, bytearrays and strings. (Like it is for growable vectors)
We `.resize` it to the actual number of elements afterwards anyways, and an array of 4 `%Pos` still fits into one or two cache lines.

phischu: You are totally right, done.

jiribenes (approved):

PR #1356, opened 2026-04-04
OLD: Fix parsing of negative doubles in 'json::readDouble'

### LINKED ISSUE #1355: Json readDouble incorrectly handles negative decimals
Line 115 always adds the fractional digits regardless of sign. That means -1.5 becomes -0.5

https://github.com/effekt-lang/effekt/blob/ad6f80fea7b1a23e087ecb250d0575388cb43d63/libraries/common/json.effekt#L115

JoshuaSchlucke:
Fixes #1355

### DIFF
--- a/libraries/common/json.effekt
+++ b/libraries/common/json.effekt
@@ -112,7 +112,11 @@ def readDouble(): Double / { Scan[Char], Exception[WrongFormat]} = {
       var b = 0.1
       var r = pre.toDouble
       while (optionally[Int] { must { readDigit() } } is Some(d)) {
-        r = r + b * d.toDouble
+        if (pre >= 0) {
+            r = r + b * d.toDouble
+        } else {
+            r = r - b * d.toDouble
+        }
         b = b * 0.1
       }
       exponent(r)


jiribenes:
Good catch, thanks!
I'll only merge this on Monday to avoid triggering a single bugfix release. :)

libraries/common/json.effekt
```diff
@@ -112,7 +112,11 @@ def readDouble(): Double / { Scan[Char], Exception[WrongFormat]} = {
...
       var r = pre.toDouble
       while (optionally[Int] { must { readDigit() } } is Some(d)) {
-        r = r + b * d.toDouble
+        if (pre >= 0) {
+            r = r + b * d.toDouble
+        } else {
+            r = r - b * d.toDouble
+        }
         b = b * 0.1
       }
```
jiribenes, line 121: One other (smaller) fix I can think of would be

```scala
      var b = if (pre >= 0) 0.1 else -0.1
      var r = pre.toDouble
      while (optionally[Int] { must { readDigit() } } is Some(d)) {
        r = r + b * d.toDouble
        b = b * 0.1
     }
```
(then we don't have to check `pre` over and over)

libraries/common/json.effekt
jiribenes, line 324: We should probably test the `json::readDouble` here by just having a negative double field (and then changing the check file in `examples/stdlib/json.check` [?])

jiribenes:
Closed in favour of #1359

PR #1357, opened 2026-04-08
Code position effect

marzipankaiser:
This allows functions (and effects etc) to require a `codepos` effect from their context, which is implicitly handled at each call site of such a function and returns a unique string for that call site.

This is (ab-)using effects for something similar to the implicit parameters in 
> Daan Leijen, Tim Whiting: Syntactic Implicit Parameters with Static Overloading

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Typer.scala
+++ b/effekt/shared/src/main/scala/effekt/Typer.scala
@@ -62,9 +62,6 @@ object Typer extends Phase[NameResolved, Typechecked] {
         Context in {
           Context.withUnificationScope {
             flowingInto(builtins.toplevelCaptures) {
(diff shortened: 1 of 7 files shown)


marzipankaiser:
This now has the following (rough) semantics:
- If a function has a `codepos` effect, it is implicitly handled at the callsite by returning a constant String
    - except if it is in a context where `codepos` is handled already, i.e.:
        - in a `try {  .... } with codepos {...}`
        - in a `def ... : ... / { ..., codepos, ... } =`
        - in a higher-order function call that handles `codepos`

This allows "forwarding" the code position to helper functions easily.

marzipankaiser:
Superseded by #1380 .

PR #1360, opened 2026-04-10
Implement println for char

### LINKED ISSUE #1302: Add `println` for `Char`
We are currently lacking an overload of `println` for characters. I wonder what the behavior should be?

martin-ilgner:
Also, me and Philipp decided that we should make js and chez behave similar to llvm in the `show` function

Fixes #1302

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -48,7 +48,7 @@ jobs:
 
       - name: Run internal tests
         run: |
-          mill project.testRemaining
+          mill -j1 project.testRemaining
 
       - name: Assemble fully optimized JS file
         run: mill project.assembleJS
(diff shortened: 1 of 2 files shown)


martin-ilgner:
Wth is that error?

marvinborner (approved):
Looks good to me, although making `show` undefined may break some things, let's see!

.github/workflows/ci-pr.yml
```diff
@@ -48,7 +48,7 @@ jobs:
 
       - name: Run internal tests
         run: |
-          mill project.testRemaining
+          mill -j1 project.testRemaining
```
jiribenes, line 51: fwiw, this is good for debugging, but it would likely make our [internal] tests _very_ slow :(

marvinborner: Nope doesn't seem too slow. No good fix though. Just a hotfix.

PR #1361, opened 2026-04-10
Add `random::shuffle` and `random::pick`

### LINKED ISSUE #1307: Add 'random::shuffle' for shuffling arrays
Into the `random` module, add a `shuffle` method that takes an array and shuffles it using the `random` effect.
https://github.com/effekt-lang/effekt/blob/main/libraries/common/random.effekt

Feel free to consult [Wikipedia on the commonly-used algorithm by Fisher and Yates](https://en.wikipedia.org/wiki/Fisher–Yates_shuffle); I'd personally strongly prefer the forward variant :)

Then add a test `examples/stdlib/random/shuffle.effekt` that tests this newly added function (with a fixed seed).
See [the other stdlib tests](https://github.com/effekt-lang/effekt/tree/main/examples/stdlib) for documentation.

DennisKleinhans:
Add the two functions `shuffle` and `pick` to the `random` module in the stdlib according the issue #1307


examples/stdlib/random/pick.effekt
```diff
@@ -0,0 +1,8 @@
+import random
+import array
+
+def main() = {
+  val arr = array::build(10) { i => i }
+  with minstd(1337);
+  println(pick(arr))
```
jiribenes: Could we run this a few times, please? Just to see what happens.

```suggestion
  10.repeat { println(pick(arr)) }
```

I'd also like to see the behaviour for an empty and a singleton array, just to cover our bases :)

examples/stdlib/random/shuffle.effekt
```diff
@@ -0,0 +1,24 @@
+import random
+import array
+
+def main() = {
```
jiribenes, line 5: I'd like to see that we're avoiding this case: https://en.wikipedia.org/wiki/Fisher–Yates_shuffle#Sattolo's_algorithm
Could we try the ABCD example there to make sure that we're not just permuting in a cycle?

examples/stdlib/random/pick.effekt
DennisKleinhans: Maybe we can also introduce a `pickOption` to always avoid an `OutOfBounds` for empty arrays. What do you think?

jiribenes: idk, we could also panic if needed (which is consistent with what arrays do now); though then we can't test for it
wdyt, @phischu?

examples/stdlib/random/pick.effekt
```diff
@@ -0,0 +1,17 @@
...
+
+  val arr = array::build(10) { i => i }
+  val singleton = array::build(1) { i => i }
+  val empty = array(0, 0)
+
+  with minstd(1337);
+  10.repeat{ println(pick(arr)) }
+  println(pick(singleton))
+
+  println(result[Int, OutOfBounds] { pick(empty) }.toOption)
```
jiribenes: it's somewhat useful to have `println("")` in-between individual tests
since we can then recognise which part was a test failure in 🙏

examples/stdlib/random/pick.effekt
phischu: Yes, it should panic and we should document this as a precondition.

libraries/common/random.effekt
```diff
@@ -138,6 +141,29 @@ def randomDouble(): Double / random =
...
+/// Shuffle an array in-place using the Fisher-Yates algorithm.
+def shuffle[A](arr: Array[A]): Unit / random = {
+  val n = arr.size
+  var i = 0
+  while (i + 1 < n) {
+    val j = randomInt(i, n - 1)
+    arr.swap(i, j)
+    i = i + 1
+  }
+}
```
jiribenes: in the spirit of using more high-level features, we might want to consider using either `stream::for` & `stream::range`:
```suggestion
  for { range(0, n - 1) } { (i: Int) =>
    val j = randomInt(i, n - 1)
    arr.swap(i, j)
  }
}
```

or `effekt::each`:
```suggestion
  each(0, n - 1) { i =>
    val j = randomInt(i, n - 1)
    arr.swap(i, j)  
  }
}
```

examples/stdlib/random/pick.effekt
phischu: Perhaps we should `fail` instead of panicking, naja, let's see what users say.

PR #1362, opened 2026-04-10
Add :compile and :watch commands to REPL

marvinborner:
This significantly improves upon workflows that do not use the visual studio code extension.

Due to the caching already implemented in the compiler architecture, watching/compiling repeatedly in the REPL will not require compilation of files that did not change.

TODO: also track dependencies in `:watch`

### DIFF
--- a/effekt/jvm/src/main/scala/effekt/Repl.scala
+++ b/effekt/jvm/src/main/scala/effekt/Repl.scala
@@ -10,6 +10,8 @@ import effekt.util.Version.effektVersion
 import kiama.util.{Console, REPL, Range, Source, StringSource}
 import kiama.parsing.{NoSuccess, ParseResult, Success, Input}
 
+import java.nio.file.Paths
+
 class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
 
   private implicit lazy val context: Context with IOModuleDB = driver.context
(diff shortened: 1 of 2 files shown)


effekt/jvm/src/main/scala/effekt/Repl.scala
```diff
@@ -115,6 +119,34 @@ class Repl(driver: Driver) extends REPL[Tree, EffektConfig, EffektError] {
...
+          if (current != last && last != 0L)  {
+            println(s"\n--- ${f} changed: running ${filename} ---")
+            driver.compileFile(filename, config)
+            files = driver.sources.keys.toList // TODO: doesn't include dependencies
+            // println(s"listening on ${files}")
+          }
+          f -> current
+        }.toMap
+      }
+    }
```
jiribenes, line 149: fwiw Java has a `java.nio.file.WatchService` in the stdlib for this purpose. :)

marvinborner: I know and this was my initial implementation. It really messed up sbt though, watching files even after the process exited (and I wasn't able to fix it). Also it needed a lot of imports and can only listen on directories. This is a lot simpler.

marvinborner:
@jiribenes (or anyone else), do you have any idea on how to get the full path to dependencies (possibly even without running the frontend explicitly)? `Include` in `.includes` only contains the basename without directory information.

b-studios:
Is this ready for review? It still says "draft" :)

Issue #821, opened 2025-02-07
List/collection patterns


jiribenes:
## Motivation

Allow 
```scala
[1, 2, 3] match {
  case [1] =>
  case [1, 2] =>
  case [] =>
  case [1, ..rest] =>
  case [1, 2, ..rest] =>
  case [head, ..rest] =>
  case _ =>
}
```
as syntax sugar for:
```scala
[1, 2, 3] match {
  case Cons(1, Nil()) =>
  case Cons(1, Cons(2, Nil())) =>
  case Nil() =>
  case Cons(1, rest) =>
  case Cons(1, Cons(2, rest)) =>
  case Cons(head, rest) =>
  case _ =>
}
```

### Reasoning

I thought about this syntax specifically in order to also _later_ allow:
1. merging two lists/collections as `[..xs, ..ys]`
2. record updates #557:
```scala
record Foo(a: Int, b: Int)
val default = Foo(12, 9) // or: Foo(a = 12, b = 9)
val x = Foo(a = 10, ..default)
println(x) // Foo(10, 9)
```

### Questions

Feel free to bikeshed:
1. `first :: rest` vs `[first, ..rest]`: I like the latter because I feel like it's easier to scale it for patterns like `[1, 2, 3, ..rest]`, on the other hand `::` is definitely more friendly for LLMs coming from Scala
2. the exact syntax for `..` (`...`, `*`): here I took inspiration from C# / Rust / JS / Swift(ish) / Zig(ish)
3. should it be `..rest` or `rest @ ..` ala Rust? (or something else like `rest..`?)
4. do we need/want this at all?

jiribenes:
If you'd like to implement this as a "good first issue" for, say, a Hackathon, here's what to do:

1. Add a case for `[` in `Parser.matchPattern` Ala the one for tuples (`case (`), only for exact lists, no `..` splats 
https://github.com/effekt-lang/effekt/blob/06d6deea59bbb839d60db8cd561e56498a6fe88e/effekt/shared/src/main/scala/effekt/Parser.scala#L1003-L1021

2. Immediately desugar it into `Cons` and `Nil` calls, something like the following; similar to list literals:
```scala
  private def NilPattern: MatchPattern =
    TagPattern(IdRef(List(), "Nil", span().synthesized), Nil, span().synthesized)

  private def ConsPattern(head: MatchPattern, tail: MatchPattern): MatchPattern =
[...]

PR #1363, opened 2026-04-10
List Literal Patterns

martin-ilgner:
Starting to implement #821


jiribenes (changes requested):
This doesn't (yet?) do the `..` part of the issue https://github.com/effekt-lang/effekt/issues/821, so we can't match on tails :(

and I'd like to see a few more pos tests like `if ([None(), Some(42) is [head, Some(x)]) { ... }` and `with val [x, y] = foo { ... }` and `foo { case [x, y, z, a] => ... }` :)

martin-ilgner:
- how do we make sure that there's only one "tail" pattern? where's the latest we can fail here

I made it so that after a tail pattern, `]` is expected

- do we add it as a token, or "just" eat two `.` tokens?

There is already a `..` token

- what's `case [..rest] => ...`? error or the whole list?

Whole list, why not

- how do we write an ignored splat? can we just write `[head, ..]` or do we need `[head, .._]`?

Write `[head, ..]`

- trailing commas?

Not after a splat. (Could change this, but I think there is no argument for it, as splat is always last)

martin-ilgner:
What do you mean with `with val [x, y] = foo { ... }`? Is it even possible/not a compile error because length of list is unknown? Maybe I could do `val [x, ..y] = lst`

martin-ilgner:
@jiribenes

jiribenes:
> What do you mean with with `val [x, y] = foo { ... }`? Is it even possible/not a compile error because length of list is unknown?

Ah, sorry, you can do `val-else` (`val [x, y] = foo { ... } else panic("ohno")`) and `with-val-else` :)
https://github.com/effekt-lang/effekt/blob/1bae57feef98e6660c552d2bd6b40028007ae2ab/examples/pos/with_val_and_else.effekt#L20-L26

martin-ilgner:
I also rewrote manyTrailing. Should be enough pos tests for now 🙏

PR #1364, opened 2026-04-10
Track depth of typed holes in lexer

### LINKED ISSUE #1301: Greater-than sign after splice ending throws error
Say you have `val x = 42` and want to print `"<" ++ x.show ++ ">"` but using splices.

Ideally, you could write `s"<${x.show}>"`, but Effekt throws the error "Expected a splice ending brace `}` but got }>" due to the related syntax for holes via `<{ ... }>`.

[Reproduction](https://effekt-lang.org/playground.html?playground=S0lNU8hNzMzT0FSwVajmUlAoS8xRqACyTYyAnIKizLySnDyNYiUbleoKveKM%2FPJaOyVNrloA&repl=bWFpbigp)

LTeuse:
In Lexer, check if we are inside a hole before parsing `}>` to fix issue #1301.

Resolves #1301


jiribenes (commented):
Thanks!

Please add a few tests into LexerTests: https://github.com/effekt-lang/effekt/blob/8acd841b794293d57a3e3ea000d32e32fdc4a9ee/effekt/jvm/src/test/scala/effekt/LexerTests.scala#L232


(like the one from the issue)

I'm also very interested in nested holes like
```
<{ 
  val x = 42 + <" here's a string ">
  println(x)
  <>
}>
```
(and similar)

Issue #1308, opened 2026-03-03
Type annotations for patterns


jiribenes:
Currently, we can't write the `: Int` part of
```scala
expr match {
  case Cons(x: Int, _) => x // doesn't parse, "Expected ) but got :"
  ...
}
```

similarly, we can't use it in `if (expr is Some(x: String))` and `val Some(b: Bool) = expr`.

This is especially annoying for tuple parameters
```scala
// argument list works
def f { g: (Int, String) => Unit }: Unit = g(0, "")
f { (x: Int, y) => () } // works (and reports an error if the annotation is wrong)

// tuple does not
def f { g: ((Int, String)) => Unit }: Unit = g((0, ""))
f { case (x: Int, y) => () } // doesn't parse, "Expected ) but got :"
```

However, I do **not** want to use it for runtime matching like in Scala:
```scala
// not this, please
expr match {
  case n: Int => ...
  case s: String => ...
  case _ => panic("ohno")
}
```

jiribenes:
One specific case where this bothers me is when using `for`:
```scala
for { "abc".each } { c =>
  <> // ^ Effects need to be fully known, but effect emit[A]'s type parameter(s) A could not be inferred.
}
```

You can annotate the type in two places:
1. on the `for` as a type parameter, which I generally find ugly
   ```scala
    for[Char] { "abc".each } { c =>
      <>
    }
    ```
2. on the binder itself:
    ```scala
    for { "abc".each } { (c: Char) =>
      <>
    }
    ```

However, for tuples (like you get when iterating over pairs of things, such as `(key, value)` pairs from a `Map`)
```scala
for { [(0, "a"), (1, "b"), (2, "c")].each } { case (i, s) =>
  <> // ^ Effects need to be fully known, but effect emit[Tuple2[A, B]]'s type parameter(s) A, B could not be inferred.
}
```

you can only use the variant 1:
```scala
for[(Int, String)] { [(0, "a"), (1, "b"), (2, "c")].each } { case (i, s) =>
  <> // ^ Effects need to be fully known, but effect emit[Tuple2[A, B]]'s type parameter(s) A, B could not be inferred.
}
```

but **not** variant 2 [!]:
```scala
for { [(0, "a"), (1, "b"), (2, "c")].each } { case (i: Int, s: String) => // parse error: "Expected ) but got :"
  <> 
}
```

jiribenes:
[...]

PR #1366, opened 2026-04-10
Type annotation for patterns

martin-ilgner:
Tries to resolve the issues mentioned in #1308


phischu:
Great! Please add a test file `examples/pos/pattern_annotations.effekt` with the examples from the linked issue.

martin-ilgner:
Done. But there is a problem:
``` scala
str match {
  case "hello" => "world"
  case s: String => s
  case _ => "bye"
}
```
is wrongly parsed. It takes the match arrow as part of a function type. But still fails with `Expected => but got keyword case`

martin-ilgner:
Should I also try to understand the Typer and use the parsed TypeAnnotation?

Edit: Too complicated

PR #1346, opened 2026-03-25
fix: add missing io/promise import in IO tour (#1345)

L4XB:
Fixes #1345

## Problem
The IO tour (`examples/tour/IO.effekt.md`) uses `promise()` and `.await` in `swapConcurrently()` and `concurrentSwap()`, but never imports `io/promise`. This causes:

  > Cannot find a function named `promise`.

  ## Fix
  Added `import io/promise` to:
  - the visible `effekt:ignore` import block (documentation for users)
  - the hidden `effekt:prelude:hide` block (actual execution context)

PR #1367, opened 2026-04-10
import io/promise to io tour (#1345)

LTeuse:
(Rewrite PR #1346, to fix issue #1345)
Added import `io/promise` to:
- the visible `effekt:ignore` import block (documentation for users)
- the hidden `effekt:prelude:hide` block (actual execution context)


jiribenes:
This really needs to be tested on the website, it likely acts differently than the local runs :/

marvinborner (approved):
Works for me, good work!

PR #1369, opened 2026-04-10
Fix position for unreachable case warning in irrefutable 'val-else' #1339

LTeuse:
### Problem
```scala
def main() = {
  val (a, b) = (1, 2) else panic("unreachable")
}
```
report `Unreachable case` at `main` instead of else (see #1339)

---

### Fix
- `Parser.scala`: Changed the `endPos` of the Match expression to get position after the `else` branch. Removed `Synthesized` flag from the span so `ErrorReporter` do not skips it.
- `Exhaustivity.scala`: Changed the reporter to anchor the error to the body of clauses instead of whole match pattern

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/ParserTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/ParserTests.scala
@@ -581,6 +581,45 @@ class ParserTests extends munit.FunSuite {
     assertEquals(parseStmts(source.content), expected)
   }
 
+  test("val-else") {
+    val (source, pos) =
+      raw"""val (left, right) = list else { return 0 }; return left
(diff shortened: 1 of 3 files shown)


jiribenes (changes requested):
Please add some tests for this 🙏 

`ParserTests` have a nice DSL for spans that you could use to verify them:
https://github.com/effekt-lang/effekt/blob/48eb56bf624868cf481a67d4de6f78488472ea4b/effekt/jvm/src/test/scala/effekt/ParserTests.scala#L223-L238

EDIT: The PR currently seems to have broken the following test:
`effekt.ParserTests.Pattern-matching val parses with correct span  0.154s `

jiribenes (approved):

jiribenes:
Thanks!

Issue #891, opened 2025-03-18
Spurious "Type A escapes through type List[A] inferred for the result of the if statement."


jiribenes:
I've seen this multiple times locally, but thought that it's a LSP-related issue.

But now it seems to have broken the build on `master`:
https://github.com/effekt-lang/effekt/actions/runs/13927195740/job/38974717008#step:11:932

For posterity, here's an excerpt from the log:
```
==> X effekt.JavaScriptTests.examples/neg/multiple_main.effekt (js)  1.961s munit.FailException: /home/runner/work/effekt/effekt/effekt/jvm/src/test/scala/effekt/EffektTests.scala:145
144:           |----------------------------------------------
145:           |${ messages.map(m => messaging.formatMessage(m)).distinct.mkString("\n\n") }""".stripMargin)
146:
Error: ot find an message [error] line 1: Multiple main functions

Compilation resulted in the following messages
----------------------------------------------
Error:  /home/runner/work/effekt/effekt/libraries/common/list.effekt:382:3: Type A escapes through type List[A] inferred for the result of the if statement.


Error:  /home/runner/work/effekt/effekt/libraries/common/list.effekt:382:3: Cannot compile dependency: /home/runner/work/effekt/effekt/libraries/common/list
```

PR #1370, opened 2026-04-10
Track qualified signature for extern definitons

timsueberkrueb:
This PR replaces the extern body for the "vm" target by an automatically generated `QualifiedSignature`.


timsueberkrueb:
Currently failing because for some reason, we get a lookup of `infixGte` with another id than is collected initially.
```
infixGte_220 -> Some(Builtin(effekt::infixGte(Int, Int),effekt.core.vm.Builtin$package$$$Lambda/0x00007c5b473ad1a0@5b83c5c9))
infixGte_5885 -> None
infixGte_5885
```

timsueberkrueb:
TODO @timsueberkrueb clean up core parser code path

effekt/shared/src/main/scala/effekt/core/Tree.scala
```diff
@@ -134,9 +135,36 @@ case class Property(id: Id, tpe: BlockType) extends Tree
...
+
+  def apply(name: Id, tparams: List[Id], vparams: List[ValueParam], bparams: List[BlockParam]): QualifiedSignature = {
+    import effekt.symbols.ErrorMessageInterpolator
+    val tps = if (tparams.isEmpty) "" else s"[${tparams.map(tp => pp"${tp}").mkString(", ")}]"
+    val valueParams = vparams.map { p => pp"${p.tpe}" }.mkString(", ")
+    val vps = if valueParams.isEmpty then "" else s"($valueParams)"
+    val bps = bparams.map { b => s"{${pp"${b.tpe}"}}" }.mkString("")
+    val ps = if (vps.isEmpty && bps.isEmpty) "()" else s"$vps$bps"
+    QualifiedSignature(s"${name}$tps$ps")
+  }
```
timsueberkrueb: This is an unfortunate duplication, but we have one invocation from core parser (with core trees needing to be printed) and one from source-to-core transformer with source symbols being printed.
We could always print the transformed core, but unfortunately, this prints the Barendregt ids, so the resulting identifier is not stable.
Therefore, the code path from core parser is useless at the moment.

timsueberkrueb: I replaced this with a "not-yet-implemented" placeholder. Open to better suggestions.

marzipankaiser (approved):
Once you decided on a way to handle the core parser, this looks good to me.

Also, since we don't use this in core tests so far, core.Parser might not be the most important thing (so merging this as-is and solving the core.Parser case some other day would be fine, too, IMHO)

(Tests are failing in other PRs with the same error, too - this might just be flaky 🤔 - maybe re-run them)

timsueberkrueb:
CI failing with

```
[180-0] ==> X effekt.JavaScriptTests.examples/neg/issue512.effekt (js)  0.311s munit.FailException: /home/runner/work/effekt/effekt/effekt/jvm/src/test/scala/effekt/EffektTests.scala:159
[180-0] 158:           |----------------------------------------------
[180-0] 159:           |${ messages.map(m => messaging.formatMessage(m)).distinct.mkString("\n\n") }""".stripMargin)
[180-0] 160:
Error:  Could not find an message [error] line 7: r1
[180-0] 
[180-0] Compilation resulted in the following messages
[180-0] ----------------------------------------------
Error: [180-0] [error] /home/runner/work/effekt/effekt/libraries/common/list.effekt:363:3: Type B escapes through type List[B] inferred as return type.
[180-0] 
[180-0] 
Error: [180-0] [error] /home/runner/work/effekt/effekt/libraries/common/list.effekt:363:3: Cannot compile dependency: /home/runner/work/effekt/effekt/libraries/common/list
```

Is this yet another flaky issue?

jiribenes:
Yep. https://github.com/effekt-lang/effekt/issues/891

PR #1371, opened 2026-04-10
Move `readDouble` into the `scanner` module

### LINKED ISSUE #1313: Move 'json::readDouble' to 'scanner', refactor it, and test it
Many student projects had to use FFI to convert a String to a Double.
However, there _is_ a `readDouble` in the `json` module already that could "just" be moved into `scanner` (and refactored to use `fail` only like `readInteger` does):
https://github.com/effekt-lang/effekt/blob/da946a7cc7ed7d5ff95d9329210d97b656f31d65/libraries/common/json.effekt#L93-L125

Moreover, we should probably test it by putting a bunch of interesting examples into a new `examples/stdlib/scanner/readDouble.effekt` file.

DennisKleinhans:
Moved the `readDouble` function from the `json` module into the scanner module


jiribenes (approved):
If the tests pass, LGTM

PR #1372, opened 2026-04-10
Extern type representations

### LINKED ISSUE #1132: Extern type representations
For an extern type, allow specifying the target representation.

For example in #1058, we have a TCP handle as the following, since `extern type` is always a `%Pos` on LLVM:
```scala
type Connection = Int
```

Instead, I'd prefer to specify external type representation directly:
```
extern type Connection =
  llvm "%Int"
  js "number"
```

Which is then used/usable to specify the type:
```scala
(c: Connection) ~> (%Int c)           // llvm
(c: Connection) ~> (c /* : number */) // js
```

Ideally, we'd reuse the checks used by target-specific externs (`extern def ... = js "..."`) for exhaustivity, etc.

marzipankaiser:
This parses right hand sides for external types and threads them through to the backends.
In the backends, currently just generates a comment with the type declaration.

Resolves #1132.

Drivebys:
- Disallows `extern def ...() =` without right hand sides and with a `=` (allows it without a `=`)
- Does reachability analysis and deadcode for externs and extern types
- Some other minor stuff

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
+++ b/effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
@@ -52,7 +52,6 @@ class Deadcode(reachable: Map[Id, Usage])
           // We need to keep "show", "showBuiltin" & "infixPlusPlus" for generating show definitions (see #1123)
           case e: Extern.Def if used(e.id) || List("show", "showBuiltin", "infixPlusPlus").contains(e.id.name.name) => e
           case e: Extern.Include => e
-          // We currently do not have usage information on extern types, so we have to keep all of them
           case e: Extern.Data if used(e.id) => e
         },
         // drop unreachable definitions
--- a/effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
+++ b/effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
@@ -43,15 +43,17 @@ class Deadcode(reachable: Map[Id, Usage])
   override def rewrite(m: ModuleDecl): ModuleDecl = m match {
     case ModuleDecl(path, includes, declarations, externs, definitions, exports) =>
       ModuleDecl(path, includes,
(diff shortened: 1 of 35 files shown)


effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
```diff
@@ -43,15 +43,17 @@ class Deadcode(reachable: Map[Id, Usage])
...
+          case d if used(d.id) => rewrite(d)
+        },
         // drop unreachable externs
         m.externs.collect {
           // We need to keep "show", "showBuiltin" & "infixPlusPlus" for generating show definitions (see #1123)
           case e: Extern.Def if used(e.id) || List("show", "showBuiltin", "infixPlusPlus").contains(e.id.name.name) => e
           case e: Extern.Include => e
           // We currently do not have usage information on extern types, so we have to keep all of them
-          case e: Extern.Data => e
+          case e: Extern.Data if used(e.id) => e
```
jiribenes: Is the comment now deprecated? 👀

marzipankaiser: Yes.
```suggestion
          case e: Extern.Data if used(e.id) => e
```

marzipankaiser:
Note: TODOs that we should address at some other time:
- [ ] Use the right hand sides in LLVM (at least)
    - For this, we'd need to come up with some syntax *within* the extern string (or allow multiple strings somehow) to get name, size, erasers etc
    - See #1378
- [ ] Use `extern type`s for some of the base types like Double (but how to handle Literals?) (was tried previously in #213)
- [x] Allow right-hand sides for `extern interface`s, too (this should be slightly easier now, but is not yet there)
    - [x] Note: We do not preserve `extern interface` in `core.Transformer`, so they are not passed on to the backends.
- [ ] Make the right-hand sides more consistent (easier once we actually use them for more than comments)

Issue #1374, opened 2026-04-13
Generate generic representation


phischu:
Currently, we generate a definition of `show` for each datatype it is used at (#1123).

We want to replace that with generating definitions `repr` for getting a tree representation of values of datatypes.

Using this we can then not only write `show` but also `equals` and `format` and perhaps other things.

Should help us address #334

jiribenes:
I have a WIP version of this, the biggest blocker was the representation.
Originally, I used the "Down with Show" posts (https://harry.garrood.me/blog/down-with-show-part-3/) to represent it as a labelled rose tree, but that presented a number of challenges:
1. we'd need smart constructors even for the compiler
2. we'd need to define it _after_ we have defined `list`, which complicates things
3. the compiler would need a lot of magic variables: at the very least: `repr`, `reprBuiltin`, `reprMakeApp`, `reprOpaque` and `reprConsField`, _and_ it would need to construct lists in the pass which is really annoying

The alternative is to use effects -- `.repr` is then actually creating a labelled stream (equivalent to a rose tree):
```scala
/// Rose tree as a datatype
record Tree[A](node: A, children: List[Tree[A]])

/// Rose tree as an effect v1
effect rose[A, R](node: A) { children: => R / rose[A, R] }: R

/// Rose tree as an effect v2
effect rose[A, R](node: A) { children: => R }: R
// perhaps swap to `Unit`, `R` here helps to make the implementation correct

namespace Repr {
  type Label {
    // primitives
    Int(n: Int)
    ...
    // constructors
    App(name: String)
    Opaque(name: String)
    Literal(name: String)
    ...
```

Then we can `reify` it into a `Tree` which can be `reflect`ed back.
And methods like `.toString` (`.repr.toString`) and `.pretty` (`.repr.pretty`) can probably work on the stream, so we needn't reify it for the compiler.

The nice thing about this is that the compiler only has to know the `Repr::Label` type and the `rose` effect, and both can be directly in `effekt.effekt` w/o dependency issues, whereas `Tree`, `reify`, `reflect` etc can be in some more advanced `repr.effekt` module :)

jiribenes:
My timeline would be to first add a new `Repr` pass (I have a prototype locally), then use both the old `Show` pass and the new `Repr` pass, figure out the ergonomics (perhaps requires `inline`), only then remove the `Show` pass.

PR #1379, opened 2026-04-15
Reduce special cases in Deadcode

jiribenes:
Since the optimizer calls `Deadcode(main)` explicitly, this also means that all unused show-related functions get removed when optimising :)

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
+++ b/effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
@@ -49,8 +49,7 @@ class Deadcode(reachable: Map[Id, Usage])
         },
         // drop unreachable externs
         m.externs.collect {
-          // We need to keep "show", "showBuiltin" & "infixPlusPlus" for generating show definitions (see #1123)
-          case e: Extern.Def if used(e.id) || List("show", "showBuiltin", "infixPlusPlus").contains(e.id.name.name) => e
+          case e: Extern.Def if used(e.id) => e
           case e: Extern.Include => e
           case e: Extern.Data if used(e.id) => e
           case e: Extern.Interface if used(e.id) => e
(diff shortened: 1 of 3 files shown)


effekt/shared/src/main/scala/effekt/core/optimizer/Deadcode.scala
```diff
@@ -80,8 +79,11 @@ object Deadcode extends Phase[CoreTransformed, CoreTransformed] {
     input match {
       case CoreTransformed(source, tree, mod, core) =>
         val term = Context.ensureMainExists(mod)
+        // when ran "directly" (i.e., before the 'Show' pass),
+        // we add the functions required by subsequent passes
+        val required = Set(term) ++ Show.requiredNames(core)
```
jiribenes, line 84: I fully intend this to be
```scala
val required = Set(term) ++ Show.requiredNames(core) ++ Repr.requiredNames(core)
```

when #1374 gets done.

effekt/shared/src/main/scala/effekt/generator/chez/ChezSchemeCPS.scala
```diff
@@ -37,7 +37,7 @@ class ChezSchemeCPS extends Compiler[String] {
     Frontend andThen Middleend
   }
 
-  lazy val Optimized = allToCore(Core) andThen Aggregate andThen Optimizer andThen core.Show map {
+  lazy val Optimized = allToCore(Core) andThen Aggregate andThen Deadcode andThen core.Show andThen Optimizer map {
```
jiribenes, line 40: cc @martin-ilgner: are there any problems with doing this? what was the rationale for the original ordering & composition of the passes? 🤔

jiribenes:
I'll merge this. The worst case scenario is that Effekt Plots show that my reshuffled Chez CPS pipeline is ridiculous, and in that case, I'll revert.

effekt/shared/src/main/scala/effekt/generator/chez/ChezSchemeCPS.scala
martin-ilgner: I think I copy pasted from js backend

PR #1380, opened 2026-04-15
Name-based implicits in Effekt

marzipankaiser:
This implements implicits similar to
> Daan Leijen, Tim Whiting: Syntactic Implicit Parameters with Static Overloading

It also implements their use to get source positions in the code.

### Syntax and Intended Behaviour
Putting a `?` in front of a value/block parameter makes it potentially implicit. If *trailing* implicit parameters are missing, an argument is generated as follows (e.g.):
- value argument called `$x`: `$x`
- value argument called `sourcePosition`: `SourcePosition("somefilename.effekt",10,3,10,6)` (corresponding to the source position of the called function, ~~without~~ with argument list)
- value argument called `callId`: A statically unique integer for *this* call (unique within an Effekt compiler run).
- value argument of type `(...) => ... at {...}`: instantiated to `box {...}` with whatever the block argument would be instantiated to.
- block argument called `$foo` of type `(A){B=>C} => D` : `{ (a: A){b: B => C} => $foo(a){b} }` (eta-expanded to allow for this call to also receive implicit arguments)
- block argument with an interface type, called `$x`: `$x`

### Implementation strategy
- In `Namer`, 
    - find the set of all implicit parameters that any of the overloads could require (i.e. for all functions of that name, all implicit parameters) and generate the source for them already.
    - then name-resolve them, but collect errors and just store them (for now)
- in `Typer`,
    - instantiate the arguments (i.e., currently, hardcoded copy of the block literal and corresponding annotations, or just copying them), refreshing them and typechecking them against their respective parameters
        - only happens if at this point, no errors were reported in this case (otherwise, aborts)
    - annotate the instantiated arguments
- in `ExplicitCapabilities`, actually change the source to pass the annotated arguments explicitly.

### TODO and limitations
- [x] ~~Check if this works for more than the existing examples (find at least the worst bugs)~~ The tests should be testing a good part of possible uses now.
- [x] This *will* cause *Typer* to run indefinitely or stack-overflow if the search is not terminated by a type/name error. We could try to catch those cases somehow, but this is nontrivial (and will reject valid programs).
    - Opinions?
    - The recursion is ~~quite indirect (and not actual recursion) in typer, and~~ (we can check for recursive uses when checking implicit arguments after instantiating them - this is separate now anyway) at a point where we don't necessarily have the concrete types (due to unification), so checking for "decreasing type size" or similar is potentially hard.
    - Now aborts if, at the same name, we have 10 levels where the expected type does not get smaller.
- [x] Syntax bikeshedding
- [ ] If merging this, we might want to change #1381 / #1123 s.t. they provide only the definitions for base and recursive cases (not for all types used etc).


effekt/shared/src/main/scala/effekt/symbols/symbols.scala
```diff
@@ -203,12 +204,52 @@ enum Binder extends TermSymbol {
...
+  }
+  case class ImplicitVar(kind: String, name: String, content: source.Var) extends ImplicitStencil[source.Var]
+  case class SourcePosition(content: source.Call) extends ImplicitStencil[source.Call] {
+    def name = "sourcePosition"
+    def kind = "value argument"
+    override def explanation =
+      Some(
+        """Implicit sourcePosition will call
+          |  SourcePosition(file, start_line, start_col, end_line, end_col)
+          | with the respecitve values for the source position of the call.
```
jiribenes: found a typo while skimming changes
```suggestion
          | with the respective values for the source position of the call.
```

marzipankaiser: thx, will fix.

marzipankaiser:
@jiribenes You might be interested in the change from the last commit: I changed it so implicts of boxed type `(...) => R at {...}` will be instantiated to `box {$foo}` where `$foo` is what a block argument of the same name would be instantiated to, so `?compare: (A, A) => Ordering at {}` will be filled with `box { (arg0: A, arg1: A) => compare(arg0, arg1) }`.

effekt/shared/src/main/scala/effekt/source/GenerateImplicitArgs.scala
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  /**
+   * Called from [[Typer]] to get a fresh instance of the given implicit value argument.
+   *
+   * Also annotates all symbols for the returned code correctly where necessary.
+   */
+  def instantiateImplicitValue(v: ImplicitStencil, tpe: symbols.ValueType)(using Context): source.ValueArg = {
+    def intArg(v: Long, name: Option[String] = None): source.ValueArg =
+      ValueArg(name, Literal(v, builtins.TInt, Span.missing), Span.missing)
```
marzipankaiser, line 306: Writing name-resolved source Terms by hand is somewhat annoying.
If we do so more (i.e. if we generate more stuff in source than we do here already), we might want to consider extracting this function, `generateResolvedId` etc into a somewhat nice library for it. I did not do for the relatively small amount of code here, but decided to at least extract some helpers.

effekt/shared/src/main/scala/effekt/source/GenerateImplicitArgs.scala
```diff
@@ -0,0 +1,371 @@
...
+
+  // Instantiation (during typer)
+  // ============================
+
+  /**
+   * Counter for integer ids for calls in the source code, to be passed as callId.
+   */
+  private var nextCallId: Long = 0
+
+  private def generateResolvedId(sym: symbols.Symbol)(using Context): (source.IdDef, source.IdRef) = {
```
marzipankaiser, line 227: (Here's `generateResolvedId`, mentioned in the other comment)

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -926,7 +976,7 @@ object Namer extends Phase[Parsed, NameResolved] {
     case _: source.FunctionType => s"a second-class function type ${t.sourceOf}"
     case _: source.Effectful => s"a type-and-effect annotation ${t.sourceOf}"
 
-    // THESE TWO SHOULD NEVER BE USER-VISIBLE!
+    // THESE THREE SHOULD NEVER BE USER-VISIBLE!
```
b-studios:
```suggestion
    // THESE TWO SHOULD NEVER BE USER-VISIBLE!
```

effekt/shared/src/main/scala/effekt/Typer.scala
```diff
@@ -1163,19 +1163,19 @@ object Typer extends Phase[NameResolved, Typechecked] {
...
+      case sym: Operation => (List(sym), Map.empty[BlockSymbol, ImplicitContext])
       case s => Context.panic(s"Not a valid method: ${s} : ${s.getClass.getSimpleName}")
     }
 
     val Result(recvTpe, recvEffs) = checkExprAsBlock(receiver, None)
 
     val interface = recvTpe.asInterfaceType
     // filter out operations that do not fit the receiver
-    val candidates = methods.filter(op => op.interface == interface.typeConstructor)
+    val candidates = methods.filter( op => op.interface == interface.typeConstructor )
```
b-studios, line 1178:
```suggestion
    val candidates = methods.filter(op => op.interface == interface.typeConstructor)
```

b-studios (approved):
Wow!

PR #1381, opened 2026-04-15
Add a 'Repr' pass for generating generic representations

### LINKED ISSUE #1374: Generate generic representation

jiribenes:
Resolves #1374

It's still pretty ad-hoc and I'm planning to continue working on this during downtimes :)


effekt/shared/src/main/scala/effekt/core/Repr.scala
```diff
@@ -0,0 +1,345 @@
...
+      val labelTpe = ValueType.Data(labelDecl.id, Nil)
+      val ifaceTpe: BlockType.Interface = BlockType.Interface(ifaceDecl.id, List(labelTpe, Type.TUnit))
+
+      val tparamSubst: Map[Id, ValueType] = (ifaceDecl.tparams zip List(labelTpe, Type.TUnit)).toMap
+      val opTpe = Type.substitute(op.tpe, tparamSubst, Map.empty)
+
+      val capId = Id("$rose")
+      val capParam = BlockParam(capId, ifaceTpe, Set.empty)
+
+      RoseMetadata(op.id, ifaceTpe, opTpe, capParam)
```
jiribenes, line 80: This is wrong, it generates a single param id that gets reused in multiple functions, which makes the LLVM backend uneasy

PR #1383, opened 2026-04-17
Replace partial functions by dynamic dispatch

b-studios:
In this PR I am trying to get rid of more usages of the macro by explicating the traversal in `Tree.Rewrite` manually. Instead of specifying the special cases in partial functions, now users have to override the specific `rewrite` method for the nonterminal of interest and call `super.rewrite(...)` for the homomorphic cases.


phischu (approved):
Yes, good!

PR #1384, opened 2026-04-21
Experiment: annotate hot paths on LLVM

jiribenes:
My local devenv is broken so I can't benchmark this; we should really benchmark each commit individually against `main` to see if we get some perf.

There are three main changes:
1. ~~assume RC ≥ 0 (not obvious to LLVM since we use a whole `i64`)~~ _EDIT: removed this via a force push for now_, [diff here](https://github.com/effekt-lang/effekt/compare/f881e45e8318c8f912d7f65976907ee537913dd1..001f6b1c56d20465cd09612c492124d260587f4d)
2. when switching on RC, assume that it's 0 for the hot path to be quicker
3. ~~assume that a to-be-erased value is non-null~~ _EDIT: not true, we even emit them `null` sometimes..._ [diff here](https://github.com/effekt-lang/effekt/compare/001f6b1c56d20465cd09612c492124d260587f4d..b6f26efdd54f2f38aa365d883956e19f7896e891)
4. dumping a bunch of "this function doesn't do bad stuff" annotations all over the place

I _think_ these are all valid assumptions / hot paths that we should prefer.
_EDIT: The crossed out changes above kept the CI in a busy loop somehow 🔁_

### DIFF
--- a/libraries/llvm/rts.ll
+++ b/libraries/llvm/rts.ll
@@ -324,7 +324,7 @@ define private %StackPointer @stackDeallocate(%Stack %stack, i64 %n) {
     ret %StackPointer %newStackPointer
 }
 
-define private i64 @nextPowerOfTwo(i64 %x) {
+define private i64 @nextPowerOfTwo(i64 %x) nounwind nosync willreturn readnone {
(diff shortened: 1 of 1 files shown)


jiribenes:
subsumed by #1446

PR #1386, opened 2026-04-24
Improve buggy stdlib bundling on ScalaJS

jiribenes:
ScalaJS' pickler has some limit on string sizes which the recent stdlib additions changes overshoot by a lot on my computer.
This was preventing me from building the project locally.

~~The fix is horrible: we just chunk the string instead. Perhaps there's a better fix.~~

We generate a `stdlib-resources.json` file which is then loaded by JS, thus removing the `!!!MULTILINEMARKER!!!` hack and generating huge Scala files altogether. :)

This ought to:
- [ ] be tested with the website
- [x] ~~be ported to `mill`~~ (this is ~~also solvable~~ _solved_ by #1387)
- [ ] be reconciled with a bundler or something like it to ensure that the json is bundled

---

```
[info] compiling 111 Scala sources to /nix/var/nix/builds/nix-63932-2468348747/effekt-src/effekt/js/target/scala-3.3.6/classes ...

  unhandled exception while running pickler on /nix/var/nix/builds/nix-63932-2468348747/effekt-src/effekt/js/target/scala-3.3.6/src_managed/main/Resour>

  An unhandled exception was thrown in the compiler.
  Please file a crash report here:
  https://github.com/scala/scala3/issues/new/choose
  For non-enriched exceptions, compile with -Xno-enrich-error-messages.


     while compiling: /nix/var/nix/builds/nix-63932-2468348747/effekt-src/effekt/js/target/scala-3.3.6/src_managed/main/Resources.scala
        during phase: pickler
                mode: Mode(ImplicitsEnabled,ReadPositions)
     library version: version 2.13.16
    compiler version: version 3.3.6
            settings: -Werror true -Xsemanticdb true -bootclasspath /nix/var/nix/builds/nix-63932-2468348747/tmp.aM6MNA3Y1w/project/.coursier/https/rep>

[error] ## Exception when compiling 111 sources to /nix/var/nix/builds/nix-63932-2468348747/effekt-src/effekt/js/target/scala-3.3.6/classes
[error] java.lang.ArrayIndexOutOfBoundsException: arraycopy: last destination index 3955373 out of bounds for char[2097152]
[error] java.base/java.lang.System.arraycopy(Native Method)
[error] scala.Array$.copy(Array.scala:110)
```


jiribenes:
I might have a better solution: generate a JSON `{ ..."<path>": "<file contents>" }` and then import it from JavaScript.

build.sbt
```diff
@@ -275,8 +278,8 @@ lazy val effekt: CrossProject = crossProject(JSPlatform, JVMPlatform).in(file("e
 
     assembleJS := {
       (Compile / clean).value
-      (Compile / compile).value
-      val jsFile = (Compile / fullOptJS).value.data
+      (Compile / fullLinkJS).value
+      val jsFile = (Compile / fullLinkJS / scalaJSLinkerOutputDirectory).value / "main.js"
       val outputFile = (ThisBuild / baseDirectory).value / "out" / "effekt.js"
       IO.copyFile(jsFile, outputFile)
     },
```
jiribenes, line 285: this doesn't look like it actually embeds the `stdlib-resources.json`, we might need a bundler or also copy that file onto the website...

PR #1388, opened 2026-04-27
Allow qualified use mentioning current module

### LINKED ISSUE #1337: Allow qualified use of a function in current module
Currently, it's impossible to use a function from the current module in a qualified way.
I've encountered this mostly in collection interfaces like `list`: each collection has a `collect` function, and we have an unwritten convention of using `collect` in a qualified manner, i.e. as `list::collect`.
But this convention can't be followed when I want to use `list::collect` in the `list` module itself as that's a error in Namer.

Thus I'd like to propose to allow such qualified use: in a module `foo` with a function `bar`, I can use _both_ `bar()` and `foo::bar()` to refer to the function.

Small repro:
```scala
module foo

def bar(): Int = 42

def quux() = {
  val n = bar() // currently works!
  val m = foo::bar() // error: Cannot find a function named `foo::bar`
  n + m
}
```

jiribenes:
Resolves #1337, quoting:

> Currently, it's impossible to use a function from the current module in a qualified way. I've encountered this mostly in collection interfaces like `list`: each collection has a `collect` function, and we have an unwritten convention of using `collect` in a qualified manner, i.e. as `list::collect`. But this convention can't be followed when I want to use `list::collect` in the `list` module itself as that's a error in Namer.
> 
> Thus I'd like to propose to allow such qualified use: in a module `foo` with a function `bar`, I can use _both_ `bar()` and `foo::bar()` to refer to the function.
> 
> Small repro:
> ```scala
> module foo
> 
> def bar(): Int = 42
> 
> def quux() = {
>   val n = bar() // currently works!
>   val m = foo::bar() // error: Cannot find a function named `foo::bar`
>   n + m
> }
> ```


phischu (approved):
Yes! This is very important!

PR #1389, opened 2026-04-27
Use iterative deepening to speed up implicit resolution

marzipankaiser:
When resolving certain forms of implicits (example below), typer takes really long to resolve them.
This can be aleviated by using iterative deepening up until the maximum depth (similar to Koka, but with a maximum, so we always terminate).

## Example
Warning: This will take *very* long to typecheck on current main.
```
type TypeTag[T] {
  Int()
  Bool()
  String()
  List[A](elem: TypeTag[A])
  Option[A](elem: TypeTag[A])
  Pair[F,S](fst: TypeTag[F], snd: TypeTag[S])
}

def typeTag(): TypeTag[Int] = Int()
def typeTag(): TypeTag[Bool] = Bool()
def typeTag(): TypeTag[String] = String()
def typeTag[A]{ ?typeTag: => TypeTag[A] }: TypeTag[List[A]] = List(typeTag())
def typeTag[A]{ ?typeTag: => TypeTag[A] }: TypeTag[Option[A]] = Option(typeTag())
def typeTag[F,S]{ ?typeTag: => TypeTag[F] }{ ?typeTag: => TypeTag[S] }: TypeTag[(F,S)] = Pair(typeTag(), typeTag())

def main() = {
  val x: TypeTag[List[(Bool, Int)]] = typeTag()
  ()
}
```
(Note that this is ambiguous, but still...)

## Alternatives
Explicitly passing the type paramters to the `TypeTag` constructors does solve this in this instance, too.


marzipankaiser:
Feel free to merge if you find an actual use case / more people run into this.

jiribenes:
It might be worth revisiting this after we add `?compare` to maps and sets, and can thus _compare_ compiler perf before this and after? 👀

marzipankaiser:
Merged into #1396 .

PR #1392, opened 2026-04-28
Add support for qualified declarations

### LINKED ISSUE #1223: Add qualified declarations
Allow the following
```scala
def foo::bar(...) = <>
```

to stand for:
```
namespace foo { def bar(...) = <> }
export foo::bar // we can't write this yet in the surface language!
```

So that you can call `bar` with both `foo::bar` and "just" `bar`.

Later on, we could enforce this for overloaded terms (so that there's always at least one "distinguishing" call),
such as `int::show`, `string::show`, ....

---

This might need some tweaks in Lexer & Parser, then "unwrap" in Namer, similarly to what https://github.com/effekt-lang/effekt/pull/1092 did:
```scala
          // define in namespace ...
          Context.namespace(interfaceId.name) {
            Context.define(id, opSym)
          }
          // ... and bind outside
          Context.bind(opSym)
          opSym
```

jiribenes:
Resolves #1223, especially useful for overloading disambiguation, see examples.

This will likely need a syntax highlighter update in effekt-vscode.


effekt/shared/src/main/scala/effekt/source/Tree.scala
```diff
@@ -255,11 +255,14 @@ sealed trait Id extends Tree {
...
-case class IdDef(name: String, span: Span) extends Id {
+case class IdDef(path: List[String], name: String, span: Span) extends Id {
   def clone(using C: Context): IdDef = {
-    IdDef(name, span)
+    IdDef(path, name, span)
   }
 }
+object IdDef {
+  def apply(name: String, span: Span): IdDef = IdDef(Nil, name, span)
+}
```
jiribenes, line 266: this is for backcompat

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -1048,14 +1050,31 @@ trait NamerOps extends ContextOps { Context: Context =>
...
+
+    def defineInPath(p: List[String]): Unit = p match {
+      case Nil => baseCase
+      case ns :: rest => namespace(ns) { defineInPath(rest) }
+    }
+
+    // define in all suffixes
+    path.tails.foreach { suffix =>
+      defineInPath(suffix)
+    }
```
jiribenes, line 1065: it might be a good idea to add a `namespace(nss..)` to make this faster
in general, this is probably super wasteful, we keep switching namespaces just to add a single thing

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -1048,23 +1052,43 @@ trait NamerOps extends ContextOps { Context: Context =>
...
+  }
+
   private[namer] def define(id: Id, s: TermSymbol): Unit = {
     assignSymbol(id, s)
-    scope.define(id.name, s)
+    val path = id match { case d: source.IdDef => d.path; case _ => Nil }
```
jiribenes: we should likely have some helper on `Id` for this

effekt/shared/src/main/scala/effekt/Namer.scala
```diff
@@ -1048,23 +1052,43 @@ trait NamerOps extends ContextOps { Context: Context =>
...
+    defineInPaths(path, scope.define(id.name, s))
   }
 
   private[namer] def bind(s: Capture): Unit = bind(s.name.name, s)
 
   private[namer] def bind(name: String, s: Capture): Unit = scope.define(name, s)
 
-  private[namer] def bind(s: TermSymbol): Unit = scope.define(s.name.name, s)
+  private[namer] def bind(s: TermSymbol): Unit = {
+    val path = s.name match { case QualifiedName(prefix, _) => prefix; case _ => Nil }
```
jiribenes: dtto for symbols

phischu (approved):
Nice, like I said I would rather have the name not be available unqualified, but ok.

jiribenes:
> Nice, like I said I would rather have the name not be available unqualified, but ok.

Since we don't have `export`, we couldn't use it for overloading like in the attached examples. :(

libraries/common/effekt.effekt
```diff
@@ -128,11 +128,16 @@ extern def panic(msg: String) at io: Nothing =
...
   chez "(void)"
   llvm "ret %Pos poison"
 
+/// Returns a textual representation of a value for debugging purposes.
+/// The output is backend-specific, works only on JS and Chez backends.
+///
+/// Prefer using `show` instead when possible.
 extern def genericShow[R](value: R) at {}: String =
   js "$effekt.show(${value})"
   chez "(show_impl ${value})"
```
jiribenes, line 143: Here's a drive-by I made, now error messages "which show did you mean?" should mention `effekt::*generated*::show` as one of the options.

jiribenes: Needs #1395 to actually be able to see the docs on hover though 😅

PR #1394, opened 2026-04-28
Add debug flag and separate type checking and inference

b-studios:
In sbt you can now call
```
debug
```
to trigger the debug mode
and
```
release
```
to trigger the release mode.

`effekt.util.Debug` contains a few debugging helpers that are disabled during release:

```scala
inline def debug(inline f: => Unit): Unit =
  inline if DEBUG then f else ()

inline def debug[T](inline f: => T)(inline otherwise: => T): T =
  inline if DEBUG then f else otherwise

inline def assert(inline assertion: Boolean): Unit =
  inline if DEBUG then Predef.assert(assertion)

inline def assert(inline assertion: Boolean, inline msg: => String): Unit =
  inline if DEBUG then Predef.assert(assertion, msg)
```

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/core/VMTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/core/VMTests.scala
@@ -236,7 +236,7 @@ class VMTests extends munit.FunSuite {
 
   val are_we_fast_yet: Seq[(File, Option[Summary])] = Seq(
     examplesDir / "benchmarks" / "are_we_fast_yet" / "bounce.effekt" -> Some(Summary(
-      staticDispatches = 20202,
+      staticDispatches = 15202,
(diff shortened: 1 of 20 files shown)


effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -678,15 +678,15 @@ class VMTests extends munit.FunSuite {
...
       allocations = 106,
       closures = 23,
       variableReads = 164,
       variableWrites = 51,
-      resets = 28,
-      shifts = 0,
-      resumes = 0
+      resets = 29,
+      shifts = 3,
+      resumes = 3
```
b-studios, line 689: This worries me a bit, but it might be unavoidable since the size computation changed a bit

PR #1395, opened 2026-04-28
LSP: Show documentation on hover for externs

jiribenes:
For some reason, we currently don't + the code is a bit of a mess.
Let's support this, docs are cool!

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Intelligence.scala
+++ b/effekt/shared/src/main/scala/effekt/Intelligence.scala
@@ -113,24 +113,22 @@ trait Intelligence {
   } yield decl
 
   def getDefinitionOf(s: Symbol)(using C: Context): Option[Tree] = s match {
-    case u: UserFunction => Some(u.decl)
-    case u: Binder       => Some(u.decl)
-    case d: Operation    => C.definitionTreeOption(d.interface)
-    case a: Anon         => Some(a.decl)
-    case m: Module       => Some(m.decl)
+    case u: UserFunction   => Some(u.decl)
+    case u: Binder         => Some(u.decl)
+    case d: Operation      => C.definitionTreeOption(d.interface)
+    case a: Anon           => Some(a.decl)
+    case m: Module         => Some(m.decl)
+    case e: ExternFunction => Some(e.decl)
+    case t: TypeSymbol     => Some(t.decl)
+    case t: TermSymbol     => Some(t.decl)
     case u => C.definitionTreeOption(u)
   }
 
   def getDocumentationOf(s: Symbol)(using C: Context): Doc =
-    getDefinitionOf(s).asInstanceOf[Option[Any]] match {
-      case Some(p: Product) =>
-        p.productElementNames.zip(p.productIterator)
-          .collectFirst {
-            case ("doc", Some(s: String)) => s
-            case ("info", source.Info(Some(s: String), _, _)) => s
-          }
-      case _ => None
-  }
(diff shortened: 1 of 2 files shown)


effekt/shared/src/main/scala/effekt/Intelligence.scala
```diff
@@ -113,24 +113,22 @@ trait Intelligence {
...
+    case u: UserFunction   => Some(u.decl)
+    case u: Binder         => Some(u.decl)
+    case d: Operation      => C.definitionTreeOption(d.interface)
+    case a: Anon           => Some(a.decl)
+    case m: Module         => Some(m.decl)
+    case e: ExternFunction => Some(e.decl)
+    case t: TypeSymbol     => Some(t.decl)
+    case t: TermSymbol     => Some(t.decl)
     case u => C.definitionTreeOption(u)
   }
```
jiribenes, line 120: This is still whack, some of the cases must overlap, right?

effekt/shared/src/main/scala/effekt/Intelligence.scala
```diff
@@ -113,24 +113,22 @@ trait Intelligence {
...
-          .collectFirst {
-            case ("doc", Some(s: String)) => s
-            case ("info", source.Info(Some(s: String), _, _)) => s
-          }
-      case _ => None
-  }
+    getDefinitionOf(s).flatMap {
+      case d: source.Def => d.doc
+      case _             => None
+    }
```
jiribenes, line 131: I don't understand the idea for the original code here, but `doc` is a property _solely_ on `source.Def`s, why not use it?

effekt/shared/src/main/scala/effekt/Intelligence.scala
jiribenes: They do. Fixed.

effekt/shared/src/main/scala/effekt/Intelligence.scala
jiribenes: cc @marvinborner bb2c2c1, why? :D

marvinborner: What about `ModuleDecl`, `Constructor`, `Operation`, is this also handled by this then? Otherwise mostly future-proofing I suppose.

jiribenes: ah, I missed those, thanks! I'll revert it since it seems like the other fix works too (but it didn't before [?])
apologies

marvinborner: No problem, I'd also like a better solution, this seems very hacky.

effekt/shared/src/main/scala/effekt/Intelligence.scala
```diff
@@ -113,11 +113,9 @@ trait Intelligence {
   } yield decl
 
   def getDefinitionOf(s: Symbol)(using C: Context): Option[Tree] = s match {
-    case u: UserFunction => Some(u.decl)
-    case u: Binder       => Some(u.decl)
-    case d: Operation    => C.definitionTreeOption(d.interface)
```
jiribenes, line 118: This one ^ is for a longer debate, but I think we should show something like:
```
<interface docs>
<operation docs>
```
(or the other way around) when hovering over an operation
and similarly for type constructors.

But I feel like that should happen somewhere else, the `.doc` property should be set to that in the first place (by parser already?)

PR #1396, opened 2026-04-30
Speedup implicits

marzipankaiser:
Speeds up implicits by just storing the Scope in namer (and not trying to resolve anything yet).
Now also contains the iterative deepening.

This might slightly change semantics of the implicitly passed parameters compared to the "desugared" version, since it does not replicate the post-filtering in Namer (`resolveFunctionCalltarget`).


b-studios:
Fantastic, this brings down the times of namer to almost the old state:

On nette effekte:

```
before (main):
- namer: 1828.06 ms
- typer: 1821.26 ms

after (this pr): 
- namer: 205.51 ms
- typer: 1939.00 ms
```

🎉

b-studios (approved):
Thanks, that's great!

PR #1402, opened 2026-06-08
Implement easy C externs

mattisboeckle:
Enables writing extern definitions such as
```scala
extern def glfwCreateWindow(width: C::Int, height: C::Int, title: C::String, monitor: C::Ptr, share: GLFWwindow): GLFWwindow = c "glfwCreateWindow"
```
which has to match a linked c function that has the correct name and types.

In this specific case

```c 
GLFWwindow* glfwCreateWindow(int width, int height, const char* title, GLFWmonitor* monitor, GLFWwindow* share);
```

This will automatically generate a wrapper function for said C function, which can then be called from effekt.
There are some selected types and utility functions in `effekt.effekt`


mattisboeckle:
@b-studios ping

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -27,42 +30,63 @@ object Transformer {
...
+
+        emit(Verbatim(s"declare ${returnLLVMType} @${cFunctionName}(${parameterLLVMTypes.mkString(", ")})"))
+        defineCWrapperFunction(functionName, transformedParameters, transformedReturnType) {
+          emit(Comment(s"C wrapper function for '${cFunctionName}'"))
+          // TODO: Check calling convention here
+          emit(Call("ccall", Ccc(), transformedReturnType, ConstantGlobal(cFunctionName), transformedParameters.map { case Parameter(typ, name) => LocalReference(typ, name) }))
+
+          transformedReturnType match {
+            case VoidType() => RetVoid()
+            case _          => Ret(LocalReference(transformedReturnType, "ccall"))
```
b-studios, line 62: This is always just generating a single call, right? What is the purpose of the `defineCWrapperFunction` then? It is only called here, right?

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -389,6 +413,10 @@ object Transformer {
             shareValues(List(value), freeVariables(rest))
             emit(Call(name.name, Ccc(), transform(name.tpe), ConstantGlobal("coerceNegPos"), List(transform(value))))
             eraseValues(List(name), freeVariables(rest))
+          case (machine.Type.CTpe(machine.CType.Void), machine.Type.Positive()) =>
+            ()
+          case (machine.Type.Positive(), machine.Type.CTpe(machine.CType.Void)) =>
+            ()
```
b-studios, line 419: Do these show up? This sounds dangerous since it violates that we should not use C::Void anywhere

effekt/shared/src/main/scala/effekt/machine/PrettyPrinter.scala
```diff
@@ -36,6 +36,7 @@ object PrettyPrinter extends ParenPrettyPrinter {
     case Type.Byte()         => "Byte"
     case Type.Double()       => "Double"
     case Type.Reference(tpe) => toDoc(tpe) <> "*"
+    case Type.CTpe(name)    => s"CType(${name})"
```
b-studios, line 39: Probably, I would just print name here. But this really does not matter :)

effekt/shared/src/main/scala/effekt/machine/Transformer.scala
```diff
@@ -88,7 +90,47 @@ object Transformer {
...
-  def transform(body: core.ExternBody[core.Expr])(using ErrorReporter): machine.ExternBody[Variable] = body match {
+  val validCTypes = List("ptr", "i64", "double", "float", "void")
+
+  def parseCType(tpe: String)(using ErrorReporter): Option[CType] = tpe match {
+    case "ptr"    => Some(machine.CType.Ptr)
+    case "i64"    => Some(machine.CType.I64)
+    case "double" => Some(machine.CType.Double)
+    case "float"  => Some(machine.CType.Float)
+    case "void"   => Some(machine.CType.Void)
+    case o        =>  None
```
b-studios:
```suggestion
    case o        => None
```

effekt/shared/src/main/scala/effekt/machine/Tree.scala
```diff
@@ -221,6 +230,7 @@ enum Type {
   case Byte()
   case Double()
   case Reference(tpe: Type)
+  case CTpe(tpe: CType)
```
b-studios, line 233: We should bikeshed this constructor name here. Same actually for `CType`. The strings on the right hand side of a c extern actually are llvm type names. So `CType` is a bit of a misnomer, or am I wrong?

libraries/common/effekt.effekt
```diff
@@ -794,4 +794,65 @@ effect splice[A](x: A): Unit
...
+  extern def Double(value: effekt::Double) at {}: Double = llvm "ret %Double ${value}"
+  extern def Float(value: effekt::Double) at {}: Float = llvm """
+    %trunc = fptrunc %Double ${value} to float
+    ret float %trunc
+  """
+  extern def Bool(value: effekt::Bool) at {}: Int = llvm """
+    %bool = extractvalue %Pos ${value}, 0
+    ret i64 %bool
+  """
+  extern def Nullptr() at {}: Ptr = llvm "ret ptr null"
```
b-studios: What about simply `Null`?

b-studios (commented):
Great work!

What I would love to see is a simple test, but I understand that might be difficult to integrate into our testing infrastructure?

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
mattisboeckle: The coercion also happens to return values of functions.
As we cannot coerce `void` or even pass it to a function as an argument we cannot use the existing infrastructure and need to handle it somewhere earlier. So there were two possible solutions.
1. Special case `CTpe(Void)` in machine transformer `perhapsUnbox` and never generate the coercion
2. Special case here and do this no-op coercion

I thought the latter looked slightly better

effekt/shared/src/main/scala/effekt/machine/Tree.scala
mattisboeckle: Yes. I thought more along the lines of types _for_ c externs instead of c types directly, but I am open to suggestions

libraries/common/effekt.effekt
mattisboeckle: You are right. `Nullptr` would be the C++ equivalent, but if we keep following LLVM names it should be `Null`
Updating it

mattisboeckle:
To reliably test this I would need to link against some c file in the testing environment. This leads back to the config file / command line debate and I would rather not hack around that for now.

effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
mattisboeckle: Just readability and consistency with `defineFunction` and `defineLabel`, but it is technically only called here. Yes.

b-studios (approved):

PR #1403, opened 2026-06-15
Add hex int syntax

mattisboeckle:
Allows for literals with exactly eight digits in hexadecimal format for Integers.
Works alongside the already existing syntax for byte literals in hex format.

Example
```scala
val a = 0xff00ff00ff
val b = 0x00000000
...
```


jiribenes:
I was wondering if we could have a new `Word` primitive type that would:
1. use `0wDEADBEEFCAFEBABE` literals (saw that SML had those, I'm willing to bikeshed the syntax),
2. be 64-bit unsigned on _all_ platforms (this would likely include some JS shim, or "polyfill" as they call it -- either wrapping a pair of `number`s or using `bigint`),
3. supported only casts into/from existing numeric types, and
4. all bitwise operations would need to be done on `Word`s or `Byte`s, that is no more bitshifts on `Int`s.

This would be useful for both FFI _and_ for more bit-twiddling, bitset-ey structures, etc. WDYT?

PR #1406, opened 2026-06-17
Fix argument count check

### LINKED ISSUE #1404: Crash because of uncaught wrong argument count in a 'resume' call
_Originally reported by @LeoZieger_

The following program ([Playground](https://effekt-lang.org/playground.html?playground=NY5BDoIwEEX3PcWXVRsW7knwEuoBahmgKbSkjhJDuLuj1eW8N3kzmVzKHS4pUNRGUd%2BTY8w2kDZNwUp11GPIdhmx4TYlFxq0J1yjZxy%2Fu9jRKoDzq3htsHoei9vEAJnuj1miMuylOFsvJ9H%2B02KedgKnIKxLvyeELtlHnqKuhsQffUCFusZAkbJ35zGtWqgxan8D&repl=bWFpbigp)) has a `resume` which ought to take a `Token` and return a `Unit`, but it does not report an error when it gets 0 arguments:
```scala
record Token()
effect make(): Token

def graph { block: => Unit / make } =
  try block() with make {
    resume() // <- notice no args!
    // ^ hover over and see `{ resume: Token => Unit }`
  }

def main() = graph {
  val tok = do make()
  println("got tok! " ++ genericShow(tok))
}
```

then somewhere inside Normaliser (even on the website), we get the actual error:
```
java.lang.AssertionError: assertion failed:
  Wrong number of value arguments:
    List(ValueParam(a,Data(Token,List())))
      vs
    List()
```

So the Typer [?] is likely overly permissive, allowing a bad program to proceed to compilation where it is rightfully rejected.

marzipankaiser:
Fixes #1404 .

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Typer.scala
+++ b/effekt/shared/src/main/scala/effekt/Typer.scala
@@ -1331,8 +1331,9 @@ object Typer extends Phase[NameResolved, Typechecked] {
     extension[A,B](self: Aligned[A, B])
       def fillImplicit[R <: A, C](use: List[C])(by: (Int, C) => Option[R]): (Aligned[A,B], List[R]) = {
         val useActually = use.drop(self.matched.length)
+        val definitelyMissing = self.missing.drop(useActually.length)
         val (remaining, impl) = self.missing.zip(useActually).zipWithIndex.partitionMap{ case ((a,c), i) => by(i, c).map((_, a)).toRight(a) }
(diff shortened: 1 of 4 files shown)


jiribenes:
Just to be safe, could you also add the other, simpler repro, as well? (the HOF calling a block with not enough args) 🙏

marzipankaiser:
@jiribenes Sure. You mean [That one](https://github.com/effekt-lang/effekt/issues/1404#issuecomment-4727230505)?

PR #1408, opened 2026-06-29
Add missing show binding

mattisboeckle:
We were missing a binding for Show, in the condition of an `if`.

For example the following would produce an error.
```scala
if (show(1) == show(1)) {
  ...
} else {
  ...
}
```


effekt/shared/src/main/scala/effekt/core/Show.scala
```diff
@@ -162,7 +162,7 @@ object Show extends Phase[CoreTransformed, CoreTransformed] {
       case If(cond, thn, els) =>
         val cond_ = transform(cond)
         ctx.withBindings {
-          If(transform(cond), transform(thn), transform(els))
+          If(transform(cond_), transform(thn), transform(els))
```
b-studios:
```suggestion
          If(cond_, transform(thn), transform(els))
```

b-studios (approved):

PR #1409, opened 2026-07-06
Memory managed C::Obj in C FFI

marvinborner:


effekt/shared/src/main/scala/effekt/generator/llvm/Transformer.scala
```diff
@@ -817,10 +832,12 @@ object Transformer {
 
   val sharePositive = ConstantGlobal("sharePositive");
   val shareNegative = ConstantGlobal("shareNegative");
+  val shareCObject = ConstantGlobal("effekt_share");
```
b-studios, line 835: Nitpick: do you think share and erase is the correct terminology at the right level for this API? How do C programmers typically think? What would be their terminology for this?

marvinborner: Personally I like share/erase, and I think it's fairly intuitive, but I'm also open to different naming. Instead of share/erase, better terminology might be ref/unref ([GObject](https://docs.gtk.org/gobject/method.Object.ref.html)), clone/drop (rust), incref/decref [python](https://docs.python.org/3/c-api/refcounting.html), acquire/release.

jiribenes: IMO as long as the operation is called share/erase in Machine and papers, it should also be called that in the FFI 😇

b-studios (approved):

PR #1412, opened 2026-07-19
Add 'string::utf8ByteLength'

jiribenes:
`string::utf8ByteLength` is `char::utf8ByteCount` but for whole strings: efficient on NodeJS and LLVM, but I also included a version written in Effekt because I don't know what Chez does...


examples/stdlib/string/utf8bytelength.check
```diff
@@ -0,0 +1,20 @@
...
+§é💩: 8 (internal: 8, bytearray: 8)
+©: 2 (internal: 2, bytearray: 2)
+¢: 2 (internal: 2, bytearray: 2)
+€: 3 (internal: 3, bytearray: 3)
+漢: 3 (internal: 3, bytearray: 3)
+अ: 3 (internal: 3, bytearray: 3)
+😀: 4 (internal: 4, bytearray: 4)
+𝄞: 4 (internal: 4, bytearray: 4)
+𐍈: 4 (internal: 4, bytearray: 4)
+مرحبا: 10 (internal: 10, bytearray: 10)
```
jiribenes, line 18: I'm a bit worried about the LTR/RTL switch here, but hopefully it's fine

examples/stdlib/string/utf8bytelength.effekt
```diff
@@ -0,0 +1,11 @@
+def one(str: String) = {
+  val bytes = str.utf8ByteLength
+  val bytes2 = str.internal::utf8ByteLength
+  val bytes3 = str.bytearray::fromString.size
+  println(s"${str}: ${bytes.show} (internal: ${bytes2.show}, bytearray: ${bytes3.show})")
+}
```
jiribenes, line 6: for the record: these three methods of computing the utf8 byte length should be the same, so the tests try all of them :)

PR #1414, opened 2026-07-28
Improve CRLF support in lexer

jiribenes:
This change should help with resolving #1183, though I can't confirm it's solved completely while being on a Mac.

I also changed the Windows smoke tests to run the case studies -- hopefully this means a slightly bigger subset of the language being checked on Windows CI :)

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/ParserTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/ParserTests.scala
@@ -32,7 +32,8 @@ object SpanSyntax {
     }
 
     def sourceAndPositions: (Source, Seq[Int]) = {
-      val lines = content.stripMargin.split("\n").toBuffer
+      val normalized = content.stripMargin.replace("\r\n", "\n")
+      val lines = normalized.split("\n", -1).toBuffer
       val positions = scala.collection.mutable.ArrayBuffer[Int]()
       var lineIdx = 0
       var lineBytePos = 0
@@ -230,7 +231,8 @@ class ParserTests extends munit.FunSuite {
 
     val textWithoutSpan =
       raw"""There is some content here.
-           |And here.""".stripMargin
+           |And here.
+           |""".stripMargin.replace("\r\n", "\n")
 
     assertEquals(span.from, 4)
     assertEquals(span.to, 33)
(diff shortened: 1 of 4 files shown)


jiribenes:
https://github.com/effekt-lang/effekt/blob/e15eca1292c5059fbcb4f3cf9fc5e6639352f0a9/effekt/shared/src/main/scala/effekt/Lexer.scala#L330-L335
is likely also a source of inconsistencies (if it doesn't "magically" work out _somehow_)

effekt/shared/src/main/scala/effekt/Lexer.scala
```diff
@@ -711,7 +713,7 @@ class Lexer(source: Source) extends Iterator[Token] {
       TokenKind.Comment(comment)
 
   private def shebang(): TokenKind =
-    advanceWhile { (curr, _) => curr != '\n' }
+    advanceWhile { (curr, _) => curr != '\n' && curr != '\r' }
     val command = getCurrentSlice(skipAfterStart = 2) // Remove `#!`
     TokenKind.Shebang(command)
 }
```
jiribenes, line 719: Though I don't expect too many CRLF enjoyers using shebangs, we should have a consistent "until the end of the line" predicate, hence this funny-looking change.

martin-ilgner:
And there are still some tests failing that probably should work, like `effekt.ParserTests.Correct multiline span`

jiribenes:
Yeah, I would assume that that's the `Position.advance` mentioned above ^ [?]

jiribenes:
@martin-ilgner if you have the time in the future, could you please try to run the tests again on this branch with an additional patch that changes:
https://github.com/effekt-lang/effekt/blob/e15eca1292c5059fbcb4f3cf9fc5e6639352f0a9/effekt/jvm/src/test/scala/effekt/ParserTests.scala#L34-L36

into:
```scala
def sourceAndPositions: (Source, Seq[Int]) = {
  val normalized = content.stripMargin.replace("\r\n", "\n")
  val lines = normalized.split("\n", -1).toBuffer
  val positions = = scala.collection.mutable.ArrayBuffer[Int]() 
  ...
```

I looked at the ParserTests and the `Position.advance` and it feels like the tests are wrong[er]

martin-ilgner:
@jiribenes I fixed the test (also had to normalize the expected value). I'm still confused by the regex above though.

effekt/jvm/src/test/scala/effekt/ParserTests.scala
jiribenes, line 239: alt. we could use `assertNoDiff(source.content, textWithoutSpan)`

jiribenes:
> I'm still confused by the regex above though.

WDYM? 🤔 
the `.split("\n", -1)` is Java's "split newlines and keep [trailing] empty lines"

PR #1415, opened 2026-07-31
Chez-CPS: Implement threads that can use handlers of the main thread

martin-ilgner:
*with many restrictions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@ licenses/
 .bsp/
 bin
 .metals/
+project/metals.sbt
+project/project
 .vscode/
 .bloop/
 *.tgz
--- a/effekt/shared/src/main/scala/effekt/Typer.scala
+++ b/effekt/shared/src/main/scala/effekt/Typer.scala
@@ -158,7 +158,11 @@ object Typer extends Phase[NameResolved, Typechecked] {
             usingCapture(capt)
             Result(vtpe, Pure)
         }
-        case b: BlockSymbol => Context.abort("Expected an expression, but got a block.")
+        case b: BlockSymbol => Context.lookup(b) match {
+          case (btpe, capt) =>
+            usingCapture(capt)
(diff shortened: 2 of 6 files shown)


martin-ilgner:
Example Program:
```scala
extern type Thread
extern type ThreadBoundary

extern chezCPS """
; Block, Boundary -> b
(define (with-boundary prog b)
  (prog (make-meta-cont top-level-k
                        (gensym "thread")
                        (create-store)
                        (make-thread-boundary (thread-boundary-mutex b)
                                              (thread-boundary-rest b)))
        top-level-k))
"""

extern def fork[T](boundary: ThreadBoundary){p: => T} at async: Thread =
  chezCPS "(k (fork-thread (lambda () (with-boundary ${p} ${boundary}))) ks)"

extern def initializeThreadBoundary() at async: ThreadBoundary =
  chezCPS "(k (make-thread-boundary (make-mutex) ks) ks)"

extern def join(thread: Thread): Unit =
  chezCPS "(thread-join ${thread})"

// For debugging
extern def displayMetaCont() at async: Unit =
  chezCPS """
    (begin
      (display "Metacontinuation:\n")
      (display ks)
      (display "\n\n")
      (k (void) ks))
  """

// Rules:
// - Threads accessing a handler of the main thread have to pass a thread boundary
// - The same thread boundary has to be used by all threads that use this handler/part of the main thread metacontinuation
// - The handler HAS TO RESUME
// - the main thread has to join
def main(): Unit = {
  try {
    val boundary = initializeThreadBoundary()
    val t1 = fork(boundary){
      println("Thread 1")
      do emit("1")
      do emit("2")
      println("Thread 1 terminated!")
    }
    val t2 = fork(boundary){
      println("Thread 2")
      do emit("11")
      do emit("22")
    }
    // do emit("main")
    join(t1)
    join(t2)
    println("Joined!")
  } with emit[String] { x =>
    resume(())
    println(x)
  }
}
```

PR #1417, opened 2026-08-01
Fix contification of blocks that are passed as an argument

jiribenes:
The non-minimal repro outright crashes, this one "only" miscomputes.

```scala
def twice(x: Int) { f: Int => Int }: Int = f(f(x))

def go(): Int = {
  def helper(x: Int): Int = x + 1
  val b = twice(10) {helper}   // value use
  helper(b)                    // tail call
}
```

`twice(10){helper}` = 12, `helper(12)` = 13.

but with `--no-optimize`, it returns `11` instead of `13`:

```diff
==> X effekt.JavaScriptTests.examples/pos/contify_block_used_as_value.effekt (js-noopt)  0.337s 
diff assertion failed
=> Obtained
    """|11
       |""".stripMargin
=> Diff (- obtained, + expected)
-11
+13
```


b-studios (approved):
The new analysis domain for Occurrences does make sense to me, thanks!

Issue #1385, opened 2026-04-21
Miscompilation of closure capture on JS


jiribenes:
The following program (v1) is miscomplied on JS, whereas v2 is compiled correctly.
This likely has something to do with contification / the direct-style transform on JS as Chez-CPS seems to be correct.

```scala
def main() = {
  var adders = []

  each(1, 4) { x =>
    println(x) // <- this line only on v1
    // println(s"debug ${x.show}") // <- this line only on v2
    adders = Cons(box { (y: Int) => y + x }, adders)
  }

  adders.foreach { adder =>
    println(adder(3))
  }
}
```

```sh
bash-5.3$ effekt --backend=js v1.effekt v2.effekt
1
2
3
7
7
7

debug 1
debug 2
debug 3
6
5
4

bash-5.3$ effekt --backend=llvm v1.effekt v2.effekt
1
2
3
6
5
4

debug 1
debug 2
debug 3
6
5
4

bash-5.3$ effekt --backend=chez-cps v1.effekt v2.effekt
1
2
3
6
5
4

debug 1
debug 2
debug 3
6
5
4
```

jiribenes:
Yeah, in v1, we mutate `i_1` instead of using `tmp_i_1`:
```js
function loop_2(i_1, ks_9, k_5) {
  loop_3: while (true) {
    if ((i_1 < (4))) {
      ...
      adders_0.set(new Cons_0((y_1, ks_8, k_4) => () => k_4((y_1 + i_1), ks_8), s_2));
      // ^ 2. this should use `tmp_i_1`
      ...
      const tmp_i_1 = i_1; // <- 1. this should be hoisted up
      i_1 = (tmp_i_1 + (1));
      continue loop_3;
...
```

which doesn't happen in v2 since we're not direct-style there:
```js
function loop_11(i_10, ks_42, k_29) {
  if ((i_10 < (4))) {
    const v_r_35 = (new Uint8Array((128)));
    const tmp_i_10 = i_10;
    function l_4(v_r_37, ks_41, k_24) {
      const v_r_36 = $effekt.println(v_r_37);
      const s_22 = adders_1.value;
      adders_1.set(new Cons_0((y_3, ks_40, k_23) => () => k_23((y_3 + tmp_i_10), ks_40), s_22));
      return loop_11((tmp_i_10 + (1)), ks_41, k_24);
    }
```

jiribenes:
Here's a diff of the CPS IR
```diff
[...]

PR #1421, opened 2026-08-09
Direct style2

b-studios:


b-studios:
It is starting to look good!

image
image
image

image


There is some performance regressions in the case studies that I need to investigate.

jiribenes:
It would be nice to know whether https://github.com/effekt-lang/effekt/issues/1385 and other programs like it tagged with `label:area:js AND label:bug` are working after these changes 👀

PR #1423, opened 2026-08-10
Add basic sharing (CSE) to Normalizer

### LINKED ISSUE #1083: Use information about previous pattern matches to optimize

jiribenes:
Resolves #1083 

This is a smaller part ofa bigger change linking the existing Normalizer to #1127, but already implementing a bad version of CSE (sharing pure expressions + propagating known information + unconditional hoisting) seems to be worth it -- just look at the VMTests :)


jiribenes:
(besides VMTests,) the test failure is that the following extern claims to be pure, but is not :D
=> #1424
https://github.com/effekt-lang/effekt/blob/8d87218e6c72cb0cd1805327f585e99a4ce1df10/libraries/common/process.effekt#L31-L36

effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
```diff
@@ -37,9 +37,56 @@ object Normalizer { normal =>
...
+    case _: Expr.Make => true
+    case _ => false
+  }
+
+  /** Are two equal values of this type interchangeable or could someone observe their identity? */
+  private def transparent(tpe: ValueType)(using C: Context): Boolean = tpe match {
+    case ValueType.Data(name, targs) => !C.decls.externDatas.contains(name) && targs.forall(transparent)
+    case ValueType.Var(_) => false
+    case ValueType.Boxed(_, _) => false
   }
```
jiribenes, line 90: This is super annoying, but I don't see a better way...

PR #1424, opened 2026-08-10
Fix process::default by marking it as impure

jiribenes:
See #1423 -- my cool CSE inlines it otherwise :D


jiribenes:
I'm not that confident in this anymore... What should be the capture? What do we assume for extern types?

jiribenes:
Essentially, we need to always ensure that even getters for mutable types are `io`, which is a bit weird...

jiribenes:
I'm closing this, see #1423 for the discussion -- adding `transparent` is probably a better fix, at least for now, I guess...

PR #1427, opened 2026-08-14
Tidy up stdlib 'map' and 'set'

jiribenes:
- Remove useless `.reverse` in `.toList` methods
- Fix `collect` not respecting the ordering of key-value pairs
- Use typed holes instead of panics to avoid `io` capture
- Use name-based implicits for `?compare` and `?show`

### DIFF
--- a/examples/stdlib/map/collect.check
+++ b/examples/stdlib/map/collect.check
@@ -1,0 +1,17 @@
+m1 := map::collect { stream() }
+m1.toList.foreach === m1.foreach
+1 -> uno
+2 -> two
+3 -> three
+1 -> uno
+2 -> two
+3 -> three
+----
+m2 := list::collect { stream() }.map:fromList
+m2.toList.foreach === m2.foreach
+1 -> uno
+2 -> two
+3 -> three
+1 -> uno
+2 -> two
+3 -> three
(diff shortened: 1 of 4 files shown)


examples/stdlib/map/collect.check
```diff
@@ -0,0 +1,17 @@
+m1 := map::collect { stream() }
+m1.toList.foreach === m1.foreach
+1 -> uno
+2 -> two
+3 -> three
+1 -> uno
+2 -> two
+3 -> three
```
jiribenes, line 8: previously this returned
```
1 -> one
2 -> two
3 -> three
```

even though the version below returned the correct thing, overriding `1 -> one` by `1 -> uno`

libraries/common/map.effekt
```diff
@@ -156,37 +156,48 @@ def filter[K, V](m: Map[K, V]) { shouldKeep: V => Bool }: Map[K, V] =
...
+def foreachReversed[K, V](m: Map[K, V]) { action: (K, V) => Unit }: Unit =
+  internal::foreachReversed(m.tree) {action}
+
 /// Convert a map `m` into a list of (key, value) pairs.
 ///
 /// O(N)
 def toList[K, V](m: Map[K, V]): List[(K, V)] = {
   var acc = Nil()
-  m.foreach { (k, v) =>
+  m.foreachReversed{ (k, v) =>
```
jiribenes:
```suggestion
  m.foreachReversed { (k, v) =>
```

jiribenes:
I'll merge this since it's technically just a refactor. If it somehow negatively influences effect-plots, I'll revert.

PR #1430, opened 2026-08-14
Add 'benchmarks/other/unify_dictionary.effekt' to VMTests

jiribenes:
The actual values might be off, this is copy-pasted from a different endeavour where I'd like to have `unify_dictionary.effekt` in the VMTests for comparison (`bytearray`-heavy)

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/vm/Builtin.scala
+++ b/effekt/shared/src/main/scala/effekt/core/vm/Builtin.scala
@@ -211,9 +211,16 @@ lazy val bytes: Builtins = Map(
 
   // Conversion
   // ----------
+  builtin("effekt::toInt(Byte)") {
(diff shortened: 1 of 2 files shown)


effekt/shared/src/main/scala/effekt/core/vm/Builtin.scala
```diff
@@ -298,7 +305,7 @@ lazy val arrays: Builtins = Map(
 )
 
 lazy val bytearrays: Builtins = Map(
-  builtin("bytearray::unsafeAllocate(Int)") {
+  builtin("bytearray::allocate(Int)") {
```
jiribenes, line 308: This was _again_ stale. Can we figure out a mechanism to make sure such renames are propagated somehow?

PR #1432, opened 2026-08-17
Add 'slow::find' to 'list' stdlib module

martin-ilgner:


examples/stdlib/list/find.effekt
```diff
@@ -0,0 +1,11 @@
+module examples/pos/list/find
```
jiribenes, line 1: 👀 the path seems wrong

libraries/common/list.effekt
```diff
@@ -820,6 +820,18 @@ def partition[A](l: List[A]) { pred: A => Bool }: (List[A], List[A]) = {
   (lefts.reverse, rights.reverse)
 }
 
+/// List functions that we discourage to use, but should still be available
+namespace slow {
+  /// Find the first element in a list which satisfies a given predicate.
+  /// If there is no such element, returns None.
+  ///
```
jiribenes, line 827: Other functions in this module also have the tests as "doctests"

libraries/common/list.effekt
```diff
@@ -820,6 +820,18 @@ def partition[A](l: List[A]) { pred: A => Bool }: (List[A], List[A]) = {
...
 }
 
+/// List functions that we discourage to use, but should still be available
+namespace slow {
+  /// Find the first element in a list which satisfies a given predicate.
+  /// If there is no such element, returns None.
+  /// 
+  /// O(N)
+  def find[A](list: List[A]){ pred: A => Bool }: Option[A] = list match {
+    case Cons(x, xs) => if (pred(x)) Some(x) else find(xs){ pred }
```
jiribenes: nit

```suggestion
    case Cons(x, xs) => if (pred(x)) Some(x) else find(xs){pred}
```

examples/stdlib/list/find.effekt
jiribenes: it's possible it's also wrong in the other files in this folder (via some refactoring)
if so, then it's fine and should be a follow-up issue 🙏

martin-ilgner: yes, its also wrong in the other files

libraries/common/list.effekt
martin-ilgner: What does this mean?

examples/stdlib/list/find.effekt
martin-ilgner: https://github.com/effekt-lang/effekt/issues/1433

libraries/common/list.effekt
jiribenes: The other functions of this module always have documentation tests as a part of their doc comments. Although we can't check them directly (yet), we mirror the tests from `stdlib/list/<functionname>` above the definition -- then they're visible in the generated documentation and on hover.

martin-ilgner: Added 👍

jiribenes (approved):

jiribenes:
Thanks!

PR #1434, opened 2026-08-19
Improve sharing on aliased values (CSE)

jiribenes:


jiribenes:
The default branch removal (https://github.com/effekt-lang/effekt/pull/1434/commits/fa967a9b390bca12c1daa2c3941537b76838270e) is only a code size thing, really, so it doesn't belong into this PR.

PR #1435, opened 2026-08-21
Speed-up array and bytearray

jiribenes:
- most array and bytearray operations on LLVM become much faster
- string comparison gets much faster on JS, speeding up all programs using `Map` with string keys greatly

### DIFF
--- a/effekt/jvm/src/test/scala/effekt/core/VMTests.scala
+++ b/effekt/jvm/src/test/scala/effekt/core/VMTests.scala
@@ -1230,10 +1230,10 @@ class VMTests extends munit.FunSuite {
     )),
 
     examplesDir / "benchmarks" / "other" / "unify.effekt" -> Some(Summary(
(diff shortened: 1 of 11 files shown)


effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -1230,10 +1230,10 @@ class VMTests extends munit.FunSuite {
     )),
 
     examplesDir / "benchmarks" / "other" / "unify.effekt" -> Some(Summary(
-      staticDispatches = 2506958,
+      staticDispatches = 6662383,
       dynamicDispatches = 1460350,
       patternMatches = 3850477,
-      branches = 2791752,
+      branches = 13982006,
```
jiribenes, line 1236: This looks worse, but it's really just the comparison being done in Effekt as opposed to via FFI...

PR #1437, opened 2026-08-21
Refactor and speed-up the 'dictionary' module

jiribenes:
The 'dictionary' used to be always dense, which made it very slow when iterated through / when looking through it. Therefore I added `Sparse` nodes/branches for cases where there's only `≤2` children (checking with a real dictionary, `hamlet.txt` and `bee-movie.txt`) -- now it's much faster.


jiribenes:
cc @phischu: before this and #1435, the immutable (tree-)`map` was always faster
after #1435 and this change, the `dictionary` now wins in most kinds of benchmarks I tried that involve a string-ey key
I refactored a lot along the way just so that I can understand what's going on

libraries/common/dictionary.effekt
```diff
@@ -21,40 +33,114 @@ record Tagged[A](key: ByteArray, value: A)
...
+  /// MUST NOT exceed 4 (`Int` is guaranteed only 32 bits ~> 32 / 8 = 4 bytes).
+  val maxSparse = 2
+
+  /// The `j`-th branching byte of a packed key set.
+  def byteAt(keys: Int, j: Int): Int = keys.bitwiseShr(8 * j).bitwiseAnd(255)
+
+  /// Index of `b` among the `count` keys, or -1 if it is not one of them.
+  def find(keys: Int, count: Int, b: Byte): Int = {
+    def go(j: Int): Int =
+      if (j >= count) -1
```
jiribenes, line 47: I'm not a fan of the sentinel here. I'd love to use `/ fail` or something, but the two call-sites are very different in how they use this.

phischu:
This is simpler and in my experiments even faster. What do you think?

jiribenes:
I tried something like this before (that's why the max factor was an adjustable constant), but it was much slower on JS IIRC. I can benchmark it later though and see.

jiribenes:
What I did measure is that on LLVM and JS (using the dictionary on `hamlet.txt` or something like that), 4 is always worse than 2...
EDIT: not _always_.

jiribenes:
I added a new standalone benchmark (because I didn't want to commit `hamlet.txt` or `bee-movie.txt`), but I promise it exercises the same paths that real text does + it distinguishes different variants I tried more than just what unify does.

One fun part is that on JS in this added benchmark, the tree`map` wins over all dictionary variants I've tried.

phischu:
Great, thank you!

PR #1439, opened 2026-08-25
Avoid allocating options in 'map' and 'set'

jiribenes:
`getOrElse`, `get`, `contains` in `map`/`set` used to allocate an Option every time, even though we only need to do so for `contains`

Let's introduce a `getWith` primitive and implement the other functions using it:
```scala
def getWith[K, V, R](m: Tree[K, V], compare: (K, K) => Ordering at {}, k: K) { found: V => R } { missing: => R }: R =
```

In all programs I have, `getOrElse`, `get` and `contains` all inline their `found` and `missing` calls :)
(in other words, the specialisation produces the best possible `getOrElse`, `get` and `contains` functions)


effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -1232,11 +1232,11 @@ class VMTests extends munit.FunSuite {
...
       dynamicDispatches = 1460350,
-      patternMatches = 3850477,
+      patternMatches = 3743980,
       branches = 13982006,
-      pushedFrames = 1970337,
-      poppedFrames = 1970337,
-      allocations = 2039967,
+      pushedFrames = 1863840,
+      poppedFrames = 1863840,
+      allocations = 1933470,
```
jiribenes, line 1239: We avoid ~100k allocations in `unify` :)

marvinborner (approved):
Looks good to me :shipit:

jiribenes:
Thanks for the speedy review, merging!

PR #1440, opened 2026-08-26
Nonuniform representation

phischu:
Finally we reap the benefits of monomorphization and use different representations for different types. Positions where we still need boxing, to be addressed in the future:

- Type variables in extern types `Array[T]`, `Ref[T]`, `Signal[A, B]`, `spawn[T]`.
- Answer types of regions

### DIFF
--- a/effekt/shared/src/main/scala/effekt/core/Mono.scala
+++ b/effekt/shared/src/main/scala/effekt/core/Mono.scala
@@ -978,7 +978,7 @@ object Mono extends Phase[CoreTransformed, CoreTransformed] {
           } else {
             targs.toList.sortBy(variantKey).foreach { variant =>
               val name =
-                if variant.isEmpty || id == core.Type.ResumeSymbol || id == core.Type.PromptSymbol then id
+                if variant.isEmpty || id == core.Type.ResumeSymbol || id == core.Type.PromptSymbol || id.isInstanceOf[symbols.ExternInterface] then id
                 else functionNames.fresh(preferredMonoName(id, variant))
               monoFunNames += ((id, variant) -> name)
             }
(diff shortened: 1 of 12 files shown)


b-studios:
> Positions where we still need boxing, to be addressed in the future:

What would be the right thing here to do?

Another question I have: how do you currently represent an empty closure? Is it allocated?

phischu:
> Another question I have: how do you currently represent an empty closure? Is it allocated?

Empty in the sense of no methods, or empty in the sense of no environment? If the environment happens to be empty there is no allocation. This is orthogonal to the number of methods.

phischu:
> What would be the right thing here to do?

Regarding "Type variables in extern types": open an issue and address separately.

Regarding "Boxed blocks": I am addressing it right now.

Regarding "Answer types of regions": Delete regions from the language.

b-studios:
Sorry, I didn't mean process but "what would be the right semantics?".

Regarding closures: yes, you're right. An empty environment is orthogonal, I see this now. This is of course symmetric to not having a payload on a Pos and only storing the tag. Which, however, is now also enabled by monomorphization.

b-studios:
Weird, the anf case study panics for me with

```
CLet(x, CLet(x1, CRet(CLit(42)), CLet(x2, CApp(g, CVar(x1)), CApp(f, CVar(x2)))), CRet(CVar(x)))
let x = let x1 = return 42 in let x2 =
      g(x1)
    in f(x2) in return x
----
PANIC: (null) not implemented yet
```
but this should've been caught by the CI, no?

b-studios:
Performance looks great for those where boxing was the main limitation:

image

jiribenes:
I'd be interested in the `other/` benchmarks because they're often data structure heavy, so (hopefully) specialized repr. ought to help them a lot

b-studios:
image

image

image

[comparison.html](https://github.com/user-attachments/files/31560089/comparison.html)

Doesn't look like a big win for "other"... Probably still dominated by other things.

phischu:
So, can we merge this or what?

b-studios (approved):

PR #1443, opened 2026-08-28
Fix scoping in block literals in 'Show'

### LINKED ISSUE #1442: 'Show' pass produces wrong output for 'show's used in impure positions
The following program ([playground](https://effekt-lang.org/playground.html?playground=RY2xCgIxEET7fMWQKuEaT6wCa2%2FtFwi38RZze2ICIpJ%2FNzmLq3YH5s2bOCKKTu7Bn4BreYnePb5YJOf2BtAZFy2oYTtkAInobRDBzpzSOtpOnI6o4JR5p51HNaYrlptoS4RnE5SkbnP%2BcYthQJ7Xtxt9Hzqg%2Bh8%3D&repl=bWFpbigp)):
```scala
def find(key: String) { missing: => Int }: Int =
  if (key == "hello1") { 42 } else { missing() }

def main() = println(find("hello" ++ show(1)) { 0 })
```

crashes on both LLVM and JS with some variant of "`Can't find variable: showApp_0`".

I think the following definition of `withBindings` is at fault:
https://github.com/effekt-lang/effekt/blob/de6641b3298d824fc95ab2ba9d20951bb6e9e44f/effekt/shared/src/main/scala/effekt/core/Show.scala#L36-L39

(`outer.toList` is evaluated _before_ the block)

jiribenes:
Resolves #1442
(see the failing tests on the first commit for the bug itself)


jiribenes:
I need to think about this a bit more -- feel free to overtake this anybody if you have the time.

PR #1445, opened 2026-08-30
Fix extern calling convention on AArch64

### LINKED ISSUE #1411: Segfault on LLVM in 'filesystem::readFile'
I only tested this on a Mac.

```scala
import io/error
import io/filesystem

def main() = {
  with on[IOError].panic
  println("before readFile")
  val s = readFile("data.txt") // any non-empty file
  println("after readFile")      // LLVM never gets here
  println(s.substring(0, 12))
}
```
```sh
$ effekt --backend=js repro.effekt
before readFile
after readFile
<here are the contents of data.txt>

$ effekt --backend=llvm repro.effekt
before readFile
[error] Process exited with non-zero exit code 138.
```

I tried minimising a bit: the only difference I can observe is a local vs global definition of the max chunk size:
```scala
import io
import io/error
import io/filesystem
import bytearray

// on LLVM, global works, local does not
// val chunkSizeGlobal = 1048576
// lines marked with a `// !` point to lines using the local chunk size

def readFile(path: String): String / Exception[IOError] = {
  val file = openForReading(path);
  with on[IOError].finalize { close(file) }
  val chunkSize = 1048576 // !
  var buffer = bytearray::allocate(chunkSize) // !
  var offset = 0
  def go(): String = {
    read(file, buffer, offset, chunkSize, -1) match { // !
      case 0 => buffer.resize(offset).toString
      case n =>
        offset = offset + n
        if (offset + chunkSize > buffer.size) { buffer = buffer.resize(2 * buffer.size) } // !
        go()
    }
  }
  go()
}

def main() = {
  with on[IOError].panic
  println(readFile("data.txt").substring(0, 5))
  println("ok")
}
```
```sh
$ effekt --backend=js repro2.effekt
<here are the contents of data.txt>
ok

$ effekt --backend=llvm repro2.effekt
[error] Process exited with non-zero exit code 139.
```

`--no-optimize` makes no difference.
I have not investigated further though.

jiribenes:
Resolves #1411 

TL;DR: calls that are not tail calls must be marked as such because of inlining

This _might_ be an upstream issue, but when we inline a ccc call into a tailcc function, we automatically tail, even though we should not. The resulting bad sibcall (tail call) is _not_ present in x86-64 codegen, because there's a special check in ISelLowering/CallLowering, but the bad call is present on AArch64 (which does _not_ have such a check in its ISel-/Call-Lowering), which makes us MacBook people crash in `input-output/` test (which this also fixes).


jiribenes:
Perf was not affected by this change: 
(https://github.com/effekt-lang/effekt-plots/issues/5#issuecomment-5468788015)

> # Backend outliers detected!
> 
> Configuration: `N=10`, significant if `modified z-score > 10.00`
> 
> * 🔼 in backend: llvm, file: `are_we_fast_yet/storage`, time: 0.890s (μ=0.836s, σ=0.017s, median=0.835s, MAD=0.0042s, mz=13.06)

PR #1446, opened 2026-08-30
Improve LLVM RTS: fix declarations, add assumption annotations

jiribenes:
Subsumes #1384

1. assume RC >= 0 (this seems to help a lot in some benchmarks like `bounce`: LLVM then can reason that `rc+1>0`)
2. const char* -> bytearray conversion was off-by-one on empty strings (see the commit description by clicking on `[...]` below)
3. fix memcpy declaration (clang couldn't pair it with the libc version and thus infer what the function does)
4. remove unused `%limit`-related computations from the RTS
5. fix exit declaration (same reason as memcpy, though this one actually lead to miscompilations)

### DIFF
--- a/libraries/llvm/rts.ll
+++ b/libraries/llvm/rts.ll
@@ -162,7 +162,7 @@ define private void @shareObject(%Object %object) alwaysinline {
 
     next:
     %objectReferenceCount = getelementptr %Header, ptr %object, i64 0, i32 0
-    %referenceCount = load %ReferenceCount, ptr %objectReferenceCount, !alias.scope !14, !noalias !24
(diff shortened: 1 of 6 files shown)


libraries/llvm/rts.ll
```diff
@@ -117,8 +117,7 @@ declare double @log1p(double)
 ; Intrinsic versions of the following two only added in LLVM 19
 declare double @atan(double)
 declare double @tan(double)
-declare void @print(i64)
```
jiribenes, line 120: this one is a drive-by: nothing uses `@print(i64)` as far as I can tell...

jiribenes:
I don't consider this controversial, so I'm merging this PR
if it worsens any perf on effekt-plots / if there's any dissent, I'm happy to revert it :)

PR #1447, opened 2026-09-01
Refactor Core inliner

### LINKED ISSUE #1444: Crash (OOM) on unbounded substitution when optimisations turned on
The following (heavily minimised) program crashes / never terminates _unless_ compiled with `--no-optimize`:
```
java.lang.OutOfMemoryError: Java heap space
  at effekt.core.substitutions$Substitution.shadowCaptures(Tree.scala:1044)
  at effekt.core.substitutions$.substitute(Tree.scala:1140)
```

```scala
def collectString { s: => Unit / { emit[Char], stop } }: String = {
  var a = bytearray::allocate(64)
  var i = 0
  boundary {
    try s() with emit[Char] { c =>
      if (i + 1 > a.size) { a = a.resize(2 * (i + 1)) }
      a.set(i, c.toInt.toByte)
      i = i + 1
      resume(())
    }
  }
  a.resize(i).toString
}

// note: more clauses here lead to an exponential blow-up (<12 are OK, 12 takes two minutes to optimise, >12 OOMs)
def isSpecial(c: Char): Bool =
  c == '*' || c == '_' || c == '`' || c == '[' || c == ']' || c == '(' || c == ')' ||
  c == '{' || c == '}' || c == '.' || c == '-' || c == '\'' || c == '"' || c == '\n'

def readWhile { pred: Char => Bool }: Unit / { emit[Char], peek[Char], next[Char], stop } = {
  val c = do peek[Char]()
  if (pred(c)) { do emit(c); do next[Char](); readWhile {pred} }
}

def lex(input: String): Unit / emit[String] =
  feed(input) {
    peeking[Char] {
      loop {
        do peek[Char]() match {
          case '*' => do next[Char](); do emit("*")
          case c   =>
            do next[Char]()
            do emit(collectString { do emit(c); readWhile { ch => not(isSpecial(ch)) } })
        }
      }
    }
  }

effect tree[R](tag: Int) { children: () => R }: R

// needs to keep state across children [!]
def intoCount { program: () => Unit / tree[Unit] }: Int = {
  var n = 0
  try { program() } with tree[Unit] { tag =>
   resume { {children} =>
      val outer = n
      n = 0
      children()
      n = outer + n + 1
    }
  }
  n
}

def main() =
  println(intoCount { for[String] { lex("a") } { _t => do tree(1) { () } } }.show)
```

jiribenes:
Adapted from #1127 
Resolves #1444 

Code size goes down (by 25% on larger projects), benchmarks become faster (some much faster, some "just" faster).

See descriptions in the individual commits for more information.
(I tried very hard to split this into reasonable-sized commits :))


effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -1278,19 +1278,19 @@ class VMTests extends munit.FunSuite {
...
+      allocations = 1146603,
       closures = 9,
       variableReads = 896183,
       variableWrites = 226310,
-      resets = 384,
-      shifts = 147456,
-      resumes = 147456
+      resets = 0,
+      shifts = 0,
+      resumes = 0
```
jiribenes, line 1293: I'm very proud of this one, by the way: `emit` and `set::collect` and `map::collect` become fused :)

effekt/shared/src/main/scala/effekt/core/optimizer/InliningPolicy.scala
```diff
@@ -0,0 +1,139 @@
...
+/** Inline a callee used exactly once, one with a known block argument, or one small enough. */
+class Unique(threshold: Int) extends InliningPolicy {
+  def apply(site: CallSite)(using Context): Boolean = site.boundBy match {
+    // an anonymous callee cannot be recursive (has no name)
+    case None => site.bargs.exists { b => b.isInstanceOf[BlockLit] }
+    case Some(callee) =>
+      !Normalizer.isRecursive(callee.id) &&
+        (Normalizer.isOnce(callee.id) || site.callee.body.size <= threshold)
+  }
+}
```
jiribenes, line 29: This is the previous policy: no idea if we should keep it, but it's much easier to read and understand it (and it might be helpful for debugging), hence why I didn't remove it myself.

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -300,12 +300,12 @@ class VMTests extends munit.FunSuite {
...
-      staticDispatches = 2022,
+      staticDispatches = 1146,
       dynamicDispatches = 0,
       patternMatches = 0,
-      branches = 3347,
-      pushedFrames = 1106,
-      poppedFrames = 1090,
+      branches = 2711,
+      pushedFrames = 230,
+      poppedFrames = 222,
```
b-studios, line 308: Wow!

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
```diff
@@ -548,15 +548,15 @@ class VMTests extends munit.FunSuite {
       dynamicDispatches = 0,
       patternMatches = 0,
       branches = 210,
-      pushedFrames = 68,
-      poppedFrames = 67,
+      pushedFrames = 1,
+      poppedFrames = 1,
```
b-studios, line 552: This one is also really, really good

effekt/jvm/src/test/scala/effekt/core/VMTests.scala
b-studios: indeed!!

PR #1449, opened 2026-09-02
Print positions on failing tests

marzipankaiser:
This changes the stdlib test library to print the source position of the failing assert.

### DIFF
--- a/examples/stdlib/acme.effekt
+++ b/examples/stdlib/acme.effekt
@@ -40,6 +40,7 @@ import result
 import scanner
 import seq
 import set
+import sourcepos
 import stream
 import string
 import stringbuffer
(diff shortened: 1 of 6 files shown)


marzipankaiser:
TODO: 
- [x] Fix check files.
- [x] ACME.

marzipankaiser:
Scheme doesn't support reading files, so it won't work with this approach for scheme 😞.

We could:
- Compute the string during compile-time and encode it into the SourcePos (or a "sourcetext" or sth).
- Somehow skip/ignore it for scheme
- Implement file reading for scheme

jiribenes:
tbh, I'd be fine if we do not show the source by default

Issue #1272, opened 2025-12-17
Optimizing through useless constructs / obfuscated code


jiribenes:
`examples/pos/with_val_else` uses a lot of features to just print the following two lines:
```txt
Read value: 42
Read value: 100
```

It uses bidirectional effects (`read` & `stop`), makes and deconstructs useless lists, modifies a mutable variable under a reset, uses a `while-let`, makes and deconstructs `Option`s, etc.

Here's the whole file:
https://github.com/effekt-lang/effekt/blob/68bf2a8892bd59ab0c4d05ced7b6ea51b043009a/examples/pos/with_val_else.effekt#L5-L18

In an ideal world, I'd like it to be optimised down to:
```scala
def main() = {
  println("Read value: 42")
  println("Read value: 100")
}
```

<details>
  <summary>Here's a screenshot of the core IR after the existing optimiser has ran</summary>

<img width="1912" height="1446" alt="Image" src="https://github.com/user-attachments/assets/d17e2569-bc1c-4a7a-9e7d-07cdfa59fca3" />

</details>

Issue #1064, opened 2025-07-01
Further optimising stream fusion


jiribenes:
I was trying out "stream fusion" (example from https://gitlab.haskell.org/ghc/ghc/-/wikis/sequent-core#cases-where-we-win), can we do better here somehow?
At the very least, I'd like this program to not be worse with a new optimiser, cc @dvdvgt.

[Playground link](https://effekt-lang.org/playground.html?playground=hZFfa4MwFMXf%2FRTnMYHQtd1%2FRze6vY697amUIRppmEYxcXOI331JjGlxhT0Uiveec373hOc5TzV4KfRuuydfSdHyGFsa410KHUUZz5GLQvPGjoWsWx3jVSi7TdHjk%2FP6yQiwecRzVRUYwhgb9BFgHZRueFKecRhjcDEBGI3bQJno9OD0QJoojjdREGpTCD1%2BfKmkIh1Dp9woq5wP6eiDjyRmYtaHyIO0MqAY%2BPF%2FbKUzjNMrXJxufoKAGPsRZ8C30Ieg6tEZL7ePwNZw1ZacEEpH8MCiQq9ne%2FyP709xgXOinMN5MNdch0Rm6BYud97dEflU8mHXZiN3zFSqfy5%2Fl28jZJiDvJvncw892difMbO9lImQ9qlRN0LqQpLdimHNcMlwxXDNcMNwy3DHcM%2BwWu4X87yyysjaGGywxEB%2FAQ%3D%3D&repl=bWFpbigp)

```scala
effect emit[A](value: A): Unit

def filter[A](input: List[A]) { keep?: A => Bool }: List[A] = {
  def stream[A](input: List[A]): Unit / emit[A] = input match {
    case Nil() => ()
    case Cons(x, xs) => do emit(x); stream(xs)
  }

  def unstream[A] { stream: => Unit / emit[A] }: List[A] =
    try { stream(); Nil() } with emit[A] { x =>
      Cons(x, resume(()))
    }

  def sfilter[A] { keep?: A => Bool } { stream: => Unit / emit[A] }: Unit / emit[A] = 
    try stream() with emit[A] { 
      case x and x.keep? => do emit(x); resume(())
      case _ => resume(())
    }

  unstream {
    sfilter { x => x.keep? } {
      stream(input)
    }
  }
}

def main() = println([1, 2, 3, 4, 5, 6, 7, 8, 9, 10].filter { x => x.mod(2) == 0 })
```

Core IR:

[...]

Issue #995, opened 2025-05-14
Missing reset-and-immediately-shift-abort optimisation


jiribenes:
```scala
effect fail(): Nothing
def main() = try {
  do fail();
  println("after")
} with fail { println("failed") }
```

produces the following optimised core with a `reset` immediately followed by a `shift`:
```scala
interface fail_3421 {}

def main_3422() = reset { {p_3561} => 
  shift(p_3561) { {k_3584} => 
    let tmp_3586 = println_1("failed")
    return tmp_3586
  }
}
```

But we can clearly eliminate this, right?
```scala
// ~>
def main_3422() = {
  let tmp_3586 = println_1("failed")
  return tmp_3586
}
```

Issue #992, opened 2025-05-14
Redundantly creating data just to destroy it, using only the tag


jiribenes:
_Extracted from #988_

Full standalone program:
```scala
import exception
import result

def digitValue(char: Char): Int / Exception[WrongFormat] =
  if (char >= '0' && char <= '9') {
    char.toInt - '0'.toInt
  } else {
    wrongFormat("Not a valid digit: '" ++ char.toString ++ "' in base 10")
  }

def isDigitBad(char: Char): Bool =
  result[Int, WrongFormat] { digitValue(char) } match {
    case Success(_) => true
    case Error(_, _) => false
  }

// ^ equivalent to:
// def isDigitBad2(char: Char): Bool =
//   try { Success(digitValue(char)) } with Exception[WrongFormat] {
//     def raise(e, msg) = Error(e, msg)
//   } match {
//     case Success(_) => true
//     case Error(_, _) => false
//   }

def isDigitGood(char: Char): Bool =
  try { digitValue(char); true } with Exception[WrongFormat] {
    def raise(e, msg) = false
  }

def main() = println(isDigitGood('1')) // feel free to swap for `isDigitBad`
```

---

Resulting optimised core of `println(isDigitBad('1'))`
```scala
def main_3427() = {
  val v_r_3920: Result_1953[Int_405, WrongFormat_552] = reset { {p_3936} => 
    if (infixGte_3146(49, 48)) {
      val v_r_3985: Int_405 = if (infixLte_3140(49, 57)) {
        return infixSub_105(toInt_3132(49), toInt_3132(48))
      } else {
        shift(p_3936) { {k_4042} => 
          return make Result_1953[Int_405, WrongFormat_552] Error_1971(make WrongFormat_552 WrongFormat_611(), infixConcat_35(infixConcat_35("Not a valid digit: '", toString_3130(49)), "' in base 10"))
        }
      };
      return make Result_1953[Int_405, WrongFormat_552] Success_1976(v_r_3985)
    } else {
      shift(p_3936) { {k_4043} => 
        return make Result_1953[Int_405, WrongFormat_552] Error_1971(make WrongFormat_552 WrongFormat_611(), infixConcat_35(infixConcat_35("Not a valid digit: '", toString_3130(49)), "' in base 10"))
      }
    }
  };
  v_r_3920 match {
    case Success_1976 { (a_3960: Int_405) => 
      let tmp_4040 = println_1("true")
      return tmp_4040
    }
[...]

PR #1451, opened 2026-09-03
Add all 'optimizer-wishlist' programs into VMTests

jiribenes:
Adds the programs from all known programs from issues tagged with `optimizer-wishlist`, namely:
- #972 
- #992 
- #995 
- #1064 
- #1272 
- #842 (since it detects bad inlining well)
- and a program of my own which aborts over a var in a region

... and a few drive-bys to make my job easier down the line.

Adds only like two seconds to the VMTests run, so it's cheap enough :)


jiribenes:
Since this is just a building block for further work on optimization, I'll merge it as soon as the test pass.

Issue #1128, opened 2025-09-23
RFC: Raw multi-line strings delimited by a prefix token


jiribenes:
> [!NOTE]
> This is a request for comments in very early stages, feel free to comment/nitpick below.

Inspired by [Zig](https://ziglang.org/documentation/master/#Multiline-String-Literals) and [MoonBit](https://docs.moonbitlang.com/en/latest/language/fundamentals.html#string), use a prefix operator for denoting a multi-line string, which has a few advantages: easier to lex [untested?], each token spanning at most a single line [ => ability to lex each line separately ], and the allowing the user to have control over indentation of the prefix.

The main disadvantage is that this spends more of the syntax weirdness budget on something unnecessary, confusing researchers, students, and LLMs alike.

## A preliminary design

Multi-line strings are denoted with `#|`, are _raw_ (no escapes), but allow splices: 
```scala
val greeting = 
    #|Hello ${name}!
      #|Welcome to our application.
    #| Today is ${do datetime().day.format} and the emoji of the day is #|--|#...\nCool, huh?

// ~> output:

//Hello Maxine!
//Welcome to our application.
// Today is Tuesday and the emoji of the day is #|--|#...\nCool, huh?
```

We can also support multi-line splices by adding a `$|` token that denotes currently being in a splice:
```scala
val template = 
    #|<div class="user-card">
    #|  <h2>${user.name}</h2>
    #|  <p>Items: <ul>${
    $|    user.items.map { item =>
    $|      "<li>${item.title}</li>"
    $|    }.join("\n")</ul>
    $|  }</p>
    #|</div>
```

and we can even nest multi-line strings in multi-line splices, were we so inclined (number of `$|`'s marks the nesting depth of the "current" splice)
```scala
def messageFor(user: User): String = 
    #|Welcome ${user.name}!
    #|
    #|Your personalized content: ${
    $|  generateContent(
    $|    #| Dear ${user.name},
    $|    #| Here's your nested message: ${
    $|    $|   processData(user.preferences)
    $|    $| }
    $|    #| Best regards!
    $|  )
    $|}
    #|Thanks for visiting!
```

A nice thing is that we can still support multi-line splices, but we can hope that our users won't need to learn `$|` and the nesting rules unless they really, really want to.

Together with this, I'd also be in favour of banning multi-line splices in single-line strings completely + revising hole string syntax to be [visually] compatible.

### An alternative design

[...]

PR #1452, opened 2026-09-04
Feature/Experiment: Line-based raw strings

marzipankaiser:
This is a shot at implementing a proof-of-concept for #1128.

It allows raw strings with the following syntax:
```
val someString = #|function thisIsTheString() {
                 #|  return "${
                 $|     "allowing splices"
                 $|  }";
                 #|}
```

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Parser.scala
+++ b/effekt/shared/src/main/scala/effekt/Parser.scala
@@ -13,6 +13,7 @@ import kiama.util.{Position, Range, Source}
 
 import scala.annotation.{tailrec, targetName}
 import scala.collection.mutable.ListBuffer
+import scala.collection.mutable
 import scala.language.implicitConversions
 import scala.util.boundary
 import scala.util.boundary.break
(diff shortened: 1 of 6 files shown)


effekt/shared/src/main/scala/effekt/Parser.scala
```diff
@@ -844,6 +856,19 @@ class Parser(tokens: Seq[Token], source: Source) {
...
     nonterminal:
       expect("string literal") {
+        case RawStr(fst) =>
+          val res = mutable.StringBuilder(fst)
+          @tailrec
+          def go(): String = {
+            peek.kind match {
+              case t@RawStr(s) =>
+                skip()
+                res.addOne('\n'); res.append(s)
```
jiribenes: there should be some way to get the exact Newline token at this position and use its value, then it doesn't break on Windows horribly

marzipankaiser: Newlines should now be properly preserved :)

PR #1453, opened 2026-09-05
Optimize both tail resumptions and tail aborts, eliminate unobservable delimiter-likes

### LINKED ISSUE #992: Redundantly creating data just to destroy it, using only the tag

### LINKED ISSUE #995: Missing reset-and-immediately-shift-abort optimisation

### LINKED ISSUE #1272: Optimizing through useless constructs / obfuscated code

jiribenes:
Resolves #1272 
Resolves #995 
Resolves #992 

This PR eliminates delimiters and delimiter-likes that cannot be observed.

The TL;DR is that a delimiter is only needed when control can leave differently from how it arrived.
If every exit of a shift's body transfers control to where the delimiter would have,
the capture is not observable and the whole handler can go :)
We add some commuting conversions ("crossings"), but we cannot move a stack segment (`Var`) across a delimiter when observed.

`RemoveTailResumptions` now removes both tail resumptions and tail non-resumptions, sharing a bunch of code about what it means to be in tail position.\*

### Results

- many `optimizer-wishlist` programs are optimised to 0 handlers (#992, #995, #1272)
  - many more were added in this very PR and subsequently optimized to 0 
- many real programs are 3-5% faster and smaller
- `casestudies/parser` loses many resets and closures, `even_odd_goto` becomes a loop, `parsing_dollars` loses all delimcc
  - 13 VM baselines move
- optimizing is now up to ~20% slower (but only +1% of total compile time) 
- the interesting benchmarks on effekt-plots are _mostly_ unchanged
- the following program gets optimised to have no delimcc (even though it uses variables, pull-streams, etc.)
```scala
def parse(s: String): Int = {
  with feed(s)
  var acc = 0
  loop { acc = 10 * acc + (do next[Char]().toInt - 48) }
  acc
}

def main() = println(parse("12345"))
```

_\* `tailResumptive` is "every exit resumes, so the shift is a call", `removeTailAborts` is "the exit is the answer, the prompt is the next frame, so the abort is a return"_


effekt/shared/src/main/scala/effekt/core/optimizer/Optimizer.scala
```diff
@@ -40,7 +40,8 @@ object Optimizer extends Phase[CoreTransformed, CoreTransformed] {
     val onceLimit = Context.config.maxOnceInlineSize().toInt
     val policy = Default(
       threshold = Context.config.maxInlineSize().toInt,
-      onceLimit = Option.when(onceLimit >= 0)(onceLimit)) // negative means no limit
+      onceLimit = Option.when(onceLimit >= 0)(onceLimit), // negative means no limit
+      carryingLimit = Context.config.maxCarryingInlineSize().toInt)
```
marzipankaiser, line 44: Does it make sense to also allow `carryingLimit` to be `None` for consistency with `onceLimit`?
Not really important, was just wondering on skimming it.

effekt/shared/src/main/scala/effekt/core/optimizer/RemoveTailResumptions.scala
```diff
@@ -11,71 +11,162 @@ object RemoveTailResumptions {
...
-      case Stmt.Hole(tpe, span) => true
+  /**
+   * Tail positions of a statement, storing only its rewrite.
+   *
+   * @param transparent blocks whose tail calls are tail positions too
+   */
+  case class TailPositions(rewrite: (Stmt => Stmt) => Stmt, skipped: Free, transparent: Set[Id] = Set.empty) {
+    def forall(p: Stmt => Boolean): Boolean = {
+      var holds = true
+      rewrite { child => holds &&= p(child); child }
```
marzipankaiser, line 31: ~~Is this using `rewrite` just as a `foreach` over the children?
This `forall` function somehow feels as if it could be a general utility... - It only uses rewrite from the context, and that only to iterate over the "children".
(We could always move it on the second use, though.)~~

I understand, reading further and understanding more about `TailPositions`. I think some documentation about what `rewrite` is would have helped.

effekt/shared/src/main/scala/effekt/core/optimizer/RemoveTailResumptions.scala
```diff
@@ -11,71 +11,162 @@ object RemoveTailResumptions {
...
-      case Stmt.Get(ref, annotatedCapt, tpe, id, body) => tailResumptive(k, body)
-      case Stmt.Put(ref, annotatedCapt, value, body) => tailResumptive(k, body) && !freeInExpr(value)
-      case Stmt.Reset(BlockLit(tparams, cparams, vparams, bparams, body)) => false
-      case Stmt.Shift(prompt, k, body) => stmt.tpe == Type.TBottom
-      case Stmt.Resume(k2, body) => k2.id == k // what if k is free in body?
-      case Stmt.Hole(tpe, span) => true
+  /**
+   * Tail positions of a statement, storing only its rewrite.
+   *
+   * @param transparent blocks whose tail calls are tail positions too
```
marzipankaiser, line 26: Maybe documenting `rewrite` and `skipped` here would help to understand the code?

effekt/shared/src/main/scala/effekt/core/optimizer/RemoveTailResumptions.scala
```diff
@@ -11,71 +11,162 @@ object RemoveTailResumptions {
...
+   * A statement stands in tail position when the enclosing computation's answer is its answer.
+   *
+   * @param crossesVar whether a `Var` preserves tail position.
+   *   It does preserve it for a handler that never resumes, but does **not** for one that observes it.
+   *   (in other words: crossing a [[Var]] is not always semantics-preserving)
+   */
+  def tailPositions(stmt: Stmt, crossesVar: Stmt.Var => Boolean): Option[TailPositions] = stmt match {
+    // a block that is only ever tail-called adds no frame ~> its own tail positions are tail positions here (and so is every call to it)
+    case Stmt.Def(id, BlockLit(tps, cps, vps, bps, inner), body)
+      if tailCalledOnly(id, body, crossesVar) && tailCalledOnly(id, inner, crossesVar) =>
```
marzipankaiser, line 47: That condition does not *exactly* match the comment above, does it?
Here, we are also checking that id is tail-recursive IIUC? I assume this is the case since non-tail calls break this for the "second level" calls, and we don't want to distinguish them?

effekt/shared/src/main/scala/effekt/core/optimizer/RemoveTailResumptions.scala
```diff
@@ -11,71 +11,162 @@ object RemoveTailResumptions {
...
-    case Stmt.Get(id, annotatedTpe, ref, annotatedCapt, body) => Stmt.Get(id, annotatedTpe, ref, annotatedCapt, removeTailResumption(k, tpe, body))
-    case Stmt.Put(ref, annotatedCapt, value, body) => Stmt.Put(ref, annotatedCapt, value, removeTailResumption(k, tpe, body))
+  /**
+   * A statement stands in tail position when the enclosing computation's answer is its answer.
+   *
+   * @param crossesVar whether a `Var` preserves tail position.
+   *   It does preserve it for a handler that never resumes, but does **not** for one that observes it.
+   *   (in other words: crossing a [[Var]] is not always semantics-preserving)
+   */
+  def tailPositions(stmt: Stmt, crossesVar: Stmt.Var => Boolean): Option[TailPositions] = stmt match {
```
marzipankaiser, line 44: IIUC, this does this for one level (not recursively)?
So, for `val x = {val y = b}` it gives us `val y = b` as the statement in tail position, not `b`?
Maybe that's worth mentioning in the doc comment (or on the docs for `rewrite`).

effekt/shared/src/main/scala/effekt/core/optimizer/RemoveTailResumptions.scala
```diff
@@ -11,71 +11,162 @@ object RemoveTailResumptions {
...
+      case Some(positions) =>
+        !positions.skipped.contains(k) &&
+          positions.forall(tailResumptive(k, _, transparent ++ positions.transparent))
+
+      // each leaf has to account for every occurrence of `k` in its subterms
+      // (the cases above handle this with `skipped`)
+      case None => stmt match {
+        case Stmt.Resume(k2, body) => k2.id == k && !body.free.contains(k)
+        // an abort never returns, so it resumes vacuously; **unless** it resumes `k`!
+        case Stmt.Shift(_, _, body) => stmt.tpe == Type.TBottom && !body.free.contains(k)
```
marzipankaiser, line 141: I'm not sure I understand. Does this case mean that we're *not* actually going to end by resuming `k`? Then it would be good to change the doccomment of this function to also note that possibility (and maybe the name, if we have a better one).

marzipankaiser (approved):
Looks OK to me.
I would welcome some more documentation in a few places (see comments) to make it easier to understand (at least for me), though.

effekt/shared/src/main/scala/effekt/core/optimizer/Optimizer.scala
jiribenes: I _think_ that miscompiles on LLVM where we assume that `new Foo { def foo() = 42 }.foo()` appears nowhere.

effekt/shared/src/main/scala/effekt/core/optimizer/RemoveTailResumptions.scala
jiribenes: Yeah, I don't know how to describe this better 😅

PR #1455, opened 2026-09-05
Add 'unreachable' to LLVM, assume switch is well-formed

jiribenes:
Every `match { ... } /* no else */` gets transformed into `match { ... } else unreachable` on LLVM.

First commit taken from #1434.


jiribenes:
I had this ready even before #1440 landed, but now we should be able to finally upstream this.
I _think_ I ran the benchmarks, but I don't have the results written down anywhere (D'oh!)

jiribenes:
`stdlib/set/unique` fails in CI, but works on my machine, gotta investigate
```diff
==> X effekt.StdlibLLVMTests.examples/stdlib/set/unique.effekt (llvm)  3.201s munit.ComparisonFailException: /home/runner/work/effekt/effekt/effekt/jvm/src/test/scala/effekt/EffektTests.scala:120
119:        test(s"${f.getPath} (${backendName})") {
120:          assertNoDiff(run(f, true), expected)
121:        }
diff assertion failed
=> Diff (- obtained, + expected)
+Cons(Effekt, Cons(after, Cons(and, Cons(around, Cons(ask, Cons(better, Cons(but, Cons(can, Cons(do, Cons(ever, Cons(faster, Cons(fellow, Cons(for, Cons(harder, Cons(hour, Cons(is, Cons(it, Cons(language, Cons(make, Cons(makes, Cons(man, Cons(more, Cons(my, Cons(never, Cons(not, Cons(of, Cons(over, Cons(programmers, Cons(programs, Cons(so, Cons(stronger, Cons(than, Cons(the, Cons(together, Cons(us, Cons(we, Cons(what, Cons(will, Cons(work, Cons(world, Cons(you, Cons(your, Nil()))))))))))))))))))))))))))))))))))))))))))
-[error] Process exited with non-zero exit code 139.
-[error] Valgrind log:
-==5961== Jump to the invalid address stated on the next line
-==5961==    at 0x0: ???
-==5961==    by 0x4037FFF: ??? (in /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2)
-==5961==    by 0x1FFEFFF49F: ???
-==5961==    by 0x12BAEF: main (main.c:38)
-==5961==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
-==5961== ∙
-==5961== ∙
-==5961== Process terminating with default action of signal 11 (SIGSEGV)
-==5961==  Bad permissions for mapped region at address 0x0
-==5961==    at 0x0: ???
-==5961==    by 0x4037FFF: ??? (in /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2)
-==5961==    by 0x1FFEFFF49F: ???
-==5961==    by 0x12BAEF: main (main.c:38)
```

effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala
```diff
@@ -122,25 +125,32 @@ ${indentedLines(instructions.map(show).mkString("\n"))}
...
-    case ConstantInt(n)                     => s"i64 $n"
-    case ConstantByte(n)                    => s"i8 $n"
-    case ConstantDouble(n)                  => s"double $n"
-    case ConstantAggregateZero(tpe)         => s"${show(tpe)} zeroinitializer"
-    case ConstantNull(tpe)                  => s"${show(tpe)} null"
-    case ConstantArray(memberType, members) => s"[${members.length} x ${show(memberType)}] [${commaSeparated(members.map(show))}]"
-    case ConstantInteger8(b)                => s"i8 $b"
+  def show(operand: Operand): LLVMString = {
+    val (typ, name) = showOperand(operand)
+    s"${show(typ)} $name"
```
jiribenes, line 153: I don't like this: some LLVM instructions have only a single type, not a type at every operand...

effekt/shared/src/main/scala/effekt/core/optimizer/Normalizer.scala
```diff
@@ -280,7 +296,7 @@ object Normalizer { normal =>
         val normalized = normalize(scrutinee)
         Stmt.Match(normalized, tpe, clauses.map { case (tag, clause) =>
           tag -> normalize(clause)(using selecting(normalized, tag, clause))
-        }, default.map(normalize))
+        }, default.filter(_ => !covers(normalized.tpe, clauses)).map(normalize))
     }
```
jiribenes, line 300: TODO(@jiribenes): `foo match { /* no cases */ } else { ... }` should be rewritten to `...` somewhere around here.

effekt/shared/src/main/scala/effekt/generator/llvm/PrettyPrinter.scala
```diff
@@ -122,25 +125,32 @@ ${indentedLines(instructions.map(show).mkString("\n"))}
...
+    case Ret(operand) =>
       s"ret ${show(operand)}"
     case Switch(operand, defaultDest, dests) =>
       def destAsFragment(dest: (Int, String)) = s"i64 ${dest._1}, label ${localName(dest._2)}";
       s"switch ${show(operand)}, label ${localName(defaultDest)} [${spaceSeparated(dests.map(destAsFragment))}]"
     case CondBr(condition, trueDest, falseDest) =>
       s"br ${show(condition)}, label ${localName(trueDest)}, label ${localName(falseDest)}"
+    case Unreachable() =>
+      s"unreachable"
+  }
```
jiribenes, line 137: after this is live, both JS and LLVM have a concept of `unreachable`
we should at least consider pushing this into Core [and CPS / Machine], then we can use it "sooner".

jiribenes:
Rebased on top of #1463, perhaps we'll know more now / the error will be gone [?]

jiribenes:
> ERROR:nix_fast_build:BUILD: 17 successes, 1 failures
> ERROR:nix_fast_build:Failed attributes: .#packages.x86_64-linux.tracing-effective-rays

(so it works on the community build, the failure is preexisting)

PR #1459, opened 2026-09-09
Add support for stream literals

marzipankaiser:
This adds support for stream literals such that:
```
foo#[a,b,c]
```
is desugared to
```
foo{
  do emit(a)
  do emit(b)
  do emit(c)
}
```

and
```
#[a,b,c]
```
is desugared to
```
locally {
  do emit(a)
  do emit(b)
  do emit(c)
}
```

Note that this is more permissive than `[a,b,c].each`, which requires a, b and c to have the same type.

### DIFF
--- a/effekt/shared/src/main/scala/effekt/Lexer.scala
+++ b/effekt/shared/src/main/scala/effekt/Lexer.scala
@@ -93,6 +93,7 @@ enum TokenKind {
   case `(`
   case `)`
   case `[`
+  case `#[`
   case `]`
   case `,`
   case `'`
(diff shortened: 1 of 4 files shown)


marzipankaiser:
If we add them, adding something like https://www.reddit.com/r/ProgrammingLanguages/comments/1wmjqm0/comment/pb88smg/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button
would be mostly a small-ish parser change, if we were interested 🤔

Issue #1264, opened 2025-12-16
Add an override flag specifying which exact Clang binary to use


jiribenes:
-  'EFFEKT_CLANG_PATH' env var
- '--clang-path' config option

jiribenes:
This would help unblocking effekt-nix which currently doesn't work with LLVM builds because of this ^.

jiribenes:
For posterity, effekt-nix now explicitly binds every clang as `clang-XY` [see below] just to avoid this, so a flag or something would be preferable :)
```sh
for v in 18 19 20 21; do
  if [ ! -e $out/bin/clang-$v ]; then
    ln -sf $out/bin/clang $out/bin/clang-$v
  fi
done
```

PR #1462, opened 2026-09-11
Try to link libuv even when it's not found

jiribenes:
Emitting `Cannot find libuv on ${OS}; please use --clang-libraries and --clang-includes to configure the paths for the libuv dylib and header files, respectively.` is morally correct, however very often Clang can help us out by finding the actual `libuv` library by itself when we pass `-luv` to it. So let's let it help us, there's not much we can lose here (and it helps on my machine / with Nix packaging).


jiribenes:
Related to #1264

PR #1465, opened 2026-09-14
Switch order of Mono & Show passes

mattisboeckle:
Previously `Show` used to complain about too much type indirection if used in a polymorphic context. By moving monomorphization before `Show` and keeping the names the same we can swap their order in the LLVM pipeline.

In debug mode the type/function names will still have their types as suffixes for easier comparison


effekt/shared/src/main/scala/effekt/core/Mono.scala
```diff
@@ -1321,18 +1306,17 @@ object Mono extends Phase[CoreTransformed, CoreTransformed] {
...
-        val monoData: ValueType.Data = ValueType.Data(names.fresh(preferredMonoName(dataName, tpes)), List.empty)
+        val monoData: ValueType.Data = ValueType.Data(fresh(preferredMonoName(dataName, tpes)), List.empty)
         monoTypeNames += ((dataName, tpes) -> monoData)
         monoData
       })
     }
 
     private def preferredMonoName(baseId: Id, tpes: Vector[GroundType]): String =
-      baseId.name.name + tpes.map(typeName).mkString
+      baseId.name.name + util.debug(tpes.map(typeName).mkString)("")
```
jiribenes, line 1319: 👀

